PR #2, opened 2018-07-17
Add django integration

untitaker:
API same as raven except that capture_exception can be used normally


untitaker:
cc @mitsuhiko

PR #54, opened 2018-09-06
fix: Filter log records depending on client config

untitaker:


untitaker:
testing is imo covered by existing tests for capture_internal_exception since it's all the same logic now, but lmk if you want more

PR #63, opened 2018-09-14
docs: Build docs on Travis

tonyo:
...so they can later be published by "craft".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ pip-log.txt
 .idea
 .eggs
 venv
+.venv
 .vscode/tags
 .pytest_cache
 .hypothesis
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,14 +28,20 @@ matrix:
     - python: "3.8-dev"
       dist: xenial
       sudo: true
-    - python: "3.6"
+
+    - name: Linting
+      python: "3.6"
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -15,6 +15,7 @@ pip-log.txt
 .idea
 .eggs
 venv
+.venv
```
untitaker: hmm, where does this venv come from

tonyo: It was created when I ran `make test`

untitaker: oh, ok 👍 just saw that the entire .gitignore is preemptive anyway

.travis.yml
```diff
@@ -28,14 +28,20 @@ matrix:
...
+        - pip install tox
       script: tox -e linters
     - python: "3.6"
-      env: DIST=1
-      script: make dist
-      after_success:
-        - npm install -g @zeus-ci/cli
-        - zeus upload -t "application/zip+wheel" dist/*
+      name: Distribution packages
+      install: false
```
tonyo: A bit of optimization: I don't think we need Rust and tox for this job.

untitaker: :+1:

tonyo:
@untitaker Ready for review 📜

PR #80, opened 2018-09-25
fix(build): Merge coverage reports

untitaker:
Don't overwrite coverage reports with each other

### DIFF
--- a/scripts/runtox.sh
+++ b/scripts/runtox.sh
@@ -1,8 +1,13 @@
 #!/bin/sh
+
+# Usage: sh scripts/runtox.sh py3.7 <pytest-args>
+# Runs all environments with substring py3.7 and the given arguments for pytest
+set -xe
 if [ -z "$1" ]; then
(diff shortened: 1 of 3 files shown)


scripts/runtox.sh
```diff
@@ -1,8 +1,13 @@
...
+# Runs all environments with substring py3.7 and the given arguments for pytest
+set -xe
 if [ -z "$1" ]; then
     searchstring="$(echo py$TRAVIS_PYTHON_VERSION | sed -e 's/pypypy/pypy/g' -e 's/-dev//g')"
 else
     searchstring="$1"
 fi
 
-exec tox -e $(tox -l | grep $searchstring | tr '\n' ',')
+coverage erase
```
markstory: Do you need to erase coverage here and in travis.yml?

untitaker: Good catch, fixed. Got confused in my own chaos

PR #82, opened 2018-09-25
ci: Add "registry" target to Craft config

tonyo:
@untitaker Before the next release, please check that your `craft` version is at least `0.5.1`.


untitaker:
Thanks, will backfill the missing versions from the registry asap

PR #94, opened 2018-09-29
feat: Django sql queries

untitaker:


untitaker:
@mitsuhiko mostly copied from the old raven SDK, except that all sql breadcrumbs have proper annotations if parameters get trimmed

PR #101, opened 2018-10-02
ref: Integration options now live on the client

### LINKED ISSUE #100: Possible double reporting in 0.3.7
We have a simple test that verifies the Sentry SDK is capturing Flask errors because we had some configuration/PyPy issues in the past:

https://github.com/getsentry/snuba/blob/fd44ec8880240575ee2cb05bfc87e4cfd1d0e853/tests/test_api.py#L928-L936

As of this writing we don't pin the `sentry-sdk` version:

https://github.com/getsentry/snuba/blob/fd44ec8880240575ee2cb05bfc87e4cfd1d0e853/setup.py#L80

The test started failing because it seems that as of `0.3.7` (I tried `0.3.6` and wasn't able to reproduce the issue) the `transport` is being called twice for the same exception, causing the number of reported issues to be `2` instead of `1`.

Maybe this is a false alarm but I think our integration is relatively simple so this may be signs of a larger issue?

untitaker:
fix #100

### DIFF
--- a/sentry_sdk/integrations/atexit.py
+++ b/sentry_sdk/integrations/atexit.py
@@ -32,15 +32,12 @@ class AtexitIntegration(Integration):
             callback = default_shutdown_callback
         self.callback = callback
 
-    @classmethod
-    def install(cls):
+    @staticmethod
+    def install():
         @atexit.register
         def _shutdown():
-            if cls.main is None:
-                return
-
-            main_client = Hub.main.client
             logger.debug("atexit: got shutdown signal")
-            if main_client is not None:
+            atch = AtexitIntegration.attachment_from_hub(Hub.main)
+            if atch is not None:
                 logger.debug("atexit: shutting down client")
-                main_client.close(shutdown_callback=cls.main.callback)
+                atch.client.close(shutdown_callback=atch.integration.callback)
--- a/sentry_sdk/integrations/atexit.py
+++ b/sentry_sdk/integrations/atexit.py
@@ -32,11 +32,15 @@ class AtexitIntegration(Integration):
             callback = default_shutdown_callback
         self.callback = callback
 
-    def install(self):
+    @classmethod
+    def install(cls):
         @atexit.register
         def _shutdown():
+            if cls.main is None:
+                return
+
             main_client = Hub.main.client
             logger.debug("atexit: got shutdown signal")
             if main_client is not None:
                 logger.debug("atexit: shutting down client")
-                main_client.close(shutdown_callback=self.callback)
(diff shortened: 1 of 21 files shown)


sentry_sdk/integrations/excepthook.py
```diff
@@ -19,14 +20,15 @@ def install(self):
...
-                with_locals=hub.client.options["with_locals"],
-                mechanism={"type": "excepthook", "handled": False},
-            )
-            hub.capture_event(event, hint=hint)
+        if ExcepthookIntegration.current is not None:
+            with capture_internal_exceptions():
+                hub = Hub.current
+                event, hint = event_from_exception(
+                    (exctype, value, traceback),
+                    with_locals=hub.client.options["with_locals"],
```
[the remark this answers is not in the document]
mitsuhiko: I'm mostly concerned because the thing is nullable and every time I read such code I'm not sure if the programmer validated that it's not none here. Rust broke me to look at nullables in a different way now :) Performance i'm not that much concerned.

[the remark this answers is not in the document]
untitaker: Ok, I know what you mean but this instance of nullability in particular should be caught by tests. I don't think adding another layer of indirection is really going to help.

sentry_sdk/scope.py
```diff
@@ -152,7 +160,8 @@ def _drop(event, cause, ty):
                     return _drop(event, processor, "error processor")
                 event = new_event
 
-        for processor in self._event_processors:
+        for processor in chain(self._event_processors, global_event_processors):
```
HazAT: Shouldn't it be `chain(global_event_processors, self._event_processors)`?

untitaker: I imagine that every event processor looks like this:

```
if "key" not in event:
    event["key"] = "bar"
```

and not like this:

```
event["key"] = "bar"
```

In the first example, the first event processor sets `"key"` and every other event processor skips their logic. So the *first processor wins*.

In the second example the *last processor wins*, as its the only one whose value does not get overridden.

The reason is that scopes are applied to the event before event processors are, and data set on the scope should win over data set by integrations.

I wonder if we should apply event processors to the event before we apply scope data.

sentry_sdk/integrations/atexit.py
```diff
@@ -32,11 +32,12 @@ def __init__(self, callback=None):
             callback = default_shutdown_callback
         self.callback = callback
 
-    def install(self):
+    @staticmethod
+    def install():
```
untitaker: @mitsuhiko why is this a staticmethod?

PR #115, opened 2018-10-09
feat: Add debug log support in client init

mitsuhiko:


untitaker:
Ok, then why not when the client is bound to a hub? Integrations are ineffective anyway if the client isn't bound to a hub.

mitsuhiko:
What's the benefit of doing that? I just worry this is going to make things worse for no obvious benefit.

untitaker:
I'm just suggesting it as alternative to this one. I don't see drawbacks for either.

untitaker:
@mitsuhiko any opinion on my proposal? I don't really care anymore

mitsuhiko:
@untitaker i am considering it. i think debug logging is a bit stupid right now for more than one reason so it would be good to change it. eg: debug messages from the transport are suppressed.

untitaker:
I don't think we can make a single API change that fixes both cases in case that's what you're aiming for.

mitsuhiko:
@untitaker no, but when making the logger explicit (eg: per hub) this PR becomes pointless.

untitaker:
I'm not a big fan of this idea because it's incompatible with how we use
logging. dictConfig in sentry (where we explicitly install a sentry handler
on sentry_sdk.errors

untitaker:
@mitsuhiko I'm actually fine with this as-is. Do you have any concerns?

mitsuhiko:
WFM. We can still clean it up later. It's just that it does not solve the issue that transports don't log.

PR #119, opened 2018-10-10
fix: Make the main hub the current hub for main thread

mitsuhiko:


untitaker:
What if `sentry_sdk` is imported from a non-main thread?

PR #123, opened 2018-10-12
fix: Fix concurrency issues in sanic

### LINKED ISSUE #120: asyncio concurrency issues
@mitsuhiko  discovered that passing around hubs does not work at all and we likely leak state between requests

untitaker:
Fix #120

### DIFF
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -36,17 +36,19 @@ class SanicIntegration(Integration):
         async def sentry_handle_request(self, request, *args, **kwargs):
             hub = Hub.current
             if hub.get_integration(SanicIntegration) is None:
-                response = old_handle_request(self, request, *args, **kwargs)
-            else:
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/sanic.py
```diff
@@ -36,17 +36,19 @@ def setup_once():
...
             if hub.get_integration(SanicIntegration) is None:
-                response = old_handle_request(self, request, *args, **kwargs)
-            else:
-                weak_request = weakref.ref(request)
+                return old_handle_request(self, request, *args, **kwargs)
+
+            weak_request = weakref.ref(request)
 
-                with hub.push_scope() as scope:
+            with Hub(hub) as hub:
```
untitaker: I couldn't think of a testcase that would show `copy_context` is necessary.

untitaker: Ah, I assume sanic already has one context per request

PR #127, opened 2018-10-18
fix: Do not capture handled exceptions in sanic

abn:
Ignore exceptions with error handlers defined by the sanic application.

Proposing this change in behavior when handling exceptions thrown by a Sanic application as the current implementation triggers issue events for "known" and gracefully handled exceptions and does not allow these to be suppressed at the client side. An example of this, is a case where a client request to a websocket route causes an `InvalidHandshake` exception which in turn, is handled by sanic to raise an `InvalidUsage` exception. These exceptions, if a handler is registered explicitly in the application, should not create an issue event as this can lead to a large number of events generated for a heavily used websocket endpoint. Additionally, in any such scenarios an event can be captured by either explicitly using a `capture_message` or initializing the  logging integration and logging an error.

The proposed change retains capturing any errors that may be encountered during the error handler execution.

### DIFF
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -68,10 +68,10 @@ class SanicIntegration(Integration):
         old_error_handler_lookup = ErrorHandler.lookup
 
         def sentry_error_handler_lookup(self, exception):
-            _capture_exception(exception)
             old_error_handler = old_error_handler_lookup(self, exception)
 
             if old_error_handler is None:
+                _capture_exception(exception)
                 return None
 
             if Hub.current.get_integration(SanicIntegration) is None:
--- a/tests/integrations/sanic/test_sanic.py
+++ b/tests/integrations/sanic/test_sanic.py
@@ -7,7 +7,7 @@ from sentry_sdk import capture_message, configure_scope
 from sentry_sdk.integrations.sanic import SanicIntegration
 
 from sanic import Sanic, request, response
-
+from sanic.exceptions import InvalidUsage
 
 @pytest.fixture
 def app():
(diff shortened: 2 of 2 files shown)


untitaker:
We intentionally did it this way across all framework integrations because an errorhandler for a 500 Internal Server error would suppress all errors.

I would recommend configuring the SDK separately for ignored errors (using [`before_send`](https://docs.sentry.io/learn/filtering/?platform=python#before-send)) or just muting them in the UI if your plan can afford it

abn:
@untitaker I can understand your decision to handle it this way. But, it does feel a bit counter intuitive to consider an error to be not handled, when it has explicitly been done so in the application. 

In the case of sanic, when an error handler is defined for `SanicException`, I would expect all child exceptions to be handled and any exception that does not derive from `SanicException` to be reported. Which as far as I can tell, does not change with this fix. Verified that with this case (assuming this represents the case you mentioned).

```py
def test_error_in_errorhandler_500(sentry_init, app, capture_events):
    sentry_init(integrations=[SanicIntegration()])
    events = capture_events()

    @app.route("/error")
    def myerror(request):
        raise ValueError("oh no")

    @app.exception(SanicException)
    def myhandler(request, exception):
        return response.json({}, 500)

    request, response = app.test_client.get("/error")
    assert response.status == 500

    assert len(events) == 1

    exception, = events[0]["exception"]["values"]
    assert exception["type"] == "ValueError"
    assert any(
        frame["filename"].endswith("test_sanic.py")
        for frame in exception["stacktrace"]["frames"]
    )
```
Could this behavior be something that can be configured using a flag when initializing the integration? Would appreciate your thoughts on this.

In any case, we might mute for now and use the `before_send` mechanism as clients hammering endpoints with either bad request methods and/or bad requests is causing a fair amount of noise (~17k events overnight).

untitaker:
If I want to have a custom error response for all kinds of 500 Internal Server Errors, I would register a route `@app.exception(Exception)`. I assume most people have a route of this kind. Such a route would swallow up all errors, such that nothing gets sent to sentry.

abn:
@untitaker for our current issue I think your solution in b50b0b3 would do the trick. Thanks for that one.

PR #145, opened 2018-10-24
changed logger.warn to logger.warning to remove DeprecationWarning

petergaultney:
The call to logger.warn causes `DeprecationWarning` when not running in an AWS environment (e.g. local testing environments). `logger.warning` has always been the preferred and documented logger method, and in Python 3.3 and above, `logger.warn` is deprecated.


untitaker:
Thanks!

PR #156, opened 2018-10-28
fix: Fix confusing/buggy push_scope behavior with callbacks

untitaker:


HazAT:
lol, two approves are better than one

Issue #159, opened 2018-10-30
exiting `with push_scope()` fails if new client is bound


FelixLC:
Similarly to https://github.com/getsentry/sentry-python/issues/147, we started to get `popped wrong scope` assertion errors in a Falcon app sending jobs to `redis-queue` after updating to sentry-sdk from raven

Our `Python 3.5.2` stack:

```
rq==0.12.0
redis==2.10.6
sentry-sdk==0.5.2
```

The worker is initialized like this:

```python
    listen = [l.value for l in constants.QueueList]
    conn = redis.from_url(os.environ.get(
        'REDIS_WORKER', 'redis://redis_worker:6379'))

    sentry_sdk.init(sentry_key,
                    environment=environment,
                    release=release,
                    integrations=[RqIntegration()])
    with Connection(conn):
        worker = Worker(list(map(Queue, listen)))
        worker.work(logging_level='WARNING')
```

If it's the same kind of issue, would a change from:

```python
            with hub.push_scope() as scope:
                scope.add_event_processor(_make_event_processor(weakref.ref(job)))
                rv = old_perform_job(self, job, *args, **kwargs)
```

to

```python
            with Hub(hub) as hub:
                with hub.configure_scope() as scope:
                    scope.add_event_processor(_make_event_processor(weakref.ref(job)))
                    rv = old_perform_job(self, job, *args, **kwargs)
```

in https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/rq.py#L28 do the job ?

If not, I'd be happy to provide more info

untitaker:
Hi, are you using the regular worker class? If yes I don't understand how this can happen conceptually, as each job runs in its own process as far as I understand. Could you give me example code that defines a job and executes it?

untitaker:
Hi, are you using the regular worker class? If yes I don't understand how this can happen conceptually, as each job runs in its own process as far as I understand. Could you give me example code that defines a job and executes it?

FelixLC:
@untitaker I am using the standard worker, yes:

`from rq import Connection, Queue, Worker`

And defining a job with the `job` decorator:

```python
from rq.decorators import job

@job(constants.CONTEXT,
     connection=redis.from_url(constants.REDIS_WORKER_URL))
def patch_page_count_content_string(meta: CommunicationRequestMeta):
    '''Perform actual patch on API'''
    page_count_patch = meta.add_page_count_extension()
    content_string_patch = meta.add_content_string()

[...]

PR #160, opened 2018-10-30
Fix: Fix async jobs in RQ

FelixLC:
Jobs performing async calls (SQL queries or HTTP calls) got `popped wrong scope assertion errors` after dealing with their tasks.
See issue https://github.com/getsentry/sentry-python/issues/159


untitaker:
closing because of open ending of #159

PR #165, opened 2018-11-09
Fix collections import DeprecationWarning

BeyondEvil:
Import ABCs from `collections` is deprecated and will stop working in python 3.8.
https://github.com/python/cpython/blob/master/Lib/collections/__init__.py#L49

Starting with python 3.3 any imported ABCs should come from `collections.abc`
https://docs.python.org/3/library/collections.abc.html

The warning looks like this:
```
/lib/python3.7/site-packages/sentry_sdk/utils.py:8: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Mapping, Sequence
```

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -5,7 +5,6 @@ import logging
 
 from contextlib import contextmanager
 from datetime import datetime
-from collections import Mapping, Sequence
 
 from sentry_sdk._compat import (
     urlparse,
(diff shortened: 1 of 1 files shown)


BeyondEvil:
I fixed the linting error (blacks trailing comma... *sigh*)

But the other 4 failures (all the same `The command "sh scripts/download-semaphore.sh" failed and exited with 1 during .`) I'm not sure I have any control over. 🤷‍♂️

Issue #169, opened 2018-11-14
Django integration can't handle psycopg2's SQL string composition


bmdavi3:
psycopg2 2.7 introduces the psycopg2.sql module, which "contains objects and functions useful to generate SQL dynamically, in a convenient and safe way."

http://initd.org/psycopg/docs/sql.html

Unfortunately, format_sql() doesn't know how to work with a 'Composed' object.  So when we try to make use of this in a Django application, we get tracebacks from the sentry_sdk.

Here's a simple management command that illustrates the problem.

```
from django.core.management.base import BaseCommand
from django.db import connection
from psycopg2 import sql


class Command(BaseCommand):
    def handle(self, *args, **options):
        cursor = connection.cursor()

        cursor.execute(sql.SQL("""
            SELECT
                %(my_param)s
            """), {
            'my_param': 10,
        })
```

Running it produces the following error:

```
bdavis@257db183b518:/ph$ python manage.py example_error
Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 316, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 353, in execute
    output = self.handle(*args, **options)
  File "/ph/ph/phdb/management/commands/example_error.py", line 14, in handle
    'my_param': 10,
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 100, in execute
    return super().execute(sql, params)
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/integrations/django/__init__.py", line 257, in execute
    record_sql(sql, params)
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/integrations/django/__init__.py", line 224, in record_sql
    real_sql, real_params = format_sql(sql, params)
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/integrations/django/__init__.py", line 206, in format_sql
    sql = sql % conv
[...]

PR #170, opened 2018-11-15
Handle psycopg2's string composition

bmdavi3:
Something like this would probably work for #169

### DIFF
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -8,6 +8,19 @@ from django import VERSION as DJANGO_VERSION
 from django.core import signals
 
 try:
+    import psycopg2.sql
+
+    def sql_to_string(sql):
+        if isinstance(sql, psycopg2.sql.SQL):
+            return sql.string
+        return sql
+except ImportError:
+    def sql_to_string(sql):
+        return sql
+
+import psycopg2.sql
+
+try:
     from django.urls import resolve
 except ImportError:
     from django.core.urlresolvers import resolve
@@ -203,6 +216,7 @@ def format_sql(sql, params):
         # convert sql with named parameters to sql with unnamed parameters
         conv = _FormatConverter(params)
         if params:
+            sql = sql_to_string(sql)
             sql = sql % conv
             params = conv.params
         else:


sentry_sdk/integrations/django/__init__.py
```diff
@@ -7,6 +7,19 @@
...
+
+    def sql_to_string(sql):
+        if isinstance(sql, psycopg2.sql.SQL):
+            return sql.string
+        return sql
+except ImportError:
+    def sql_to_string(sql):
+        return sql
+
+import psycopg2.sql
```
untitaker: I think this line is not intentional?

untitaker:
apart from one comment lgtm! I am not sure how to test this though. I'd like to

sentry_sdk/integrations/django/__init__.py
bmdavi3: Ha.  Correct, will update.

sentry_sdk/integrations/django/__init__.py
```diff
@@ -7,6 +7,19 @@
...
+try:
+    import psycopg2.sql
+
+    def sql_to_string(sql):
+        if isinstance(sql, psycopg2.sql.SQL):
+            return sql.string
+        return sql
+except ImportError:
+    def sql_to_string(sql):
+        return sql
```
scottsexton: I'd prefer the full try/except/else blocks so there's less in the try:
```
try:
    import psycopg2.sql
except ImportError:
    def sql_to_string(sql):
        return sql
else:
    # Handle psycopg2 sql object strings (or some such comment)
    def sql_to_string(sql):
        if isinstance(sql, psycopg2.sql.SQL):
            return sql.string
        return sql
```

untitaker: it's fine, just a function declaration whose errors are caught

bmdavi3:
Let me see if I can come up with a test to exercise this.

untitaker:
@bmdavi3 can you run `black` over that file, I think it's misformatted

bmdavi3:
I'm not sure if they are what you had in mind, but I did add a couple tests in the last commit.  They're not super laser focused, in that they don't call format_sql() directly, or the try / except block itself even, but they do end up hitting the call to sql_to_string(sql), and would fail without the changes introduced in this p.r.

untitaker:
Yeah, just noticed... they break pypy. I'll figure out a fix. Thanks!

PR #176, opened 2018-11-21
Add the stack option back to the Logging integration

jrideout:
In Raven, you could do this:

```
logger.warning('things happened', extra={'stack': True})
```

And the stack would be collected. This adds back that feature.

### DIFF
--- a/sentry_sdk/integrations/logging.py
+++ b/sentry_sdk/integrations/logging.py
@@ -7,6 +7,7 @@ from sentry_sdk.hub import Hub
 from sentry_sdk.utils import (
     to_string,
     event_from_exception,
+    current_stacktrace,
     capture_internal_exceptions,
 )
 from sentry_sdk.integrations import Integration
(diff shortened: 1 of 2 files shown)


untitaker:
@jrideout thanks! I believe we want to somehow expose this on the scope instead of having special options on the logger though. That approach did not scale well in Raven.

untitaker:
@jrideout we have these scopes to override default values:

```
with push_scope() as scope:
    scope.fingerprint = ['foobar']

    capture_exception()  # custom fingerprint

capture_exception()  # no custom fingerprint
```

I wonder whether we can use scopes here too:

```
with push_scope() as scope:
    scope.attach_stacktrace = True

    logger.error(...)


```

I realize this is not optimal but I want to get rid of the weird arguments on logging calls

jrideout:
What about using `exc_info=True` for the non-exception use case as well? That would give a consistent user experience.

untitaker:
I like that idea, not sure if we have this info. @mitsuhiko what do you think?

untitaker:
Let's keep this open, I will check it out tomorrow

jrideout:
oops, I don't know how I closed this, I didn't mean too!

untitaker:
Thanks!

PR #202, opened 2018-12-12
fix: Dont crash on composed psycopg2 queries

### LINKED ISSUE #201: Django integration can't handle psycopg2's SQL string `.format`
Relates to #169

psycopg2 2.7 introduces the psycopg2.sql module, which "contains objects and functions useful to generate SQL dynamically, in a convenient and safe way."

http://initd.org/psycopg/docs/sql.html

Unfortunately, format_sql() doesn't know how to work with a 'Composed' object. So when we try to make use of this in a Django application, we get tracebacks from the sentry_sdk.

Here's a simple management command that illustrates the problem.

```
from django.core.management.base import BaseCommand
from django.db import connection
from psycopg2 import sql


class Command(BaseCommand):
    def handle(self, *args, **options):
        cursor = connection.cursor()

        cursor.execute('drop table if exists my_test_table')
        cursor.execute('create table my_test_table (id int)')
        cursor.execute(sql.SQL("""
            SELECT
                1
            FROM
                {my_table} 
            """).format(my_table=sql.Identifier('my_test_table')), {
            'foo': 'foo', #This is needed to trigger the bug.  In our prod code we use this to format into our 
            # query qith %(foo)s syntax, omitted here for clarity 
        })
```

Running it with sentry_sdk's DjangoIntegration() enabled produces the following error:

```
$ ./manage.py test-sentry
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 316, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 353, in execute
    output = self.handle(*args, **options)
  File "/ph/ph/app/management/commands/test-sentry.py", line 18, in handle
    'foo': 'foo',
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 100, in execute
    return super().execute(sql, params)
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/integrations/django/__init__.py", line 273, in execute
    record_sql(sql, params)
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/integrations/django/__init__.py", line 240, in record_sql
    real_sql, real_params = format_sql(sql, params)
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/integrations/django/__init__.py", line 222, in format_sql
    sql = sql % conv
TypeError: unsupported operand type(s) for %: 'Composed' and '_FormatConverter'
$
```

untitaker:
Fix #201

### DIFF
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -13,6 +13,8 @@ try:
     def sql_to_string(sql):
         if isinstance(sql, psycopg2.sql.SQL):
             return sql.string
+        elif isinstance(sql, psycopg2.sql.Composed):
+            return "".join(x.string for x in sql)
         return sql
 
 
@@ -239,13 +241,14 @@ def record_sql(sql, params):
     hub = Hub.current
     if hub.get_integration(DjangoIntegration) is None:
         return
-    real_sql, real_params = format_sql(sql, params)
+    with capture_internal_exceptions():
+        real_sql, real_params = format_sql(sql, params)
 
-    if real_params:
-        try:
-            real_sql = format_and_strip(real_sql, real_params)
-        except Exception:
-            pass
+        if real_params:
+            try:
+                real_sql = format_and_strip(real_sql, real_params)
+            except Exception:
+                pass
 
     hub.add_breadcrumb(message=real_sql, category="query")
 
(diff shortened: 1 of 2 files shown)


untitaker:
#203 is better

PR #203, opened 2018-12-12
Use cursor.mogrify when available 

mlennon-lumere:
Instead of trying to reimplement sql formatting, use cursor.mogrify when it is available.  This is implemented by psycopg2 precisely in order to get the SQL as it would be sent to the server via `cursor.execute`

http://initd.org/psycopg/docs/cursor.html#cursor.mogrify
Fix for #201


untitaker:
@mlennon-lumere I pushed what I have right now but I probably won't get around to it this week. As stated in the linked issue the problem is that none of this is properly tested.

mlennon-lumere:
I think there could be tests added here.  But the reason I'd prefer this approach instead of what's in #202 is that with the approach in #202, you'd have to have a case to handle each and every thing that psycopg2 could throw at you.

For example, `sql.Placeholder` instances, or `sql.Literal` instances, or other classes that psycopg2 hasn't even developed yet.

By standing on their shoulders with `mogrify`, `sentry-python` should be able to get their updates and keep feature-parity for much less effort

untitaker:
Yes I closed #202 because I think this is the superior approach. It's still harder to test though. A test like "assert that our integration calls mogrify" without actually having a real psycopg cursor isn't good enough I think

mlennon-lumere:
Is this still being worked on?  Cheers

untitaker:
Sorry, we've been really busy with non-sdk related stuff recently. I'll try to make some room for this. As far as I remember I was unhappy with the tests.

mlennon-lumere:
We're hoping to get these changes upstream, so we can stop relying on our fork.  Do you need something else to get this done, or is simple patience the way to go?

Thanks!

untitaker:
I really want to merge this with decent tests only. Right now we don't have any other customer/user complaining about this so it's kind of low-priority

mlennon-lumere:
What sort of unit tests do you believe are missing?  I'm hoping to see this get in, our org needs it to be able to keep using sentry.

Looking at the tests output is difficult, as many of the test runs on travis failed because of hitting an API rate limit.  Can you guide as to what you'd like to see?

Thanks

untitaker:
@mlennon-lumere I merged master into this branch, I hope the errors are now clearer. We recently added a `CONTRIBUTING.md` which should make it easier for you to set up a local dev env.

This has been a long time ago, but I believe the issue here was that we configured a sqlite database while testing postgres specific stuff. This has already caused some issues in the past, but so far we've been able to get away with it even though the query crashed halfway through.

tests/integrations/django/test_basic.py
```diff
@@ -186,49 +186,80 @@ def test_sql_queries(sentry_init, capture_events):
...
             """SELECT count(*) FROM people_person WHERE foo = %(my_foo)s""",
             {"my_foo": 10},
         )
 
     capture_message("HI")
-
     event, = events
 
     crumb, = event["breadcrumbs"]
-    assert crumb["message"] == ("SELECT count(*) FROM people_person WHERE foo = 10")
```
untitaker: I don't think we want this change. Why does this happen?

mlennon-lumere: Yeah, I'm not sure why it's getting formatted funny in the message with the bytestring like that.  Fair enough that it's not what's desired

untitaker: I'll try to debug this.

mlennon-lumere:
I'm not sure what's going on with the celery tests breaking - thanks for having a look and fixing the bytestring issues, postgres config

untitaker:
Seems unrelated. Thanks for the PR!

mlennon-lumere:
Thanks for merging - when can we expect a release?  

Thanks again

untitaker:
I will make one tomorrow morning, i.e. 10 hours from now.

untitaker:
0.7.8 is out!

PR #206, opened 2018-12-14
Clear scopes for wsgi requests and celery tasks

dimrozakis:
This prevents breadcrumbs produced during process startup from being
attached to events that occur during wsgi requests and celery tasks handling.


untitaker:
I get what you are trying to do but this also means that an import-time `scope.set_tag("server_name", "foobar")` (modifying the top-level scope for the main hub and therefore all future threads) will just no longer work. I think we need to figure out something else there. Having an API for clearing just breadcrumbs might be an option, or just logging a breadcrumb when a transaction has started such that people know which ones are important. Anyway, I think it's necessary to collect ideas on this first.

Issue #208, opened 2018-12-14
do not recursively create events while processing events


mmerickel:
1. Configure LoggingIntegration and PyramidIntegration.
1. Put `log.error` statements in authentication policy's `authenticated_userid` function.
1. Trigger some unrelated error during a request.
1. Observe infinite recursion as the event processor in PyramidIntegration invokes `request.authenticated_userid` which triggers creation of new events, until the process crashes.

It seems that `capture_internal_exceptions` should also temporarily disable the creation of new events by the hub.

mmerickel:
I'm not sure exactly how to approach it right now. My initial impression is to add a flag to the scope and check for it in `client._should_capture`. I'd then modify `capture_internal_exceptions` to set that flag but turn it off before invoking the `debug_hub` so that `capture_internal_exceptions` can still catch / create events.

Another, possibly better, way to look at it is to have `client._prepare_event` manage a flag on the scope around the processors. This would prevent any events from being emitted while processing an event, but that might prevent `capture_internal_exceptions` from emitting errors it catches unless I work around that somehow.

untitaker:
I would attach this to capture_internal_exceptions.

PR #210, opened 2018-12-15
fix event recursion

mmerickel:
fixes https://github.com/getsentry/sentry-python/issues/208

I think this is the simplest and most comprehensive fix but let me know what you think.

I looked at fixing it directly in `capture_internal_exceptions` but it felt super awkward. It looked like this, with the import inside to avoid recursive imports between `utils.py` and `hub.py`. This PR does not suffer from that issue and will only avoid creating events while processing one.

```diff
diff --git a/sentry_sdk/utils.py b/sentry_sdk/utils.py
index 6388f96..fadd4bc 100644
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -39,12 +39,24 @@ def _get_debug_hub():
 
 @contextmanager
 def capture_internal_exceptions():
+    from .api import configure_scope
+
+    # avoid any code inside generating new events
+    with configure_scope() as scope:
+        old_scope = scope
+        old_capture = scope._should_capture
+        scope._should_capture = False
+
     try:
         yield
     except Exception:
         hub = _get_debug_hub()
         if hub is not None:
             hub._capture_internal_exception(sys.exc_info())
+    finally:
+        with configure_scope() as scope:
+            if scope is old_scope:
+                scope._should_capture = old_capture
```

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -156,6 +156,9 @@ class Client(object):
         return False
 
     def _should_capture(self, event, hint, scope=None):
+        if scope is not None and not scope._should_capture:
+            return False
+
         if (
             self.options["sample_rate"] < 1.0
             and random.random() >= self.options["sample_rate"]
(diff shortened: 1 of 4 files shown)


mmerickel:
Looks like I exposed a sentry_sdk bug in older Pyramid versions when send_default_pii is enabled. I'll fix that.

mmerickel:
There seems to be some unrelated issues in travis.

untitaker:
Thanks!

Issue #212, opened 2018-12-17
sdk do NOT support dsn with subpath


vincenthcui:
By some reason, we can't use a absolute domain for sentry, so we use subpath instead like `company.io/sentry/project_id`. But `sentry-python` think project_id as int, and think the dsn above is invalid.

Exception:

```python
sentry_sdk.utils.BadDsn: Invalid project in DSN (u'sentry/848')
```

untitaker:
Which version of sentry (server) are you using? I am told that we just don't support subpath installations anymore since at least 2017: https://github.com/getsentry/sentry/pull/5174

untitaker:
We're going to backport this anyway due to proxies running in front of sentry.

vincenthcui:
@untitaker Version 9.0.0

untitaker:
Will be fixed in the next release. Thanks for your patience!

mitsuhiko:
Reopening because we need to revisit this. Not supporting subpaths was an intentional change and is already not supported in a lot of environments.

mitsuhiko:
@vincenthcui how do you have sentry deployed that sentry is at a non root path? That should not work.

mitsuhiko:
@untitaker If we want to bring it back we need to do it properly. This is not documented behavior and it's not supported by relay or any other new SDK.

mitsuhiko:
This is the current spec on how a DSN is supposed to be parsed: https://docs.sentry.io/development/sdk-dev/overview/#parsing-the-dsn

wedamija:
This change breaks Sentry's usage of `Auth` and is causing a subset of tests to fail. I'll pin to an older version for now, but if you need to introduce a new value into `Auth`'s constructor could you make it optional?

(See https://github.com/getsentry/sentry/blob/da920dda23598ef3137b2b829cbce48570daa99c/src/sentry/utils/sdk.py#L136 for usage)

untitaker:
Alright let's just revert this then

wedamija:
Thanks @untitaker, could you ping me on the next pypi release? I'll unpin the sdk version once it's out

untitaker:
@wedamija 0.6.6 is released

PR #213, opened 2018-12-17
Dsn project_id may not be int

vincenthcui:
support sentry in sub path like `company.io/sentry/project_id`

Refer to sentry-javascript.

#212 

@getsentry/python


untitaker:
This will not fix the issue at all. Deploying your application with this will just send broken requests that are discarded by the server (at best).

Thanks for raising this issue though, it seems that the raven behavior slipped through the cracks when rewritng the SDK

PR #226, opened 2018-12-21
fix: No boxed testing

untitaker:


untitaker:
We now rely on boxing all over the place. Still need a solution for benchmarking, but this can't be it.

PR #229, opened 2019-01-03
add means to capture locals only once

bucko909:
`with_locals` can be very expensive to run, and we've found our software can get into trouble when exceptions with complex locals are triggered in a tight loop, however, the information contained in the locals is often extremely useful for debugging. This commit allows the application to specify a means of selecting if locals should be captured, with a sample method which captures only the first for a given traceback. The cost of hashing the stack is quite small compared to the cost of stringifying the locals. Clearly the code can leak quite badly if used in a project with code generation, but in general the set of tracebacks should be small.

Note this can be quite awkward as Sentry doesn't really have a means of searching for a capture with locals, though one can normally just hunt backwards for the first instance.

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -32,6 +32,39 @@ epoch = datetime(1970, 1, 1)
 logger = logging.getLogger("sentry_sdk.errors")
 
 
+class CapturePicker(object):
+    @staticmethod
+    def should_capture(tb):
+        raise NotImplementedError()
+
+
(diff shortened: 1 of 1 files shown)


bucko909:
Eventually I'll defeat the linting, but I don't understand the coverage errors (or the remaining test ones) as I've clearly not committed an extra 6% to the codebase.

untitaker:
I am closing this in favor of #247 which should allow you to skip `repr()` calls.

PR #233, opened 2019-01-10
Add option to specify client certificate

maresac:
See [urllib3 docs](https://urllib3.readthedocs.io/en/latest/advanced-usage.html#custom-ssl-certificates-and-client-certificates)

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -28,6 +28,7 @@ DEFAULT_OPTIONS = {
     "debug": False,
     "attach_stacktrace": False,
     "ca_certs": None,
+    "cert_file": None,
 }
 
 
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -18,7 +18,7 @@ except ImportError:
     from urllib import getproxies
 
 
-def _make_pool(parsed_dsn, http_proxy, https_proxy, ca_certs):
+def _make_pool(parsed_dsn, http_proxy, https_proxy, ca_certs, cert_file):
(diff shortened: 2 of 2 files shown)


maresac:
Is there a possibility to re-run the pipeline without exceeding the API rate limit?

untitaker:
Sorry about that CI failure, will fix.

Did you manage to do something similar in Raven? We're a bit hesitant to support this as it's not a popular usecase and would prefer if you subclassed the `HTTPTransport` class yourself to provide this functionality.

Right now that class is not very flexible but I'll prepare another PR for that

untitaker:
Could you check if this fits your usecase? https://github.com/getsentry/sentry-python/pull/235

ckoehn:
This will definitely work, but still sad that you don't want to support a feature which can be implemented by just adding 4 lines of code.

untitaker:
@ckoehn The reason I want to avoid having more options on the client is that we try to align those across SDKs. There are many options on `urllib3.PoolManager` that might be useful to some people (connection timeouts?), but adding individual support for each of them ends up being a lot of code.

Using a client certificate to authenticate the SDK against the server is very non-standard. The "auth" mechanism we intended for people to use was the key in the DSN. Let's say somebody added something to their nginx/apache cfg to check for HTTP Basic auth header. Would you expect me to merge the PR that adds corresponding username/password options to the SDK?

ckoehn:
Fair point. You didn't mention that you want to align the options across SDKs and I was just curious why you rejected such a simple change :wink:

PR #239, opened 2019-01-18
Handle iteration errors during frame traversal

### LINKED ISSUE #238: Event is dropped if Exception occurs during `object_to_json`
This may happen if Sequence is Django queryset that has incorrect field in `filter()` expressions.

leotrubach:
Fixes #238

Not sure if this is the best way to show it, i think it is better than showing just empty list. But maybe there could be some better way to show that exception occured during traversal.

There is more general problem that it is probably good to not iterate over queryset, so DB is not queried. Any ideas on how to do that?

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -311,6 +311,8 @@ def object_to_json(obj):
     def _walk(obj, depth):
         if depth < 4:
             if isinstance(obj, (list, tuple)):
+                # It is not safe to iterate over another sequence types as this may raise errors or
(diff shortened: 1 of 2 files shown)


untitaker:
I believe it would be appropriate to only traverse builtin collections (check for `isinstance(x, (tuple, list))`) and just stringify the rest like arbitrary objects

untitaker:
Cool, could you add an explanatory comment as well?

leotrubach:
@untitaker Do you mean comment to code or commit message?

untitaker:
in the code :) thanks!

untitaker:
Awesome thanks!

PR #240, opened 2019-01-19
Another place where errors may arise during frame traversal

leotrubach:


untitaker:
Thanks, could you add the same comment there?

untitaker:
Thanks!

PR #242, opened 2019-01-23
WIP port aiohttp transports from raven_aiohttp

tonal:
I do not understand how to correct close this transports

### DIFF
--- a/sentry_sdk/integrations/aiohttp_transports.py
+++ b/sentry_sdk/integrations/aiohttp_transports.py
@@ -1,0 +1,205 @@
+# -*- codong: utf-8 -*-
+from abc import abstractmethod
+import asyncio
+from datetime import datetime, timedelta
+import gzip
+import io
+import json
+
+from aiohttp import ClientSession
+from sentry_sdk import Transport, VERSION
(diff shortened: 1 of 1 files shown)


untitaker:
Why do you need the aiohttp transport?

untitaker:
@tonal interesting. Do you no longer experience issues with this transport? We could try to figure out how to set a different transport from integrations (we might need this for AWS Lambda anyway) and then use this.

Thanks for working on this! I still don't understand what's happening though. Why does asyncio freeze up when something spawns its own thread?

What are the speed differences you see? Does your transport send stuff in parallel? This is something we explicitly don't want, because you would get rate-limited anyway.

cancan101:
Any updates on this or an async transport? I am using starlette and would prefer to use non blocking / async transports.

untitaker:
@cancan101 as said the transport should be non-blocking because it runs on a dedicated OS thread. Of course there are better ways to achieve parallelism in async but this method works in both async/sync and we only spawn one extra thread.

untitaker:
I'm going to close this because it is WIP and inactive for quite a while now. I would probably be open to having one alternative transport if the speed is really so much better though. But before that I'd like to see it working.

PR #249, opened 2019-02-04
feat: Deep vars

untitaker:


untitaker:
Let's revisit when we decide to do this in more sdks

PR #261, opened 2019-02-12
Add missing await in AioHttpIntegration.

### LINKED ISSUE #260: Half installed AioHttpIntegration causes aiohttp to crash
If I call:
```python
sentry_sdk.integrations.setup_integrations(
        [sentry_sdk.integrations.aiohttp.AioHttpIntegration()])
```
after `sentry_sdk.init()` the `_handle` method of `aiohttp.web.Application` gets replaced but the integration does not get registered in the client. This causes the replaced `_handle` ro run into a codepath where there as a `await` missing. This gives an exception in every request:
```
ERROR:aiohttp.server:Unhandled exception                                                                                                                                                                                                                             
  File "/usr/local/lib/python3.7/site-packages/aiohttp/web_protocol.py", line 447, in start                                                                                                                                                                          
    await resp.prepare(request)                                                                                                                               
AttributeError: 'coroutine' object has no attribute 'prepare'                                                                                                                                                                                                        
/usr/local/lib/python3.7/site-packages/xxx/base.py:151: RuntimeWarning: coroutine 'Application._handle' was never awaited                              
  self._loop.run_forever()     
```

This will not get logged to sentry at all, because the `aiohttp.server` logger gets ignored by (half-)installing the integration (see #259).

janLo:
The replaced _handle method of aiohttp.web.Application is a coroutine
itself. So it has to await the call to the original handle otherwise its
never awaited.

This fixes #260

Signed-off-by: Jan Losinski <losinski@wh2.tu-dresden.de>

### DIFF
--- a/sentry_sdk/integrations/aiohttp.py
+++ b/sentry_sdk/integrations/aiohttp.py
@@ -32,7 +32,7 @@ class AioHttpIntegration(Integration):
             async def inner():
                 hub = Hub.current
                 if hub.get_integration(AioHttpIntegration) is None:
-                    return old_handle(self, request, *args, **kwargs)
+                    return await old_handle(self, request, *args, **kwargs)
 
                 weak_request = weakref.ref(request)
 
--- a/tests/integrations/aiohttp/test_aiohttp.py
+++ b/tests/integrations/aiohttp/test_aiohttp.py
@@ -53,3 +53,22 @@ async def test_403_not_captured(sentry_init, aiohttp_client, loop, capture_event
     assert resp.status == 403
 
     assert not events
+
+
+async def test_half_initialized(sentry_init, aiohttp_client, loop, capture_events):
+    sentry_init(integrations=[AioHttpIntegration()])
+    sentry_init()
+
+    async def hello(request):
+        return web.Response(text="hello")
+
+    app = web.Application()
+    app.router.add_get("/", hello)
+
+    events = capture_events()
+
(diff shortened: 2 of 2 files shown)


untitaker:
I admit I don't understand the async story in Python. I assumed directly returning a coroutine in an async function would be equivalent.

Could you write a test for this? You should be able to do this:

```
init(integrations=[AioHttpIntegration()])
init()
```

and copy the rest from an existing test.

janLo:
The problem is, that if you call a Coroutine (something with `async def`) you get a coroutine object back. To get the actual return (or exception or whatever is the result of the coroutine) you need to `await` it. So returning a coroutine from a coroutine (without the `await` inside) means, that you need to write `await` twice.

untitaker:
nvm, yes, brainfart. I think I was too much in Rust's async support where using a bare `return` means you would have to return an actual future.

janLo:
Btw, urllib3 and certifi seems to be missing in the test requirements

untitaker:
I assume you did this:

`pip install -r test-requirements.txt`

while I usually do this in addition:

`pip install -e .`

untitaker:
try this: `sh scripts/runtox.sh aiohttp`

There are a lot of "optional test requirements" because some things don't work in certain python versions.

janLo:
There is now a unittest that fails without the fix and works with it :)

untitaker:
Awesome, thanks!

Issue #268, opened 2019-02-19
SDK client silently ignores capturing if server responds with 413 (big payload)


amureki:
Greetings fellows,

we recently by accident discovered that we are not getting all exceptions in Sentry.

### Current versions
Python 3.6.7
Django 2.1.7
Celery 4.1.1
Sentry SDK 0.7.2

### Current behavior
If the exception will have a big context (I do not know the exact limits of it), Sentry server (sentry.io) responds with HTTP code 413.
The code in `sentry_sdk.transport.HttpTransport#_send_event` is raising `ValueError("Unexpected status code")` in this case. If you do not set `debug=True` during the Sentry SDK initialization, you would not know that and you would be completely unaware of having issues in the code.
This issue today took us some time to discover.

### Expected behaviour (IMHO)
I believe, Sentry SDK should try to do its best to send exceptions. 
In case of a too big payload, it maybe could retry the request without the full payload, but just an exception name with some note "Your context is too big, could not be shown" or something like this.
In the worst case, we should get this kind of events in logs even without debug turned off.

What do you think?

untitaker:
There's some internal discussion around this, and I personally believe a better fix is to loosen the hard limits on event payload size to a level where they're no longer a problem in practice (because for all other events we already aggressively trim the event data to make it acceptably small). This presents some challenges around scaling that we haven't figured out yet.

In case this never happens I still believe that the fix should be purely serverside.

amureki:
@untitaker hey Markus,

thanks for the update.
We did clean the event causing troubles, but I want to be sure about possible future cases, that is why I raised this ticket.

If you could update us whenever you will have some decisions on this one, it would be great. 👍

kobybum:
We are also losing errors because of this. Until this is fixed, this may cause serious production issues for the companies that depend on you for error management.

We will truncate our messages for the meantime, but consider that a server-side solution will have to decompress the gzip encoded message, and will cause serious server load.

You should consider a hotfix for this, before continuing the discussions of a higher level solution.

untitaker:
@kobybum if you tell me which values you are truncating we can start to do that in the SDK instead.

kobybum:
Thanks for the quick response!

I'm currently working on a solution. It's hard to pinpoint exactly what should be truncated - in our case, we pass around huge dictionaries, and have huge SQL queries.
The SQL queries are recorded as breadcrumbs, and the locals are sent as-is to the server.
[...]

PR #279, opened 2019-03-04
Stacktrace: Pass with_locals parameter.

kobybum:
Currently, `sentry_sdk` is ignoring the `with_locals` option when collecting stacktrace info.
This makes events collect local variables, which in our case creates huge events which are dropped by sentry.

This PR passes the client options to `current_stacktrace` to avoid the problem.

Related to #268

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -118,7 +118,7 @@ class Client(object):
             with capture_internal_exceptions():
                 event["threads"] = [
                     {
-                        "stacktrace": current_stacktrace(),
+                        "stacktrace": current_stacktrace(self.options["with_locals"]),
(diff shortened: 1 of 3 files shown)


untitaker:
Thanks, this is actually a bug in the SDK (independent of the discussion at #268)! Could you add a test for this in `tests/test_client.py`?

kobybum:
Wrote a test, unfortunately don't have time to setup a local dev environment to test, feel free to take the PR over

untitaker:
Ah, I overlooked that. I will, and sorry for the inconvenience!

kobybum:
Sure man, thanks for the quick replies!

untitaker:
Thanks!

PR #284, opened 2019-03-08
fix: Wsgi is not supposed to handle general data as JSON

shenek:
Note that this change will make the code which processes raw data in `extract_into_event` reachable
https://github.com/getsentry/sentry-python/blob/741332900b0d7761c15990096d61171e56de1c25/sentry_sdk/integrations/_wsgi_common.py#L47

### DIFF
--- a/sentry_sdk/integrations/_wsgi_common.py
+++ b/sentry_sdk/integrations/_wsgi_common.py
@@ -84,7 +84,10 @@ class RequestExtractor(object):
 
             return data
 
-        return self.json()
+        if self.is_json():
+            return self.json()
+
+        return None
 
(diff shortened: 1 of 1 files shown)


untitaker:
Could you explain what your issue was? I believe I am missing a lot of context here

shenek:
Hi, I read the code to hastily and I misunderstood it. I thought that function
```python
def parsed_body(self):
...
    return self.json()
```
is trying to parse JSON no matter which `Content-Type` it has. I haven't noticed that `Content-Type` checking is done within `json()` function. So that only requests with a proper `Content-Type` header are being parsed as JSON.

Sorry for bothering you.

PR #286, opened 2019-03-08
fix: Wsgi - valid JSONs ([], {}) are evaluated as raw data

shenek:
It should be evaluated as empty JSON object/array

Note that according to RFC 4627 top level elements can be empty objects or arrays


untitaker:
Thanks!

PR #288, opened 2019-03-10
feat: Add integration for parsing GNU-style backtrace dumps

untitaker:
Value proposition: ClickHouse sends us error messages with stacktrace information. Ideally we would parse this stacktrace information into actual stacktraces. Here is an integration for exactly that. The testcase is ripped straight from https://sentry.io/organizations/sentry/issues/856036064

ClickHouse calls [`backtrace_symbols`](https://linux.die.net/man/3/backtrace_symbols). It already returns a formatted string/list of strings. ClickHouse then parses this string and [replaces symbols with demangled versions](https://github.com/yandex/ClickHouse/blob/1e22b308d9cd974803920c90d968ca89ad36c84f/dbms/src/Common/StackTrace.cpp). I didn't bother with parsing addresses and offsets as I don't see much value in that information (not sure if it's used in grouping) and some of those return addresses (brackets at the end) seem wrong:

```
23. clickhouse-server() [0xa3c68cf]
24. /lib/x86_64-linux-gnu/libpthread.so.0(+0x8184) [0x7fdb108c5184]
25. /lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7fdb0fee003d]
```

I'm open to alternative naming suggestions. As far as I can tell this function has its origins in glibc. `LibExecInfoBacktraceIntegration` seemed bulky. Perhaps `GlibcBacktraceIntegration` to avoid conflict with GDB's output format, which seems entirely different? I'm generally unsure of `backtrace_symbols`'s origin. It seems to have started as a glibc thing but there are ports for BSD and Darwin. It isn't POSIX though.

When discussing the idea of this with @mitsuhiko there was the idea to have a single `NativeStacktraceIntegration` where we would add new kinds of stacktrace formats over time. However, in order to not break grouping between minor versions, we can't actually add new variants to that "super-integration" without having the user explicitly opt into each of them (just the same reason why we can't make this a default integration). Therefore I see no point in having one integration for multiple formats, and the naming problem persists.

cc @alex-hofsteede

### DIFF
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -141,19 +141,23 @@ class DjangoIntegration(Integration):
         @add_global_event_processor
         def process_django_templates(event, hint):
             # type: (Dict[str, Any], Dict[str, Any]) -> Dict[str, Any]
-            if hint.get("exc_info", None) is None:
+            exc_info = hint.get("exc_info", None)
+
+            if exc_info is None:
                 return event
 
-            if "exception" not in event:
+            exception = event.get("exception", None)
+
+            if exception is None:
                 return event
 
-            exception = event["exception"]
+            values = exception.get("values", None)
 
-            if "values" not in exception:
+            if values is None:
                 return event
 
             for exception, (_, exc_value, _) in zip(
-                exception["values"], walk_exception_chain(hint["exc_info"])
+                values, walk_exception_chain(exc_info)
             ):
                 frame = get_template_frame_from_exception(exc_value)
                 if frame is not None:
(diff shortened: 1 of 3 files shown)


untitaker:
I wonder if we can share this code across SDKs somehow. There's nothing in this integration that couldn't apply to nodejs or C#. I considered writing it in Haxe.

untitaker:
@bruno-garcia adding you because of what we discussed the last retros about SDK work.

bruno-garcia:
@untitaker do you have a screenshot of what it looks like in Sentry the mix callstack? If you do, could you paste in the PR, please?

untitaker:
Screenshot 2019-03-18 at 11 26 02

https://sentry.io/share/issue/ac4f29d3b10141a291d83fc1db35a505/

untitaker:
@mitsuhiko reminded me to set platform to native, wil fix

bruno-garcia:
@untitaker we ahve new stacktrace rendering now? :)

untitaker:
@bruno-garcia I updated the screenshot

untitaker:
One shitty aspect of this right now is that every frame is annotated as inline frame because there are no instruction addrs

jan-auer:
@untitaker We can change that quite easily. Should actually.

PR #292, opened 2019-03-15
ref: Do not use test classes for tornado tests

untitaker:
@nickbaum you seem to know what you're doing. Can you recommend something to test Tornado apps with? This solution I have right now is working fine but it seems hard to maintain

### DIFF
--- a/scripts/runtox.sh
+++ b/scripts/runtox.sh
@@ -14,4 +14,4 @@ else
     searchstring="$1"
 fi
 
-exec $TOXPATH -p auto -e $(tox -l | grep "$searchstring" | tr '\n' ',') -- "${@:2}"
+exec $TOXPATH -p auto -e $($TOXPATH -l | grep "$searchstring" | tr '\n' ',') -- "${@:2}"
--- a/tests/integrations/tornado/test_tornado.py
+++ b/tests/integrations/tornado/test_tornado.py
@@ -8,6 +8,31 @@ from tornado.web import RequestHandler, Application, HTTPError
 from tornado.testing import AsyncHTTPTestCase
 
 
+@pytest.fixture
+def tornado_testcase(request):
+    # Take the unittest class provided by tornado and manually call its setUp
+    # and tearDown.
+    #
+    # The pytest plugins for tornado seem too complicated to use, as they for
+    # some reason assume I want to write my tests in async code.
+    def inner(app):
+        class TestBogus(AsyncHTTPTestCase):
+            def get_app(self):
+                return app
+
+            def bogustest(self):
+                # We need to pass a valid test method name to the ctor, so this
+                # is the method. It does nothing.
+                pass
+
+        self = TestBogus("bogustest")
+        self.setUp()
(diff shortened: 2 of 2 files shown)


nickbaum:
Haha, I know barely enough to be dangerous, but I don't think the AsyncHTTPTestCase requires any async code, it just supports it. The following code seems to work to trigger an exception in the test case, perhaps you see how you'd integrate the Sentry testing into that?

```
import json
import unittest
import tornado.web
import tornado.testing


class TestHandler(tornado.web.RequestHandler):
    def get(self):
        raise Exception("GET")

    def post(self):
        raise Exception("POST")


def make_app():
    return tornado.web.Application([
        (r"/test", TestHandler),
    ])


class TestCase(tornado.testing.AsyncHTTPTestCase):
    def get_app(self):
        return make_app()

    def test_homepage(self):
        response = self.fetch('/test')
        self.assertEqual(response.code, 500)
        response = self.fetch('/test', method='POST', body=json.dumps({"arg1": "value1"}))
        self.assertEqual(response.code, 500)


if __name__ == '__main__':
    unittest.main()
```

@bdarnell (Tornado's maintainer) might have some pointers. Ben, Markus is looking for the best way to maintain Sentry's integration with Tornado, if you have any advice.

untitaker:
I'm interested in using pytest fixtures for the tests as opposed to unittest classes, because injecting my existing pytest fixtures into AsyncHTTPTestCase-based tests is painful.

Anyway, what i am doing now works well enough for me, just wondering if there's a better way.

>but I don't think the AsyncHTTPTestCase requires any async code

I was talking about pytest-tornado and pytest-asyncio when I said that. I have nothing against async code even, but I just didn't get anything other than what I have right now to run.

bdarnell:
In general I recommend either fully embracing async (in this case using pytest-tornado or pytest-asyncio. It's not clear what obstacles you ran into there), or keeping it at arms length by running the tornado server on a separate thread (as, I assume, you are doing with the synchronous frameworks). What you have here looks like an awkward middle ground. 

I think it's also possible to make a more targeted test fixture for your use case that doesn't rely on AsyncHTTPTestCase. Something like this (I don't have much experience with pytest and I'm not sure how its fixtures work, but it's also not clear that they're necessary here):

```
def tornado_fetch(app, path):
    async def inner():
        # might need a try/finally for `sock.close()`
        sock, port = tornado.testing.bind_unused_port()
        server = tornado.httpserver.HTTPServer(app)
        server.add_socket(sock)
        client = tornado.httpclient.AsyncHTTPClient()
        return await client.fetch(urllib.parse.urljoin('http://127.0.0.1:%d' % port, path))
    with contextlib.closing(tornado.ioloop.IOLoop()) as loop:
        return loop.run_sync(inner)
```

`tornado_fetch(app, path)` should be equivalent to your `tornado_testcase(app).fetch(path)`.

nickbaum:
Thanks for the reply, @bdarnell!

Markus, I'm not at all familiar with these test frameworks, so I'm afraid I won't be much further help on this.

Issue #289, opened 2019-03-13
support aiohttp and sanic for python 3.6


aCLr:
Looks like aiohttp and sanic integrations python 3.6 can be supported, because of [backported contextvars library](https://pypi.org/project/contextvars/)

untitaker:
afaik the issue was that this contextvars backport doesn't work with asyncio from the stdlib at all (leaking context all over the place).

We have leak tests in the sanic test suite, so if the testsuite works with that backport this would be interesting to investigate again

aCLr:
Unfortunatelly, you are right :(
[Here](https://github.com/MagicStack/contextvars/issues/2) is an issue about asyncio support.

untitaker:
Just curious, is something holding you on Python 3.6? Distro packages?

aCLr:
Sure :)

PR #293, opened 2019-03-16
use contextvars backport for aiohttp and sanic support in py3.6

aCLr:
Ok, found a workaround for [this](https://github.com/getsentry/sentry-python/issues/289) issue. All tests are passed for py3.7 and py3.6.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ format: .venv
 .PHONY: format
 
 test: .venv
-	@$(VENV_PATH)/bin/tox -e py2.7,py3.7
+	@$(VENV_PATH)/bin/tox -e py2.7,py3.7,py3.6
 .PHONY: test
 
 test-all: .venv
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ format: .venv
 .PHONY: format
 
 test: .venv
-	@$(VENV_PATH)/bin/tox -e py2.7,py3.7,py3.6
+	@$(VENV_PATH)/bin/tox -e py2.7,py3.7
 .PHONY: test
 
 test-all: .venv
--- a/sentry_sdk/integrations/aiohttp.py
+++ b/sentry_sdk/integrations/aiohttp.py
@@ -9,7 +9,7 @@ from sentry_sdk.integrations._wsgi_common import _filter_headers
 from sentry_sdk.utils import (
     capture_internal_exceptions,
     event_from_exception,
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ format: .venv
 .PHONY: format
 
 test: .venv
-	@$(VENV_PATH)/bin/tox -e py2.7,py3.7
+	@$(VENV_PATH)/bin/tox -e py2.7,py3.7,py3.6
```
untitaker: Please undo this, this target is only meant to test a few versions and to be fast, not all of them.

sentry_sdk/utils.py
```diff
@@ -867,8 +867,12 @@ def realign_remark(remark):
         value=rv, metadata={"len": rv_original_length, "rem": rv_remarks}
     )
 
-
 try:
+    if not PY2 and sys.version_info < (3, 7):
+        try:
```
untitaker: Instead of checking `sys.modules` at some completely other place, can we export constants here that tell you whether context vars work with asyncio, and use them in integrations?

test-requirements.txt
```diff
@@ -5,3 +5,4 @@ tox==3.7.0
 Werkzeug==0.14.1
 pytest-localserver==0.4.1
 pytest-cov==2.6.0
+aiocontextvars==0.2.1
```
untitaker: Please move this to `tox.ini`, and just for sanic targets. You want to add a line to the tox sanic target here: https://github.com/getsentry/sentry-python/blob/c359d695e6ae06ed1569ee04ba85af81446425e0/tox.ini#L23

Makefile
aCLr: Hm, I've thought that you want to test it for both versions of python. `aiocontextvars` not used in py3.7.

untitaker: @aCLr right, but `make test` doesn't run the whole testsuite, `make test-all` does. The test matrix you want is in `tox.ini`.

aCLr: Oh, I see, ok

sentry_sdk/integrations/aiohttp.py
```diff
@@ -61,7 +65,7 @@ def setup_once():
 
                     return response
 
-            return await asyncio.create_task(inner())
```
untitaker: This is not a good idea. It will leak context between requests.

untitaker:
Thanks!

PR #298, opened 2019-03-20
fix: `RuntimeError: dictionary changed size during iteration` raised within extract_locals

shenek:
* affects only python3 because items() in python3 doesn't return list
* note that this situation is actually a simplified example how environ is treated in Bottle framework

### DIFF
--- a/tests/integrations/wsgi/test_wsgi.py
+++ b/tests/integrations/wsgi/test_wsgi.py
@@ -12,6 +12,26 @@ def crashing_app():
     return app
 
 
+@pytest.fixture
+def crashing_env_modifing_app():
+    class TooSmartClass(object):
+        def __init__(self, environ):
+            self.environ = environ
+
+        def __repr__(self):
+            if "my_representation" in self.environ:
+                return self.environ["my_representation"]
+
+            self.environ["my_representation"] = "<This is me>"
+            return self.environ["my_representation"]
+
+    def app(environ, start_response):
+        environ["tsc"] = TooSmartClass(environ)
+        1 / 0
+
+    return app
+
+
 def test_basic(sentry_init, crashing_app, capture_events):
     sentry_init(send_default_pii=True)
     app = SentryWsgiMiddleware(crashing_app)
@@ -30,3 +50,15 @@ def test_basic(sentry_init, crashing_app, capture_events):
         "query_string": "",
         "url": "http://localhost/",
     }
+
+
+def test_env_modifing_app(sentry_init, crashing_env_modifing_app, capture_events):
+    sentry_init(send_default_pii=True)
+    app = SentryWsgiMiddleware(crashing_env_modifing_app)
+    client = Client(app)
+    events = capture_events()
+
+    with pytest.raises(ZeroDivisionError):
+        client.get("/")
+
+    assert len(events) == 1  # only one exception is raised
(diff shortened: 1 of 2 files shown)


untitaker:
That's messed up. Can you link me to the code in bottle that changes the environ during `repr`?

tests/integrations/wsgi/test_wsgi.py
```diff
@@ -30,3 +51,15 @@ def test_basic(sentry_init, crashing_app, capture_events):
         "query_string": "",
         "url": "http://localhost/",
     }
+
+
+def test_env_modifing_(sentry_init, crashing_env_modifing_app, capture_events):
```
untitaker: Could you fix this name to something that doesn't have an underscore at the end?

shenek: Sure, sorry I missed that.

shenek: There supposed to be `_app`

shenek:
The code itself in bottle is not very 'visible'. And it took a quite amount of time to debug. But i can at least show you a part of my trace.

So lets start with `repr`: https://github.com/bottlepy/bottle/blob/0.12.16/bottle.py#L1386
Here you call `self.url` which calls `self.urlparts`.
see https://github.com/bottlepy/bottle/blob/0.12.16/bottle.py#L1252
The code itself still looks ok (no modifications to environ are done).
But there is this `DictProperty` decorator.
It creates a new element `bottle.request.urlparts` in environ when it is called for the first time.
see https://github.com/bottlepy/bottle/blob/0.12.16/bottle.py#L166

untitaker:
Amazing. I'll include this in 0.7.8. Thanks!

shenek:
You're welcomed.

untitaker:
FYI, with #368 this will soon spit out a placeholder instead of the value. Using `list(dict.items())` causes severe performance issues for some of our customers. Let me know if you have a better suggestion.

shenek:
> FYI, with #368 this will soon spit out a placeholder instead of the value. Using `list(dict.items())` causes severe performance issues for some of our customers. Let me know if you have a better suggestion.

I'm not sure, but perhaps something like this could help:
```diff
diff --git a/sentry_sdk/_compat.py b/sentry_sdk/_compat.py
index 871995e..5dbf158 100644
--- a/sentry_sdk/_compat.py
+++ b/sentry_sdk/_compat.py
@@ -1,14 +1,28 @@
 import sys
+import copy
 
 if False:
     from typing import Optional
     from typing import Tuple
     from typing import Any
     from typing import Type
+    from typing import Callable
 
 
 PY2 = sys.version_info[0] == 2
 
+
+def safe_iteritems_wrapper(unsafe_iter):
+    # type: (Callable[[dict], None]) -> Callable[[dict], None]
+    def wrapped(x):
+        try:
+            return unsafe_iter(x)
+        except RuntimeError:
+            return copy.copy(x).items()
+
+    return wrapped
+
+
 if PY2:
     import urlparse  # noqa
 
@@ -52,6 +66,9 @@ else:
         raise value
 
 
+iteritems = safe_iteritems_wrapper(iteritems)
+
+
 def with_metaclass(meta, *bases):
     class metaclass(type):
         def __new__(cls, name, this_bases, d):
```

untitaker:
Do you think an easy fix for this could be that we just serialize the event breadth-first?

shenek:
> Do you think an easy fix for this could be that we just serialize the event breadth-first?

I would doubt that. The problem is that you are modifying the dict which are you currently iterating.
```
              A
             /\  \
current -> A1 A2 (B1) <- inserted element
          / \
        AA1 AA2
```
I guess that it would only have failed sooner. I'll try to run some test for that later.

untitaker:
Nevermind, `list(dict.items())` is basically what I meant with breadth-first. But I figured out a way to restore the old behavior without impacting perf, all good.

PR #300, opened 2019-03-21
feat: Bottle integration

shenek:
After some struggle I was finally able to prepare this MR, which provides some basic integration with Bottle web framework.

I was mostly inspired by Flask and Pyramid integrations source code.
The code itself doesn't handle any session/login logic and it contains a few bottle specific stuff.

In a couple of days I plan to do some real life testing and I tend to prepare some kind of a tutorial.

Right now I will be grateful for any feedback or advice.


sentry_sdk/integrations/bottle.py
```diff
@@ -0,0 +1,165 @@
...
+
+def _make_request_event_processor(app, request, integration):
+    # type: (Bottle, LocalRequest, BottleIntegration) -> Callable
+    def inner(event, hint):
+        # type: (Dict[str, Any], Dict[str, Any]) -> Dict[str, Any]
+
+        # if the request is gone we are fine not logging the data from
+        # it.  This might happen if the processor is pushed away to
+        # another thread.
+        if request is None:
```
untitaker: This can never be `None`, it will always be the same instance of `bottle.LocalRequest`.

AIn Flask this exists because we make a weakref of the current request object, such that when a user copies the scope explicitly to a new thread or even into a job queue, the original request data is still there as long as the request is still alive (and not the current request from a thread that probably doesn't have request data). I am not sure if we can get a request object without threadlocals from a `LocalRequest`, but it doesn't seem so.

sentry_sdk/integrations/bottle.py
```diff
@@ -0,0 +1,165 @@
...
+            return event
+
+        try:
+            if integration.transaction_style == "endpoint":
+                event["transaction"] = request.route.name or getattr(
+                    request.route.callback, "func_name", request.route.callback.__name__
```
untitaker: I think the function `transaction_from_function` is useful here. It creates a full path including module etc

tox.ini
```diff
@@ -19,6 +19,7 @@ envlist =
     {pypy,py2.7}-django-1.6
 
     {pypy,py2.7,py3.5,py3.6,py3.7,py3.8}-flask-{1.0,0.11,0.12,dev}
+    {pypy,py2.7,py3.5,py3.6,py3.7,py3.8}-bottle-{0.12.13,0.12.14,0.12.15,0.12.16}
```
untitaker: I would prefer to only test `0.x` versions, not every bugfix release.

shenek:
Thanks for the review. I'll try to process all the comments shortly. Also thanks for the documentation hint/link.

tox.ini
shenek: No problem. Should be fixed by now.

sentry_sdk/integrations/bottle.py
shenek: Thanks for the hint. The function seems quite handy here.

sentry_sdk/integrations/bottle.py
shenek: You're right. Thanks for the explanation.

sentry_sdk/integrations/bottle.py
```diff
@@ -0,0 +1,165 @@
...
+
+                def patched_call(self, *args, **kwargs):
+                    try:
+                        old_call(*args, **kwargs)
+                    except Exception as exception:
+                        hub = Hub.current
+                        event, hint = event_from_exception(
+                            exception,
+                            client_options=hub.client.options,
+                            mechanism={"type": "bottle", "handled": self.app.catchall},
```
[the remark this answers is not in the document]
shenek: Thanks for the clarification. I set `handled` to `False`.

untitaker:
Almost there.

* We can ignore OSX, it's broken on Travis
* Could you run formatting? `CONTRIBUTING.md` should have some commands to enable this

sentry_sdk/integrations/bottle.py
```diff
@@ -0,0 +1,165 @@
...
+            hub = Hub.current
+            integration = hub.get_integration(BottleIntegration)
+            if integration is None:
+                return old_app(self, environ, start_response)
+
+            # monkey patch method self(Bottle).router.match -> (route, args)
+            # to monkey patch route.call
+            old_match = self.router.match
+
+            def patched_match(self, *args, **kwargs):
```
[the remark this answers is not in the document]
shenek: I agree the patching done in this way is not really nice and a bit confusing. I tried to patch the other functions outside of `sentry_patched_wsgi_app` and I was more or less successful.

Just instead of `Route.call` (which is a cached_property) I  managed to patch `Router._make_callback`. The tests on my machine seems to work.

shenek:
I just pushed two other fixups. The first updates the patching a bit and the second should reformat the source code using `make format`.

untitaker:
LGTM, however I can't release without docs. If you need assistance let me know

shenek:
Thanks. I'll look into.

shenek:
I just managed to prepare a pull request for the docs. see https://github.com/getsentry/sentry-docs/pull/867

untitaker:
@shenek I would plan to merge + release this as-is.

>In a couple of days I plan to do some real life testing and I tend to prepare some kind of a tutorial.

Any updates here?

shenek:
Thanks. 

I tried to hook some sample application against a production sentry instance and it seems that it captures events.

sentry-captured

sentry-bottle-detail


I plan to do some more test during next week.

untitaker:
Fantastic, I'll release tomorrow

untitaker:
@shenek so this is out as 0.7.9: https://sentry.io/for/bottle/

shenek:
Super. Thank you very much for your help.

PR #302, opened 2019-03-24
feat: Support recent Sanic versions

untitaker:


untitaker:
ref https://github.com/huge-success/sanic/issues/1532

PR #304, opened 2019-03-25
Set up CI with Azure Pipelines [skip ci]

tonyo:


tonyo:
An example pipeline provided by Azure.

PR #306, opened 2019-03-26
self-protect against broken Mapping implementations

crepererum:


sentry_sdk/utils.py
```diff
@@ -372,11 +372,16 @@ def object_to_json(obj, remaining_depth=4, memo=None):
                 ]
 
             if isinstance(obj, Mapping):
+                try:
```
untitaker: Instead of this, use `capture_internal_exceptions` around the entire function.

crepererum: It's not really an internal exception. It is broken user code somewhere in the stack the shouldn't lead to the fact that the entire stack-capturing breaks. Also I don't really understand what your suggestion is. I still have to return some value from the function and some catch-all value for all possible problems is likely not too helpful.

untitaker: I am suggesting that you replace this:

```python
def object_to_json(...):
   return <body>
```

with:

```python
def object_to_json(...):
    rv = '<broken repr>'
        rv = <body>
    return rv
```

Even if it's not our SDK that is buggy but we still want to allow the user to debug this if they really want the values.

crepererum:
should be fixed now

untitaker:
Thanks!

PR #307, opened 2019-03-26
Add badges to Readme

peleccom:
Add pypi package version badge
Add package licence Badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 
 # sentry-python - Sentry SDK for Python
 
+[![PyPi page link -- version](https://img.shields.io/pypi/v/sentry-sdk.svg)](https://pypi.python.org/pypi/sentry-sdk)
 [![Build Status](https://travis-ci.com/getsentry/sentry-python.svg?branch=master)](https://travis-ci.com/getsentry/sentry-python)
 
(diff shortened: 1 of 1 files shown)


untitaker:
If you think some information is missing from the README I would suggest to add regular links. My experience with them is that their text is cryptic and that they impact page load perf.

peleccom:
You could add just pypi version badge only to be aligned with sdk package for other languages, python-raven had this information too.

Currently, Readme file of this repository is poor. There are no module usage examples. No current version information. The first impression was it is not usable at all.

untitaker:
@peleccom Thanks for the feedback. I didn't want to duplicate information found elsewhere. I will add a small example though and I agree the current version does make sense to show.

untitaker:
trying to figure out what to do here:

* I actually like the version badge now
* https://sentry.io/for/python/ is the perfect readme for the sdk, but I don't know if I should just copypaste it into the code. Problem is that for each framework you need to do more things, so the basic example without integrations is often useless

untitaker:
Thanks for your input, I think the README should be good now. I am still not sure if this will be misleading as a user often has to set up integrations to see anything.

PR #315, opened 2019-04-02
Type hint fixes

### LINKED ISSUE #313: Incorrect signature for capture_exception
The type hint for `capture_exception()` implies only `ValueError`s can be passed in, which causes a type check warning in IDEs such as PyCharm when you pass in a non-ValueError exception.

https://github.com/getsentry/sentry-python/blob/62e3a647e5787957dcabc889bce5d53538dc423e/sentry_sdk/api.py#L53-L54

I suppose the hint should be more like

```python
# type: (BaseException) -> Optional[str] 
```

akx:
Fixes #313

### DIFF
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -10,6 +10,7 @@ if False:
     from typing import Optional
     from typing import overload
     from typing import Callable
+    from typing import Dict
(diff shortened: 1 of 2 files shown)


untitaker:
Awesome, thanks!

PR #317, opened 2019-04-03
Bottle redirect fix

shenek:
I found out that a sentry record is created when a Bottle redirect is performed.

This MR should fix that.

I also included a small tox.ini update.

### DIFF
--- a/sentry_sdk/integrations/bottle.py
+++ b/sentry_sdk/integrations/bottle.py
@@ -109,10 +109,8 @@ class BottleIntegration(Integration):
 
                 try:
                     res = prepared_callback(*args, **kwargs)
-                except HTTPResponse as exception:
-                    if exception.status == 500:
-                        capture_exception(exception)
-                    raise exception
+                except HTTPResponse:
+                    raise
                 except Exception as exception:
                     capture_exception(exception)
                     raise exception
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/bottle.py
```diff
@@ -94,16 +99,23 @@ def patched_make_callback(self, *args, **kwargs):
...
                         exception,
                         client_options=hub.client.options,
                         mechanism={"type": "bottle", "handled": False},
                     )
                     hub.capture_event(event, hint=hint)
+
+                try:
+                    res = prepared_callback(*args, **kwargs)
+                except HTTPResponse as exception:
+                    if exception.status == 500:
```
untitaker: Are those exceptions even interesting?

sentry_sdk/integrations/bottle.py
```diff
@@ -94,16 +99,23 @@ def patched_make_callback(self, *args, **kwargs):
                 return prepared_callback
 
             def wrapped_callback(*args, **kwargs):
-                try:
-                    res = prepared_callback(*args, **kwargs)
-                except Exception as exception:
+                def capture_exception(exception):
                     hub = Hub.current
```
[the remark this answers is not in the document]
shenek: You're right. Thanks for noticing.

sentry_sdk/integrations/bottle.py
shenek: It seems not. I'll skip all. Thanks.

PR #337, opened 2019-04-13
Fix for 'data' key in logger's extra

### LINKED ISSUE #336: Unified sentry-sdk integration does not have support to add stack trace in python logger using 'stack': True in extra dict.
Migration from raven to unified sentry sdk, affected extended functionalities to python logging provided by raven. _extra_from_record - excludes keywords 'stack' and 'data'. Is there a known workaround?

gagan-chawla:
Fixes #336

### DIFF
--- a/sentry_sdk/integrations/logging.py
+++ b/sentry_sdk/integrations/logging.py
@@ -106,7 +106,6 @@ COMMON_RECORD_ATTRS = frozenset(
     (
         "args",
         "created",
-        "data",
         "exc_info",
         "exc_text",
         "filename",
--- a/sentry_sdk/integrations/logging.py
+++ b/sentry_sdk/integrations/logging.py
@@ -134,19 +134,11 @@ COMMON_RECORD_ATTRS = frozenset(
 
 def _extra_from_record(record):
     # type: (LogRecord) -> Dict[str, None]
-    extra = getattr(record, 'data', None)
-    if not isinstance(extra, dict):
-        if extra:
-            extra = {'data': extra}
-        else:
-            extra = {}
-
-    for k, v in vars(record).items():
-        if k in COMMON_RECORD_ATTRS or k.startswith("_"):
-            continue
-        extra[k] = v
-
-    return extra
+    return {
+        k: v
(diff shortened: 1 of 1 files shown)


untitaker:
Can we preserve the current behavior in this SDK? I feel like Raven's old behavior added an unnecessary layer of complexity (accessing `data` first), but maybe I am missing something.

I would've expected that you just remove the `data` key from the const.

gagan-chawla:
> Can we preserve the current behavior in this SDK? I feel like Raven's old behavior added an unnecessary layer of complexity (accessing `data` first), but maybe I am missing something.
> 
> I would've expected that you just remove the `data` key from the const.

I thought of just removing `data` from COMMON_RECORD_ATTRS before, but keeping Raven's old behaviour in mind, this way we'd unwrap `data` dict before showing it in sentry's Additional Data section.

If we still wanna preserve the current behaviour, I guess removing `data` from constant also resolves the issue.

untitaker:
So what I want to understand is why you prefer Raven's behavior over sentry-sdk's behavior. I am open to either one, but we have to keep in mind that changing this behavior is a breaking change within sentry-sdk's version timeline.

gagan-chawla:
As I migrated from raven to new sentry-sdk, my projects stopped logging info as I was following old Raven's way to log additional data. This change is what I did to kinda cope with it.

I guess Raven did unwrapping of `data` to make it easier to integrate logger's components-
Since Raven provided two main extended functionalities to logging - 'extra' = {**data**: {...}, **stack**: True}, so it was convenient to maintain a `data` dict (that also performed unwrapping similar to 'extra') and to explicitly mention `stack` True when required, rather than adding `stack` key along with other data in a dict and pass it to 'extra' directly. 

But as you mentioned earlier, `exc_info` is used for callstack in sentry-sdk, I think unwrapping `data` doesn't serve any purpose now. I'm inclined towards removing `data` from constant as well.

untitaker:
Good point. I agree and would prefer to keep sentry-sdk's current behavior wrt data unwrapping.

untitaker:
Thanks!

PR #339, opened 2019-04-13
Provide simple means of stacktraces for non-error messages

### LINKED ISSUE #338: Provide a means for capturing stack trace for messages that are not errors
... without relying on the logging integration.

SDK v0.7.10

Currently, the only way to do this is via the logging integration and issuing a log at the configured level or higher with `exc_info=True`. I don't think this is a good substitute because:
- I don't want to classify these events as errors (the end goal being errors should drive alerts)
- Sometimes there are interesting or unexpected events that I want to debug more and leveraging Sentry's stack traces with locals helps in this regard
- I don't want to lower our logging integration level to warning or similar, as that may spam our project with a lot of warnings/events that are not interesting or useful

Ideally, the Sentry SDK could add an additional argument to the and `capture_message` APIs to make this simple from a user perspective:

```python
sentry_sdk.capture_message("Unexpected event", level="warning", stack_trace=True)
```

I'm trying to get this working locally with the SDK, and it's leading to incomplete event data (see picture) and also leading to pretty hairy code.
```python
sentry_sdk.capture_event({"message": "oops 2", "level": "warning", "threads": [{"stacktrace": sentry_sdk.utils.current_stacktrace(with_locals=True), "crashed": False, "current": True}])
```

This was more or less copied from the code I see for adding stack traces in `Client._prepare_event`.

Screenshot from 2019-04-13 14-39-24

goodspark:
Resolves #338

When I run equivalent code in a shell:
Screenshot from 2019-04-13 15-10-42

### DIFF
--- a/sentry_sdk/hub.py
+++ b/sentry_sdk/hub.py
@@ -13,6 +13,7 @@ from sentry_sdk.utils import (
     event_from_exception,
     logger,
     ContextVar,
+    current_stacktrace,
 )
 
 
@@ -267,8 +268,8 @@ class Hub(with_metaclass(HubMeta)):  # type: ignore
             return rv
         return None
 
-    def capture_message(self, message, level=None):
-        # type: (str, Optional[Any]) -> Optional[str]
+    def capture_message(self, message, level=None, stacktrace=False):
+        # type: (str, Optional[Any], bool) -> Optional[str]
         """Captures a message.  The message is just a string.  If no level
         is provided the default level is `info`.
         """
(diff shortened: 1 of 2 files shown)


tests/test_client.py
```diff
@@ -52,7 +52,7 @@ def test_transport_option(monkeypatch):
     assert str(Client(transport=transport).dsn) == dsn
 
 
-def test_http_proxy(monkeypatch):
+def test_http_proxy():
```
goodspark: These were unused

sentry_sdk/hub.py
```diff
@@ -267,16 +268,19 @@ def capture_event(self, event, hint=None):
...
         is provided the default level is `info`.
         """
         if self.client is None:
             return None
         if level is None:
             level = "info"
-        return self.capture_event({"message": message, "level": level})
+        event = {"message": message, "level": level}
+        if stacktrace is True:
+            event["stacktrace"] = current_stacktrace(self.client.options["with_locals"])
```
goodspark: Possible improvement: Also take a `with_locals` to override the client's options. If `None`, use the client's options?

Issue #327, opened 2019-04-08
Cannot be run wtih -OO flag


cramosc:
After installing the version 0.7.9. I cannot run sentry with the python flag -OO (discard docstrings). This is because in `sentry_sdk/integrations/__init__.py` the following was addded:
`iter_default_integrations.__doc__ += "\n- `{}`".format(import_string)`
which in case of running with -OO `__doc__` is `None` and you get the exception:
`TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str'`

untitaker:
Yeah, this code has been around for a while. I would accept a patch but we don't support `-O` on Python.

medns:
I have the same issue

untitaker:
Thanks for the PR @medns. Out of curiosity: Why do you use that flag?

allanlei:
We migrated from https://github.com/getsentry/raven-python, which works with `-O` and `-OO`, and encountered this problem.  The [migration guide](https://forum.sentry.io/t/switching-to-sentry-python/4733) also does not mention this.

The PR https://github.com/getsentry/sentry-python/pull/341 works with `-O`, but the error still happens with `-OO`.

untitaker:
@allanlei Since you edited your comment, do you encounter any issues with this SDK right now?

allanlei:
No problems with 0.7.12

untitaker:
Cool. I'll close this issue then.

PR #340, opened 2019-04-15
fix: TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str' when Python with `-O` flag

medns:
fix: TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str' when Python with `-O` flag

ref #327


untitaker:
FYI @medns it's possible to commit to the same branch so you don't need to make new PRs for every change.

PR #345, opened 2019-04-19
clear event and error processors from scope when clear() is called

baaym:
## Description
User-set scope-local event and error processors are not cleaned up when the `clear()` function is called. This creates a situation where only new entries are added, but never cleaned up. This could create a memory leak for long-running processes that assume the scope is cleaned on every `clear()`.

This merge request fixes the issue by clearing the `_event_processors` and `_error_processors` lists in the scope when `clear()` is called.

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -127,6 +127,9 @@ class Scope(object):
         self.clear_breadcrumbs()
         self._should_capture = True
 
+        self._event_processors = []
+        self._error_processors = []
+
     def clear_breadcrumbs(self):
         # type: () -> None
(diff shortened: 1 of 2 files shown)


untitaker:
This should not cause memory leaks. Resource management typically happens by pushing and popping scopes, not writing into a singleton scope and clearing it like in Raven.

`clear()` is only there if the user wants to reset their own scope data and can't pop the scope yet. `push_scope`, `pop_scope` and adding event processors is typically driven by integrations, and that's the reason it's not part of clearing the scope.

I am open to changing this behavior as I can see that clearing the scope would be expected to clear processors, but it should not be because of memory leaks in typical use.

PR #346, opened 2019-04-20
feat(integration): Add Falcon integration

jmagnusson:
Adds support for the [Falcon web framework](https://github.com/falconry/falcon). Any feedback or suggestions would be greatly appreciated.

### DIFF
--- a/sentry_sdk/integrations/falcon.py
+++ b/sentry_sdk/integrations/falcon.py
@@ -1,0 +1,162 @@
+from typing import Any, Callable, Dict
+
+import falcon
+import falcon.api_helpers
+import sentry_sdk.integrations
+from sentry_sdk.hub import Hub
+from sentry_sdk.integrations._wsgi_common import RequestExtractor
+from sentry_sdk.integrations.wsgi import SentryWsgiMiddleware
+from sentry_sdk.utils import capture_internal_exceptions, event_from_exception
+
+
+class FalconRequestExtractor(RequestExtractor):
+
+    def env(self):
+        return self.request.env
+
+    def cookies(self):
+        return self.request.cookies
+
+    def raw_data(self):
+        # As request data can only read once we won't make this available
+        # to Sentry.
+        # TODO(jmagnusson): Figure out if there's a way to support this
+        return None
+
+    def form(self):
+        return None  # No such concept in Falcon
+
+    def files(self):
+        return None  # No such concept in Falcon
+
+    def json(self):
+        # We don't touch falcon.Request.media as that can raise an exception
+        # on non-JSON requests.
+        return self.request._media
+
+
+class SentryFalconMiddleware(object):
+    """Captures exceptions in Falcon requests and send to Sentry"""
+
+    def process_request(self, req, resp, *args, **kwargs):
+        hub = Hub.current
+        integration = hub.get_integration(FalconIntegration)
+        if integration is None:
+            return
+
+        with hub.configure_scope() as scope:
+            scope.add_event_processor(
+                _make_request_event_processor(req, integration)
+            )
+
+
+class FalconIntegration(sentry_sdk.integrations.Integration):
+    identifier = "falcon"
+
+    transaction_style = None
+
+    def __init__(self, transaction_style="uri_template"):
+        # type: (str) -> None
+        TRANSACTION_STYLE_VALUES = ("uri_template", "path")
+        if transaction_style not in TRANSACTION_STYLE_VALUES:
+            raise ValueError(
+                "Invalid value for transaction_style: %s (must be in %s)"
+                % (transaction_style, TRANSACTION_STYLE_VALUES)
+            )
+        self.transaction_style = transaction_style
+
+    @staticmethod
+    def setup_once():
+        # type: () -> None
+        _patch_wsgi_app()
+        _patch_handle_exception()
+        _patch_prepare_middleware()
+
+
+def _patch_wsgi_app():
+    original_wsgi_app = falcon.API.__call__
+
+    def sentry_patched_wsgi_app(self, env, start_response):
+        hub = Hub.current
+        integration = hub.get_integration(FalconIntegration)
+        if integration is None:
+            return original_wsgi_app(self, env, start_response)
+
+        sentry_wrapped = SentryWsgiMiddleware(
+            lambda envi, start_resp: original_wsgi_app(self, envi, start_resp)
+        )
+        with hub.push_scope() as scope:
+            scope._name = "falcon"
+            resp = sentry_wrapped(env, start_response)
+
+        return resp
+
+    falcon.API.__call__ = sentry_patched_wsgi_app
+
+
+def _patch_handle_exception():
+    original_handle_exception = falcon.API._handle_exception
+
+    def sentry_patched_handle_exception(self, *args):
+        # NOTE(jmagnusson): falcon 2.0 changed falcon.API._handle_exception
+        # method signature from `(ex, req, resp, params)` to
+        # `(req, resp, ex, params)`
+        if isinstance(args[0], Exception):
+            ex = args[0]
+        else:
+            ex = args[2]
+
+        was_handled = original_handle_exception(self, *args)
+
+        hub = Hub.current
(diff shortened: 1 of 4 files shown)


sentry_sdk/integrations/falcon.py
```diff
@@ -0,0 +1,162 @@
...
+        if integration is None:
+            return
+
+        with hub.configure_scope() as scope:
+            scope.add_event_processor(
+                _make_request_event_processor(req, integration)
+            )
+
+
+class FalconIntegration(sentry_sdk.integrations.Integration):
```
untitaker: Could you rewrite this import to import `Integration` into this namespace? (like other integrations)

sentry_sdk/integrations/falcon.py
```diff
@@ -0,0 +1,162 @@
...
+    def sentry_patched_wsgi_app(self, env, start_response):
+        hub = Hub.current
+        integration = hub.get_integration(FalconIntegration)
+        if integration is None:
+            return original_wsgi_app(self, env, start_response)
+
+        sentry_wrapped = SentryWsgiMiddleware(
+            lambda envi, start_resp: original_wsgi_app(self, envi, start_resp)
+        )
+        with hub.push_scope() as scope:
```
untitaker: You do not have to push a scope here as the WSGI middleware already does this (it makes a new hub which is close enough.

Setting `scope._name` is not strictly necessary, but you can do this in the falcon middleware.

sentry_sdk/integrations/falcon.py
```diff
@@ -0,0 +1,162 @@
...
+    def form(self):
+        return None  # No such concept in Falcon
+
+    def files(self):
+        return None  # No such concept in Falcon
+
+    def json(self):
+        # We don't touch falcon.Request.media as that can raise an exception
+        # on non-JSON requests.
+        return self.request._media
```
untitaker: I think this is problematic as it will send request data with an event only once it has been accessed. If exceptions are your concern I think it's best to catch them down here.

sentry_sdk/integrations/falcon.py
```diff
@@ -0,0 +1,162 @@
...
+    def sentry_patched_handle_exception(self, *args):
+        # NOTE(jmagnusson): falcon 2.0 changed falcon.API._handle_exception
+        # method signature from `(ex, req, resp, params)` to
+        # `(req, resp, ex, params)`
+        if isinstance(args[0], Exception):
+            ex = args[0]
+        else:
+            ex = args[2]
+
+        was_handled = original_handle_exception(self, *args)
```
untitaker: Is this `True` if the user registered a generic 500 error page with Falcon?

What I am getting at is that in many frameworks there is no clear way to know whether the user intended to actually suppress an exception by registering an error handler. I see that you have this testcase where you call `capture_exception` in the errorhandler again, but this means that the integration will not capture anything out of the box for the average application.

For Pyramid we ignore its equivalent of `was_handled` and capture exceptions even if there's an errorhandler. It's unfortunate and people have been complaining about it, but in the face of ambiguity I try to err on the side of sending too much than sending too little.

sentry_sdk/integrations/falcon.py
```diff
@@ -0,0 +1,162 @@
...
+class FalconRequestExtractor(RequestExtractor):
+
+    def env(self):
+        return self.request.env
+
+    def cookies(self):
+        return self.request.cookies
+
+    def raw_data(self):
+        # As request data can only read once we won't make this available
```
untitaker: I think it's ok if you return a placeholder string to just tell the user that "raw data" was part of the request.

In the end we will not send the raw payload to sentry anyway as we can't strip sensitive data from arbitrary blobs (we also don't send *contents* of file uploads for that reason).

sentry_sdk/integrations/falcon.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def process_request(self, req, resp, *args, **kwargs):
+        hub = Hub.current
+        integration = hub.get_integration(FalconIntegration)
+        if integration is None:
+            return
+
+        with hub.configure_scope() as scope:
+            scope.add_event_processor(
+                _make_request_event_processor(req, integration)
```
untitaker: Can you create a weakref for the request here? If someone tries to clone the scope to somewhere else outside of the request handler we want to avoid holding onto too many resources.

Flask, Django etc do this.

untitaker:
Also I think this integration doesn't work for Python 2, see CI.

sentry_sdk/integrations/falcon.py
jmagnusson: Unfortunately no, as [falcon.request.Request uses __slots__](https://github.com/falconry/falcon/blob/master/falcon/request.py#L401) and does not contain `__weakref__`, [which it must](https://docs.python.org/3.7/reference/datamodel.html#slots). I don't know how big the implication of this is. Perhaps it's okay for now?

sentry_sdk/integrations/falcon.py
jmagnusson: Fixed

sentry_sdk/integrations/falcon.py
jmagnusson: Fixed

sentry_sdk/integrations/falcon.py
jmagnusson: Correct.

Unless an internal error handler is added by the user, all requests will be captured in Sentry. Personally I think it's sensible to not have sentry add any error handlers and let the user decide if the event should still be captured in Sentry. But I can adapt as it's your library and not mine 😉

jmagnusson: So just let me know if I should change this or not 😊

jmagnusson:
Thanks for your useful feedback @untitaker, much appreciated ❤

sentry_sdk/integrations/falcon.py
untitaker: >Personally I think it's sensible to not have sentry add any error handlers and let the user decide if the event should still be captured in Sentry

I assume you would expect the user to decide this by putting/not putting `capture_exception  in their handler?

That's fair but generally we want framework integrations to be useful when the user is just starting to use Sentry, so just calling `init()` should demonstrate the *value* of Sentry to the user. This has been the expectation for every other integration and that's also how we market all of our SDKs. Given that, we have two choices:

* Capture too much, and risk capturing some useless issues
* Capture nothing, which makes the user think our product is broken

The latter results in more support tickets ;)

So yeah please change it. You can still use `before_send` to drop certain exceptions.

sentry_sdk/integrations/falcon.py
untitaker: Yes, it's fine not to do it.

tox.ini
```diff
@@ -24,6 +24,8 @@ envlist =
 
     {pypy,py2.7,py3.5,py3.6,py3.7,py3.8}-bottle-0.12
 
+    {pypy,py2.7,py3.5,py3.6,py3.7,py3.8}-falcon-1.4
```
untitaker: I overlooked this.. you need to write `falcon-{1.4,dev}` to actually test the dev version. Bottle is broken here as well.

sentry_sdk/integrations/falcon.py
jmagnusson: Thanks, I guess it makes sense in the context of Sentry's core goals. Updated code as suggested.

tox.ini
jmagnusson: I've just changed it to 2.0 actually. 2.0 is in its third release candidate so testing 3.0 dev feels a bit early, right?

jmagnusson:
@untitaker docs PR can be found here: https://github.com/getsentry/sentry-docs/pull/924

untitaker:
Will release `0.7.11` soon.

untitaker:
Thanks for this awesome integration @jmagnusson!

jmagnusson:
No probs - Thank you for the great feedback 😀

untitaker:
0.7.11 is released and the docs are deployed as well: https://docs.sentry.io/platforms/python/falcon/

Issue #344, opened 2019-04-18
Does this client truncate long strings and lists?


Siecje:
In the raven config there was `'list_max_length'` and `'string_max_length'` but they don't exist for the new client.

I see there is this function for strings but I don't see one for lists.
https://github.com/getsentry/sentry-python/blob/2f5487235848e3bda9eae150facb0bb9e0c490a4/sentry_sdk/utils.py#L242-L248

untitaker:
We have this: https://github.com/getsentry/sentry-python/blob/263a0bd92bdf85c6b23c889c8cd5a28adfd81ea2/sentry_sdk/utils.py#L823

People have reported that this is too lenient, but we haven't figured out why yet.

Siecje:
So if I have a list with 1000 items I will only get the first 20 items?

https://github.com/getsentry/sentry-python/blob/263a0bd92bdf85c6b23c889c8cd5a28adfd81ea2/sentry_sdk/utils.py#L844

untitaker:
Yes exactly. We intend to make this configurable but keep in mind that you will likely run into serverside limits as well.

rhcarvalho:
The original question has been answered, closing to clean up the issue tracker.

PR #352, opened 2019-04-25
[#344] Add max_depth, max_breadth, max_string_length options

Siecje:


Siecje:
This purpose of this PR is to increase the payload sent to Sentry. I guess you could decrease the defaults to prevent losing information which I noticed happens with large variables.

I was using this script with this branch.
```Python
import sentry_sdk

sentry_sdk.init(
    "",
    max_depth=1000,
    max_breadth=1000,
    max_string_length=1000,
)

def foo():
    long_list = list(range(1001))
    long_string = "a" * 1001
    1/0

foo()
```

With the limits set high I only get `long_list`. I don't get `long_string`. 
If I set them to what appears to be the server limits. 512 for strings and 206 for lists. I do get both variables.

rybakm:
Is this request being worked on? If not, is there any chance for this issue (and maybe other "data trimming related") to be fixed in the future? It would be great to be able to set limit for string length etc. in config instead of monkey patching.

sl0thentr0py:
we will revisit data truncation separately in a coherent way, closing this since very old

Issue #360, opened 2019-05-08
Possibility to send error through the shell.


tolgahanuzun:
Normally I used the `raven` library for sentry. As of yesterday, I decided to move to the new library. (sentry-sdk) I tried to test the error via shell to test. But nothing happened.

When I examined the codes I saw an `if` to block it. I think this decision can be made optional. If you think so, I have prepared a PR. You can check and ask for edits..

(It also gave a great pleasure to visit the codes inside the library. :) )

Pr link : https://github.com/getsentry/sentry-python/pull/361

untitaker:
Keeping this open because the feature request is still valid. I don't want to have this as option on `init` but if people want to, we can have it as option on the integration (`init(integrations=[ExcepthookIntegration(always_run=True)])`)

PR #361, opened 2019-05-08
Add shell control in options

tolgahanuzun:
Description : https://github.com/getsentry/sentry-python/issues/360

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -38,7 +38,6 @@ if False:
             "attach_stacktrace": bool,
             "ca_certs": Optional[str],
             "propagate_traces": bool,
-            "is_shell_active": bool
         },
         total=False,
     )
@@ -72,7 +71,6 @@ DEFAULT_OPTIONS = {
     "attach_stacktrace": False,
     "ca_certs": None,
     "propagate_traces": True,
-    "is_shell_active": False,
 }
 
 
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -38,6 +38,7 @@ if False:
             "attach_stacktrace": bool,
             "ca_certs": Optional[str],
             "propagate_traces": bool,
+            "is_shell_active": bool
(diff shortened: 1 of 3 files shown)


untitaker:
Thanks for your PR! Would you mind making this an option on the integration, like the options on e.g. `FlaskIntegration`?

tolgahanuzun:
I think the integration could be this way. Isn't that right?

```python3
import sentry_sdk
sentry_sdk.init(
    'http.....',
    debug=True,
    is_shell_active=True,
)
```

untitaker:
@tolgahanuzun My point is that this global setting is not used by any other integration. To configure `transaction_style` for the `FlaskIntegration`, you run this:

```python
sentry_sdk.init(
    ...,
    integrations=[FlaskIntegration(transaction_style="...")]
)
```

And I think it should work the same for this integration, because it's really only something the integration uses.

untitaker:
You can send events with `capture_event` and other integrations. I have trouble imagining where you'd be fine with a `NameError` being sent to Sentry:

```
>>> foo = 42
>>> fooo
NameError: ...
```

untitaker:
>In every situation, this should be a configurable parameter, not hard coded.

We will not add configurable options without a reason, especially not on the client object. You write:

>We need to send events to Sentry working in REPL.

And I don't understand why that is desirable. A REPL session is full of typos and broken code that you then rerun immediately. Why do you want to add all of that into Sentry?

In any case it seems nowhere near a popular enough usecase to have this as param on `init`.

untitaker:
>We need to send events to Sentry working in REPL.

Again, why is this desirable behavior?

tolgahanuzun:
`In any case it seems nowhere near a popular enough usecase to have this as param on init.`
Well then, thank you. :) You can close Pr.

untitaker:
@tolgahanuzun To be clear I would be happy to have this on the integration object, but we are more careful with direct args on `init`

PR #368, opened 2019-05-16
ref: Performance improvements when serializing events

untitaker:


untitaker:
Large perf difference for breadth benchtest. Requires running without `--boxed` (edit `pytest.ini`)

untitaker:
We probably should land https://github.com/getsentry/sentry-python/pull/226 at some point to get benchtests working out of the box...

Issue #379, opened 2019-05-30
SystemExit and KeyboardInterrupt no longer caught (regression from raven-python)


saifelse:
The same reasoning as getsentry/raven-python#675 applies for sentry-python,
where gunicorn raising a SytemExit(1) is something we would like to capture
in Sentry, as it is unexpected behavior from the application.

WhyNotHugo:
Huh, I've the opposite issue: raven did not log this, but the new client does.
However, I don't want this logged, since it doesn't make sense to me (the system will exit on every deployment, and that can happen way too many times).

Is there any way to avoid logging `SystemExit`? It doesn't make sense to me, and is filling my sentry issues board.

untitaker:
@WhyNotHugo can you confirm those events have the wsgi mechanism?

WhyNotHugo:
Yes, they all do.

untitaker:
Hmm, from what I know this means you abort gunicorn while it is handling a request and that might be the wrong way to shut it down.

In any case you can explicitly ignore this error like any error class with this here: https://github.com/getsentry/sentry-python/issues/149#issuecomment-434448781

WhyNotHugo:
Hmmm... let me double check, I believe we're killing gunicorn gracefully, but I might be wrong.

WhyNotHugo:
Yeah, looks like docker was killing them before gunicorn's internal grace period, I'll tweak that and it should solve it for me. Thanks.

PR #380, opened 2019-05-30
fix(wsgi): Log SystemExit (with non-zero exit code) and KeyboardInterrupt

saifelse:
Neither SystemExit nor KeyboardInterrupt subclass Exception, so we must
explicitly handle these two exception classes in addition to Exception in the
WSGI middleware.

This is notably a direct port from raven-python:
 - https://github.com/getsentry/raven-python/blob/master/raven/middleware.py
 - https://github.com/getsentry/raven-python/blob/master/tests/middleware/tests.py

Fixes: GH-379

There are notably 3 places an exception could be raised, but it looks like we're only
testing one case. If we want to test the other two, would appreciate advice 🙏🏾.

### DIFF
--- a/sentry_sdk/integrations/wsgi.py
+++ b/sentry_sdk/integrations/wsgi.py
@@ -90,9 +90,9 @@ class SentryWsgiMiddleware(object):
             except Exception:
                 reraise(*_capture_exception(hub))
             except KeyboardInterrupt:
-                reraise(*_capture_exception(self._hub))
+                reraise(*_capture_exception(hub))
             except SystemExit as e:
-                reraise(*_capture_exception(self._hub, skip_capture=e.code == 0))
+                reraise(*_capture_exception(hub, skip_capture=e.code == 0))
 
         return _ScopedResponse(hub, rv)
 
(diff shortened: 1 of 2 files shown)


saifelse:
@untitaker : thanks, back to you!

sentry_sdk/integrations/wsgi.py
```diff
@@ -192,8 +197,8 @@ def close(self):
                 self._response.close()
             except AttributeError:
                 pass
-            except Exception:
-                reraise(*_capture_exception(self._hub))
+            except BaseException:
```
untitaker: You're missing an `as e` here

saifelse: oof, thanks

saifelse:
> one change necessary, also please make sure that CI passes (see `CONTRIBUTING.md`)
looks like it was failing from zeus, so hopefully merging in master takes care of that 🤞🏾

sentry_sdk/integrations/wsgi.py
```diff
@@ -149,17 +149,22 @@ def get_client_ip(environ):
     return environ.get("REMOTE_ADDR")
 
 
-def _capture_exception(hub):
-    # type: (Hub) -> ExcInfo
+def _capture_exception(hub, e):
```
untitaker: Instead of taking an additional param here, could you read this info from `exc_info[2]`?

untitaker:
Thanks!

PR #381, opened 2019-05-30
fix(celery): Fix RecursionError

### LINKED ISSUE #265: celery integration RecursionError
Hi there,
I upgraded sentry_sdk to 0.7.0 and started getting RecursionError if there's an issue with celery task. Sentry record doesn't contain any stack trace for that but found that error in my apm system (can attach screenshot only, text data is a real mess there). I'm running celery 4.2.1 on Ubuntu 18.

untitaker:
Fix #265


thomasw:
Thanks, @untitaker. I can confirm that this solves the issue while running my test suite at least.

PR #385, opened 2019-06-05
Fix py.typed not being included in distribution

bluetech:
The `typedpkg` example in the PEP is a little misleading unfortunately.

I tested that it is now properly included and properly picked up by mypy.


untitaker:
Thanks!

PR #388, opened 2019-06-06
fix: Honor __tracebackhide__ even if frame has name

### LINKED ISSUE #387: should_hide_frame fails to consider __hide_traceback__ in any frame with a __name__
https://github.com/getsentry/sentry-python/blob/8b3de2c80e9c4bc23fe860855d63e6eb294e987d/sentry_sdk/utils.py#L211-L226

If `__name__` is in `frame.f_globals` and is a string, then no check for `__hide_traceback__` is ever performed, which effectively makes that variable unusable to hide a frame from a stacktrace.

millimat:
fix #387


untitaker:
Thanks!

PR #395, opened 2019-06-21
fix: Deal with gevent patched thread locals

### LINKED ISSUE #392: gunicorn with gevent worker - is this supported?
We see 'mixed' events in our Sentry installation. We recently activated the gevent worker for gunicorn, to get more concurrent requests (our service is highly IO dependent).

In the docs on sentry.io I couldn't find any documentation if gevent workers are supported or not.

I have seen some older tickets that state that it is not supported at all, but not sure if this is still the case? If it is not supported, how could we make it supported? I guess we would need to put tracking information into a Greenlet local? Would that be the right way to do it?

Thanks for any information.

untitaker:
Fix #392

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -734,15 +734,50 @@ def format_and_strip(
     )
 
 
-HAS_REAL_CONTEXTVARS = True
+def _is_threading_local_monkey_patched():
+    # type: () -> bool
+    try:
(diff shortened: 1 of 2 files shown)


amir-hadi:
Thank you very much for this @untitaker!

untitaker:
0.9.1 is out :)

PR #400, opened 2019-06-24
fix(docs): Fix docstring in `ignore_logger`

lobsterkatie:
Using this function will block regular event recording as well as breadcrumb recording.

### DIFF
--- a/sentry_sdk/integrations/logging.py
+++ b/sentry_sdk/integrations/logging.py
@@ -27,9 +27,10 @@ _IGNORED_LOGGERS = set(["sentry_sdk.errors"])
 
 def ignore_logger(name):
     # type: (str) -> None
-    """This disables the breadcrumb integration for a logger of a specific
-    name.  This primary use is for some integrations to disable breadcrumbs
-    of this integration.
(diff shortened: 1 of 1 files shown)


sentry_sdk/integrations/logging.py
```diff
@@ -27,9 +27,10 @@
 
 def ignore_logger(name):
     # type: (str) -> None
-    """This disables the breadcrumb integration for a logger of a specific
-    name.  This primary use is for some integrations to disable breadcrumbs
-    of this integration.
+    """This disables recording (both in breadcrumbs and as events) calls to
+    a logger of a specific name.  Among other uses, many of our integrations
+    use this to prevent their actions being recorded as breadcrumbs. Exposed
```
untitaker: > to prevent their actions being recorded as breadcrumbs

More precisely, the integrations use this to suppress spam breadcrumbs/events from the *framework* that would be sent if the integration was disabled. A classic is sending a stringified traceback as log message, or more generally a log error by the framework for every exception we already recorded via other means

untitaker: Ah, I now understand what *their* means in this context, nvm my comment.

PR #406, opened 2019-07-01
Add typings for CeleryIntegration.__init__

rik:
I have not tested this change but did something similar that the existing typing in `DjangoIntegration`.


untitaker:
Thanks!

PR #412, opened 2019-07-05
fix: Fix typings for Python2 and check in CI

bluetech:
Run mypy in --py2 mode and fix the errors. Also add it to CI
(Python3-only integrations are excluded).

### DIFF
--- a/sentry_sdk/worker.py
+++ b/sentry_sdk/worker.py
@@ -8,7 +8,7 @@ from sentry_sdk.utils import logger
 
 MYPY = False
 if MYPY:
-    from queue import Queue
+    from queue import Queue  # type: ignore # for --py2
     from typing import Any
     from typing import Optional
     from typing import Callable
--- a/sentry_sdk/_compat.py
+++ b/sentry_sdk/_compat.py
@@ -27,6 +27,11 @@ if PY2:
         return cls
 
     exec("def reraise(tp, value, tb=None):\n raise tp, value, tb")
+    if MYPY:
+
+        def reraise(tp, value, tb=None):
+            # type: (Optional[Type[BaseException]], Optional[BaseException], Optional[Any]) -> None
+            pass
 
 
 else:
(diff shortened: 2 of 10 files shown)


sentry_sdk/worker.py
```diff
@@ -8,7 +8,7 @@
 
 MYPY = False
 if MYPY:
-    from queue import Queue
+    from queue import Queue  # type: ignore # for --py2
```
untitaker: If we do this we might as well import the queue from `_compat`, which already imports the right queue depending on version

untitaker:
@bluetech what's the motivation for running mypy in python 2 mode at all? Are type completions etc broken for Python 2 users of the SDK?

bluetech:
This conflicts with the tracing support which was just merged - I'll fix that.

> what's the motivation for running mypy in python 2 mode at all? Are type completions etc broken for Python 2 users of the SDK?

No, I don't use Py2 personally, but when making the other PR I wanted to make sure I don't break Py2 users (since I think sentry-sdk wants to support it) so I added it to the CI.

sentry_sdk/worker.py
bluetech: Yes, `queue` is actually imported from `_compat` just above, but I think mypy gets confused and prefers the stdlib. I tried various things, couldn't get it to work, so just added the ignore. It's only needed for py2 though.

untitaker:
my theory is that people don't actually use mypy on python 2 projects since it's extremenly painful to use without syntax support. I would prefer to ignore this issue unless we know users who do this

bluetech:
Maybe some IDEs use it when the user uses Python 2?

Anyway, if there's actually no intention to support typings for Python 2, ignoring this is perfectly fine by me - as I said I don't actually need it.

untitaker:
Yeah, I'd much rather not. When implementing the new features in the tracing PR I noticed that mypy can already quite the nuisance and I don't think it's worth it to add it for the Python 2 codepath.

PR #413, opened 2019-07-05
feat: Improve type annotations in utils.py

bluetech:
Replace some Anys with proper types.

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -20,11 +20,13 @@ if MYPY:
     from typing import Tuple
     from typing import Type
     from typing import Union
+    from types import FrameType
+    from types import TracebackType
 
     from sentry_sdk.hub import Hub
 
     ExcInfo = Tuple[
-        Optional[Type[BaseException]], Optional[BaseException], Optional[Any]
+        Optional[Type[BaseException]], Optional[BaseException], Optional[TracebackType]
     ]
 
     Event = Dict[str, Any]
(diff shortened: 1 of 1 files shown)


untitaker:
Thanks for your continued efforts on this @bluetech!

bluetech:
@untitaker This has one failure left:

```
sentry_sdk/utils.py:398: error: Argument 2 to "filename_for_module" has incompatible type "Optional[str]"; expected "str"
```

It looks legitimate but I'm not sure. The `module` variable in `serialize_frame` can be None, but `filename_for_module` doesn't except None. But `filename_for_module` is wrapped in a `try...except Exception` so it doesn't crash, but I'm not sure if that's intended.

Would you mind taking a look?

untitaker:
Yeah, seems like i was a bit careless about the logic on that one

sentry_sdk/utils.py
```diff
@@ -358,7 +358,10 @@ def safe_repr(value):
 
 
 def filename_for_module(module, abs_path):
-    # type: (str, str) -> str
+    # type: (str, Optional[str]) -> Optional[str]
```
bluetech: Looking at `serialize_frame`, `module` can be `None` too.

untitaker: You're right, not sure why mypy didn't catch that one. I think I fixed it now though!

untitaker:
Thanks again!

PR #416, opened 2019-07-08
test: Treat warnings as errors

### LINKED ISSUE #415: Collections.abc deprecation warning
The below line generates a warning regarding importing from collections.

https://github.com/getsentry/sentry-python/blob/7c4683f0cd2a32738ba5e22c4a9bc6040ea13f15/sentry_sdk/tracing.py#L27

```
  /home/karthi/freezegun-venv/lib/python3.9/site-packages/sentry_sdk/tracing.py:27: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    class EnvironHeaders(collections.Mapping):  # type: ignore
```

This can be fixed like similar fix done in other file

https://github.com/getsentry/sentry-python/blob/44b8949fe2ff0f6319dd586d15c3e76b239971d6/sentry_sdk/serializer.py#L30-L37

untitaker:
Fix #415

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,4 @@
 [pytest]
 DJANGO_SETTINGS_MODULE = tests.integrations.django.myapp.settings
-addopts = --boxed --tb=short -Werror
+addopts = --boxed --tb=short
 markers = tests_internal_exceptions
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,4 @@
 [pytest]
 DJANGO_SETTINGS_MODULE = tests.integrations.django.myapp.settings
-addopts = --boxed --tb=short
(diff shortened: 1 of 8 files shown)


tirkarthi:
Looking at the build failure it seems some dependencies in the project also have deprecation warnings related to collections. Some of the libraries still don't have releases where this is fixed and this is also blocking the removal in CPython such that the planned removal in Python 3.8 is moved to Python 3.9 given that vendored copy of pip in Python is also incompatible : https://github.com/python/cpython/pull/10596

untitaker:
lol, okay I will try to explicitly ignore warnings in each test and see how that goes

PR #426, opened 2019-07-11
fix: Allow int keys in extra dict

vladlutkov:
If I pass dict with int keys in extra keyword argument
```
logger.warning('test', extra={1:1})
```
I'll get an exception and event will not be sent.

### DIFF
--- a/sentry_sdk/integrations/logging.py
+++ b/sentry_sdk/integrations/logging.py
@@ -139,7 +139,8 @@ def _extra_from_record(record):
     return {
         k: v
         for k, v in iteritems(vars(record))
-        if k not in COMMON_RECORD_ATTRS and not k.startswith("_")
+        if k not in COMMON_RECORD_ATTRS
+        and (not isinstance(k, str) or not k.startswith("_"))
     }
 
 
--- a/tests/integrations/logging/test_logging.py
+++ b/tests/integrations/logging/test_logging.py
@@ -61,6 +61,17 @@ def test_logging_extra_data(sentry_init, capture_events):
     )
 
 
+def test_logging_extra_data_integer_keys(sentry_init, capture_events):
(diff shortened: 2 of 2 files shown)


sentry_sdk/integrations/logging.py
```diff
@@ -139,7 +139,7 @@ def _extra_from_record(record):
     return {
         k: v
         for k, v in iteritems(vars(record))
-        if k not in COMMON_RECORD_ATTRS and not k.startswith("_")
+        if k not in COMMON_RECORD_ATTRS and not (isinstance(k, str) and k.startswith("_"))
```
untitaker: can you change this into `and (not k.startswith("_") or not isinstance(k, str)`? Not simpler but easier to read for me.

vladlutkov: this will lead to an AttributeError in `k.startswith("_")`

untitaker: sorry, of course the swapped variant `not isinstance(k, str) or not ...`

vladlutkov:
I added a test

sentry_sdk/integrations/logging.py
vladlutkov: Ok

untitaker:
Sorry but could you run `make format` and `make lint` on this? It's not properly formatted. For more information see `CONTRIBUTING.md`

untitaker:
Thanks! Do you need a release for this right now or is this just something you happened to come by?

vladlutkov:
Thanks, I'm ready to wait until the next release.
Off-topic
Does the library only support python 3 and higher?

untitaker:
You've just found a regression in the docs, please refer to
https://github.com/getsentry/sentry-docs/pull/1112.

PR #434, opened 2019-07-16
:bug: [#423] fix reference cycle

### LINKED ISSUE #423: Exception: raise OSError("handle is closed")
When I initialized sentry_sdk and used concurrent.futures.process.ProcessPoolExecutor, the exception will be raised after python exit.

```
from concurrent.futures.process import ProcessPoolExecutor

import sentry_sdk

sentry_sdk.init(dsn="")


def test():
    ...


if __name__ == "__main__":
    with ProcessPoolExecutor(max_workers=4) as worker:
        worker.submit(test)
```

The exception:
```
Error in atexit._run_exitfuncs:
  File "/Users/tony.li/miniconda3/lib/python3.7/concurrent/futures/process.py", line 101, in _python_exit
    thread_wakeup.wakeup()
  File "/Users/tony.li/miniconda3/lib/python3.7/concurrent/futures/process.py", line 89, in wakeup
    self._writer.send_bytes(b"")
  File "/Users/tony.li/miniconda3/lib/python3.7/multiprocessing/connection.py", line 183, in send_bytes
    self._check_closed()
  File "/Users/tony.li/miniconda3/lib/python3.7/multiprocessing/connection.py", line 136, in _check_closed
    raise OSError("handle is closed")
OSError: handle is closed
```

GuangTianLi:
Fixes #423 

According to the [Python Data Model](https://docs.python.org/3.7/reference/datamodel.html?#index-35), the code below will cause a reference cycle between  `__self__` and `__dict__`.

```
class Test:
    def start(self, *a, **kw):
        self.run()

    def run(self):
        ...


old_start = Test.start


def _wrap_run(old_run):
    def run(*a, **kw):
        result = old_run(*a, **kw)
        return result
    return run


def test_start(self, *a, **kw):
    self.run = _wrap_run(self.run)
    return old_start(self, *a, **kw) 


Test.start = test_start

if __name__ == "__main__":
    tmp = Test()
    tmp.start()
```
The references graph like this::

Meantime [Python Standard Library - concurrent.futures](https://github.com/python/cpython/blob/3.7/Lib/concurrent/futures/process.py#L584) used thread to manage tasks queue, therefore when sentry patched start method of `Thread`, it will cause reference cycle and prevent gc from collecting.

So in `concurrent.futures.process`, the `_queue_management_thread` in _threads_wakeups will be discarded until running a full gc collection, if not, exception will be raise when [python exited](https://github.com/python/cpython/blob/3.7/Lib/concurrent/futures/process.py#L102), because the actual manage thread has already stopped.

### DIFF
--- a/sentry_sdk/integrations/threading.py
+++ b/sentry_sdk/integrations/threading.py
@@ -33,7 +33,7 @@ class RunDescriptor:
         inside closure rather than storing reference of itself into the attribute of instance.
         """
         if instance is None:
-            return self
+            return self.class_func
 
         def run(*a, **kw):
             hub = self.parent_hub or Hub.current
@@ -56,7 +56,6 @@ class ThreadingIntegration(Integration):
     def setup_once():
         # type: () -> None
         old_start = Thread.start
-        old_run = Thread.run
 
         def sentry_start(self, *a, **kw):
             hub = Hub.current
@@ -67,7 +66,7 @@ class ThreadingIntegration(Integration):
                 else:
                     hub_ = Hub(hub)
 
-                self.__class__.run = RunDescriptor(old_run, hub_)
+                self.__class__.run = RunDescriptor(self.__class__.run, hub_)
 
             return old_start(self, *a, **kw)  # type: ignore
 
--- a/sentry_sdk/integrations/threading.py
+++ b/sentry_sdk/integrations/threading.py
@@ -16,10 +16,13 @@ if MYPY:
     from typing import Any
 
 
-class RunDescriptor:
-    def __init__(self, func, parent_hub):
-        self.class_func = func
+class PatchedInstanceMethodDescriptor(object):
+    def __init__(self, class_method, parent_hub):
+        self.origin_class_method = class_method
         self.parent_hub = parent_hub
+
+    def __set__(self, instance, value):
+        pass
 
     def __get__(self, instance, owner):
         """The descriptor which is used to patch instance method is sort of tricky and
@@ -33,13 +36,13 @@ class RunDescriptor:
         inside closure rather than storing reference of itself into the attribute of instance.
         """
         if instance is None:
-            return self.class_func
+            return self.origin_class_method
 
         def run(*a, **kw):
             hub = self.parent_hub or Hub.current
             with hub:
                 try:
-                    return MethodType(self.class_func, instance)(*a, **kw)
+                    return MethodType(self.origin_class_method, instance)(*a, **kw)
                 except Exception:
                     reraise(*_capture_exception())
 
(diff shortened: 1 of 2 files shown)


untitaker:
This is quite impressive! I don't yet understand how the reference cycle is created but I will take a closer look tomorrow. Thanks for your work!

GuangTianLi:
In the end, I found more proper way to prevent reference cycle, but it is a sort of tricky.

untitaker:
Patching the class itself will cause issues if one instantiates multiple instances of thread and starts them. We don't want to wrap `run` more than once per instance.

GuangTianLi:
I fixed it by passing the __class__ of instance into the descriptor.

shidenggui:
Is there any progress about this pr? We also meet this problem in production.

bluetech:
> I fixed it by passing the__class__ of instance into the descriptor.

IIUC, this is still not thread safe. When you do this

```py
                self.__class__.run = PatchedInstanceMethodDescriptor(
                    self.__class__.run, hub_
                )
```

it affects all the entire Thread class, not just the single instance which is intended.

bluetech:
How about something like this (untested):

```diff
diff --git a/sentry_sdk/integrations/threading.py b/sentry_sdk/integrations/threading.py
index 3bd6032..5907b45 100644
--- a/sentry_sdk/integrations/threading.py
+++ b/sentry_sdk/integrations/threading.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import
 
 import sys
 
-from threading import Thread
+from threading import Thread, current_thread
 
 from sentry_sdk import Hub
 from sentry_sdk._compat import reraise
@@ -35,26 +35,24 @@ class ThreadingIntegration(Integration):
                 else:
                     hub_ = Hub(hub)
 
-                self.run = _wrap_run(hub_, self.run)
+                # Use the function object rather than the method
+                # descriptor to avoid a reference cycle (issue #434).
+                old_run = self.run.__func__
+                def sentry_run(*a, **kw):
+                    hub = hub_ or Hub.current
+                    with hub:
+                        try:
+                            self = current_thread()
+                            return old_run(self, *a, **kw)
+                        except Exception:
+                            reraise(*_capture_exception())
+                self.run = sentry_run
 
             return old_start(self, *a, **kw)  # type: ignore
 
         Thread.start = sentry_start  # type: ignore
 
 
-def _wrap_run(parent_hub, old_run):
-    def run(*a, **kw):
-        hub = parent_hub or Hub.current
-
-        with hub:
-            try:
-                return old_run(*a, **kw)
-            except Exception:
-                reraise(*_capture_exception())
-
-    return run
-
-
 def _capture_exception():
     hub = Hub.current
     exc_info = sys.exc_info()
```

GuangTianLi:
> > I fixed it by passing the__class__ of instance into the descriptor.
> 
> IIUC, this is still not thread safe. When you do this
> 
> ```python
>                 self.__class__.run = PatchedInstanceMethodDescriptor(
>                     self.__class__.run, hub_
>                 )
> ```
> 
> it affects the entire Thread class, not just the single instance which is intended.

In my opinion, pathing `Thread.start` has already affected the entire Thread class and it has nothing with thread safe. But due to an API of Threading module - `current_thread` will access current thread instance, we can avoid a reference cycle with an easier way.

untitaker:
I really like that last fix!

untitaker:
Thanks @GuangTianLi !

PR #437, opened 2019-07-17
Fix to patched Popen regarding iterable arguments

### LINKED ISSUE #436: Problem with Popen arguments passed as map
Greetings!

I am having an issue when using the package `sorl.thumbnail` on a Django project. That package produces thumbnails from images using system installed programs (called engines.)

When `sorl.thumbnail` makes a `Popen` call here:

https://github.com/jazzband/sorl-thumbnail/blob/22ccd9781462a820f963f57018ad3dcef85053ed/sorl/thumbnail/engines/convert_engine.py#L57

The arguments that are received in your patched `Popen` are not being interpreted correctly. This is the suspect of the problem:

https://github.com/getsentry/sentry-python/blob/2c0a2ea96882adf69742bd7e79e588564ff5b85c/sentry_sdk/integrations/stdlib.py#L143

Arguments are handled in such a way that the original `popen` crashes when trying to access `args[0]`, as the `args` parameter is interpreted (or actually becomes) an empty list.

In my opinion, the problem arises because in `sorl.thumbnail` the arguments to `Popen` are passed to a `map` function first, so that a `map` object is what gets passed as arguments. I am not sure what's the effect of a `map` object being received in your `*a` of the patched `Popen`.

Here's the relevant part of the traceback I am getting:

```
File "/mnt/d/Proyectos/ginseng/.venv/lib/python3.6/site-packages/sorl/thumbnail/base.py", line 131, in get_thumbnail
    thumbnail)
  File "/mnt/d/Proyectos/ginseng/.venv/lib/python3.6/site-packages/sorl/thumbnail/base.py", line 164, in _create_thumbnail
    default.engine.write(image, options, thumbnail)
  File "/mnt/d/Proyectos/ginseng/.venv/lib/python3.6/site-packages/sorl/thumbnail/engines/convert_engine.py", line 57, in write
    p = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
  File "/mnt/d/Proyectos/ginseng/.venv/lib/python3.6/site-packages/sentry_sdk/integrations/stdlib.py", line 165, in sentry_patched_popen_init
    return old_popen_init(self, *a, **kw)
  File "/usr/lib/python3.6/subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.6/subprocess.py", line 1248, in _execute_child
    executable = args[0]
IndexError: list index out of range
```

If I remove the Sentry integration altogether from the Django project, `sorl.thumbnail` is able to effectively call `Popen` with the "mapped" arguments, and everything works.

Thanks in advance!

lorinkoz:
Wrapped patched Popen retrieval of original arguments in `itertools.tee`, to preserve original iterators. Closes #436

### DIFF
--- a/sentry_sdk/integrations/stdlib.py
+++ b/sentry_sdk/integrations/stdlib.py
@@ -2,6 +2,7 @@ import os
 import subprocess
 import sys
 import platform
+from itertools import tee
 
 from sentry_sdk.hub import Hub
 from sentry_sdk.integrations import Integration
@@ -147,7 +148,7 @@ def _install_subprocess():
 
         # do not setdefault! args is required by Popen, doing setdefault would
         # make invalid calls valid
-        args = _init_argument(a, kw, "args", 0) or []
+        args = tee(_init_argument(a, kw, "args", 0) or [])
         cwd = _init_argument(a, kw, "cwd", 10)
 
         env = None


lorinkoz:
`itertools.tee` documentation warns about the considerable amount of space it can use. Not sure if this is to be worried about, as the thing we're iterating is the arguments to `Popen`, which is "empirically" known to be limited to a reasonable amount of data.

lorinkoz:
The other solution would be to convert `args` to a list, and then reinsert the list into `a`/`kw` in a process similar to what `_init_argument` does, but opposite.

untitaker:
Thanks for your PR! Unfortunately this introduces another bug, because `tee` returns two iterables and destroys the input. I hope this example illustrates what is happening:

```python
from itertools import tee

old_iter = iter((1, 2, 3))
new_iter = tee(old_iter)

# should: 1 2 3
# is: <itertools.tee object at 0x10b99b830> <itertools.tee object at 0x10b99b8c0>
print(" ".join(map(str, new_iter)))
```

lorinkoz:
Hmm, it worked on my end, but I might have not tested it thoroughly in all possible scenarios. Apologies, I'll leave the issue up to you, then. Thanks for your feedback :)

lorinkoz:
@untitaker If you can decide what's the best way to fix this, I could try doing it myself and submitting a new PR. I am not very fond of converting `args` to a list and reinserting it into the `a`/`kw`, but if that's the best approach, please let me know.

untitaker:
I really apprechiate this! I can't think of any other way than to insert
args back into the a/kw. Just not the `tee` approach.

lorinkoz:
Okay thanks, will submit a new PR. And totally sorry about the dumb `tee` approach.

PR #444, opened 2019-07-28
fix: types for logging integration args

sbdchd:
The sdk supports disabling the LoggingIntegration by passing `None` as
arguments; however, the current types require ints.

This behavior is noted in the docs:
https://docs.sentry.io/platforms/python/logging/#options

```
sentry_sdk.init(integrations=[LoggingIntegration(level=None, event_level=None)])
```

edit: when adding `# type: ignore` I found a second issue with the argument `integrations` which mypy recommended as typing as `Sequence` instead of `List`. This fixed the type error.

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -8,6 +8,7 @@ if MYPY:
     from typing import Type
     from typing import Dict
     from typing import Any
+    from typing import Sequence
 
     from sentry_sdk.transport import Transport
     from sentry_sdk.integrations import Integration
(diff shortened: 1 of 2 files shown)


untitaker:
What do you pass to integrations other than a list? I'd like to prevent people from passing in non-resettable iterators because I make no guarantees about how often I iterate over the sequence.

untitaker:
Nvm sequence seems fine for this.

bluetech:
The changes LGTM.

---

With regards to `List` vs `Sequence`, the rules I try to follow is:

For inputs, e.g. function arguments, prefer `Sequence`, because it accepts more types, and provides more guarantees to the caller, namely that the sequence is not mutated (`append`, `pop`, `del` and friends).

For outputs, give the specific type (e.g. `List`), so the caller can do whatever they want with it. One exception is that if the API wants to retain the option to return some other sequence type, then it should use a more general type.

Also, when taking a callback, the situation is reversed.

untitaker:
Thanks!

PR #446, opened 2019-08-01
Beam integration

tiopi:
Beam integration requires #439. New files include `beam.py`, `test_beam.py` and test additions to tox.

### DIFF
--- a/sentry_sdk/integrations/beam.py
+++ b/sentry_sdk/integrations/beam.py
@@ -11,7 +11,8 @@ from sentry_sdk.integrations import Integration
 from sentry_sdk.integrations.logging import ignore_logger
 
 WRAPPED_FUNC = "_wrapped_{}_"
-INSPECT_FUNC = "_inspect_{}"
+INSPECT_FUNC = "_inspect_{}" # Required format per apache_beam/transforms/core.py
+USED_FUNC = "_sentry_used_"
 
 
 class BeamIntegration(Integration):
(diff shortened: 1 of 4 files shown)


sentry_sdk/integrations/beam.py
```diff
@@ -0,0 +1,147 @@
...
+
+        ignore_logger("root")
+        ignore_logger("bundle_processor.create")
+
+        function_patches = ["process", "start_bundle", "finish_bundle", "setup"]
+        for func_name in function_patches:
+            setattr(
+                DoFn,
+                INSPECT_FUNC.format(func_name),
+                patched_inspect_process(DoFn, func_name),
```
Zylphrex: this no longer just patch the `_inspect_process` method, consider renaming it

sentry_sdk/client.py
```diff
@@ -74,23 +74,37 @@ class _Client(object):
     """
 
     def __init__(self, *args, **kwargs):
-        # type: (*Optional[str], **Any) -> None
+        # type: (*Any, **Any) -> None
```
untitaker: Not sure what happened here but this diff seems to be the result of merge conflict resolution and not actually intended.

sentry_sdk/integrations/beam.py
```diff
@@ -0,0 +1,147 @@
...
+        client = hub.client
+        event, hint = event_from_exception(
+            exc_info,
+            client_options=client.options,
+            mechanism={"type": "beam", "handled": False},
+        )
+        hub.capture_event(event, hint=hint)
+
+
+def raiseException(client):
```
untitaker: For consistency, could you use snake_case here as well?

tests/test_client.py
```diff
@@ -359,28 +350,6 @@ def callback(scope):
     assert not calls
 
 
-@pytest.mark.parametrize("debug", (True, False))
```
untitaker: This seems like a botched merge as well

PR #447, opened 2019-08-04
Fix detection of gevent threading.local monkey-patch

bluetech:
The `_` in `_threading` seems like a mistake and always returns False.
Removing it leads to the detection working properly:

```py
Python 3.7.4 (default, Jul 16 2019, 07:12:58)
[GCC 9.1.0] on linux
>>> import gevent.monkey
>>> gevent.monkey.is_object_patched('_threading', 'local')
False
>>> gevent.monkey.is_object_patched('threading', 'local')
False
>>> gevent.monkey.patch_all()
True
>>> gevent.monkey.is_object_patched('_threading', 'local')
False
>>> gevent.monkey.is_object_patched('threading', 'local')
True
```

### DIFF
--- a/tests/utils/test_contextvars.py
+++ b/tests/utils/test_contextvars.py
@@ -1,0 +1,58 @@
+import pytest
+import random
+import time
+
+import gevent
+
+
+from sentry_sdk.utils import _is_threading_local_monkey_patched
+
+
+def test_gevent_is_patched():
+    gevent.monkey.patch_all()
+    assert _is_threading_local_monkey_patched()
+
+
+def test_gevent_is_not_patched():
+    assert not _is_threading_local_monkey_patched()
+
+
+@pytest.mark.parametrize("with_gevent", [True, False])
+def test_leaks(with_gevent):
+    if with_gevent:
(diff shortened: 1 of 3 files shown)


bluetech:
We are using gevent and are seeing some data "leaking" between events. I hope this will fix the problem, but I did not verify it.

untitaker:
Thanks!

tests/utils/test_contextvars.py
```diff
@@ -0,0 +1,58 @@
...
+import time
+
+import gevent
+
+
+from sentry_sdk.utils import _is_threading_local_monkey_patched
+
+
+def test_gevent_is_patched():
+    gevent.monkey.patch_all()
```
bluetech: I wonder, does this not affect all of the subsequent tests in the test run?

untitaker: We run each test in its own subprocess because the SDK itself is already so heavy on monkeypatching.

bluetech: Ha, nice.

PR #454, opened 2019-08-07
feat: Capture spans for Celery's apply_async

dcramer:


dcramer:
Just throwing this up for visibility/feedback. I think it'd be a useful stock instrumentation, but maybe there are better attributes to expose as part of it.

PR #457, opened 2019-08-08
Add user event data from logging extra

alexmick:
While migrating from `raven` to `sentry-python` I noticed my usage of the `user` key no longer populated the field on the sentry interface :
```python
logger.critical("User error", extra=dict(user=dict(id=123, email="user@example.org")))
```

This PR is an attempt to bring back this behaviour by adding the `user` key to the captured `event`. 

I tried to do my best to respect codestyle and conventions but I'm sure this can be further improved. This is more a proof of concept, wasn't sure if it was better to open an issue or and PR so I chose the option with a bit of code 😊

### DIFF
--- a/sentry_sdk/integrations/logging.py
+++ b/sentry_sdk/integrations/logging.py
@@ -145,6 +145,14 @@ def _extra_from_record(record):
     }
 
 
+def _user_from_record(record):
+    # type: (LogRecord) -> Optional[Dict[str, None]]
+    user = vars(record).get("user")
+    if not isinstance(user, dict):
+        return None
+    return user
+
+
 class EventHandler(logging.Handler, object):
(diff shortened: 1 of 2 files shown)


untitaker:
Sorry, we categorically decided not to do this, here is the rationale: https://github.com/getsentry/sentry-php/pull/844#issuecomment-510120197

There is also a vague hint as to how a workaround could look like (there are multiple solutions, such as using `before_send` or `push_scope`). Let me know if you would like to have a usage example.

alexmick:
Thanks, I was unaware that it was by design ! 

You're right, a workaround with `push_scope` by overriding `_emit()` in `EventHandler` or a `before_send` function on the sdk init work fine 👌

PR #460, opened 2019-08-09
[wip] ref: Do not add PII to SQL queries

### LINKED ISSUE #459: Use unparsed query for descriptions from Django ORM
By using the parsed query we inject params which might be sensitive or PII (which then means we'd have to strip the whole description). Right now we dont have scrubbing on spans, but if it did it would make a bunch of them fairly useless.

I think we should either:

- Change the description to be 100% unparsed (so placeholder variables)
- Change the description to be [statement type] [tables] (SELECT table1, table2)

We can still keep parameters in the context and apply data scrubbing logic there.

untitaker:
Fix #459


untitaker:
Part of #459 now

PR #466, opened 2019-08-12
ref: Next iteration of span API

untitaker:


sentry_sdk/tracing.py
```diff
@@ -123,7 +132,31 @@ def __repr__(self):
...
+        # type: () -> Span
+        hub = self.hub or sentry_sdk.Hub.current
+
+        _, scope = hub._stack[-1]
+        old_span = scope.span
+        scope.span = self
+        self._context_manager_state = (hub, scope, old_span)
+        return self
+
+    def __exit__(self, ty, value, tb):
```
kamilogorek: I assume `value` is some kind of error here? Or exception parameter? Thats why it fails the transaction?

sentry_sdk/tracing.py
```diff
@@ -183,29 +221,77 @@ def to_traceparent(self):
...
 
-    def finish(self):
-        self.timestamp = datetime.now()
-        if self._finished_spans is not None:
-            self._finished_spans.append(self)
+    def finish(self, hub=None):
+        # type: (Optional[sentry_sdk.Hub]) -> Optional[str]
+        hub = hub or self.hub or sentry_sdk.Hub.current
+
+        if self.timestamp is None:
```
kamilogorek: I think it should short-circuit here, so same transaction cannot be sent multiple times.

```
if self.timestamp is not None:
  return
```

sentry_sdk/tracing.py
```diff
@@ -303,7 +391,8 @@ def record_http_request(hub, url, method):
                     span.set_data(k, v)
 
 
-def maybe_create_breadcrumbs_from_span(hub, span):
+def _maybe_create_breadcrumbs_from_span(hub, span):
+    # type: (sentry_sdk.Hub, Span) -> None
     if span.op == "redis":
         hub.add_breadcrumb(type="redis", category="redis", data=span._tags)
     elif span.op == "http" and span.is_success():
```
kamilogorek: You don't want failed http requests to be breadcrumbed as well?

sentry_sdk/tracing.py
untitaker: `value` is an exception object while `ty` is its type and `tb` a traceback object.

sentry_sdk/tracing.py
untitaker: I agree, let's do that

sentry_sdk/tracing.py
untitaker: The old Python SDK didn't do that, and I don't want to change this UI for now.

Issue #421, opened 2019-07-10
Enforce functools.wraps on all monkeypatches


mumumumu:
I'm using [`celery-once`](https://github.com/cameronmaske/celery-once) to make sure some tasks are only run one at a time. The lock is acquired in `apply_async` and released in `after_return`. 

When I use the `CeleryIntegration` I'm not seeing any of the locks being released after the task has finished running. I'm not very familiar with the inner workings of celery, but any ideas what's going on?

untitaker:
Do you see this bug in Raven-Python? Raven uses official Celery hooks while this SDK aggressively (for lack of a better word) monkeypatches Celery to capture exceptions happening in hooks, and to work around some bugs in Celery 3.

mumumumu:
I think I've narrowed it down to an issue with `celery-once`. The library uses the task name and call arguments from the `run` method (using `inspect.getcallargs) to generate the lock key. It generates the key both when acquiring and releasing the lock. It seems the arguments before and after the task gets patched are slightly different (the wrapper takes in optional args).

I don't think this is strictly an issue with the SDK, so I'll close this for now, thanks for looking at it though.

untitaker:
@mumumumu if it's a missing functools.wraps somewhere we can fix it in the SDK as well, no big deal

mumumumu:
Thanks @untitaker, that'd be much appreciated. Adding a `@wraps(f)` to the `_wrap_task_call` `_inner` function does fix the issue for me.

untitaker:
We should do this everywhere then. Monkeypatches are quite inconsistent across the codebase.

untitaker:
Since this is blocking you I'll just get a minimal fix out of the door.

Proper wrapping will likely take more effort (possibly using wrapt instead of functools).

PR #475, opened 2019-08-20
Py2 wrapped

cameronmaske:


untitaker:
Where do you need this?

cameronmaske:
@untitaker My apologies, I accidentally opened this PR and isn't ready yet.

I'm the author of [celery-once](https://github.com/cameronmaske/celery-once/tree/master/celery_once) and this PR is related to #421.

For context, behind the scenes, celery once attempts to [inspect the arguments of the task function](https://github.com/cameronmaske/celery-once/blob/master/celery_once/tasks.py#L110) to generate a "key" to lock the task. 
While #432 fixed the issue for Python 3, for Python 2, `getcallargs` can't inspect the original function, as the `__wrapped__` attribute was only added in [>3.2](https://docs.python.org/3/library/functools.html#functools.update_wrapper). This is an attempt to backport that functionality for Python 2.

untitaker:
Got it. Conceptually I would approve this PR. Could you add a test that loads and uses celery-once? Also please move this new `wraps` into `sentry_sdk.utils`.

Ping me when it's ready :)

cameronmaske:
Hi @untitaker the code should be ready for a review. 

I'm not sure exactly why the tests are failing on Travis failed, any insight?

image

image

tox.ini
```diff
@@ -32,6 +32,8 @@ envlist =
     {pypy,py2.7,py3.5,py3.6,py3.7,py3.8}-celery-{4.1,4.2,4.3}
     {pypy,py2.7}-celery-3
 
+    {pypy,py2.7,py3.5,py3.6,py3.7}-celery_once
```
untitaker: Can we flatten this into the regular celery tests?

tests/integrations/celery/conftest.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+
+pytest.importorskip("celery")
```
untitaker: I believe moving this to `__init__` will fix your test issues. I assume it breaks a certain build that makes pytest skip all tests (and return with exitcode 1 because it didn't run anything)

cameronmaske:
@untitaker Thanks for the advice on the tests! Your suggestion are implemented and this should be ready for re-review.

untitaker:
@cameronmaske I don't quite understand the test, but it still seems to pass if I revert all your changes in `sentry_sdk`. So I think it isn't effective in testing that the bug does not exist.

untitaker:
Also sorry for the slow review!

cameronmaske:
@untitaker Just to check, with the reverted changes, were the tests run on Python 3 or 2? From my understanding, the tests should still pass on Python 3 without the changes, but fail on 2. 

The tests are a bit obtuse, but they are checking that [backend](https://github.com/cameronmaske/celery-once/blob/master/BACKEND_GUIDE.rst) is called with a correctly generated key from the example's tasks name and arguments. 
i.e.
```
def dummy_task(x, y):
   pass
```
Should become
```
qo_tests.integrations.celery.test_celery_once.dummy_task_x-1_y-1
# < Name assigned by celery to task> . (<Argument Key > <Arugment Value>, <Argument Key> <Arugment Value>,....
```
In this case, I put a dummy backend (simpler too test), but normally, this would be a redis or file backend to check the locks.

untitaker:
I was running this:

```
git checkout origin/master sentry_sdk/
tox -e py2.7-celery-3
```

...since you only use celery-once in celery 3.

untitaker:
@cameronmaske any update on this? Right now the tests don't seem useful IMO

cameronmaske:
I need to investigate this further on my end. I'm currently travelling and away from my development environment, will follow up when I return later next week.

untitaker:
@Marcelo-Theodoro I would love it if somebody took over this PR and fixed merge conflicts + fixed the tests to actually fail when the bugfix is not applied

I can't answer the first question, this is something for @cameronmaske

cameronmaske:
@untitaker Sorry for the delayed follow up. Had a chance to test this, and it appears that switching from [`getcallargs` -> `signature`](https://github.com/cameronmaske/celery-once/pull/107/files) the issue `wraps` does not occur. I'm closing this PR.

PR #478, opened 2019-08-22
Allow support for user introspection for more than just Flask-Login

matthoskins1980:
There are multiple user management modules for Flask.  This merge allows for the user to define the flask user module, the user object that stores the userinfo, and a mapping from user object attributes to the sentry user_info dict.  Code example to use flask_jwt_extended, instead of flask_login (very similar module):

```python
sentry_sdk.init(
    dsn="https://foobar@sentry.io/000000",
    integrations=[
        FlaskIntegration(
            user_module="flask_jwt_extended",
            user_object="current_user",
            user_attr_mapping={
                "username": "email",
                "email": "email"
            }
        )
    ],
    send_default_pii=True
)
```

### DIFF
--- a/sentry_sdk/integrations/flask.py
+++ b/sentry_sdk/integrations/flask.py
@@ -16,6 +16,7 @@ if MYPY:
     from sentry_sdk.integrations.wsgi import _ScopedResponse
     from typing import Any
     from typing import Dict
+    from typing import Optional
     from werkzeug.datastructures import ImmutableTypeConversionDict
     from werkzeug.datastructures import ImmutableMultiDict
     from werkzeug.datastructures import FileStorage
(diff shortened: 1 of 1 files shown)


untitaker:
I would attach the user object to the `hints` object of the event. Then the user can copy additional attributes over in `before_send`:

```python
def before_send(event, hint):
    event['user']['foo'] = hint['flask_user'].foo
    return event
```

...or do some fancier logic such as calling a method.

sentry_sdk/integrations/flask.py
```diff
@@ -26,6 +28,12 @@
 except ImportError:
     flask_login = None
 
+if flask_login is None:
```
untitaker: Those imports are not necessary when you later use `importlib` anyway. I would like to avoid adding the functionality of providing your own custom module... You can always use `configure_scope()` and friends to add user data if you want to do that. Docs are misssing examples, probably.

matthoskins1980:
Okay - thanks for the feedback.  I'll close this PR and look at your suggestions.

PR #503, opened 2019-09-19
fix: Do not capture SQL params for now

untitaker:
We have memory usage issues for very large queries, because large params are held in memory (spans or breadcrumb buffer). Do not capture SQL params for now.

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -49,6 +49,7 @@ class ClientConstructor(object):
         # DO NOT ENABLE THIS RIGHT NOW UNLESS YOU WANT TO EXCEED YOUR EVENT QUOTA IMMEDIATELY
         traces_sample_rate=0.0,  # type: float
         traceparent_v2=False,  # type: bool
+        _experiments={},  # type: Dict[str, Any]
     ):
         # type: (...) -> None
         pass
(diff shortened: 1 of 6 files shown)


amureki:
Greetings @untitaker and thank you for taking care of python Sentry SDK!

Could we elaborate a bit on the issue? Is it written anywhere with some details?

One of my team's big Django applications is using now `sentry-sdk==0.7.14`, because from version `0.8.0` we are having constant request timeouts (tried versions `0.8`, `0.10`). Sadly I am still not able to reproduce it on test servers (probably depends on the app load) and can't understand the reason for it.

So I am wondering if our issue is related to what is described here and might be solved by this PR.

untitaker:
@amureki I have no clue... please create a new issue

amureki:
@untitaker I would create a new issue when I would be able to gather more details on our case.
But I was wondering, how did you find this case you are solving here? Was it one of the customers issue? How did it come up? I am just trying to match the symptoms here...

untitaker:
@amureki Without disclosing too much customer information, the issue that triggered this was a long running celery task that fired a lot of DB queries with massive params list each. In the specific case this created an OOM

Theoretically you can have a long-running web request that does the same thing but this seems unrelated. I assume this will generally improve memory usage though, but it doesn't seem like that's your problem.

I would create a new issue anyway. having a version range where it works and doesn't work might be good enough. Thanks

PR #513, opened 2019-09-25
allow sentry to correctly parse events with `datetime`s in them

clayjonesthomas:
We recently came up against a problem where we discovered an error sentry wouldn't report. Turns out, the event had an un-stringified datetime in its breadcrumbs, and sentry was failing to json-ify the event due to that, silencing the error.

### DIFF
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -108,6 +108,11 @@ class HttpTransport(Transport):
 
         self.hub_cls = Hub
 
+    @staticmethod
+    def _convert_datetime(event):
+        if isinstance(event, datetime):
+            return str(event)
+        
     def _send_event(
         self, event  # type: Event
     ):
@@ -119,7 +124,7 @@ class HttpTransport(Transport):
 
         body = io.BytesIO()
         with gzip.GzipFile(fileobj=body, mode="w") as f:
-            f.write(json.dumps(event, allow_nan=False).encode("utf-8"))
+            f.write(json.dumps(event, allow_nan=False, default=self._convert_datetime).encode("utf-8"))
 
         assert self.parsed_dsn is not None
         logger.debug(


untitaker:
Could you tell us more about how you initialize the SDK? We have a `Serializer` class that should take care of converting those types, and I don't understand how any event can bypass its call here: https://github.com/getsentry/sentry-python/blob/9f9311c4e6fc3e899eb525c025f5b58e84d654ba/sentry_sdk/client.py#L174

clayjonesthomas:
Interesting. Here's a sanitized version of our init statement: 
```sentry_sdk.init(
            dsn='...',
            integrations=[FlaskIntegration(), AwsLambdaIntegration(),],
            environment='prod',
            send_default_pii=True,
            before_send=before_sentry, # the method that can deterministically receive datetimes
            debug=False,
        )
```

The line you link is indeed getting executed, and at that point, the event is valid JSON. After some digging, it seems to be the scope getting added to the event that is adding in datetimes. the way we do that is pretty much like this: 
```
from sentry_sdk import configure_scope

def before_sentry(event, hint):
   # json.dumps(event) works
         event = scope.apply_to_event(event, {})
       # json.dumps(event) does not work due to datetimes in the breadcrumb added to scope
    return event
```

I can continue this discussion as an issue if you'd prefer.

untitaker:
I see, I believe that your before_send hook is adding those datetimes.

Scope data should go through the serializer.

untitaker:
Ah, sorry. You should not have to call apply-to-event at all. Could you
remove the before-send hook entirely? I do not understand what it does.

clayjonesthomas:
Sorry for the confusion, that was truncated code. We use the `before_send` hook to apply several context-dependent tags, like so:
```
from sentry_sdk import configure_scope

def before_sentry(event, hint):
    # more code here
        # more code here
         scope.set_tag('is_admin', True)
         event = scope.apply_to_event(event, {})
    return event
```

To be clear, our tag setting is not what's doing this; my minimal example above is all that's needed to get datetimes in the `breadcrumbs` of an event.

untitaker:
The tag setting not, but calling apply_to_event might do this. There should
never be a need to call this function yourself. If you want to set a tag
for all events I suggest to run this at import time:

```
    scope.set_tag(k, v)

```

clayjonesthomas:
I see, good to know as it definitely sounds like we're going about this a little wrong, then. 

How would you recommend we set a tag that is context-dependent, then? Sorry to keep misleading you with poor pseudocode; assume we are only setting tags that do indeed need to be set differently depending on each event.

untitaker:
@clayjonesthomas could you fully elaborate on your usecase? This may be more or less elegant depending on the concrete usecase.

clayjonesthomas:
Sure, our usecase is ~10 tags that are set based on the specific event. A couple examples pesudocoded out below:
```
from sentry_sdk import configure_scope

def before_sentry(event, hint):
    # more code here
        # add several request-level aws-related tags
        scope.set_tag('aws.lambda.request_id', event['extra']['lambda'].get('aws_request_id'))
        # add a flag to maintain if an event came from postman (ie, not our website)
        scope.set_tag('postman', 'false')
        for header in headers.keys():
            if 'postman' in header.lower():
                scope.set_tag('postman', 'true')
                break
         # about 8 more of this type of thing
         event = scope.apply_to_event(event, {})
    return event
```

untitaker:
@clayjonesthomas Ok, so you're in aws lambda? you can call configure_scope within your request handler (completely detached from the code that handles the error) and it should just work. The tags will then be applied to all errors happening within that request.

untitaker:
For example in Flask I would do this:

```
@app.route("/")
def index():
        scope.set_tag("foo", get_foo_value(flask.request))

    ...

    raise Exception("fail")
```

Since we integrate with Flask, we will make sure your tags do not leak between requests (we "push a new scope" before a request starts and "pop" it when it ends).

clayjonesthomas:
Gotcha, that makes sense for 90% of our use case, thanks for clarifying. 

We also have tags set based on the error itself, eg `scope.set_tag('thrown_module', thrown_module)`, where we do some parsing of the stacktrace to get the most recent in-house module in the trace. would you recommend we just edit the event manually in that case, avoiding `scope` entirely?

untitaker:
For this I would use before_send:

```
def before_send(event, hint):
    event.setdefault("tags", {})['thrown_module'] = ...
    return event
```

Using scopes in before_send wreaks havoc. I've seen a few people attempt that, I am considering preventing this on a technical level such that they don't shoot their foot. wdyt?

clayjonesthomas:
Works fine for us! Considering `scope` has unintended behavior in `before_send` I think it's quite reasonable to discourage its use, coming from someone with a hole in their foot :)

PR #519, opened 2019-10-01
Add PySpark Integration

AbhiPrasad:
~This is draft PR for the purposes of reviewing the work I have so far. Much of this is still in flux, and has to be extensively tested on different executions env like Databricks/Dataproc/YARN.~

We make available the integration for both the Spark driver as well as Spark workers (which may be in different clusters depending on the execution environment). In order to add the sentry-integration, you must install the `sentry-sdk`. 

In order to start sending Sentry errors from the main Spark driver, you can simply init Sentry before you init your `SparkContext` when writing your job.

```py
import sentry_sdk
from sentry_sdk.integrations.spark import SparkIntegration

if __name__ == "__main__":
    sentry_sdk.init("___PUBLIC_DSN___", integrations=[SparkIntegration()])
    spark = SparkSession\
        .builder\
        .appName("ExampleApp")\
        .getOrCreate()
    ...
```

You can also just use `StreamingContext` in case of SparkStreaming usage or just use `SparkContext`.

This will log errors to Sentry, building breadcrumbs using 
https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/scheduler/SparkListener.scala#L216

In order to instrument workers, additional steps will have to be taken. The configuration options used to instrument the workers are only available in Spark 2.4 and above.

First, you must install the `sentry-sdk` on your workers. This can be done locally be sending the `sentry-python` egg or zip to spark using the `--py-files` cmd line option or with `sc.addPyFile()`.

Next, we have to monkey patch the spark daemon to initialize Sentry on each worker. This can be done by creating a file with the following code. 

```py
import sentry_sdk
from sentry_sdk.integrations.spark import SparkWorkerIntegration
import pyspark.daemon as original_daemon

if __name__ == '__main__':
    sentry_sdk.init("___PUBLIC_DSN___", integrations=[SparkWorkerIntegration()])
    original_daemon.manager()
```

In order to tell Spark to replace the daemon, you must set some properties, either with a custom `SparkConf` or using the `--conf` cmd line option. These are `spark.python.use.daemon=true` and `spark.python.daemon.module=sentry_patched_worker_file_name`. You must also use `--py-files` to send `sentry_patched_worker_file_name.py` to each of the workers.

https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/api/python/PythonWorkerFactory.scala#L56-L63

Both the workers and drivers share a common `application_id`, so we can easily filter for which workers are associated with what drivers.

Next Steps:
- ~Test with Databricks~
- ~Test with Dataproc~
- ~Write tests on this branch~
- ~Update documentation~ https://github.com/getsentry/sentry-docs/pull/1275


sentry_sdk/integrations/spark/__init__.py
```diff
@@ -0,0 +1,4 @@
+from sentry_sdk.integrations.spark.spark_driver import SparkIntegration
+from sentry_sdk.integrations.spark.spark_worker import sentry_worker_main
+
+__all__ = ["SparkIntegration", "sentry_worker_main"]
```
untitaker: Why do we need to export `sentry_worker_main` for users to patch in? Why can't we patch `main` ourselves?

sentry_sdk/integrations/spark/spark_driver.py
```diff
@@ -0,0 +1,174 @@
...
+    # type: () -> None
+    from pyspark import SparkContext  # type: ignore
+    from pyspark.java_gateway import ensure_callback_server_started  # type: ignore
+
+    spark_context_init = SparkContext._do_init
+
+    def _sentry_patched_spark_context_init(self, *args, **kwargs):
+        init = spark_context_init(self, *args, **kwargs)
+
+        if not getattr(self, "_sentry_is_patched", False):
```
untitaker: Is the concern that `_do_init` is called multiple times? Are there possibilities that the method is called concurrently?

sentry_sdk/integrations/spark/spark_driver.py
```diff
@@ -0,0 +1,174 @@
...
+            # So workers have access to app_name and id
+            self.setLocalProperty("app_name", self.appName)
+            self.setLocalProperty("application_id", self.applicationId)
+
+            # Initialize Sentry Listener to start recording breadcrumbs
+            hub = Hub.current
+            gw = self._gateway
+            ensure_callback_server_started(gw)
+            listener = SentryListener(hub)
```
untitaker: Instead of passing the hub around, would it be possible to access `Hub.current` in each listener method?

sentry_sdk/integrations/spark/spark_driver.py
```diff
@@ -0,0 +1,174 @@
...
+
+def patch_spark_context_init():
+    # type: () -> None
+    from pyspark import SparkContext  # type: ignore
+    from pyspark.java_gateway import ensure_callback_server_started  # type: ignore
+
+    spark_context_init = SparkContext._do_init
+
+    def _sentry_patched_spark_context_init(self, *args, **kwargs):
+        init = spark_context_init(self, *args, **kwargs)
```
untitaker: Please early return here if `Hub.current.get_integration(SparkIntegration) is None` (such that one can disable the integration effectively.

sentry_sdk/integrations/spark/spark_worker.py
```diff
@@ -0,0 +1,69 @@
...
+
+    client_options = client.options
+    mechanism = {"type": "spark", "handled": False}
+
+    exc_info = exc_info_from_error(exc_info)
+
+    exc_type, exc_value, tb = exc_info
+    rv = []
+
+    for exc_type, exc_value, tb in walk_exception_chain(exc_info):
```
untitaker: Why do you explicitly build the error chain here?

sentry_sdk/integrations/spark/spark_driver.py
AbhiPrasad: The main concern was that _do_init could be called multiple times, and I did not want to open a py4j gateway and call `addSparkListener` multiple times

sentry_sdk/integrations/spark/__init__.py
AbhiPrasad: We can't patch main ourselves because the code has already been sent to the distributed executors by the time we patch `SparkContext` (which is the earliest we can do anything). As a result, we opt to set it as a spark config option and have the users create their own file. 

I know this means extra steps for our users, I am actively looking at alternate methods.

sentry_sdk/integrations/spark/spark_worker.py
AbhiPrasad: Sorry should have given you some more context here. 

When a PySpark worker throws an exception, it doesn't raise any exceptions

https://github.com/apache/spark/blob/master/python/pyspark/worker.py#L599-L617

Instead it writes to the JVM, where there is Scala code to handle the Python error (log/start new worker/kill application etc.) 

https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/api/python/PythonRunner.scala#L618-L619

So the exception handling path here is that the worker will write to the stream and then call `sys.exit(-1)`.

The reason I am explicitly building the error chain is that I want to remove the `SystemExit`, `EOFError`, and `ConnectionResetError` exception types. `SystemExit` will always exist as `sys.exit(-1)` is being called on worker shut down (so provide no useful information to the user), while `EOFError` and `ConnectionResetError` are noisy exceptions that the workers raise frequently (and also are not useful as a result).

sentry_sdk/integrations/spark/__init__.py
AbhiPrasad: Do you think we should also have a `SparkWorkerIntegration` in this case?

sentry_sdk/integrations/spark/spark_worker.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+
+    if rv:
+        rv.reverse()
+        hint = event_hint_with_exc_info(exc_info)
+        event = {"level": "error", "exception": {"values": rv}}
+
+        taskContext = TaskContext._getOrCreate()
+
+        with configure_scope() as scope:
```
untitaker: Can you do this not as part of exception catching, but on every invocation of main? This way errors captured by the user will also have this context data

sentry_sdk/integrations/spark/spark_worker.py
untitaker: For my understanding. Are those errors part of the same exception chain that contains legitimate errors? We filter out error types in other integrations as well but never had to do this

AbhiPrasad: Yes, these errors are part of the same exception chain that contains legitimate errors.

sentry_sdk/integrations/spark/spark_worker.py
AbhiPrasad: So in this case the TaskContext is only created when the `main` function of the worker is called.

https://github.com/apache/spark/blob/master/python/pyspark/worker.py#L502-L507

Not sure the best way to patch `main` to make sure user errors have this context data.

AbhiPrasad:
Update: This is probably ready for another look through.

I went through and refactored some of the driver/worker code. I also added tests for the driver integration.

For some more context - here is a WIP branch for the docs https://github.com/getsentry/sentry-docs/compare/abhi/pyspark-docs.

I'm not sure if I should be using spark or pyspark, thoughts?

sentry_sdk/integrations/spark/spark_worker.py
untitaker: This should work:

```
    @scope.add_event_processor
    def process_event(event, hint):
        context = TaskContext.get()
        event.setdefault("tags", {}).setdefault("stageId", context.stageId())
        ...
        return event
```

I.e. never use the `scope.set_tag`

untitaker:
@AbhiPrasad spark is probably fine

sentry_sdk/integrations/spark/spark_driver.py
```diff
@@ -0,0 +1,191 @@
...
+        # type: () -> None
+        patch_spark_context_init()
+
+
+def _set_app_properties():
+    from pyspark import SparkContext
+
+    sparkContext = SparkContext._active_spark_context
+    if sparkContext:
+        sparkContext.setLocalProperty("app_name", sparkContext.appName)
```
untitaker: Can we safely do this without interfering with user-set properties? Why do we need this?

AbhiPrasad: We set a local property so that the worker has access to "app_name" and "application_id".

```python
if "app_name" in taskContext._localProperties:
    scope.set_tag("app_name", taskContext._localProperties["app_name"])
    scope.set_tag(
        "application_id", taskContext._localProperties["application_id"]
    )
```

It might be a good idea to rename it to something more sentry specific so that it won't get changed by the user. `sentry_integration_app_name`?

sentry_sdk/integrations/spark/spark_worker.py
AbhiPrasad: What is the best way to test that the tags have been set correctly?

sentry_sdk/integrations/spark/spark_driver.py
untitaker: I don't understand why this is necessary. Couldn't you set them on the scope directly?

(also I would prefer it if you used an event processor here like in the other code sample I posted)

sentry_sdk/integrations/spark/spark_worker.py
untitaker: I would send an event and assert the tags have been set correctly. See other tests in `tests/integrations/flask`, for example.

sentry_sdk/integrations/spark/spark_driver.py
AbhiPrasad: https://github.com/apache/spark/blob/master/python/pyspark/context.py#L1015-L1020

https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/SparkContext.scala#L668-L685

The driver propagates these local properties to the workers, I'm essentially using this as a way to communicate between the driver process and worker process (because they are on different clusters). We set the local property with the driver integration so that the worker integration can access it.

If we don't use `setLocalProperty` in the driver, the worker integration cannot set tags for `application_id` and `app_name` (which makes it much harder to pair worker errors with driver errors by tag in Sentry itself). I have not found any other way to access this information with a worker.

Did change to use an event processor 👍

untitaker: Makes sense!

untitaker: Hmm, could you add a short comment that this is how you transport data over the wire

sentry_sdk/integrations/spark/spark_driver.py
```diff
@@ -46,18 +55,35 @@ def _sentry_patched_spark_context_init(self, *args, **kwargs):
...
+                event.setdefault("tags", {}).setdefault(
+                    "executor.id", self._conf.get("spark.executor.id")
+                )
+                event["tags"].setdefault(
+                    "spark.submit.deployMode", self._conf.get("spark.submit.deployMode")
+                )
+                event["tags"].setdefault(
+                    "driver.host", self._conf.get("spark.driver.host")
+                )
+                event["tags"].setdefault(
```
untitaker: Can you wrap this with `with capture_internal_exceptions()`? This is random access to an API we don't control that might break.

sentry_sdk/integrations/spark/spark_worker.py
```diff
@@ -58,25 +58,37 @@ def _capture_exception(exc_info, hub):
...
-                )
-
-            if "callSite.short" in taskContext._localProperties:
-                scope.set_extra(
-                    "callSite", taskContext._localProperties["callSite.short"]
-                )
+    with configure_scope() as scope:
+
+        @scope.add_event_processor
+        def process_event(event, hint):
```
untitaker: Same here, please just wrap the entire event processor in `capture_internal_exceptions`

tests/integrations/spark/test_spark.py
```diff
@@ -1,4 +1,5 @@
 import pytest
+import sys
 
 import pdb
```
untitaker: Unused import

AbhiPrasad:
Update: Addressed review comments, fix lint errors.

Not sure why build is failing though.

untitaker:
@AbhiPrasad you had both pyspark and spark in the tox.ini which caused it to run the main test suite within the pyspark build. It's a mistake that `test_transport` even ran as part of pyspark build.

harpaj:
Hi guys, I'm right now building this branch to try it on one of our pipelines. I don't have much time for testing unfortunately, it's a busy week, but I could at least tell you if I manage to set it up initially.

harpaj:
The general setup worked flawlessly! Managed to capture an error in a Arrow UDF with a proper Python Stacktrace - perfect, that's exactly what I was hoping for.

I didn't have much time to look into details, but I noticed one thing: 
An event that gets created on the workers does not include any of the tags set on the driver.
We are setting tags like this, after the initialisation of Sentry on the driver:
```python3
        scope.set_tag("source", args["source"])
        scope.set_tag("target", args["target"])
        scope.set_tag("run-id", args["run_id"])
```
In the event form the worker, these tags are missing
spark_worker_event

In the event from the driver, these tags (and also some additional ones that this integration is setting) are present
spark_driver_event

I guess this might just be a technical limitation by this setup, but I think it warrants a note in the documentation at least.

AbhiPrasad:
@harpaj thank you for your feedback! 

Any code executed in the main spark job outside of spark actions only run the driver, that's why only the driver tagged it, again as you said just a technical limitation of this approach (not much more we can do as the driver process and worker process are running on different clusters).

You might be able to tag inside the UDF itself, because that is serialized and sent to the worker, but I have not tested that yet. This might also add a lot of overhead and slow down the UDF too much.

tox.ini
```diff
@@ -153,7 +153,7 @@ deps =
     sqlalchemy-1.2: sqlalchemy>=1.2,<1.3
     sqlalchemy-1.3: sqlalchemy>=1.3,<1.4
 
-    spark: spark==2.4.4
+    spark: pyspark==2.4.4
```
untitaker: Sorry, yes, that was me.

Issue #220, opened 2018-12-19
AIOHTTP: Attach request body


untitaker:
Methods for reading the request body are async in AIOHTTP, but event processors are not. So right now we can't attach any request bodies to the event without severely affecting perf.

untitaker:
Everybody who voted with 👍 , please suggest how to read the body in a sync function. I don't really understand why aiohttp works the way it does. Other async frameworks don't make me await on something that is just `json.parse`

sivakov512:
Any solutions here?

untitaker:
@sivakov512 no, I would prefer if the community investigated how other code deals with aiohttp for reading bodies twice

EasyLovv:
@untitaker 
The aiohttp does a cache of the body in bytes on the first read call. [BaseRequest.read()](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_request.py#L558-L573)

EasyLovv:
@untitaker 
I could be out of the concept, by from the processor you could get a loop and create async task which will get the body and send it to sentry. something like this:
`from aiohttp import web
    import asyncio

    loop = asyncio.get_event_loop()

    async def async_task(request: web.Request):
        body = await request.read()
        # send to sentry

    def processor():
        loop = asyncio.get_event_loop()
        request: web.Request = None
        loop.create_task(async_task(request))


    async def async_func():
        processor()

    loop.run_until_complete(async_func())`

untitaker:
@EasyLovv that sounds a bit dangerous. I wonder if we should just grab `request._post` and `request._read_bytes` if it's already there, and just do nothing if it isn't.

EasyLovv:
@untitaker it is also an option, would be great to have at least that. But in my view, I can't see anything dangerous, we are just reading the request body, any unsupported behavior or type could be covered by try-catch and posted to sentry with some additional "Unsupported body" tag etc.

untitaker:
@EasyLovv my issue is that I need the request body as string in the processor. I can't just spawn a task and return. So I block for `async_task` in `processor`, which ends up behaving like a blocking sync operation in `async_func`. Not cool for performance.

untitaker:
@EasyLovv do you want to make a PR for this so people can try it out?

samuelcolvin:
[...]

PR #527, opened 2019-10-09
feat: aiohttp request body

REVIMI:
Partially fixes https://github.com/getsentry/sentry-python/issues/220
In case if the request body has been read at least once, we have it in the `request._read_bytes` attribute. It allows us simply access it from non-async code.

### DIFF
--- a/sentry_sdk/integrations/aiohttp.py
+++ b/sentry_sdk/integrations/aiohttp.py
@@ -156,6 +156,9 @@ def _capture_exception(hub):
     return exc_info
 
 
+BODY_NOT_READ_MESSAGE = "[Can't show request body due to implementation details.]"
+
+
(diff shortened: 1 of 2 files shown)


untitaker:
@REVIMI could you add the following:

* Some test in `tests/integrations/aiohttp`
* Some sort of placeholder value for body in case there is none, such that a user can distinguish between "there is no body" and "body was not read". I believe this is only a check of `request.body_exists`

REVIMI:
@untitaker `request.body_exists` returns `True` even if it was `GET` request without the `body`. I used `request.can_read_body`. It returns `True` if the request has some data in the `body` but we haven't tried to read it.

REVIMI:
I'll add tests a bit later

PR #530, opened 2019-10-14
Create Scope.set_extras

Marcelo-Theodoro:
The idea is to add a method that makes easy to add more than one `extra` at the same time.

### DIFF
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -38,8 +38,10 @@ __all__ = [
     "last_event_id",
     "start_span",
     "set_tag",
+    "set_tags",
     "set_context",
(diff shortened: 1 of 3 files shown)


untitaker:
Could you add this consistently for `set_tags` etc too? We already have this in the [JS SDK](https://github.com/getsentry/sentry-javascript) and I believe it would be good if it worked the same (so close!)

untitaker:
Also please add global methods like in #540

Marcelo-Theodoro:
Sure!!
I'll work on it tomorrow, is that ok?
Thanks.

untitaker:
Of course!

PR #540, opened 2019-10-23
Global scope setters

a-d-collins:
This merge would allow the current scope's level, user, tags, context, and extras to be modified at the global scope.

NOTE: This is a follow-up to pull-request #523 which attempted to allow push_scope() to function as a decorator.

### DIFF
--- a/sentry_sdk/hub.py
+++ b/sentry_sdk/hub.py
@@ -1,10 +1,9 @@
 import copy
-import functools
 import random
 import sys
 
 from datetime import datetime
-from contextlib import contextmanager
+from contextlib import ContextDecorator, contextmanager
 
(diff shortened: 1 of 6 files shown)


sentry_sdk/hub.py
```diff
@@ -483,8 +489,6 @@ def push_scope(  # noqa
 
         return _ScopeManager(self)
 
-    scope = push_scope
```
untitaker: I think this is fine btw. We never documented this method

untitaker:
Thanks!

a-d-collins:
Of course! And this looks like it will work well with #530. Once that PR is merged, the global set_extras and set_tags methods can be created.

PR #548, opened 2019-11-08
Trytond integration

n1ngu:
Sentry SDK integration for the Trytond ERP framework


n1ngu:
Waiting for Tryton community feedback https://discuss.tryton.org/t/sentry-sdk-integration/1925

n1ngu:
Regarding the failed checks:
- the Python-3.5 fail is due to a str vs bytes issue
- the Python-3.4 fail is due to the trytond>5.0 package not being distributed for that version
- the Python-2.7 and the PyPy failures are rather unexpected because AFAIU they were out of the contributed dependency matrix

If the community agrees, I would settle for the easy solution: only supporting python>=3.6

untitaker:
@n1ngu it's ok to remove `py3.4-trytond-5.0` from the matrix by having multiple lines (see django)

n1ngu:
I still can't understand why the trytond submodule tests are being run with the Python-2.7 and PyPy interpreters. What am I missing?

untitaker:
@n1ngu have a look at `tests/integrations/*/__init__.py` :)

sentry_sdk/integrations/trytond.py
```diff
@@ -0,0 +1,69 @@
...
+
+            @app.error_handler
+            def _(app, request, e):  # type: ignore
+                error_handler(e)
+
+        else:
+            app.error_handlers.append(error_handler)
+
+
+class TrytondSentryHandler(logging.NullHandler):
```
n1ngu: @untitaker I am very interested in the insights of sentry-python members about this kind of handlers.

If I think about it, this has nothing to do with any particular framework, yet I was not capable of finding anything similar in the documentation nor the code base.

No rush, I just need some light here.

n1ngu:
I'd like to invoke @pokoli the maintainer of a former trytond raven integration

pokoli:
@n1ngu do you have any documentation on how this can be run on trytond? Which are the current steps to run the integration? I will like to test it but I do not now how to configure it.

n1ngu:
> @n1ngu do you have any documentation on how this can be run on trytond? Which are the current steps to run the integration? I will like to test it but I do not now how to configure it.

Cleaning the repo before making the PR I removed my test script :cry: but it should be as easy as the wsgi script stub depicted here https://discuss.tryton.org/t/sentry-sdk-integration/1925 and loading it with your favorite wsgi server. As discussed in that very thread, that server can't be the trytond binary since the wsgi module can't be chosen.

Soon enough, I'll also share a logging configuration to test the hacky "not TrytondCronIntegration" present here as a logging handler.

sentry_sdk/integrations/trytond.py
untitaker: Where is this handler used? It seems to me its purpose is to be called with certain logger names that we cannot know statically, but I am not sure.

Ideally we would never call `init()` from an integration, as the integration would have to provide a way to pass additional configuration to `init()` if the user needs to do that (forwarding of `**kwargs` is not really great)

pokoli: IIUC this is not required if we use [Load wsgi middleware from configuration](https://bugs.tryton.org/issue8807)

untitaker: Please just monkeypatch the wsgi application instead of injecting config values

n1ngu: This handler is *NOT* for the wsgi application. This handler is *NOT* a SentrySDK integration. This handler is to fool custom commandline tools shipped with the Tryton framework, namely trytond-admin, trytond-cron and trytond-worker. I only tested it with trytond-cron.

In order not to write wrappers for every tool, we intend to inject a `sentry_sdk.init` call inside a null handler that should be instantiated once and plugged anywhere (it i s a NullHandler!). The following logconf file can be used:

```
[loggers]
keys = root

[handlers]
keys = console, sentry

[formatters]
keys =

[logger_root]
level = INFO
handlers = console, sentry

[handler_console]
class = StreamHandler
args = (sys.stdout,)

[handler_sentry]
class=sentry_sdk.integrations.trytond.TrytondSentryHandler
args = (
    'https://37a9a6ea994a4dfe81ef0f278ab779a1@sentry.io/1824945',
    ('list', 'of', 'silenced', 'logger', 'qualnames'))
```
Then run `trytond-cron --logconf logconf.cfg --database <dbname>`

I am sorry for the misunderstanding. Obviously this PR will include adequate documentation before it is integrated!

n1ngu:
So as to wrap it here, while the documentation of this integration is not written, a wsgi.py script is needed to test it:
```
# mywsgi.py
import sentry_sdk
import sentry_sdk.integrations.trytond

sentry_sdk.init(
    "https://key@sentry.io/project",
    integrations=[sentry_sdk.integrations.trytond.TrytondWSGIIntegration()]
)

from trytond.application import app as application

# WATCH-IT! trytond>=5.4 only:
from trytond.exceptions import TrytonException
from trytond.exceptions import UserError

@application.error_handler
def _(app, request, e):
    if isinstance(e, TrytonException):
        return
    else:
        event_id = sentry_sdk.last_event_id()
        data = UserError('Custom Message', f'{event_id}\n{e}')
        return app.make_response(request, data)
```
Then run it with `$ gunicorn mywsgi` or your favorite wsgi server.

@cedk stated that a built-in trytond mechanism to use arbitrary wsgi middleware is on the way. Soon enough it will be interesting to wrap this integration in a dummy middleware and ship it (also with adequate documentation) so that it can be used in the most trytonic way without a wsgi script.

untitaker:
In my eyes there should be no reason to call `init()` from an integration. This is something we want to enforce for all stable integrations as doing otherwise requires us to replicate the API of `init()` somewhere else. If people use your logging handler to initialize the SDK, how do they configure `before_send` or any other option that takes more complex arguments than a DSN?

sentry_sdk/integrations/trytond.py
n1ngu: @untitaker my point is that this could be a not-trytond-specific feature, shipped with the core of SentrySDK so that any framework or application could benefit from it to inject `sentry_sdk.init` calls in any tools that don't allow the execution of arbitrary start-up code but use the python logging configuration system.
My fear is that it resembles the behaviour of the deprecated raven package so I am very interested in your insights.

untitaker: I understand the desire for this shortcut but I don't believe it will work sufficiently. See usage of `before_send` and more advanced passing of arbitrary python objects to `init()`

n1ngu: The handler relies on the standard Sentry LoggingIntegration being initialized by default, which will already spy on logger calls. But any later call to this handler will do nothing since it inherits a standard NullHandler!

n1ngu:
> In my eyes there should be no reason to call `init()` from an integration. This is something we want to enforce for all stable integrations as doing otherwise requires us to replicate the API of `init()` somewhere else. If people use your logging handler to initialize the SDK, how do they configure `before_send` or any other option that takes more complex arguments than a DSN?

That's a good point. This handler should be understood not as a framework integration but as a mean to inject code in tools that won't allow it. Obviously, if the default sdk initialization does not suffice for a particular purpose, neither will this handler. If this is a concern and such a feature was not acceptable for the core sdk I will retire it from this PR so we can focus on the actual trytond wsgi integration above! (but I'll still vendor it in my deployments :grin:)

Regarding complex `init()` options: The "ignored_loggers" feature is already beyond my initial will, it is present because I felt it was extremely straightforward. With a similar schema, other `init()` options could be exposed, but obviously its complexity grows as with a need for a fair amount of documentation.

That's why I'd settle for the "use this only if the default `init()` suffices you" as an out-of-the-box mechanism to power people to use Sentry with not easily-hackable tools, even if the Sentry setup is not fully customizable.

n1ngu:
@untitaker Stated with different words: this handler could be shipped alongside the EventHandler and BreadcrumbHandler in the sentry_sdk.integrations.logging module. Not to power full Sentry customization but to ease a naive Sentry initialization in honor to your own documentation:
> Usually, you don’t need this. You can use this together with default_integrations=False if you want to opt into what the Sentry Python SDK captures. However, correctly setting up logging is difficult. Also, an opt-in approach to capturing data will miss errors you may not think of on your own.
( https://docs.sentry.io/platforms/python/logging/#handler-classes )
The alternative for this kind of not easily-hackable tools is a tedious logging config setup with the EventHandler, and still with the problem of "where to state the DSN"

untitaker:
@n1ngu I would much rather separate the discussion of alternative initialization methods from getting this integration merged. The former is more controversial imo

untitaker:
@n1ngu could you open a pr against https://github.com/getsentry/sentry-docs?

untitaker:
Example PR: https://github.com/getsentry/sentry-docs/pull/867

sentry_sdk/integrations/trytond.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self):  # type: () -> None
+        pass
+
+    @staticmethod
+    def setup_once():  # type: () -> None
+        def error_handler(e):  # type: (Exception) -> None
+            if isinstance(e, TrytonException):
+                return
+            else:
+                sentry_sdk.capture_exception(e)
```
untitaker: Instead of this, could you use `event_from_exception`? The reason is that then you can set a exception `mechanism` of trytond.

sentry_sdk/integrations/trytond.py
```diff
@@ -0,0 +1,35 @@
...
+    identifier = "trytond_wsgi"
+
+    def __init__(self):  # type: () -> None
+        pass
+
+    @staticmethod
+    def setup_once():  # type: () -> None
+        def error_handler(e):  # type: (Exception) -> None
+            if isinstance(e, TrytonException):
+                return
```
untitaker: Can we also return if the integration is not enabled

```
hub = Hub.current
if hub.get_integration(TrytondWSGIIntegration) is None:
    return
```

sentry_sdk/integrations/trytond.py
```diff
@@ -0,0 +1,35 @@
...
+import sentry_sdk.integrations
+
+from trytond.exceptions import TrytonException  # type: ignore
+from trytond.wsgi import app  # type: ignore
+
+
+# TODO: trytond-worker, trytond-cron and trytond-admin intergations
+
+
+class TrytondWSGIIntegration(sentry_sdk.integrations.Integration):
```
untitaker: If this is a WSGI app then it might be useful to wrap it in the WSGI middleware: https://docs.sentry.io/platforms/python/wsgi/

This will add request data to each crash

cedk: In such case, I think it will be good to prevent to store the password from the request date when the login endpoint if called.

untitaker: Sentry removes keys called password by default, does that work for you?

cedk: Good, it is the name used by Tryton.

untitaker: Yeah that should work. Our tech for data protection tech is something that is beyond integrations anyway (using `before_send` and serverside datascrubbing): https://docs.sentry.io/error-reporting/configuration/filtering/

n1ngu: I'll check it out and I'll also assert that passwords do not get leaked.

But this also raises a concern about leaking bearer authorization headers, the default auth mechanism for Tryton APIs. Can we set the integration to skip those by default?

Also, Tryton has a pluggable auth system so you can easily tweak the login. It can be used for 2FA with temporary secrets, but the default password login can be completely overridden by other systems that could request secrets in arbitrary keys (am I wrong, @cedk ?). I'd vote for completely masking request/response bodies for the Tryton login endpoint. Is that feasible?

untitaker:
I think it would be fine to omit the request data as well if you think it
cannot be useful in the context of trytond. I do not know what Tryton does

sentry_sdk/integrations/trytond.py
cedk: Yes the authentication mechanism can be completely customized. But I do not see why someone would use another key than 'password' to receive a secret password.
For temporary token like TOTP, I do not see big issue to get them logged as they are temporary.
But of course, if it is even better to not log any data from the login endpoint if it is doable.

sentry_sdk/integrations/trytond.py
n1ngu: I still don't understand that mechanism but I trust it will prove useful in the future. I'm about to upload the changes you requested.

sentry_sdk/integrations/trytond.py
n1ngu: Done. Please give it a look.

It does filter out the password field and as @cedk said, one time secrets shouldn't be a concern and other login systems shouldn't be using a field different than password

sentry_sdk/integrations/trytond.py
untitaker: I realize the API is suboptimal and should be part of `capture_exception`. However right now we make it a policy to set an exception [mechanism](https://docs.sentry.io/development/sdk-dev/event-payloads/exception/#exception-mechanism) for every integration so a user can see why an error was captured.

untitaker:
I tried to push this but failed bc the PR is not editable for me:

```diff
diff --git a/sentry_sdk/integrations/trytond.py b/sentry_sdk/integrations/trytond.py
index 152e879..c1a669a 100644
--- a/sentry_sdk/integrations/trytond.py
+++ b/sentry_sdk/integrations/trytond.py
@@ -4,10 +4,14 @@ import sentry_sdk.hub
 import sentry_sdk.utils
 import sentry_sdk.integrations
 import sentry_sdk.integrations.wsgi
+from sentry_sdk._types import MYPY
 
 from trytond.exceptions import TrytonException  # type: ignore
 from trytond.wsgi import app  # type: ignore
 
+if MYPY:
+    from typing import Any
+
 
 # TODO: trytond-worker, trytond-cron and trytond-admin intergations
 
@@ -25,12 +29,19 @@ class TrytondWSGIIntegration(sentry_sdk.integrations.Integration):
 
         def error_handler(e):  # type: (Exception) -> None
             hub = sentry_sdk.hub.Hub.current
+
             if hub.get_integration(TrytondWSGIIntegration) is None:
                 return
             elif isinstance(e, TrytonException):
                 return
             else:
-                (event, hint) = sentry_sdk.utils.event_from_exception(sys.exc_info())
+                # If an integration is there, a client has to be there.
+                client = hub.client  # type: Any
+                event, hint = sentry_sdk.utils.event_from_exception(
+                    sys.exc_info(),
+                    client_options=client.options,
+                    mechanism={"type": "trytond", "handled": False},
+                )
                 hub.capture_event(event, hint=hint)
 
         # Expected error handlers signature was changed
```

could you apply it manually?

sentry_sdk/integrations/trytond.py
untitaker: lgtm

sentry_sdk/integrations/trytond.py
```diff
@@ -0,0 +1,55 @@
...
+
+            if hub.get_integration(TrytondWSGIIntegration) is None:
+                return
+            elif isinstance(e, TrytonException):
+                return
+            else:
+                # If an integration is there, a client has to be there.
+                client = hub.client  # type: Any
+                event, hint = sentry_sdk.utils.event_from_exception(
+                    e,
```
n1ngu: Switched back to build the event from the caught exception instead of sys.exc_info() because it feels more elegant to me. Does using sys.exc_info() have any advantatge I am missing?

untitaker: Under Python 2 this will cause you to lose the traceback, but it should be fine in this context.

untitaker:
fwiw this is blocked on the doc changes

n1ngu:
My bad. I am aware the ball is in my court. I can only spend here some spare work hours, plus I've been on vacation. But hopefully I'll have the documentation sorted in the short term!

untitaker:
Gonna get this out and document later

PR #551, opened 2019-11-11
Fixed the serialization of byte-string objects in Python 3

### LINKED ISSUE #550: Bytes Objects serialized as list of integers in Python 3
Using `sentry_sdk` version 0.13.0 in Python 3.7.3, strings encoded as `bytes` objects are serialized as lists of integers:

```
>>> from sentry_sdk.serializer import serialize
>>> serialize(b'123abc')
[49, 50, 51, 97, 98, 99]
>>> serialize(b'123abc', should_repr_strings=True)
['49', '50', '51', '97', '98', '99']
```

This doesn't seem like the expected behavior, instead it could be:
```
>>> serialize(b'123abc')
'123abc'
>>> serialize(b'123abc', should_repr_strings=True)
"b'123abc'"
```

I was able to reproduce the issue with Python 3.7.3, 3.6.7, 3.5.6 and 3.4.10. The serialization seems to behave properly with Python 2.7.

xvillaneau:
Fixes #550

The serializer was returning lists of integers when attempting to serialize `bytes`. On top of that, the behavior of `sentry_sdk.utils.safe_repr` was decoding non-printable characters, and generally not behaving as expected. 

Both issues were fixed for Python 3. `safe_repr` still has the bug in Python 2, but the solution for that will not be straightforward.

### DIFF
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -36,10 +36,16 @@ if PY2:
     # Importing ABCs from collections is deprecated, and will stop working in 3.8
     # https://github.com/python/cpython/blob/master/Lib/collections/__init__.py#L49
     from collections import Mapping, Sequence
+
+    serializable_str_types = string_types
+
 else:
     # New in 3.3
     # https://docs.python.org/3/library/collections.abc.html
     from collections.abc import Mapping, Sequence
+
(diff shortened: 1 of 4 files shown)


untitaker:
Solid work, thanks!

PR #566, opened 2019-11-25
fix skipped None

syurchenko:
Please answer question in [Issue 567](https://github.com/getsentry/sentry-python/issues/567) to fix failing tests

### DIFF
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -275,23 +275,18 @@ def serialize(event, **kwargs):
                     break
 
                 str_k = text_type(k)
-                if v is None:
-                    rv_dict[str_k] = v
-                    i += 1
-                else:
-                    v = _serialize_node(
-                        v,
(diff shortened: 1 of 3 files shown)


sentry_sdk/serializer.py
```diff
@@ -275,19 +275,23 @@ def _serialize_node_impl(
...
-                    v,
-                    segment=str_k,
-                    should_repr_strings=should_repr_strings,
-                    is_databag=is_databag,
-                    remaining_depth=remaining_depth - 1
-                    if remaining_depth is not None
-                    else None,
-                    remaining_breadth=remaining_breadth,
-                )
-                if v is not None:
```
untitaker: Instead of doing this if-else, could you just add `v` here unconditionally? `_serialize_node` of `None` is `None`

syurchenko: thanks!

untitaker:
thanks!

PR #571, opened 2019-12-03
Fix 'asyncio.CancelledError' capturing for 'AioHttpIntegration'

nickgashkov:
This PR disables capturing of `asyncio.CancelledError` in `AioHttpIntegration`.

`aiohttp` will raise `asyncio.CancelledError` if clients interrupt request ([ref.](https://aiohttp.readthedocs.io/en/stable/web_advanced.html#web-handler-cancellation)) and this behaviour is considered normal.

In the meanwhile, such aborted requests are getting captured:

Screen Shot 2019-12-03 at 5 35 35 PM

### DIFF
--- a/sentry_sdk/integrations/aiohttp.py
+++ b/sentry_sdk/integrations/aiohttp.py
@@ -82,6 +82,8 @@ class AioHttpIntegration(Integration):
                         except HTTPException as e:
                             span.set_http_status(e.status_code)
                             raise
+                        except asyncio.CancelledError:
+                            raise
                         except Exception:
                             # This will probably map to a 500 but seems like we
                             # have no way to tell. Do not set span status.
--- a/sentry_sdk/integrations/aiohttp.py
+++ b/sentry_sdk/integrations/aiohttp.py
@@ -83,6 +83,7 @@ class AioHttpIntegration(Integration):
                             span.set_http_status(e.status_code)
                             raise
                         except asyncio.CancelledError:
+                            span.set_status("cancelled")
                             raise
                         except Exception:
                             # This will probably map to a 500 but seems like we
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/aiohttp.py
```diff
@@ -82,6 +82,8 @@ def setup_once():
                         except HTTPException as e:
                             span.set_http_status(e.status_code)
                             raise
+                        except asyncio.CancelledError:
```
untitaker: Can you invoke `span.set_status("cancelled")` here?

nickgashkov: Sure thing.

tests/integrations/aiohttp/test_aiohttp.py
```diff
@@ -120,6 +123,28 @@
...
+    async def hello(request):
+        raise asyncio.CancelledError()
+
+    app = web.Application()
+    app.router.add_get("/", hello)
+
+    events = capture_events()
+    client = await aiohttp_client(app)
+
+    with suppress(ServerDisconnectedError):
```
untitaker: please use `pytest.raises` here unless you are saying this error does not happen all the time

nickgashkov: I've used `suppress` to avoid testing `aiohttp` here.

At the moment, `ServerDisconnectedError` is always raised, but this behavior can be changed in the future and will cause tests to fail even though `sentry-python` operates normally.

But, if you'd like to use `pytest.raises` instead, let me know :)

untitaker: ok, let's stick with this

untitaker:
Thanks!

PR #572, opened 2019-12-04
fix(apm): Continuation of a trace from another span should use its span id as the parent span id

dashed:
Correction based on what I see on the JS SDK:

https://github.com/getsentry/sentry-javascript/blob/01cac4ff6c09d7cf4b5e9d6cf595b095a8d036f3/packages/apm/src/span.ts#L202-L207

------

The incorrect mapping (red arrows) I found from my local development. The blue arrow is the supposedly correct mapping.

Screen Shot 2019-12-03 at 11 04 51 PM copy

## TODO

- [x] test?

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -236,7 +236,7 @@ class Span(object):
         else:
             sampled = None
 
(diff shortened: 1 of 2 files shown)


untitaker:
Thx

PR #573, opened 2019-12-04
feat: Instrument ASGI under Django 3.0

untitaker:
Django 3.0 has been released and we need this for proper scope management, less leaks and better request data.


sentry_sdk/integrations/django/__init__.py
```diff
@@ -314,6 +316,30 @@ def sentry_patched_asgi_handler(self, receive, send):
...
+    if not HAS_REAL_CONTEXTVARS:
+        # We better have contextvars or we're going to leak state between
+        # requests.
+        #
+        # We cannot hard-raise here because Django may not be used at all in
+        # the current process.
+        logger.warning(
+            "We detected that you are using Django 3. To get proper "
+            "instrumentation for ASGI requests, the Sentry SDK requires "
+            "Python 3.7+ or the aiocontextvars package from PyPI."
```
rhcarvalho: Is it enough to `import aiocontextvars` for side-effects in https://github.com/getsentry/sentry-python/blob/85c15e2e6d9707e43a412f0b49a530aab8925624/sentry_sdk/utils.py#L741-L759?

I'm new to that, please bear with me :)

sentry_sdk/integrations/django/asgi.py
```diff
@@ -0,0 +1,30 @@
+"""
+Instrumentation for Django 3.0
+
+Since this file contains `async def` it is conditionally imported depending on
+Django version.
```
rhcarvalho: Does this refer to https://github.com/getsentry/sentry-python/pull/573/commits/adb3fd3d770a1db1e4c62ebc8f3a7e1d41391940#diff-f7309d947f9ef191e39296742d71f72aR322?

If so, that's less so about the Django version and more about the existence of an import path.

sentry_sdk/integrations/django/__init__.py
```diff
@@ -314,6 +316,30 @@ def sentry_patched_asgi_handler(self, receive, send):
...
+        )
+
+    from sentry_sdk.integrations.django.asgi import patch_django_asgi_handler_impl
```
rhcarvalho: In a wrong combination of Python version and Django, could this potentially blow up?
Maybe wrap this in a `try... except SyntaxError`?

tests/integrations/django/asgi/test_asgi.py
```diff
@@ -1,16 +1,24 @@
...
+from tests.integrations.django.myapp.asgi import channels_application
 
+APPS = [channels_application]
+if django.VERSION >= (3, 0):
+    from tests.integrations.django.myapp.asgi import asgi_application
 
+    APPS += [asgi_application]
+
+
+@pytest.mark.parametrize("application", APPS)
```
rhcarvalho: Why add this now when there is only one (unused) parameter?

sentry_sdk/integrations/django/asgi.py
untitaker: I am referring to the conditional import of *this* file in `sentry_sdk.integrations.django.__init__`. This file is invalid syntax on Python 2.

sentry_sdk/integrations/django/__init__.py
untitaker: I don't think this can blow up because `django.core.handlers.asgi` already would contain invalid syntax (and only exists in Django 3), so the first few lines in this function will return early

fwiw we do test all minor versions of Django in tox.ini, in all Python versions that work with the SDK and Django.

tests/integrations/django/asgi/test_asgi.py
untitaker: The `application` used to be imported from `tests.integration.django.myapp.asgi` at the top of this file and is now a pytest parameter. It is used in line 25 of this file which didn't change and is therefore outside of this diff.

`APPS` contains all the ASGI apps we want to test. If we are running with Django 2 installed, it will contain only one item, but if Django 3 is installed, it will contain both `asgi_application` and `channels_application`.

In `tox.ini` we install a lot of different Django versions and here we try to figure out what we can actually test under that Django version.

sentry_sdk/integrations/django/__init__.py
untitaker: `aiocontextvars` is a third-party package that offers a stand-in replacement for Python 3.7's `contextvars` in the stdlib. It is generally preferred over `threading.local` because it actually works with asyncio, *however*, in applications that use gunicorn/gevent/eventlet, `threading.local` works better than either contextvars or aiocontextvars. The reason for *that* is because gevent monkeypatches the threading module to behave correctly, but is not doing the same thing for contextvars or aiocontextvars.

This is why this logic became that contrived. We test this by having multiple virtualenvs with gevent/eventlet installed/not installed and initialized/not initialized.

untitaker: note: check if aiocontextvars actually works

sentry_sdk/integrations/django/asgi.py
untitaker: note: clarify what the actual condition is

sentry_sdk/integrations/django/__init__.py
untitaker: It did work, but is unreadable. Follow up is https://github.com/getsentry/sentry-python/pull/574

sentry_sdk/integrations/django/asgi.py
untitaker: updated, thanks!

sentry_sdk/integrations/django/asgi.py
```diff
@@ -0,0 +1,31 @@
...
+if MYPY:
+    from typing import Any
+
+
+def patch_django_asgi_handler_impl(cls):
+    # type: (Any) -> None
+    old_app = cls.__call__
+
+    async def sentry_patched_asgi_handler(self, scope, receive, send):
+        # type: (Any, Any, Any, Any) -> Any
```
rhcarvalho: I'm not very familiar with the type annotations. What's the gain of doing it in this file with every type being `Any`?

untitaker: That's just my way of opting out of type annotations. I just feel like it's not worth the effort here bc of the interaction with django, a module unknown to mypy.

rhcarvalho: What would be the difference if this line is not there altogether?

untitaker: I configured mypy to complain about missing type annotations. This way a contributor is forced to think about types of parameters but can still opt out relatively easily. If mypy wasn't configured that way it would assume all parameters and return values to be `Any` on functions that do not have type annotations. This way it's at least explicit.

rhcarvalho: Thanks for explaining :)

Issue #568, opened 2019-11-26
Sentry for Spark on Databricks


kutny:
Hi,

I wonder how to correctly integrate Sentry with **Databricks** https://databricks.com/ (Spark as a Service from the Spark founders).

I've spent some time trying to integrate Sentry with the Databricks IPython-base shell, but still not sure if it works correctly. I was not able to integrate SparkWorkerIntegration to the Spark worker nodes and also the Spark Streaming errors are not caught to Sentry...

Has anyone any experience with the Databricks integration?

Thanks

untitaker:
cc @AbhiPrasad

AbhiPrasad:
Hey thanks for reaching out @kutny.

We've done some early testing with Databricks when we initially developed the Spark integration. 

Can you provide some more details on your cluster setup? What Spark Config and client libraries (python packages) does the cluster use? Also, are you running an initialization script of any kind?

I'm also interested in hearing about your use case with Databricks in more detail. Do you mind reaching out to me at aprasad@sentry.io so we can chat about it?

kutny:
Hi @AbhiPrasad, sure! We use the newest Databricks Runtime (DBR) 6.1 with Python 3 only for both batch processing and streaming.

As described in [your documentation](https://blog.sentry.io/2019/11/12/sentry-for-data-error-monitoring-with-pyspark), I tried to initialize Sentry before creating the SparkContext/SparkSession. I found it pretty challenging as when I initialized Sentry in the first cell of the Databricks notebook, it is already too late as SparkSession is AFAIK initialized before the initialization of the notebook environment.

So I tried to monkey-patch the /dbfs/tmp/[random-timestamp]/PythonShell.py library, which handles the SparkSession initialization by adding the following code at the very beginning of the PythonShell.py file:

```
import os, sys, subprocess
subprocess.run([sys.executable, '-m', 'pip', 'install', 'sentry-databricks'], check=True)
from sentry_databricks.notebook import initSentry
initSentry(os.environ['SENTRY_DSN'], os.environ['APP_ENV'], [os.environ['APP_ROOT_NAMESPACE']])
```

As you can see I have created the [sentry-databricks python package](https://pypi.org/project/sentry-databricks/) to simplify the Sentry initialization from Databricks.

After some testing, it works well even with the SparkWorkerIntegration enabled, see my Databricks config:
[...]

PR #578, opened 2019-12-19
ref(spark): Track SparkContext after init

AbhiPrasad:
## Caution 🚧 - WIP
Hey I'm back :). As Mike says, you can leave Sentry, but you can never leave open-source.

Alright so this PR primarily addresses #568, specifically the problem with initializing Sentry after creating a `SparkContext`

Problem: When using Databricks shell (kinda like a running a Jupyter notebook on top of a Spark cluster), the SparkContext is already instantiated. This means that we can't monkey patch SparkContext init anymore.

Fix: I now check to see if an active spark context exists, if it does, I forgo monkeypatching the init and just use that active spark context
```python
sparkContext = SparkContext._active_spark_context
if sparkContext:
    _set_sentry_metadata(sparkContext)
else:
    patch_spark_context_init()
```

As a reminder, `patch_spark_context_init` patches spark's `_do_init` function -> https://github.com/apache/spark/blob/master/python/pyspark/context.py#L129

cc @kutny @andyzzhao


AbhiPrasad:
cc @untitaker - Still WIP, but I would appreciate some guidance on where to go for this:

> There seems to be a problem though, the add event processor is not working (failing test below)

> https://github.com/getsentry/sentry-python/pull/578/files#diff-788d0bac84002e50dec8ea05b94f9ea4R243-R245

> When I create a SparkContext before initing Sentry, the event processor does not seem to get added and I'm not sure why.

> Any idea where where I can start debugging this?

dashed:
@AbhiPrasad welcome back.

untitaker:
I don't know what will happen to this but I don't see anybody driving this.

PR #593, opened 2020-01-17
fix: Enforce max queue length in transport

### LINKED ISSUE #586: Memory leak introduced in 0.12.3
Dear maintainers,

I recently noticed that my application containers started going out of memory fairly quickly. Upon investigation, I found that recent upgrade of sentry-sdk was causing this.

Tested with a few versions of sentry-sdk using Python 3.7.6:
sentry-sdk==0.12.2 -> memory usage stays at ~70 MiB, does not raise

sentry-sdk==0.12.3
sentry-sdk==0.13.5
sentry-sdk @ master
-> memory usage starts from ~70MiB and raises steadily until reaching memory limit and crashes container.

Please see if you can reproduce this behaviour. If not, I'll try to put up a minimal reproducible example, but this should be fairly simple to see. My application does not do anything special really, just receives some millions of messages over MQ and out of those, some few thousands end up getting logged to Sentry due to exceptions occurring. This is using pretty much the default Sentry config, no integrations.

Thanks for your help!

untitaker:
fix #586

### DIFF
--- a/sentry_sdk/worker.py
+++ b/sentry_sdk/worker.py
@@ -86,10 +86,14 @@ class BackgroundWorker(object):
 
     def kill(self):
         # type: () -> None
+        """
+        Kill worker thread. Returns immediately. Not useful for
+        waiting on shutdown for events, use `flush` for that.
+        """
         logger.debug("background worker got kill request")
         with self._lock:
             if self._thread:
-                self._queue.put(_TERMINATOR)
+                self._queue.put_nowait(_TERMINATOR)
                 self._thread = None
                 self._thread_for_pid = None
 
--- a/sentry_sdk/worker.py
+++ b/sentry_sdk/worker.py
@@ -93,7 +93,11 @@ class BackgroundWorker(object):
         logger.debug("background worker got kill request")
         with self._lock:
             if self._thread:
-                self._queue.put_nowait(_TERMINATOR)
+                try:
+                    self._queue.put_nowait(_TERMINATOR)
+                except queue.Full:
+                    logger.debug("background worker queue full, kill failed")
+
                 self._thread = None
                 self._thread_for_pid = None
 
--- a/sentry_sdk/worker.py
+++ b/sentry_sdk/worker.py
@@ -22,7 +22,7 @@ class BackgroundWorker(object):
     def __init__(self):
         # type: () -> None
         check_thread_support()
-        self._queue = queue.Queue(-1)  # type: Queue[Any]
+        self._queue = queue.Queue(30)  # type: Queue[Any]
         self._lock = Lock()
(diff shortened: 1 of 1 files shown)


sentry_sdk/worker.py
```diff
@@ -89,7 +89,7 @@ def kill(self):
         logger.debug("background worker got kill request")
         with self._lock:
             if self._thread:
-                self._queue.put_nowait(_TERMINATOR)
+                self._queue.put(_TERMINATOR)
```
untitaker: Not sure about this part. I think it's legit to block a little more on flush/close, even beyond the timeout, if the queue is completely full.

rhcarvalho: (sorry, fat fingered the resolve button on my phone)

Other SDKs have a timeout on operations like this. Is kill used like "close"?

untitaker: `kill` is an internal method that is supposed to be nonblocking. `close` calls `flush` with a timeout and then `kill`. I will actually revert this change for now.

rhcarvalho:
Running some tests locally :eyes:

sentry_sdk/worker.py
```diff
@@ -86,6 +86,10 @@ def start(self):
 
     def kill(self):
         # type: () -> None
+        """
+        Kill worker thread. Returns immediately. Not useful for
+        waiting on shutdown for events, use `flush` for that.
+        """
```
rhcarvalho: In theory now `kill` may raise `queue.Full`.

I'll consider people should never call this directly, but instead use `Hub.current.client.close()`, which does:

```python
           self.flush(timeout=timeout, callback=callback)
           self.transport.kill()
```

The first `flush` should empty the queue (may not empty if the timeout is reached), then `transport.kill()` should be able to put the `_TERMINATOR`.

Since there are some "shoulds", how about we handle `queue.Full` explicitly? In that case we can log a debug message. I don't see a clear way out without changing the current termination mechanism.

Thoughts?

sentry_sdk/worker.py
```diff
@@ -22,7 +22,7 @@ class BackgroundWorker(object):
     def __init__(self):
         # type: () -> None
         check_thread_support()
-        self._queue = queue.Queue(-1)  # type: Queue[Any]
+        self._queue = queue.Queue(30)  # type: Queue[Any]
```
rhcarvalho: :+1:

sentry_sdk/worker.py
untitaker: Yeah I did the same thing as with the other put_nowait now

untitaker:
Thanks!

untitaker:
CI is unrelated to this build

PR #609, opened 2020-01-25
fix: Update to pytest 4

untitaker:
we can't upgrade to pytest 5 yet because of an odd interaction with Flask.


untitaker:
Also can't update to pytest 4 because of old pytest-django, because of old django. Oh well!

PR #618, opened 2020-02-10
feat: add cloudwatch logs URL to lambda additional data

flyinbutrs:
As discussed on discord, this adds an additional "extra" for cloudwatch logs URL, log group, and log stream.

image

Not sure what if any tests are required for this change. Thoughts?

### DIFF
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -1,3 +1,5 @@
+from datetime import datetime, timedelta
+from os import environ
 import sys
 
 from sentry_sdk.hub import Hub, _should_send_default_pii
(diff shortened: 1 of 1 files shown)


sentry_sdk/integrations/aws_lambda.py
```diff
@@ -214,3 +224,32 @@ def _get_url(event, context):
...
+
+    Arguments:
+        context {LambdaContext} -- context from lambda handler
+
+    Returns:
+        str -- AWS Console URL to logs.
+    """
+    start_time_str = (start_time - timedelta(seconds=1)).isoformat().split(".")[0]
+    end_time_str = (datetime.now() + timedelta(seconds=2)).isoformat().split(".")[0]
+    url = ''.join([
```
untitaker: Instead of this please do

```
url = (
    "string 1"
    "string 2"
)
```

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -214,3 +224,32 @@ def _get_url(event, context):
...
+    """
+    Generates a CloudWatchLogs console URL based on the LambdaContext object
+
+    Arguments:
+        context {LambdaContext} -- context from lambda handler
+
+    Returns:
+        str -- AWS Console URL to logs.
+    """
+    start_time_str = (start_time - timedelta(seconds=1)).isoformat().split(".")[0]
```
untitaker: would prefer a strftime with explicit args here

untitaker:
oh yeah please also add tests for this

sentry_sdk/integrations/aws_lambda.py
flyinbutrs: That syntax will only work with quoted strings, half of these are variables/functions. I can do a multiline string with `.format` at the end, is that what you want?

sentry_sdk/integrations/aws_lambda.py
flyinbutrs: Ok, why? Happy to do it, just curious. could also do this:

`.replace(microsecond=0).isoformat()` if you object to the `.split('.')[0]`

sentry_sdk/integrations/aws_lambda.py
flyinbutrs: (fixed)

untitaker: yeah sorry I didn't even think about that. `.format` is good

sentry_sdk/integrations/aws_lambda.py
untitaker: just personal preference

untitaker:
Awesome, thanks!

flyinbutrs:
Oh, uh... sorry, wasn't done. I just pushed up the tests.

untitaker:
Damn, sorry. Could you PR those separately?

PR #625, opened 2020-02-18
feat: Auto-enabling integrations behind feature flag

untitaker:
The idea is that you can write `init(_experiments={"auto_enabling_integrations": True})` to get maximal instrumentation.

Unsolved questions:

* How to opt out of integrations?

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -107,7 +107,7 @@ class _Client(object):
                 self.options["integrations"],
                 with_defaults=self.options["default_integrations"],
                 with_auto_enabling_integrations=self.options["_experiments"].get(
-                    "auto_enabling_integrations"
+                    "auto_enabling_integrations", False
                 ),
             )
         finally:
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -106,6 +106,9 @@ class _Client(object):
             self.integrations = setup_integrations(
                 self.options["integrations"],
                 with_defaults=self.options["default_integrations"],
+                with_auto_enabling_integrations=self.options["_experiments"].get(
+                    "auto_enabling_integrations"
+                ),
             )
         finally:
             _client_init_debug.set(old_debug)
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -58,7 +58,7 @@ class SanicIntegration(Integration):
                 " or aiocontextvars package"
             )
 
-        if VERSION.startswith("0.8."):
+        if SANIC_VERSION.startswith("0.8."):
             # Sanic 0.8 and older creates a logger named "root" and puts a
             # stringified version of every exception in there (without exc_info),
             # which our error deduplication can't detect.
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -45,7 +45,7 @@ class SanicIntegration(Integration):
         try:
             version = tuple(map(int, SANIC_VERSION.split(".")))
         except (TypeError, ValueError):
-            raise DidNotEnable("Unparseable Sanic version: {}".format(VERSION))
+            raise DidNotEnable("Unparseable Sanic version: {}".format(SANIC_VERSION))
 
         if version < (0, 8):
             raise DidNotEnable("Sanic 0.8 or newer required.")
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -28,7 +28,7 @@ if MYPY:
     from sentry_sdk._types import Event, EventProcessor, Hint
 
 try:
-    from sanic import Sanic, __version__ as VERSION
+    from sanic import Sanic, __version__ as SANIC_VERSION
     from sanic.exceptions import SanicException
     from sanic.router import Router
     from sanic.handlers import ErrorHandler
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -9,14 +9,9 @@ from sentry_sdk.utils import (
     event_from_exception,
     HAS_REAL_CONTEXTVARS,
 )
-from sentry_sdk.integrations import Integration
+from sentry_sdk.integrations import Integration, DidNotEnable
 from sentry_sdk.integrations._wsgi_common import RequestExtractor, _filter_headers
 from sentry_sdk.integrations.logging import ignore_logger
-
-from sanic import Sanic, __version__ as VERSION
-from sanic.exceptions import SanicException
-from sanic.router import Router
-from sanic.handlers import ErrorHandler
 
 from sentry_sdk._types import MYPY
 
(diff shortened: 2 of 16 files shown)


rhcarvalho:
:eyes: 

> to get maximal instrumentation.

@untitaker does that mean the integrations that it enables may change over time?

> How to opt out of integrations?

Good question. What's the motivation for this new option? The PR description gave the what, we're missing the why.

untitaker:
@rhcarvalho  This was asked for in the context of APM where people would have to enable a lot of small integrations to get a meaningful span tree. Generally it's nice if people have to think about as little as possible (so, not about which integrations are necessary) when enabling the SDK.

The semver compatibility is another thing. Yes, you could upgrade to a new version of the SDK and just get more integrations enabled automatically. Similar effects were already observable to some degree as we added more features to integrations that the user already explicitly enabled (breadcrumbs for django sql queries for example). I think we will just avoid this problem and make sure that new integrations or changes to existing ones don't break fundamental things like grouping, ever (like we already do)

sentry_sdk/client.py
```diff
@@ -106,6 +106,9 @@ def _init_impl(self):
             self.integrations = setup_integrations(
                 self.options["integrations"],
                 with_defaults=self.options["default_integrations"],
+                with_auto_enabling_integrations=self.options["_experiments"].get(
+                    "auto_enabling_integrations"
```
rhcarvalho:
```suggestion
                    "auto_enabling_integrations", False
```

sentry_sdk/integrations/sanic.py
```diff
@@ -32,17 +27,33 @@
 
     from sentry_sdk._types import Event, EventProcessor, Hint
 
+try:
+    from sanic import Sanic, __version__ as VERSION
```
rhcarvalho:
```suggestion
    from sanic import Sanic, __version__ as SANIC_VERSION
```

For consistency only.

sentry_sdk/integrations/sanic.py
```diff
@@ -32,17 +27,33 @@
...
+
 
 class SanicIntegration(Integration):
     identifier = "sanic"
 
     @staticmethod
     def setup_once():
         # type: () -> None
+        try:
+            version = tuple(map(int, VERSION.split(".")))
```
rhcarvalho:
```suggestion
            version = tuple(map(int, SANIC_VERSION.split(".")))
```

sentry_sdk/integrations/sanic.py
```diff
@@ -32,17 +27,33 @@
...
+        except (TypeError, ValueError):
+            raise DidNotEnable("Unparseable Sanic version: {}".format(VERSION))
```
rhcarvalho:
```suggestion
            raise DidNotEnable("Unparseable Sanic version: {}".format(SANIC_VERSION))
```

tests/test_basics.py
```diff
@@ -37,6 +38,31 @@ def error_processor(event, exc_info):
     assert event["exception"]["values"][0]["value"] == "aha! whatever"
 
 
+def test_auto_enabling_integrations_does_not_break(sentry_init, caplog):
+    caplog.set_level(logging.DEBUG)
+
+    sentry_init(_experiments={"auto_enabling_integrations": True}, debug=True)
```
rhcarvalho: Since integrations are enabled only once globally, isn't [`sentry_init`](https://github.com/getsentry/sentry-python/blob/dd85044c0ec76b19bc7f5b2984c3d4dcc47567ed/tests/conftest.py#L182-L197) influenced by the other tests?

tests/test_basics.py
```diff
@@ -37,6 +38,31 @@ def error_processor(event, exc_info):
     assert event["exception"]["values"][0]["value"] == "aha! whatever"
 
 
+def test_auto_enabling_integrations_does_not_break(sentry_init, caplog):
```
rhcarvalho: What's the intention of this test? "Does not break" is too open ended.

Seems like it checks that none of the integrations in the list get enabled.

tests/test_basics.py
```diff
@@ -37,6 +38,31 @@ def error_processor(event, exc_info):
...
+        "sentry_sdk.integrations.django.DjangoIntegration",
+        "sentry_sdk.integrations.flask.FlaskIntegration",
+        "sentry_sdk.integrations.bottle.BottleIntegration",
+        "sentry_sdk.integrations.falcon.FalconIntegration",
+        "sentry_sdk.integrations.sanic.SanicIntegration",
+        "sentry_sdk.integrations.celery.CeleryIntegration",
+        "sentry_sdk.integrations.rq.RqIntegration",
+        "sentry_sdk.integrations.aiohttp.AioHttpIntegration",
+        "sentry_sdk.integrations.tornado.TornadoIntegration",
+        "sentry_sdk.integrations.sqlalchemy.SqlalchemyIntegration",
```
rhcarvalho: Suggestion: extract this sequence such that the tests stay in sync with the implementation.

rhcarvalho:
> * How to opt out of integrations?

How to opt-out of a default integration today, say we don't want `sentry_sdk.integrations.argv.ArgvIntegration`?

tests/test_basics.py
untitaker: That they don't get enabled, but by break I rather meant that the DidNotEnable and ImportErrors are caught down properly. I'll rename

untitaker:
@rhcarvalho we don't have a solution for that either. We tell you to disable `default_integrations` and enumerate all integrations explicitly

rhcarvalho:
> We tell you to disable `default_integrations` and enumerate all integrations explicitly

Fair enough.

---

I wonder what would it look like if we simply "promoted" those now "auto-enabling" integrations to be *default integrations*. Meaning there would be only one concept: either an integration is part of the default set or not.

Default integrations are enabled on a best-effort basis, such that integrations for modules you don't use would just not be enabled (e.g. the Flask integration, though being default on, would not be enabled in a typical Django project).

To disable any integration, only a single answer: enumerate them explicitly, overriding the default set.

But maybe the auto-enabling integrations have some other characteristic I didn't account for?

untitaker:
>I wonder what would it look like if we simply "promoted" those now "auto-enabling" integrations to be default integrations.

This is the goal, and if it weren't for auto-integration being behind a feature flag it would already work

tests/test_basics.py
untitaker: The monkeypatches themselves leak, but not integration settings or which integrations are really enabled for the current client. The code we patch in should not execute unless the integration is marked as enabled on the current hub.

Some tests we mark with `pytest.mark.forked` to execute them in the subprocess as that's the most surefire way to make sure tests do not make other tests fail. This affects test times negatively so we use it sparingly. There's a CLI option too if you need to debug a state leak.

rhcarvalho:
@untitaker my bad, I overlooked that in the sanic integration there was already a `VERSION` variable before, and the rename broke it... https://github.com/getsentry/sentry-python/pull/625/files#diff-9118f4d8201c6ac1bb83687985d21737R61

https://travis-ci.com/getsentry/sentry-python/jobs/288969634

sentry_sdk/client.py
```diff
@@ -106,6 +106,9 @@ def _init_impl(self):
             self.integrations = setup_integrations(
                 self.options["integrations"],
                 with_defaults=self.options["default_integrations"],
+                with_auto_enabling_integrations=self.options["_experiments"].get(
```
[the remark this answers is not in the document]
untitaker: I added in-code documentation in `consts.py` but believe adding "real" documentation may give the impression anything in there is stable (despite any disclaimers).

For example we removed Relay from the main docs as it was alpha, even though each page of its documentation had a red "alpha" banner on top.

tests/test_basics.py
untitaker: Yeah this makes sense

PR #631, opened 2020-02-20
fix: Use monotonic clock to compute durations

rhcarvalho:
In summary, care must be taken when computing durations. Monotonic
clocks are not subject to system clock adjustments or system clock skew.
The difference between any two chronologically recorded time values is
guaranteed to never be negative.

The same guarantee above does not exist for the difference between two
calls to datetime.now() and friends.

More details and rationale see PEP 418.

Resources:

PEP 418 -- Add monotonic time, performance counter, and process time functions
https://www.python.org/dev/peps/pep-0418/

PEP 564 -- Add new time functions with nanosecond resolution
https://www.python.org/dev/peps/pep-0564/

---

**Know limitation**

At present, we only address timestamp computation within a single span.
A more complete patch should compute all durations in a transaction using a monotonic clock and set all timestamps relative to a single reading of a wall clock time source.

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -320,8 +320,8 @@ class Span(object):
             return None
 
         try:
-            durationSeconds = time.perf_counter() - self._start_timestamp_monotonic
-            self.timestamp = self.start_timestamp + timedelta(seconds=durationSeconds)
+            duration_seconds = time.perf_counter() - self._start_timestamp_monotonic
+            self.timestamp = self.start_timestamp + timedelta(seconds=duration_seconds)
         except AttributeError:
             self.timestamp = datetime.utcnow()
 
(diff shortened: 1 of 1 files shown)


rhcarvalho:
Related: https://github.com/getsentry/sentry-javascript/pull/2441

rhcarvalho:
@untitaker could you please have a look?

untitaker:
💯

rhcarvalho:
> how to configure a linter that would prevent you from writing `durationSeconds`.

Shame on me :man_facepalming: 

Thanks for noticing! I can only blame on doing the JavaScript and Python changes simultaneously :smile:

untitaker:
I think the bigger shame is that there is no lint against this and that I couldn't figure out how to configure one

PR #636, opened 2020-02-27
ref: Introduce linter for proper naming conventions

untitaker:


sentry_sdk/integrations/bottle.py
```diff
@@ -42,7 +45,6 @@ class BottleIntegration(Integration):
     def __init__(self, transaction_style="endpoint"):
         # type: (str) -> None
 
-        TRANSACTION_STYLE_VALUES = ("endpoint", "url")
```
rhcarvalho: Hmm is it worth changing this (and in the other integrations)? Could break downstream code.

sentry_sdk/_compat.py
```diff
@@ -59,12 +59,12 @@ def reraise(tp, value, tb=None):
 
 def with_metaclass(meta, *bases):
     # type: (Any, *Any) -> Any
-    class metaclass(type):
```
rhcarvalho: I might not be remembering well, but I thought `class metaclass` was common practice?

sentry_sdk/integrations/spark/spark_driver.py
```diff
@@ -106,99 +106,101 @@ def process_event(event, hint):
 
 
 class SparkListener(object):
-    def onApplicationEnd(self, applicationEnd):
+    def onApplicationEnd(self, applicationEnd):  # noqa: N802,N803
```
rhcarvalho: Maybe apply this to the whole class or module/file?
Doing it per method is very noisy :(

Also could we have a note like "# noqa: ...; ignore camelCase in methods names and attributes because this code is inherited from somewhere..."

.flake8
```diff
@@ -1,8 +1,10 @@
 [flake8]
 ignore = 
   E203, E266, E501, W503, E402, E731, C901, B950, B011,
-  B014  // does not apply to Python 2
+  B014,  // does not apply to Python 2
+  N812,  // i don't care
+  N804   // conflicts with B902
```
rhcarvalho: Could we add a link to https://pypi.org/project/pep8-naming/?

Would be easier to figure out what the codes mean. A quick note on their meaning wouldn't hurt either?!

sentry_sdk/integrations/bottle.py
untitaker: How would the downstream code depend on this?

sentry_sdk/_compat.py
untitaker: I honestly don't know. Are you saying that metaclasses are supposed to be snakecase?

sentry_sdk/integrations/spark/spark_driver.py
untitaker: I am not sure how to specify it on a class level. I would like to avoid doing it on a module level for sure asssuming we do not restructure the modules.

.flake8
untitaker: Yeah I will look into how to document this better.

sentry_sdk/integrations/bottle.py
rhcarvalho: `BottleIntegration.TRANSACTION_STYLE_VALUES`

untitaker: it's not an attribute but a local var in init (that's why the linter complained)

rhcarvalho: Yes, I misread it, sorry.

.flake8
untitaker: I updated all the docs but couldn't figure out how to truly reference lints by name

PR #644, opened 2020-03-09
fix(tracing) Omit the top level status tag

markstory:
In the product side we don't really want people to search by this tag as it is far more expensive than the `transaction.status` property which is indexed separately. By not emitting this tag and only including it in the trace context we won't end up with poor performing tags for users to click on.

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -408,7 +408,7 @@ class Span(object):
         }
 
         if "status" in self._tags:
-            rv["status"] = self._tags["status"]
+            rv["status"] = self._tags.pop("status")
 
         return rv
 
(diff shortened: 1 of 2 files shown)


HazAT:
We never did any of this in JS 🤔 
Was this just a precaution string or am I missing something?
@untitaker

untitaker:
What do you mean with precaution string? We used to send this:

```
{
    "event_id": ...,
    "tags": {"status": "ok"},
    "contexts": {"trace": {"status": "ok"}}
}
```

with this change we remove the event tag

HazAT:
@untitaker OK, updated my comment.
And we never did this in JS

markstory:
I had to add a status attribute to the Span object. Without that transactions collected inside celery would not have statuses set because `get_trace_context()` is called multiple times and after the first the status tag is gone.

untitaker:
@HazAT I am looking at the UI and I do see JS transactions having a status tag. Could you please figure out what logic to apply here? Do we want to get rid of tags altogether, etc

untitaker:
From chat:

>because we do not set any, except when it fails

PR #647, opened 2020-03-13
Add instrumentation to Django's db.connect() call

mvantellingen:
This is a really interesting metric to know since this is sometimes
cause of performance issues

### DIFF
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -224,12 +224,32 @@ class DjangoIntegration(Integration):
                 id(value),
             )
 
+        _patch_db_connect()
         _patch_channels()
         patch_django_middlewares()
 
 
 _DRF_PATCHED = False
 _DRF_PATCH_LOCK = threading.Lock()
+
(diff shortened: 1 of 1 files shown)


sentry_sdk/integrations/django/__init__.py
```diff
@@ -232,6 +233,30 @@ def _django_queryset_repr(value, hint):
...
+        hub = Hub.current
+        integration = hub.get_integration(DjangoIntegration)
+        if integration is None:
+            return old_connect(self)
+        else:
+            description = "%s.connect - %s" % (
+                self.settings_dict.get("ENGINE") or "django.db",
+                self.alias,
+            )
+            with hub.start_span(op="db", description=description):
```
untitaker: Can we rename the op to `db.connect`, and move the db engine into tags?

mvantellingen: will do 👍

sl0thentr0py:
duplicate of https://github.com/getsentry/sentry-python/pull/1250
closing

PR #651, opened 2020-03-19
feat(transport): Remove absolute time stamps

mitsuhiko:
The quota system no longer supports absolute timestamps.


mitsuhiko:
Limit already gets trimmed.

PR #652, opened 2020-03-20
fix: Test transport rate limits parsing and enforcement

untitaker:
Also fix a bug where missing categories ("123::project") would not
enforce a rate limit for all categories, as they were parsed as category
"" instead of category None.


untitaker:
That should be this one (from the rust impl): https://github.com/getsentry/sentry-python/pull/652/files#diff-50acfff8adfa27fc0ebafc7c5a4b5754R71

jan-auer:
@untitaker Sorry for the ambiguity. I meant enforcement, not parsing.

PR #658, opened 2020-03-24
perf: Don't unnecessarily capture exceptions that are ignored

### LINKED ISSUE #650: Sanic Integration: Exceptions captured even when ignored on exception routes
When an exception route is exposed via Sanic and the Sentry SDK is used, the "exception" is captured regardless of whether or not that exception is ignored.

The `_capture_exception` function can be a bit expensive.

Simple exception route without Sentry integration:
```
Running 5s test @ http://localhost:8080/404
  6 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     5.86ms    1.95ms  24.45ms   72.12%
    Req/Sec     2.75k   338.93     3.27k    72.67%
  82225 requests in 5.01s, 10.90MB read
  Non-2xx or 3xx responses: 82225
Requests/sec:  16403.63
Transfer/sec:      2.17MB
```

Simple exception route with Sentry integration.
```
Running 5s test @ http://localhost:8080/404
  6 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    62.64ms   17.68ms 109.91ms   60.98%
    Req/Sec   255.03     47.26   400.00     64.55%
  7743 requests in 5.10s, 1.03MB read
  Non-2xx or 3xx responses: 7743
Requests/sec:   1517.53
Transfer/sec:    205.99KB
```

Example Code:
```py
import sentry_sdk
from sanic import Sanic
from sanic.exceptions import NotFound
from sanic.response import text
from sentry_sdk.integrations.sanic import SanicIntegration

app = Sanic()

@app.exception(NotFound)
def handle_404s(request,exception):
    return text("404 Not Found", 404)

if __name__ == "__main__":
    sentry_sdk.init(
        dsn="http://00000000000000000000000000000011:00000000000000000000000000000011@sentry/11",
        integrations=[SanicIntegration()],
        ignore_errors=[NotFound],
    )
    
    app.run(host="localhost", port=8080, debug=False, access_log=False, workers=1)
```

Flamegraph:
image

jordangarside:
Having the sentry integration enabled with Sanic results in a 10x reduction in requests per second to `@app.exception` routes, even when those route exceptions are ignored.

The main function we're trying to avoid is `exceptions_from_error_tuple`, as illustrated in the flamegraph in #650. In a relatively simple world, the sentry-python package shouldn't execute anything if the error is in the `ignored_exceptions`.

Ideally this change would live inside of `event_from_exception` [here](https://github.com/getsentry/sentry-python/blob/b7679a50f31ef63614da21a6ecda9e4ff43a5754/sentry_sdk/utils.py#L671), but I wasn't sure what would happen in the scenario that either no exception was returned or a mock exception was returned.

In theory every integration should ignore the exception passed back from `event_from_exception` and simply use the `hint` to assess whether or not `_should_capture_exception`, but since `hint` has a default value of `{}` it's possible that some cases might not actually be passing the `hint` correctly.

Fixes #650 

TLDR: This fixes it for Sanic. It would be nice to fix this generally, but I don't have enough context to know whether the change is safe.

### DIFF
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -150,6 +150,9 @@ def _capture_exception(exception):
     # If an integration is there, a client has to be there.
     client = hub.client  # type: Any
 
+    if client and exception in client.ignore_errors:
+        return
+
     with capture_internal_exceptions():
         event, hint = event_from_exception(
             exception,


sentry_sdk/integrations/sanic.py
```diff
@@ -150,6 +150,9 @@ def _capture_exception(exception):
     # If an integration is there, a client has to be there.
     client = hub.client  # type: Any
 
+    if client and exception in client.ignore_errors:
```
untitaker, line 153: Not entirely sure how this works, do you have exception instances in `ignore_errors`?

PR #673, opened 2020-04-20
feat: Add Conda to CI

untitaker:


untitaker:
seems to work fine and was just a hunch in response to a customer bug. not worth it

PR #674, opened 2020-04-20
fix: Preserve contextvars in aiohttp integration

### LINKED ISSUE #670: aiohttp integration ability to use contextvars in logger
Hi!

I was looking to add a custom [logging filter](https://docs.python.org/3/library/logging.html#logging.Filter) to my aiohttp server during exception that are catched [here](https://github.com/aio-libs/aiohttp/blob/16a49c143fc0abab75163fb78738fff3d3e17f49/aiohttp/web_protocol.py#L387).

It's useful and easy to do, but, it occurs that, in my custom logging filter, I can't find [contextvars](https://docs.python.org/3/library/contextvars.html) that I have created in my request handler task.

Two things caused this problem:
- [aiohttp handled requests in a sub-Task, and catched exceptions in the parent Task](https://github.com/aio-libs/aiohttp/blob/6a5ab96bd9cb404b4abfd5160fe8f34a29d941e5/aiohttp/web_protocol.py#L415-L416).
  → This was fixed in https://github.com/aio-libs/aiohttp/commit/9997cae (because users asked to be able to access `contextvars` -- like us). It was even [backported to aiohttp version 3.7](https://github.com/aio-libs/aiohttp/commit/29eccad84e8200b5c90856c8732da0fdbbcef904).
- [Sentry-aiohttp integration handles requests in a sub-Task too](https://github.com/getsentry/sentry-python/blob/cd646579d04e2fad6a8994304314ac52fec2f83c/sentry_sdk/integrations/aiohttp.py#L113).

Python documentation on Tasks [here](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task). One important thing is that they reset *contextvars*.

To summarize:
```
   aiohttp                               logging exception
      +                                        ^
      |                                        |
      | asyncio.create_task(handle_request())  | contextvars didn't go up again
      |                                        | (it's fixed now)
      |                                        |
      v                                        |
   Sentry                                      |
      +                                        |
      |                                        |
      | asyncio.create_task(handle_request())  | contextvars don't go up
      |                                        |
      v                                        |
I set contextvars        +---------------------+

                                   Exception
```

As long as the issue is not fixed in Sentry, I still can't use `contextvars` to log custom data using the standard Python `logging` library.
The only solution is to disable Sentry, then logging works OK with contextvars.

Any idea how to fix this in Sentry-aiohttp code?
I'd be happy to open a PR, but I'm not familiar enough with Sentry code, or Python in general, thus I need some help at least.

H--o-l:
aiohttp integration currently re-create a task to encapsulate the request
handler.
But:
- aiohttp already does it.
- contextvars created in it can't be read by aiohttp.
  It's an issue for users custom logger.

Fix #670

### DIFF
--- a/sentry_sdk/integrations/aiohttp.py
+++ b/sentry_sdk/integrations/aiohttp.py
@@ -71,46 +71,41 @@ class AioHttpIntegration(Integration):
 
         async def sentry_app_handle(self, request, *args, **kwargs):
             # type: (Any, Request, *Any, **Any) -> Any
-            async def inner():
-                # type: () -> Any
-                hub = Hub.current
-                if hub.get_integration(AioHttpIntegration) is None:
-                    return await old_handle(self, request, *args, **kwargs)
+            hub = Hub.current
+            if hub.get_integration(AioHttpIntegration) is None:
+                return await old_handle(self, request, *args, **kwargs)
 
-                weak_request = weakref.ref(request)
+            weak_request = weakref.ref(request)
 
-                with Hub(Hub.current) as hub:
-                    with hub.configure_scope() as scope:
-                        scope.clear_breadcrumbs()
-                        scope.add_event_processor(_make_request_processor(weak_request))
+            with Hub(Hub.current) as hub:
+                # Scope data will not leak between requests because aiohttp
+                # create a task to wrap each request.
+                with hub.configure_scope() as scope:
+                    scope.clear_breadcrumbs()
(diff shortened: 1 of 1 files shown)


untitaker:
You can inline `inner()` into the outer function now

H--o-l:
@untitaker Ok, something like what I just push?

untitaker:
Perfect, thanks!

PR #708, opened 2020-05-27
fix: Allow nested spans to override sampled argument

beezz:
Do not crash if nested spans/transactions specify sampled argument.


untitaker:
thanks!

PR #707, opened 2020-05-27
fix(django): Support for Django 3.1

untitaker:
* Django 3.1a1 adds more parameters to `load_middleware` which we do not really care about.

* Django 3.1a1 starts executing exception handlers in a random thread/with the wrong context. Turns out they have their own implementation of context local that is necessary to be able to find the right hub. See also https://github.com/getsentry/sentry-docs/pull/1721

More support is required for supporting async middlewares once Django 3.1 comes out but this should unbreak basic usage of the sdk.

Fix #704

PR #710, opened 2020-05-28
fix: Add missing test requirements

rhcarvalho:
They are part of tox.ini, but when running tests outside of tox, without
those, pytest will not run tests in tests/integrations/asgi.

Surprinsingly, running pytest in a parent directory collects other tests
and skips asgi tests, without error.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -7,3 +7,5 @@ pytest-cov==2.8.1
 gevent
 eventlet
 newrelic
+starlette
+requests


untitaker:
Starlette will not install on Python 2 which is why I am resorting to tox here. I don't think there's an easy way to run tests for integrations outside of tox.

The `asgi/__init__.py` is responsible for skipping those tests as I do not want pytest to even import those files as they are invalid Python 2 due to the use of async.

I saw that tox is not properly running those asgi tests which I am intending to fix in #707

rhcarvalho:
@untitaker I tried to run the tests based on what we document in CONTRIBUTING.md. No tests are run. I'm using Python 3.7.

Good point that if a dependency cannot install under Python 2, then it cannot be in the shared test-requirements.txt.

We could:

- Update the contribution guide
  - Specify that only tox is supported, if that's the case
- Split test-requirements.txt into test-requirements-py2.txt and test-requirements-py3.txt (and IIRC they could reference a common set of dependencies with a line `-r test-requirements.txt`.
- All of the above

Thoughts?

untitaker:
>Specify that only tox is supported, if that's the case

I would like that specifically for the tests in `tests/integrations/`. We could point out that the user can still use the pytest installation of the individual tox env.

untitaker:
@rhcarvalho I am closing this for now but generally we would need much more test-requirements files for integrations, probably one for each minor Python version

PR #717, opened 2020-06-09
Fix typo.

lmasikl:


lmasikl:
@untitaker I didn't any changes in that string https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/aiohttp.py#L54, because I don't is it reasonable. I catched this exception because aiohttp has not parseable version https://github.com/aio-libs/aiohttp/releases/tag/v4.0.0a1.
But string like this can fix this exception in root
`version = tuple(map(int, AIOHTTP_VERSION.split(".")[:2]))`

untitaker:
@lmasikl I agree, can you put this change into your PR.

I am surprised that people actually use alpha versions, but so be it.

untitaker:
Thanks, works for me!

PR #722, opened 2020-06-18
Pass when exception is raised trying to set the transaction name

robindboer:
When Pyramid can't match the route `request.matched_route` is set to `None`. The patched call view is throwing an AttributeException trying to set the transaction name.


untitaker:
Thanks! I will make a release asap

untitaker:
0.15.1 is out, sorry for the inconvenience!

robindboer:
Thanks for the swift response 🙌

PR #727, opened 2020-06-22
ref: Remove Hub.current is not None checks

rhcarvalho:
By [construction](https://github.com/getsentry/sentry-python/blob/6e378f18919a834d3de50b6f981e332b5094ad83/sentry_sdk/hub.py#L133-L142), Hub.current is never None, such that the expression

    Hub.current is not None

always evaluates to True.

This commit simplifies all uses of Hub.current, and in particular
chooses to write `return Hub.current.method(...)` for every method, even
when the method returns None. The intent is to make it easier to keep
the static API matching the Hub behavior. Without this, if a method
returns anything other than None the static API would silently drop it,
leading to unnecessary debugging time spent trying to identify the
culprit.

### DIFF
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -1,5 +1,4 @@
 import inspect
-from contextlib import contextmanager
 
 from sentry_sdk.hub import Hub
 from sentry_sdk.scope import Scope
@@ -72,10 +71,7 @@ def capture_event(
     **scope_args  # type: Dict[str, Any]
 ):
     # type: (...) -> Optional[str]
-    hub = Hub.current
-    if hub is not None:
-        return hub.capture_event(event, hint, scope=scope, **scope_args)
-    return None
+    return Hub.current.capture_event(event, hint, scope=scope, **scope_args)
 
 
 @hubmethod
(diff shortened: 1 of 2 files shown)


sentry_sdk/api.py
```diff
@@ -170,59 +147,37 @@ def push_scope(
...
+    return Hub.current.push_scope(callback)
 
 
 @scopemethod  # noqa
 def set_tag(key, value):
     # type: (str, Any) -> None
-    hub = Hub.current
-    if hub is not None:
-        hub.scope.set_tag(key, value)
+    return Hub.current.scope.set_tag(key, value)
```
rhcarvalho, line 156: As noted in the PR description / commit message, I intentionally decided to always use the pattern `return Hub.current.method(...)`, even when the return value is `None`. The idea behind it is to avoid accidentally returning `None` when the `Hub` method returns something else (e.g. as part of a behavior change). In other words, I find this easier to maintain than to selectively use or not the `return` keyword.

rhcarvalho:
Removed now unused import

```py
from contextlib import contextmanager
```

PR #729, opened 2020-06-22
feat: Send transactions in envelopes

rhcarvalho:
This matches [what the JS SDK does](https://github.com/getsentry/sentry-javascript/pull/2553) and what the Tracing dev docs indicates.

Related: https://github.com/getsentry/sentry-python/pull/730

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -7,11 +7,12 @@ import socket
 
 from sentry_sdk._compat import string_types, text_type, iteritems
 from sentry_sdk.utils import (
-    handle_in_app,
-    get_type_name,
     capture_internal_exceptions,
     current_stacktrace,
     disable_capture_event,
+    format_timestamp,
+    get_type_name,
+    handle_in_app,
     logger,
 )
 from sentry_sdk.serializer import serialize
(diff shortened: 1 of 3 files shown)


sentry_sdk/client.py
```diff
@@ -330,7 +330,12 @@ def capture_event(
         if session:
             self._update_session_from_event(session, event)
 
-        self.transport.capture_event(event_opt)
+        if event.get("type") == "transaction":
+            # Transactions should go to the /envelope/ endpoint.
+            self.transport.capture_envelope(Envelope(items=[event]))
```
rhcarvalho, line 335: I found no evidence that this actually sends to `/envelope/`. Too late now, going to sleep.

rhcarvalho: Sending envelopes to the `/envelope/` endpoint is https://github.com/getsentry/sentry-python/pull/730.

tests/test_tracing.py
```diff
@@ -22,7 +22,8 @@ def test_basic(sentry_init, capture_events, sample_rate):
             pass
 
     if sample_rate:
-        (event,) = events
```
untitaker, line 25: I really like this pattern! :P

rhcarvalho: I don't dislike it. The problem was the kind of output we were getting when the test fails "not enough values to unpack" or "too many values to unpack", no clue what the events were.

The fix is adding an assert so that pytest handles printing the contents of `events`. Could have kept the unpacking, but now I don't want to trigger CI again.

untitaker: that's also been a complaint from @mitsuhiko. I might make a pytest plugin that does something similar to assertion rewrites

PR #731, opened 2020-06-23
fix: Do not call before_send for transactions

rhcarvalho:
This matches the behavior with JS and [the specs](https://develop.sentry.dev/sdk/unified-api/tracing).

The change in JS was introduced in https://github.com/getsentry/sentry-javascript/pull/2600.
See notes under "Few other fixes".

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -199,7 +199,7 @@ class _Client(object):
             event = serialize(event)
 
         before_send = self.options["before_send"]
-        if before_send is not None:
+        if before_send is not None and event.get("type") != "transaction":
             new_event = None
             with capture_internal_exceptions():
                 new_event = before_send(event, hint or {})
--- a/tests/test_tracing.py
+++ b/tests/test_tracing.py
@@ -167,3 +167,16 @@ def test_no_double_sampling(sentry_init, capture_events):
         pass
 
     assert len(events) == 1
+
+
+def test_transactions_do_not_go_through_before_send(sentry_init, capture_events):
+    def before_send(event, hint):
+        raise RuntimeError("should not be called")
+
+    sentry_init(traces_sample_rate=1.0, before_send=before_send)
+    events = capture_events()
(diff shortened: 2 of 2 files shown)


rhcarvalho:
Got a test failure in `test_aws.py`:

https://travis-ci.com/github/getsentry/sentry-python/jobs/352702298#L1612-L1616

image

Quick glance at the test implementation and I guess it could be flakey.

cc @untitaker

rhcarvalho:
Rebased to fix merge conflict.

vmarkovtsev:
Is there a standard way to filter out transactions now? I've got all my backend endpoint handler entry points wrapped in spans and need to silence a few unimportant ones to save $$$ on the events quota.

rhcarvalho:
> Is there a standard way to filter out transactions now? I've got all my backend endpoint handler entry points wrapped in spans and need to silence a few unimportant ones to save $$$ on the events quota.

@vmarkovtsev you can use `sentry_sdk.scope.add_global_event_processor` and pass in a func to silence transaction events (and in fact do anything `before_send` could do). Make sure to check for `event.get("type") == "transaction"`.

This is a workaround. Keep in mind that manually sampling transactions will potentially skew the aggregate metrics you see in Sentry.

rhcarvalho:
@vmarkovtsev also worth noting that event processors run right before the transaction is sent, which means that the sampling decision has already been propagated to other services for the life time of the transaction. It might or might not be of importance to you.

If you're doing manual instrumentation, you should also be able to override the sampling decision early on by starting a transaction with `sampled=False`.

We're working on a more general solution for controlling transaction sampling from SDKs.

vmarkovtsev:
In my case, I am using `aiohttp` integration which automatically starts transactions for each incoming HTTP request. I am not sure how to set `sampled=False` then... But yeah, canceling the sampling whatsoever is better ofc.

Thanks for the workaround! It should work fine for me :+1:

PR #733, opened 2020-06-23
feat(redis): Add tags for more commands

tonyo:
When we deal with multikey commands (e.g. DEL), it's sometimes still useful to have the key as a tag. To make it less ambiguous, let's add the tag only if the command has one argument.

### DIFF
--- a/sentry_sdk/integrations/redis.py
+++ b/sentry_sdk/integrations/redis.py
@@ -9,6 +9,11 @@ from sentry_sdk._types import MYPY
 if MYPY:
     from typing import Any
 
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/redis.py
```diff
@@ -9,6 +9,11 @@
 if MYPY:
     from typing import Any
 
+SINGLE_KEY_COMMANDS = frozenset(
```
untitaker, line 12: Can you prefix those w underscore?

tonyo: hell yeah

PR #734, opened 2020-06-23
feat: Access transaction in current scope

rhcarvalho:
Specially when trying to add spans to automatically instrumented transactions, users need access to the current transaction.

This gives direct access no matter how deep the code is in the transaction/span tree.

This replicates a similar feature in JS: https://github.com/getsentry/sentry-javascript/pull/2668.

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -134,10 +134,33 @@ class Scope(object):
         """When set this overrides the default fingerprint."""
         self._fingerprint = value
 
-    @_attr_setter
+    @property
+    def transaction(self):
+        # type: () -> Any
+        # would be type: () -> Optional[Span], see https://github.com/python/mypy/issues/3004
+        # XXX: update return type to Optional[Transaction]
+        """Return the transaction (root span) in the scope."""
+        if self._span is None or self._span._span_recorder is None:
+            return None
+        try:
+            return self._span._span_recorder.spans[0]
+        except (AttributeError, IndexError):
+            return None
+
+    @transaction.setter
     def transaction(self, value):
-        # type: (Optional[str]) -> None
+        # type: (Any) -> None
+        # would be type: (Optional[str]) -> None, see https://github.com/python/mypy/issues/3004
         """When set this forces a specific transaction name to be set."""
+        # XXX: the docstring above is misleading. The implementation of
+        # apply_to_event prefers an existing value of event.transaction over
+        # anything set in the scope.
+        # XXX: note that with the introduction of the Scope.transaction getter,
+        # there is a semantic and type mismatch between getter and setter. The
+        # getter returns a transaction, the setter sets a transaction name.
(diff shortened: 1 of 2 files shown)


sentry_sdk/scope.py
```diff
@@ -134,14 +134,30 @@ def fingerprint(self, value):
...
+        # XXX: note that with the introduction of the Scope.transaction getter,
+        # there is a semantic and type mismatch between getter and setter. The
+        # getter returns a transaction, the setter sets a transaction name.
+        # Without breaking version compatibility, we could make the setter set a
+        # transaction name or transaction (self._span) depending on the type of
+        # the value argument.
         self._transaction = value
-        span = self._span
-        if span:
-            span.transaction = value
```
rhcarvalho, line 144: This is already done in `Scope.apply_to_event`. Removing from here to match the behavior with JS.

Note that setting `span.transaction` here may inadvertently promote a regular span into a transaction; we're leaning towards a clearer conceptual separation between transactions and spans. The separation should become more clear with https://github.com/getsentry/sentry-python/pull/715, that introduces the `Transaction` class and makes `Span.transaction` => `Transaction.name`.

rhcarvalho:
Dang, the current implementation runs into a MYPY limitation because the type annotation for property getter and setter is different.

https://github.com/python/mypy/issues/3004

https://travis-ci.com/github/getsentry/sentry-python/builds/173008390

sentry_sdk/scope.py
untitaker: This is not done in apply_to_event, only the trace context is being set

rhcarvalho: https://github.com/getsentry/sentry-python/blob/b539ecb9c6a8c990051ccc5d7d0d80f8723f6a3b/sentry_sdk/scope.py#L295-L296

With the difference that in JS there is no equivalent of `event.get("transaction") is None` as the first condition, but I don't want to touch that detail in this PR, as it is changing a lot of how Scope works and is unrelated to Tracing.

rhcarvalho: I give up removing this for now. The test `test_continue_from_headers` implies an expectation that setting `scope.transaction` overwrites the name of the current transaction. In JS changing the scope only changes the scope, and the scope is applied to events/transactions later.

sentry_sdk/scope.py
```diff
@@ -134,10 +134,33 @@ def fingerprint(self, value):
...
+        except (AttributeError, IndexError):
+            return None
+
+    @transaction.setter
     def transaction(self, value):
-        # type: (Optional[str]) -> None
+        # type: (Any) -> None
+        # would be type: (Optional[str]) -> None, see https://github.com/python/mypy/issues/3004
         """When set this forces a specific transaction name to be set."""
+        # XXX: the docstring above is misleading. The implementation of
```
untitaker, line 155: Feel free to fix that behavior in this PR, I don't think anybody would notice even.

Issue #735, opened 2020-06-23
RedisIntegration does not throw DidNotEnable when there is no redis module


soutys:
Most of integrations have ``ModuleDoesNotExist`` / ``ImportError`` try...except block throwing ``DidNotEnable`` exception (see: celery, flask, sqlalchemy).

``RedisIntegration`` have ``redis`` module import inside ``setup_once`` method because of monkey-patching.

This causes some sort of problems during wide-range integrations autodetection.

untitaker:
>This causes some sort of problems during wide-range integrations autodetection.

RedisIntegration is not part of auto-enabling integrations so it should not even try to import or enable by itself. https://github.com/getsentry/sentry-python/blob/3a7f4f26a40ec1a0965f4d43b6ba7b24a3a69c8b/sentry_sdk/integrations/__init__.py#L55

PR #737, opened 2020-06-24
fix: Auto-enabling Redis and Pyramid integration

untitaker:
See also #735


untitaker:
This needs some version checks put in place.

antonpirker (approved):
lgtm

PR #747, opened 2020-06-26
feat: Introduce Transaction and Hub.start_transaction

rhcarvalho:
This aligns the tracing implementation with the current JS tracing implementation, up to a certain extent.

Hub.start_transaction or start_transaction are meant to be used when starting transactions, replacing most uses of Hub.start_span / start_span.

Spans are typically created from their parent transactions via transaction.start_child, or start_span relying on the transaction being in the current scope.

It is okay to start a transaction without a name and set it later.
Sometimes the proper name is not known until after the transaction has started.

We could fail the transaction if it has no name when calling the finish method. Instead, set a default name that will prompt users to give a name to their transactions. This is the same behavior as implemented in JS.

### DIFF
--- a/sentry_sdk/hub.py
+++ b/sentry_sdk/hub.py
@@ -8,7 +8,7 @@ from contextlib import contextmanager
 from sentry_sdk._compat import with_metaclass
 from sentry_sdk.scope import Scope
 from sentry_sdk.client import Client
-from sentry_sdk.tracing import Span
+from sentry_sdk.tracing import Span, Transaction
 from sentry_sdk.sessions import Session
 from sentry_sdk.utils import (
     exc_info_from_error,
@@ -441,38 +441,88 @@ class Hub(with_metaclass(HubMeta)):  # type: ignore
     ):
         # type: (...) -> Span
         """
-        Create a new span whose parent span is the currently active
-        span, if any. The return value is the span object that can
-        be used as a context manager to start and stop timing.
+        Create and start timing a new span whose parent is the currently active
+        span or transaction, if any. The return value is a span instance,
+        typically used as a context manager to start and stop timing in a `with`
+        block.
 
-        Note that you will not see any span that is not contained
-        within a transaction. Create a transaction with
-        ``start_span(transaction="my transaction")`` if an
-        integration doesn't already do this for you.
+        Only spans contained in a transaction are sent to Sentry. Most
+        integrations start a transaction at the appropriate time, for example
+        for every incoming HTTP request. Use `start_transaction` to start a new
+        transaction when one is not already in progress.
         """
+        # TODO: consider removing this in a future release.
+        # This is for backwards compatibility with releases before
+        # start_transaction existed, to allow for a smoother transition.
+        if isinstance(span, Transaction) or "transaction" in kwargs:
+            deprecation_msg = (
+                "Deprecated: use start_transaction to start transactions and "
+                "Transaction.start_child to start spans."
+            )
+            if isinstance(span, Transaction):
+                logger.warning(deprecation_msg)
+                return self.start_transaction(span)
+            if "transaction" in kwargs:
+                logger.warning(deprecation_msg)
+                name = kwargs.pop("transaction")
+                return self.start_transaction(name=name, **kwargs)
 
-        client, scope = self._stack[-1]
+        if span is not None:
+            return span
 
         kwargs.setdefault("hub", self)
 
-        if span is None:
-            span = scope.span
-            if span is not None:
-                span = span.new_span(**kwargs)
-            else:
-                span = Span(**kwargs)
+        span = self.scope.span
+        if span is not None:
+            return span.start_child(**kwargs)
 
-        if span.sampled is None and span.transaction is not None:
+        return Span(**kwargs)
+
+    def start_transaction(
+        self,
+        transaction=None,  # type: Optional[Transaction]
+        **kwargs  # type: Any
+    ):
+        # type: (...) -> Transaction
+        """
+        Start and return a transaction.
+
+        Start an existing transaction if given, otherwise create and start a new
+        transaction with kwargs.
+
+        This is the entry point to manual tracing instrumentation.
+
+        A tree structure can be built by adding child spans to the transaction,
+        and child spans to other spans. To start a new child span within the
+        transaction or any span, call the respective `.start_child()` method.
+
+        Every child span must be finished before the transaction is finished,
+        otherwise the unfinished spans are discarded.
+
+        When used as context managers, spans and transactions are automatically
+        finished at the end of the `with` block. If not using context managers,
+        call the `.finish()` method.
+
+        When the transaction is finished, it will be sent to Sentry with all its
+        finished child spans.
+        """
+        if transaction is None:
+            kwargs.setdefault("hub", self)
+            transaction = Transaction(**kwargs)
+
+        client, scope = self._stack[-1]
+
+        if transaction.sampled is None:
             sample_rate = client and client.options["traces_sample_rate"] or 0
-            span.sampled = random.random() < sample_rate
+            transaction.sampled = random.random() < sample_rate
 
-        if span.sampled:
+        if transaction.sampled:
             max_spans = (
                 client and client.options["_experiments"].get("max_spans") or 1000
             )
-            span.init_finished_spans(maxlen=max_spans)
+            transaction.init_span_recorder(maxlen=max_spans)
 
-        return span
+        return transaction
 
     @overload  # noqa
     def push_scope(
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -5,6 +5,7 @@ from itertools import chain
 from sentry_sdk._functools import wraps
 from sentry_sdk._types import MYPY
 from sentry_sdk.utils import logger, capture_internal_exceptions
+from sentry_sdk.tracing import Transaction
 
 if MYPY:
     from typing import Any
@@ -137,8 +138,7 @@ class Scope(object):
     @property
     def transaction(self):
         # type: () -> Any
-        # would be type: () -> Optional[Span], see https://github.com/python/mypy/issues/3004
-        # XXX: update return type to Optional[Transaction]
+        # would be type: () -> Optional[Transaction], see https://github.com/python/mypy/issues/3004
         """Return the transaction (root span) in the scope."""
         if self._span is None or self._span._span_recorder is None:
             return None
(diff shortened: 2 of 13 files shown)


rhcarvalho:
I don't know what's wrong with the docs generation job https://travis-ci.com/github/getsentry/sentry-python/jobs/354722815

sentry_sdk/hub.py
```diff
@@ -441,38 +441,138 @@ def start_span(
...
+        # is already an existing transaction/span in the scope, then
+        # this is a hard to use API. We don't document nor support
+        # appending an existing span to a new transaction created to
+        # contain the span.
+
+        # TODO: consider removing this in a future release.
+        # This is for backwards compatibility with releases before
+        # start_transaction existed, to allow for a smoother transition.
+        if isinstance(span, Transaction) or "transaction" in kwargs:
+            deprecation_msg = (
```
AbhiPrasad, line 464: I don't think we need the redundant check here for `isinstance(span, Transaction) or "transaction" in kwargs`, we should just extract `deprecation_msg` to be a top level constant

sentry_sdk/hub.py
```diff
@@ -441,38 +441,138 @@ def start_span(
...
         """
-
-        client, scope = self._stack[-1]
+        # XXX: if the only way to use start_span correctly is when there
+        # is already an existing transaction/span in the scope, then
+        # this is a hard to use API. We don't document nor support
+        # appending an existing span to a new transaction created to
+        # contain the span.
+
+        # TODO: consider removing this in a future release.
```
AbhiPrasad, line 460: As an aside for another PR: we should have a standard deprecated comment somewhere like we do in JS, so we can easily see what to remove in each major version.

sentry_sdk/scope.py
```diff
@@ -182,17 +182,19 @@ def set_user(self, value):
     @property
     def span(self):
         # type: () -> Optional[Span]
```
AbhiPrasad, line 184:
```suggestion
        # type: () -> Optional[Union[Span, Transaction]]
```

sentry_sdk/tracing.py
```diff
@@ -413,6 +418,91 @@ def get_trace_context(self):
         return rv
 
 
+class Transaction(Span):
+    __slots__ = ("name",)
```
AbhiPrasad, line 422: nit: No need for the tuple right?

```suggestion
    __slots__ = "name"
```

sentry_sdk/hub.py
```diff
@@ -441,38 +441,138 @@ def start_span(
...
+        # Code such as:
+        #     hub.start_transaction(Transaction(name="foo"), name="bar")
+        #
+        # is clearly weird, but not so weird if we intentionally want to rename
+        # a transaction we got from somewhere else:
+        #     hub.start_transaction(transaction, name="new_name")
+        #
+        # Would be clearer if Hub was not involved:
+        #     transaction.name = "new_name"
+        #     with transaction:  # __enter__ => start, __exit__ => finish
```
AbhiPrasad, line 552: Wouldn't we still have to attach the transaction to the current hub, even with `with transaction:`? That way, I'd much prefer to be explicit using `hub.start_transaction` then hide what it's doing here.

sentry_sdk/hub.py
```diff
@@ -441,38 +441,138 @@ def start_span(
...
+        finished at the end of the `with` block. If not using context managers,
+        call the `.finish()` method.
+
+        When the transaction is finished, it will be sent to Sentry with all its
+        finished child spans.
+        """
+        # XXX: should we always set transaction.hub = self?
+        # In a multi-hub program, what does it mean to write
+        #     hub1.start_transaction(Transaction(hub=hub2))
+        # ? Should the transaction be on hub1 or hub2?
```
AbhiPrasad, line 539: Hmm, this is tricky. I think if we keep our current implementation, the transaction should be on `hub1` and we should not allow setting hub through `kwargs`. If we decide the change the API, we can revisit this.

sentry_sdk/tracing.py
untitaker: I am surprised to learn that works but I would like to avoid it.

sentry_sdk/hub.py
```diff
@@ -441,38 +441,138 @@ def start_span(
...
+            )
+            if isinstance(span, Transaction):
+                logger.warning(deprecation_msg)
+                return self.start_transaction(span)
+            if "transaction" in kwargs:
+                logger.warning(deprecation_msg)
+                name = kwargs.pop("transaction")
+                return self.start_transaction(name=name, **kwargs)
+
+        # XXX: this is not very useful because
```
untitaker, line 476: This is there to be consistent with start_transaction I suppose.

sentry_sdk/hub.py
```diff
@@ -441,38 +441,138 @@ def start_span(
...
-            span = scope.span
-            if span is not None:
-                span = span.new_span(**kwargs)
-            else:
-                span = Span(**kwargs)
+        span = self.scope.span
+        if span is not None:
+            return span.start_child(**kwargs)
+
+        # XXX: returning a detached span here means whatever span tree built
```
untitaker, line 494: This makes the API simpler to use as the caller does not have to check whether they are in a transaction. I don't really see why you call this not useful.

sentry_sdk/scope.py
untitaker: `Span` is a subtype of `Transaction`

sentry_sdk/hub.py
```diff
@@ -441,38 +441,138 @@ def start_span(
...
+
+        When the transaction is finished, it will be sent to Sentry with all its
+        finished child spans.
+        """
+        # XXX: should we always set transaction.hub = self?
+        # In a multi-hub program, what does it mean to write
+        #     hub1.start_transaction(Transaction(hub=hub2))
+        # ? Should the transaction be on hub1 or hub2?
+
+        # XXX: it is strange that both start_span and start_transaction take
```
untitaker, line 541: btw i think those discussions should not be done as comments in the sourcecode

sentry_sdk/scope.py
AbhiPrasad: Yup, I'll resolve this.

rhcarvalho:
> I don't know what's wrong with the docs generation job https://travis-ci.com/github/getsentry/sentry-python/jobs/354722815

Circular imports.

Removed the `SpanLike` type, making the total diff smaller and solving the circular import.

<details>

```diff
diff --git a/sentry_sdk/_types.py b/sentry_sdk/_types.py
index 7c1aaa6..8e5427c 100644
--- a/sentry_sdk/_types.py
+++ b/sentry_sdk/_types.py
@@ -12,7 +12,7 @@ if MYPY:
     from typing import Optional
     from typing import Tuple
     from typing import Type
-    from typing import Union
+
     from typing_extensions import Literal
 
     ExcInfo = Tuple[
@@ -37,7 +37,3 @@ if MYPY:
     ]
     SessionStatus = Literal["ok", "exited", "crashed", "abnormal"]
     EndpointType = Literal["store", "envelope"]
-
-    from sentry_sdk.tracing import Span, Transaction  # noqa
-
-    SpanLike = Union[Span, Transaction]
diff --git a/sentry_sdk/scope.py b/sentry_sdk/scope.py
index 0c082c4..f928063 100644
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -26,7 +26,7 @@ if MYPY:
         Type,
     )
 
-    from sentry_sdk.tracing import SpanLike
+    from sentry_sdk.tracing import Span
     from sentry_sdk.sessions import Session
 
     F = TypeVar("F", bound=Callable[..., Any])
@@ -114,7 +114,7 @@ class Scope(object):
         self.clear_breadcrumbs()
         self._should_capture = True
 
-        self._span = None  # type: Optional[SpanLike]
+        self._span = None  # type: Optional[Span]
         self._session = None  # type: Optional[Session]
         self._force_auto_session_tracking = None  # type: Optional[bool]
 
@@ -181,13 +181,13 @@ class Scope(object):
 
     @property
     def span(self):
-        # type: () -> Optional[SpanLike]
+        # type: () -> Optional[Span]
         """Get/set current tracing span or transaction."""
         return self._span
 
     @span.setter
     def span(self, span):
-        # type: (Optional[SpanLike]) -> None
+        # type: (Optional[Span]) -> None
         self._span = span
         # XXX: this differs from the implementation in JS, there Scope.setSpan
         # does not set Scope._transactionName.
diff --git a/sentry_sdk/tracing.py b/sentry_sdk/tracing.py
index 05abcf6..d390c6b 100644
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -25,9 +25,6 @@ if MYPY:
     from typing import Dict
     from typing import List
     from typing import Tuple
-    from typing import Type
-
-    from sentry_sdk._types import SpanLike
 
 _traceparent_header_format_re = re.compile(
     "^[ \t]*"  # whitespace
@@ -225,7 +222,7 @@ class Span(object):
 
     @classmethod
     def continue_from_environ(
-        cls,  # type: Type[SpanLike]
+        cls,
         environ,  # type: typing.Mapping[str, str]
         **kwargs  # type: Any
     ):
@@ -239,7 +236,7 @@ class Span(object):
 
     @classmethod
     def continue_from_headers(
-        cls,  # type: Type[SpanLike]
+        cls,
         headers,  # type: typing.Mapping[str, str]
         **kwargs  # type: Any
     ):
@@ -261,7 +258,7 @@ class Span(object):
 
     @classmethod
     def from_traceparent(
-        cls,  # type: Type[SpanLike]
+        cls,
         traceparent,  # type: Optional[str]
         **kwargs  # type: Any
     ):
```

</details>

sentry_sdk/hub.py
rhcarvalho: This can hide an entire instrumented span tree if one is not careful to guarantee a transaction is in the scope when calling `hub.start_span(op="...")`.

You told me today about a use case for intentionally omitting an entire instrumented span tree, I still keeping wrap my head around that. In other parts of the Unified API we explicitly prefer to avoid doing things when not necessary -- e.g. no-op transport/client when SDK has no DSN, callbacks are not called, etc.

I'll move these discussions into issues.

sentry_sdk/hub.py
rhcarvalho: The intention here was to have a self-contained block with the TODO note, so it is clearer what it applies to.

sentry_sdk/hub.py
rhcarvalho: Well, `start_transaction` doesn't exist outside this PR. In one of its usages, the previous `start_span` would also just return the input (sampling decision work moved to transactions only).

Keeping  it for backwards-compat.

rhcarvalho:
Removed some commentary, will move the discussion into new issues where appropriate.

Issue #748, opened 2020-06-27
stack_data, executing, and pure_eval


alexmojaki:
Hi! I've recently created a few libraries that are helpful for extracting information from frames and tracebacks, and I think they could be very useful to sentry.

- [`executing`](https://github.com/alexmojaki/executing) can identify the exact AST node being executed by a frame in many cases. It's always disappointed me that Python only points to a line in each frame in tracebacks, especially when the expression/statement at fault spans many lines. Sometimes it can be really hard to interpret. Currently [executing is used in IPython](https://github.com/ipython/ipython/pull/12150) (in master, unreleased) to highlight the node, here's what it looks like:

![ipython](https://user-images.githubusercontent.com/3627481/75476425-3e6b9280-59a3-11ea-9b6c-b9e099475a45.png)

- `executing` can also infer a function `__qualname__` from a code object, meaning your traceback can say `lorem/ipsum.py in MyClass.__init__ at line 123` instead of just `__init__`, which is much more informative.
- [`pure_eval`](https://github.com/alexmojaki/pure_eval) can safely evaluate certain AST nodes without side effects, so that if the source for a frame contains expressions like `self.foo` and `bar[key]` their values can often be shown alongside plain variables.
- [`stack_data`](https://github.com/alexmojaki/stack_data) collects data from tracebacks for the purpose of formatting and displaying them. It uses `executing` and `pure_eval` and also provides a lot of functionality of its own. For example, in this frame:

![Screen Shot 2020-06-27 at 19 57 57](https://user-images.githubusercontent.com/3627481/85928853-94e4d100-b8b0-11ea-93e9-3eb8915e333b.png)

there are lines from `complex_filter` which the user doesn't need to see, and `stack_data` knows how to filter those out by only collecting lines which belong in the current scope. [I integrated`stack_data` into IPython](https://github.com/ipython/ipython/pull/11886) which allowed removing a lot of flaky custom introspection code, fixing several bugs and making the code more maintainable.
- `stack_data` identifies the locations of variables and evaluated expressions in the source lines. An intelligent UI could use this, for example, to allow the user to hover over a variable in source code to see its value rather than scroll through a list to find it.

Overall there are lots of possibilities. What would interest you? I'm willing to make a PR and make any necessary changes to the libraries.

untitaker:
Hey @alexmojaki, this seems like a nice library!

[...]

PR #749, opened 2020-06-28
Use executing to infer code qualname

alexmojaki:
This is one of the features mentioned in https://github.com/getsentry/sentry-python/issues/748. As requested by @untitaker I've made executing an optional dependency.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,10 @@ sentry-sdk==0.10.1
 
 A major release `N` implies the previous release `N-1` will no longer receive updates. We generally do not backport bugfixes to older versions unless they are security relevant. However, feel free to ask for backports of specific commits on the bugtracker.
 
+## 0.16.1
+
+* Flask integration: Fix a bug that prevented custom tags from being attached to transactions.
+
 ## 0.16.0
 
 * Redis integration: add tags for more commands
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,9 +27,13 @@ sentry-sdk==0.10.1
 
 A major release `N` implies the previous release `N-1` will no longer receive updates. We generally do not backport bugfixes to older versions unless they are security relevant. However, feel free to ask for backports of specific commits on the bugtracker.
 
-## [Unreleased]
+## 0.16.0
 
 * Redis integration: add tags for more commands
+* Redis integration: Patch rediscluster package if installed.
+* Session tracking: A session is no longer considered crashed if there has been a fatal log message (only unhandled exceptions count).
+* **Breaking change**: Revamping of the tracing API.
+* **Breaking change**: `before_send` is no longer called for transactions.
 
 ## 0.15.1
 
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,1 +1,9 @@
+coverage:
+  status:
+    project:
+      default: false
+    patch:
+      default: false
+      python:
+        target: 90%
 comment: false
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -22,7 +22,7 @@ project = u"sentry-python"
 copyright = u"2019, Sentry Team and Contributors"
 author = u"Sentry Team and Contributors"
 
-release = "0.16.0"
+release = "0.16.1"
 version = ".".join(release.split(".")[:2])  # The short X.Y version.
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -22,7 +22,7 @@ project = u"sentry-python"
 copyright = u"2019, Sentry Team and Contributors"
 author = u"Sentry Team and Contributors"
 
-release = "0.15.1"
+release = "0.16.0"
 version = ".".join(release.split(".")[:2])  # The short X.Y version.
 
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -48,3 +48,5 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 [mypy-asgiref.*]
 ignore_missing_imports = True
+[mypy-executing.*]
+ignore_missing_imports = True
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -89,7 +89,7 @@ DEFAULT_OPTIONS = _get_default_options()
 del _get_default_options
 
 
-VERSION = "0.15.1"
+VERSION = "0.16.0"
 SDK_INFO = {
     "name": "sentry.python",
     "version": VERSION,
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -89,7 +89,7 @@ DEFAULT_OPTIONS = _get_default_options()
 del _get_default_options
 
 
-VERSION = "0.16.0"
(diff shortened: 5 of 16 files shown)


alexmojaki:
I'm getting a Django test failure, but when I try running the tests locally, I get `Apps aren't loaded yet`:

```
(venv) ➜  sentry-python git:(executing-qualname) export DJANGO_SETTINGS_MODULE=tests.integrations.django.myapp.settings
(venv) ➜  sentry-python git:(executing-qualname) python -m pytest -k django                                            
================================================================= test session starts ==================================================================
platform darwin -- Python 3.8.0, pytest-3.7.3, py-1.9.0, pluggy-0.13.1
rootdir: /Users/alexhall/Desktop/python/sentry-python, inifile: pytest.ini
plugins: celery-4.4.6, localserver-0.5.0, forked-1.1.3, cov-2.8.1
collected 415 items / 1 errors / 411 deselected / 7 skipped                                                                                            

======================================================================== ERRORS ========================================================================
_______________________________________________ ERROR collecting tests/integrations/django/test_basic.py _______________________________________________
tests/integrations/django/test_basic.py:8: in <module>
    from django.contrib.auth.models import User
venv/lib/python3.8/site-packages/django/contrib/auth/models.py:2: in <module>
    from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
venv/lib/python3.8/site-packages/django/contrib/auth/base_user.py:47: in <module>
    class AbstractBaseUser(models.Model):
venv/lib/python3.8/site-packages/django/db/models/base.py:107: in __new__
    app_config = apps.get_containing_app_config(module)
venv/lib/python3.8/site-packages/django/apps/registry.py:252: in get_containing_app_config
    self.check_apps_ready()
venv/lib/python3.8/site-packages/django/apps/registry.py:135: in check_apps_ready
    raise AppRegistryNotReady("Apps aren't loaded yet.")
E   django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================================== 7 skipped, 411 deselected, 1 error in 3.62 seconds ==================================================
```

What's the procedure for testing without tox?

alexmojaki:
I don't understand what's happening with the project coverage.

untitaker:
@alexmojaki Hey, sorry for responding so late! Would it be possible to have this be behind some sort of feature flag or integration, even? I think one could use the global event processors similar to how [gnu backtrace](https://docs.sentry.io/platforms/python/gnu_backtrace/) does it to overwrite the data. The reason I am asking is because I am still not sure if doing this sort of thing incurs performance overhead, and if so, I would hate to see it being enabled simply because some package is installed.

Now to your questions:

* Ignore code coverage, it's not set up properly
* the tests seem to pass on travis, so I think I was too late here. I am not sure what the issue was with running outside of tox?

alexmojaki:
OK, I'm happy to make this optional based on an integration or something. However I don't know how to implement that. I can't take the approach of GNU backtrace because I need the actual code object (and similarly the other features need an actual frame object) which is not available in the data by the time I reach an event processor. Being able to use the frame directly within `serialize_frame` as I've done in this PR seems like the obvious (only?) route. But I don't know how to check for the integration in there. I tried:

```python
def function_name(frame):
    # type: (FrameType) -> str
    from sentry_sdk import Hub
    from sentry_sdk.integrations.executing import ExecutingIntegration

    if Hub.current.get_integration(ExecutingIntegration) is None:
        return frame.f_code.co_name
    else:
        import executing
        return executing.Source.for_frame(frame).code_qualname(frame.f_code)
```

and apart from being awkward it doesn't work. I don't know how exactly Hub.current works, but apparently not like this. Any ideas?

untitaker:
Ah sorry, I misremembered. The "prior art" can be found in the Django integration, where we access the `hint` parameter to get access to the actual `exc_info`. We then call `walk_exception_chain` to get a list of `exc_info`s, and the frame should then be available on the traceback object.

https://github.com/getsentry/sentry-python/blob/dce439fccbd2d157d2c855c09027417155c23760/sentry_sdk/integrations/django/__init__.py#L150-L152

I acknowledge this is extremely awkward, but it still kind of keeps the API surface between the core SDK and the other integrations minimal. If one had to fully separate the integrations from the core sdk the utility functions used could be easily vendored.

untitaker:
>I don't know how exactly Hub.current works, but apparently not like this.

It's a thread local. If the executing integration is enabled once it should at least be accessible on the same thread.

alexmojaki:
Regarding performance, there are three main sources of overhead:

- `executing.Source.for_frame(frame)` (as used here to get the qualname) parses the source file and adds a bunch of metadata. I haven't measured this, but I think the cost is moderate. However, the object is cached based on the source filename, so each file is only parsed once. Once this is done, getting the qualname is a trivial lookup.
- Identifying the executing node (still coming) also has moderate overhead the first time, but again this is cached based on the instruction index. [I actually have a test that 10,000 calls complete in under a second.](https://github.com/alexmojaki/executing/blob/2f0d1c8a1ac145836c814a06bb79a62bdee7ddb1/tests/test_main.py#L154-L163) The point is that the overhead of these features is proportional to the number of unique, distinct events that an app produces. If the same exception happens many times, that's not a problem, and sentry itself will be the main source of overhead.
- pure_eval is a different story. Its work is based on runtime values and doesn't leave much room for caching. And even if it was free, it's going to add more values to serialize and send. So I think this has the greatest potential for performance problems. On the other hand, it uses the same `executing.Source.for_frame(frame)`, because it also needs to parse the file and use much of the same metadata. So if the `PureEvalIntegration` is active, even if the `ExecutingIntegration` isn't, I think it's reasonable to get the qualname anyway since there shouldn't be any performance problem.

alexmojaki:
> > I don't know how exactly Hub.current works, but apparently not like this.
> 
> It's a thread local. If the executing integration is enabled once it should at least be accessible on the same thread.

What I found is that in `test_attach_stacktrace_enabled`, I have:

```python
hub = Hub(...)
hub.capture_message("HI")
```

which leads directly to `function_name` where I check `Hub.current` which turns out to be different from `hub` above. I was hoping `hub.capture_message` would set `Hub.current` but at least where I need it to it doesn't.

untitaker:
You're right. I would ignore this error for now, I would argue the testcase is unrealistic. Hub isn't supposed to be used directly, at least not like that. The testcases using `sentry_init` are more realistic.

I can fix that testcase up for you when the time comes.

alexmojaki:
OK, this mostly works, but only if there's an actual exception. It doesn't work in cases like `test_attach_stacktrace_enabled` with `attach_stacktrace=True` and `capture_message`. I guess that's an acceptable limitation?

I don't think I can make the diff coverage 90% without jumping through unreasonable hoops.

untitaker:
Merging this now so you can rebase the other prs

PR #750, opened 2020-06-28
Extract additional expression values with pure_eval

alexmojaki:
Another part of https://github.com/getsentry/sentry-python/issues/748


untitaker:
Like with https://github.com/getsentry/sentry-python/pull/749 please make this a separate integration. btw I think this is by far the most valuable feature out of everything you've brainstormed!

PR #752, opened 2020-06-30
feat(redis): Patch rediscluster if present

beezz:
In addition to the redis and rb clients try to patch also the rediscluster library which does not use the already patched clients.

### DIFF
--- a/sentry_sdk/integrations/redis.py
+++ b/sentry_sdk/integrations/redis.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 
 from sentry_sdk import Hub
-from sentry_sdk.utils import capture_internal_exceptions
+from sentry_sdk.utils import capture_internal_exceptions, logger
 from sentry_sdk.integrations import Integration
 
 from sentry_sdk._types import MYPY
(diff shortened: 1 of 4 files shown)


sentry_sdk/integrations/redis.py
```diff
@@ -15,6 +15,27 @@
 _MULTI_KEY_COMMANDS = frozenset(["del", "touch", "unlink"])
 
 
+def _patch_rediscluster():
+    # type: () -> None
+    try:
+        import rediscluster
+    except ImportError:
+        pass
```
tonyo, line 23: Should we do a "return" here perhaps? Otherwise the code below will fail.

tonyo: Ah, you're catching everything when in the parent function. Still, would probably be nice to return, because it doesn't make a lot of sense to continue if the package wasn't imported properly.

beezz: good catch. that should definitely be a return!

sentry_sdk/integrations/redis.py
```diff
@@ -34,6 +55,11 @@ def setup_once():
             patch_redis_client(rb.clients.MappingClient)
             patch_redis_client(rb.clients.RoutingClient)
 
+        try:
+            _patch_rediscluster()
+        except Exception:
+            logger.exception("Error occured while patching `rediscluster` library")
```
tonyo, line 61: Not sure where this message will end up if we use `exception`, what about doing `logger.warning()` here?

untitaker: That should be fine. We get the log message attached even with exception

tests/integrations/rediscluster/test_rediscluster.py
```diff
@@ -0,0 +1,51 @@
+from sentry_sdk import capture_message
+import sentry_sdk.integrations
+from sentry_sdk.integrations.redis import RedisIntegration
+
+import rediscluster
+
+
+def _test_rediscluster_basic(sentry_init, capture_events, monkeypatch, clsname):
```
untitaker, line 8: Please instead of using this helper function use `pytest.mark.parametrize`.

tests/integrations/rediscluster/test_rediscluster.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def execute_command(*args, **kwargs):
+        execute_command_calls.append((args, kwargs))
+
+    monkeypatch.setattr(
+        redis_cluster_cls, "execute_command", execute_command,
+    )
+
+    # should be done by the sentry_init fixture?
+    sentry_sdk.integrations._installed_integrations = set()
```
untitaker, line 22: Why do we need to do this?

tests/integrations/rediscluster/test_rediscluster.py
beezz: How would that look like with the test if the `StrictRedisCluster` class is present?
https://github.com/getsentry/sentry-python/pull/752/files#diff-52fd8f5367c660c1f800f2cb58148e7eR49-R51

tests/integrations/rediscluster/test_rediscluster.py
beezz: I need the `RedisIntegration.setup_once` to run for both tests.

tests/integrations/rediscluster/test_rediscluster.py
beezz: Figured something out.

PR #756, opened 2020-07-08
Capturing Initialization and Timeout errors for AWS Lambda Integration 

shantanu73:
Changes:

1) Added a new wrapper decorator for post_init_error method to capture initialization error for AWS Lambda integration.
2) Modified _wrap_handler decorator to include code which runs a parallel thread to capture timeout error.
3) Modified _make_request_event_processor decorator to include execution duration as parameter.
4) Added TimeoutThread class in utils.py which is useful to capture timeout error.

### DIFF
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta
 from os import environ
 import sys
 import json
+import os
 
 from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk._compat import reraise
@@ -28,8 +29,8 @@ if MYPY:
     F = TypeVar("F", bound=Callable[..., Any])
 
 # Constants
-TIMEOUT_THRESHOLD_MILLIS = 1500  # Minimum time required to capture TimeoutError
-SECONDS_CONVERSION_FACTOR = 1000.0
+TIMEOUT_WARNING_BUFFER = 1500  # Buffer time required to send timeout warning to Sentry
+MILLIS_TO_SECONDS = 1000.0
 
 
 def _wrap_init_error(init_error):
@@ -78,31 +79,33 @@ def _wrap_handler(handler):
 
         # If an integration is there, a client has to be there.
         client = hub.client  # type: Any
-        configured_time_in_millis = context.get_remaining_time_in_millis()
+        configured_time = context.get_remaining_time_in_millis()
 
         with hub.push_scope() as scope:
             with capture_internal_exceptions():
                 scope.clear_breadcrumbs()
                 scope.transaction = context.function_name
-                scope.add_event_processor(_make_request_event_processor(event, context, configured_time_in_millis))
+                scope.add_event_processor(
+                    _make_request_event_processor(event, context, configured_time)
+                )
+                # Starting the Timeout thread only if the configured time is greater than Timeout warning
+                # buffer and timeout_warning parameter is set True.
+                if (
+                    integration.timeout_warning
+                    and configured_time > TIMEOUT_WARNING_BUFFER
+                ):
+                    waiting_time = (
+                        configured_time - TIMEOUT_WARNING_BUFFER
+                    ) / MILLIS_TO_SECONDS
+
+                    timeout_thread = TimeoutThread(
+                        waiting_time, configured_time / MILLIS_TO_SECONDS
+                    )
+
+                    # Starting the thread to raise timeout warning exception
+                    timeout_thread.start()
 
             try:
-                # Checking if parameter to check timeout is set True
-                if integration.get_check_timeout_error():
-                    # Starting the Timeout thread only if the configured time is greater than Timeout threshold value
-                    if configured_time_in_millis > TIMEOUT_THRESHOLD_MILLIS:
-                        remaining_time_in_sec = (configured_time_in_millis - TIMEOUT_THRESHOLD_MILLIS)/SECONDS_CONVERSION_FACTOR
-
-                        configured_time_in_sec = configured_time_in_millis / SECONDS_CONVERSION_FACTOR
-                        configured_time = int(configured_time_in_sec)
-
-                        # Setting up the exact integer value of configured time(in seconds)
-                        if configured_time < configured_time_in_sec:
-                            configured_time = configured_time + 1
-
-                        # Starting the thread to raise timeout warning exception
-                        timeout_thread = TimeoutThread(remaining_time_in_sec, configured_time)
-                        timeout_thread.start()
                 return handler(event, context, *args, **kwargs)
             except Exception:
                 exc_info = sys.exc_info()
@@ -131,13 +134,9 @@ def _drain_queue():
 class AwsLambdaIntegration(Integration):
     identifier = "aws_lambda"
 
-    def __init__(self, check_timeout_error=False):
+    def __init__(self, timeout_warning=False):
         # type: (bool) -> None
-        self.check_timeout_error = check_timeout_error
-
-    def get_check_timeout_error(self):
-        # type: () -> bool
-        return self.check_timeout_error
+        self.timeout_warning = timeout_warning
 
     @staticmethod
     def setup_once():
@@ -234,16 +233,29 @@ def _make_request_event_processor(aws_event, aws_context, configured_timeout):
 
     def event_processor(event, hint, start_time=start_time):
         # type: (Event, Hint, datetime) -> Optional[Event]
-        extra = event.setdefault("extra", {})
+        total_memory, used_memory, free_memory = map(
+            int, os.popen("free -t -m").readlines()[-1].split()[1:]
+        )
         remaining_time_in_milis = aws_context.get_remaining_time_in_millis()
         exec_duration = configured_timeout - remaining_time_in_milis
+
+        contexts = event.setdefault("contexts", {})
+        if (
+            isinstance(contexts, dict)
+            and "memory usage and execution duration" not in contexts
+        ):
+            contexts["memory usage and execution time"] = {
+                "Execution duration in millis": exec_duration,
+                "Memory usage in MB": used_memory,
+                "Remaining time in millis": remaining_time_in_milis,
+            }
+
+        extra = event.setdefault("extra", {})
         extra["lambda"] = {
             "function_name": aws_context.function_name,
             "function_version": aws_context.function_version,
             "invoked_function_arn": aws_context.invoked_function_arn,
-            "remaining_time_in_millis": remaining_time_in_milis,
             "aws_request_id": aws_context.aws_request_id,
-            "execution_duration": exec_duration,
         }
 
         extra["cloudwatch logs"] = {
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -1,6 +1,7 @@
 from datetime import datetime, timedelta
 from os import environ
 import sys
+import json
 
 from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk._compat import reraise
@@ -9,6 +10,7 @@ from sentry_sdk.utils import (
     capture_internal_exceptions,
     event_from_exception,
     logger,
+    TimeoutThread,
 )
 from sentry_sdk.integrations import Integration
 from sentry_sdk.integrations._wsgi_common import _filter_headers
@@ -25,6 +27,49 @@ if MYPY:
 
     F = TypeVar("F", bound=Callable[..., Any])
 
+# Constants
+TIMEOUT_THRESHOLD_MILLIS = 1500  # Minimum time required to capture TimeoutError
+SECONDS_CONVERSION_FACTOR = 1000.0
+
+
+def _wrap_init_error(init_error):
+    # type: (F) -> F
+    def sentry_init_error(*args, **kwargs):
+        # type: (*Any, **Any) -> Any
+
+        # Fetch Initialization error details from arguments
+        error = json.loads(args[1])
+
+        hub = Hub.current
+        integration = hub.get_integration(AwsLambdaIntegration)
+        if integration is None:
+            return init_error(*args, **kwargs)
+
+        # If an integration is there, a client has to be there.
+        client = hub.client  # type: Any
+
+        with hub.push_scope() as scope:
+            with capture_internal_exceptions():
+                scope.clear_breadcrumbs()
+            try:
+                # Checking if there is any error/exception which is raised in the runtime
+                # environment from arguments and, re-raising it to capture it as an event.
+                if error.get("errorType"):
+                    exc_info = sys.exc_info()
+                    reraise(*exc_info)
+            except Exception:
+                exc_info = sys.exc_info()
+                event, hint = event_from_exception(
+                    exc_info,
+                    client_options=client.options,
+                    mechanism={"type": "aws_lambda", "handled": False},
+                )
+                hub.capture_event(event, hint=hint)
+
+        return init_error(*args, **kwargs)
+
+    return sentry_init_error  # type: ignore
+
 
 def _wrap_handler(handler):
     # type: (F) -> F
@@ -37,14 +82,31 @@ def _wrap_handler(handler):
 
         # If an integration is there, a client has to be there.
         client = hub.client  # type: Any
+        configured_time_in_millis = context.get_remaining_time_in_millis()
 
         with hub.push_scope() as scope:
             with capture_internal_exceptions():
                 scope.clear_breadcrumbs()
                 scope.transaction = context.function_name
-                scope.add_event_processor(_make_request_event_processor(event, context))
+                scope.add_event_processor(_make_request_event_processor(event, context, configured_time_in_millis))
 
             try:
+                # Checking if parameter to check timeout is set True
+                if integration.get_check_timeout_error():
+                    # Starting the Timeout thread only if the configured time is greater than Timeout threshold value
+                    if configured_time_in_millis > TIMEOUT_THRESHOLD_MILLIS:
+                        remaining_time_in_sec = (configured_time_in_millis - TIMEOUT_THRESHOLD_MILLIS)/SECONDS_CONVERSION_FACTOR
+
+                        configured_time_in_sec = configured_time_in_millis / SECONDS_CONVERSION_FACTOR
+                        configured_time = int(configured_time_in_sec)
+
+                        # Setting up the exact integer value of configured time(in seconds)
+                        if configured_time < configured_time_in_sec:
+                            configured_time = configured_time + 1
+
+                        # Starting the thread to raise timeout warning exception
+                        timeout_thread = TimeoutThread(remaining_time_in_sec, configured_time)
+                        timeout_thread.start()
                 return handler(event, context, *args, **kwargs)
             except Exception:
                 exc_info = sys.exc_info()
@@ -72,6 +134,14 @@ def _drain_queue():
 
 class AwsLambdaIntegration(Integration):
     identifier = "aws_lambda"
+
+    def __init__(self, check_timeout_error=False):
+        # type: (bool) -> None
+        self.check_timeout_error = check_timeout_error
+
+    def get_check_timeout_error(self):
+        # type: () -> bool
+        return self.check_timeout_error
 
     @staticmethod
     def setup_once():
@@ -126,6 +196,10 @@ class AwsLambdaIntegration(Integration):
 
             lambda_bootstrap.to_json = sentry_to_json
         else:
+            lambda_bootstrap.LambdaRuntimeClient.post_init_error = _wrap_init_error(
+                lambda_bootstrap.LambdaRuntimeClient.post_init_error
+            )
+
             old_handle_event_request = lambda_bootstrap.handle_event_request
 
             def sentry_handle_event_request(  # type: ignore
@@ -158,19 +232,22 @@ class AwsLambdaIntegration(Integration):
             )
 
 
-def _make_request_event_processor(aws_event, aws_context):
-    # type: (Any, Any) -> EventProcessor
+def _make_request_event_processor(aws_event, aws_context, configured_timeout):
+    # type: (Any, Any, Any) -> EventProcessor
     start_time = datetime.now()
 
     def event_processor(event, hint, start_time=start_time):
         # type: (Event, Hint, datetime) -> Optional[Event]
         extra = event.setdefault("extra", {})
+        remaining_time_in_milis = aws_context.get_remaining_time_in_millis()
+        exec_duration = configured_timeout - remaining_time_in_milis
         extra["lambda"] = {
             "function_name": aws_context.function_name,
             "function_version": aws_context.function_version,
             "invoked_function_arn": aws_context.invoked_function_arn,
-            "remaining_time_in_millis": aws_context.get_remaining_time_in_millis(),
+            "remaining_time_in_millis": remaining_time_in_milis,
             "aws_request_id": aws_context.aws_request_id,
+            "execution_duration": exec_duration,
         }
 
         extra["cloudwatch logs"] = {
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -2,7 +2,6 @@ from datetime import datetime, timedelta
 from os import environ
 import sys
 import json
-import os
 
 from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk._compat import reraise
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/aws_lambda.py
```diff
@@ -25,6 +27,49 @@
...
+SECONDS_CONVERSION_FACTOR = 1000.0
+
+
+def _wrap_init_error(init_error):
+    # type: (F) -> F
+    def sentry_init_error(*args, **kwargs):
+        # type: (*Any, **Any) -> Any
+
+        # Fetch Initialization error details from arguments
+        error = json.loads(args[1])
```
untitaker, line 41: Can we move this down into line 47, no reason to do this work and potentially crash even if the integration is disabled

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -25,6 +27,49 @@
...
+
+        with hub.push_scope() as scope:
+            with capture_internal_exceptions():
+                scope.clear_breadcrumbs()
+            try:
+                # Checking if there is any error/exception which is raised in the runtime
+                # environment from arguments and, re-raising it to capture it as an event.
+                if error.get("errorType"):
+                    exc_info = sys.exc_info()
+                    reraise(*exc_info)
```
untitaker, line 59: Why reraise to catch again instead of calling `sys.exc_info` once? Does this change the stacktrace somehow?

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -126,6 +196,10 @@ def sentry_to_json(*args, **kwargs):
 
             lambda_bootstrap.to_json = sentry_to_json
         else:
+            lambda_bootstrap.LambdaRuntimeClient.post_init_error = _wrap_init_error(
```
untitaker, line 199: Note this will be executed for Python 3.7 only.

shantanu73: Yes, since for python environments 2.7 & 3.6 the runtime already handles these issues. So, I need not handle for those scenarios. In those cases runtime generates FaultException for initialization error.

sentry_sdk/integrations/aws_lambda.py
shantanu73: Sure, I'll make this change.

sentry_sdk/integrations/aws_lambda.py
shantanu73: Nope, this doesn't change the stacktrace.  
I got your point. I'll check once and capture the details without reraising.

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -234,3 +239,50 @@ def test_handler(event, context):
...
+        b'{"foo": "bar"}',
+    )
+    log_result = (base64.b64decode(response["LogResult"])).decode("utf-8")
+    expected_text = "name 'func' is not defined"
+    assert expected_text in log_result
+
+
+def test_timeout_error(run_lambda_function):
+    # Modifying LAMBDA_PRELUDE since capturing timeout error is kept optional.
+    modified_prelude = LAMBDA_PRELUDE.replace("[AwsLambdaIntegration()]", "[AwsLambdaIntegration(True)]")
```
rhcarvalho, line 264: This test could silently do the wrong thing if the string `"[AwsLambdaIntegration()]"` is not part of `LAMBDA_PRELUDE` (e.g. after an unrelated code change).

Also using a positional argument is cryptic. One needs to go through hops to figure out what the `True` value is used for.

My suggestion is to add a new keyword argument to `init_sdk` defined in `LAMBDA_PRELUDE`:

```py
def init_sdk(check_timeout_error=False, **extra_init_args):
    # ...
        integrations=[AwsLambdaIntegration(check_timeout_error=check_timeout_error)],
```

This way, instead of string replacement in the prelude, we can simply append code that uses the new argument.

```py
init_sdk(check_timeout_error=True)
```

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -73,6 +131,14 @@ def _drain_queue():
 class AwsLambdaIntegration(Integration):
     identifier = "aws_lambda"
 
+    def __init__(self, check_timeout_error=False):
```
rhcarvalho, line 134: `check_timeout_error` is a rather misleading name because it doesn't cause the integration to "check for a timeout error".

What it does is trigger a warning if you get "close enough" to the timeout.

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -25,6 +27,45 @@
 
     F = TypeVar("F", bound=Callable[..., Any])
 
+# Constants
+TIMEOUT_THRESHOLD_MILLIS = 1500  # Minimum time required to capture TimeoutError
```
rhcarvalho, line 31: This sort of makes `AwsLambdaIntegration(check_timeout_error=True)` "maybe" work as intended, depending on external state that is not obvious when you initialize the integration. Do we really need this?

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -37,14 +78,31 @@ def sentry_handler(event, context, *args, **kwargs):
...
                 scope.clear_breadcrumbs()
                 scope.transaction = context.function_name
-                scope.add_event_processor(_make_request_event_processor(event, context))
+                scope.add_event_processor(_make_request_event_processor(event, context, configured_time_in_millis))
 
             try:
+                # Checking if parameter to check timeout is set True
+                if integration.get_check_timeout_error():
+                    # Starting the Timeout thread only if the configured time is greater than Timeout threshold value
+                    if configured_time_in_millis > TIMEOUT_THRESHOLD_MILLIS:
```
rhcarvalho, line 93: If we do need this threshold, then let's write a single `if` condition == less indentation levels.

```py
if COND_A and COND_B:
   ...
```

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -37,14 +78,31 @@ def sentry_handler(event, context, *args, **kwargs):
...
+                        remaining_time_in_sec = (configured_time_in_millis - TIMEOUT_THRESHOLD_MILLIS)/SECONDS_CONVERSION_FACTOR
+
+                        configured_time_in_sec = configured_time_in_millis / SECONDS_CONVERSION_FACTOR
+                        configured_time = int(configured_time_in_sec)
+
+                        # Setting up the exact integer value of configured time(in seconds)
+                        if configured_time < configured_time_in_sec:
+                            configured_time = configured_time + 1
```
rhcarvalho, line 101: `time.sleep` can work with floating point numbers, this dance to round up to the next integer is not really needed, is it?

We could also probably get rid of `SECONDS_CONVERSION_FACTOR` and simply use subsecond timeout in the tests, perhaps?

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -158,19 +228,22 @@ def inner(*args, **kwargs):
...
-    # type: (Any, Any) -> EventProcessor
+def _make_request_event_processor(aws_event, aws_context, configured_timeout):
+    # type: (Any, Any, Any) -> EventProcessor
     start_time = datetime.now()
 
     def event_processor(event, hint, start_time=start_time):
         # type: (Event, Hint, datetime) -> Optional[Event]
         extra = event.setdefault("extra", {})
+        remaining_time_in_milis = aws_context.get_remaining_time_in_millis()
+        exec_duration = configured_timeout - remaining_time_in_milis
```
rhcarvalho, line 239: Is this guaranteed to be correct for every execution of the Lambda? Cold and warm starts?

sentry_sdk/utils.py
```diff
@@ -870,3 +872,30 @@ def transaction_from_function(func):
 
 
 disable_capture_event = ContextVar("disable_capture_event")
+
+
+class TimeoutThread(threading.Thread):
+    """Creates a Thread."""
```
rhcarvalho, line 878: This is not a useful docstring for this class.

sentry_sdk/utils.py
```diff
@@ -870,3 +872,30 @@ def transaction_from_function(func):
...
+        self.timeout_duration = timeout_duration
+        self.configured_timeout = configured_timeout
+
+    def get_timeout_duration(self):
+        # type: () -> float
+        return self.timeout_duration
+
+    def get_configured_timeout(self):
+        # type: () -> int
+        return self.configured_timeout
```
rhcarvalho, line 892: In Python it is not idiomatic to have this type of one line getters. Much simpler to access the properties directly.

sentry_sdk/utils.py
```diff
@@ -870,3 +872,30 @@ def transaction_from_function(func):
...
+
+    def run(self):
+        # type: () -> None
+        time.sleep(self.get_timeout_duration())
+        # Raising Exception after timeout duration is reached
+        raise Exception(
+            "WARNING : Function is expected to get timed out. Configured timeout duration = {} seconds".format(
+                self.get_configured_timeout())
+        )
```
rhcarvalho, line 901: We could include in the message the function name (if available) and how long the function has run for.

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -234,3 +239,50 @@ def test_handler(event, context):
...
+        def test_handler(event, context):
+            time.sleep(10)
+            return 0
+        """
+        ),
+        b'{"foo": "bar"}',
+    )
+    expected_text = "WARNING : Function is expected to get timed out. Configured timeout duration = 4 seconds"
+    if not events:
+        # In case of Python 2.7 runtime environment
```
rhcarvalho, line 279: Tests should not rely on the truthy value of `events` and infer it is Python 2.7 or 3.x.
E.g., if something is wrong in Python 3.x and `events` is the empty list, the test would go into the wrong branch of the `if...else`.

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -67,7 +67,11 @@ def inner(code, payload):
         # Check file for valid syntax first, and that the integration does not
         # crash when not running in Lambda (but rather a local deployment tool
         # such as chalice's)
-        subprocess.check_call([sys.executable, str(tmp.join("test_lambda.py"))])
+        try:
+            subprocess.check_call([sys.executable, str(tmp.join("test_lambda.py"))])
+        except Exception as e:
+            # Exception caught in case of Initialization error
+            pass
```
rhcarvalho, line 74: It doesn't look right to silence the exception here. If calling the subprocess fails for some arbitrary reason, the test should fail.

tests/integrations/aws_lambda/test_aws.py
untitaker: Also I am confused, does this mean timeout events are not actually working on python 2.7?

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -37,14 +78,31 @@ def sentry_handler(event, context, *args, **kwargs):
...
 
             try:
+                # Checking if parameter to check timeout is set True
+                if integration.get_check_timeout_error():
+                    # Starting the Timeout thread only if the configured time is greater than Timeout threshold value
+                    if configured_time_in_millis > TIMEOUT_THRESHOLD_MILLIS:
+                        remaining_time_in_sec = (configured_time_in_millis - TIMEOUT_THRESHOLD_MILLIS)/SECONDS_CONVERSION_FACTOR
+
+                        configured_time_in_sec = configured_time_in_millis / SECONDS_CONVERSION_FACTOR
+                        configured_time = int(configured_time_in_sec)
```
untitaker, line 97:
```suggestion
                        configured_time_in_sec = int(configured_time_in_millis / SECONDS_CONVERSION_FACTOR)
```

super minor nitpick but i think we don't need so many locals here

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -73,6 +131,14 @@ def _drain_queue():
 class AwsLambdaIntegration(Integration):
     identifier = "aws_lambda"
 
+    def __init__(self, check_timeout_error=False):
+        # type: (bool) -> None
+        self.check_timeout_error = check_timeout_error
+
+    def get_check_timeout_error(self):
```
untitaker, line 138: Don't need this, just access the attrs directly (same in Thread subclass)

sentry_sdk/integrations/aws_lambda.py
untitaker: We are determining the "configured" timeout by calling `get_remaining_time_in_millis` at the start of the lambda function, so this probably already excludes cold startup overhead.

sentry_sdk/utils.py
```diff
@@ -870,3 +872,30 @@ def transaction_from_function(func):
...
+    def get_configured_timeout(self):
+        # type: () -> int
+        return self.configured_timeout
+
+    def run(self):
+        # type: () -> None
+        time.sleep(self.get_timeout_duration())
+        # Raising Exception after timeout duration is reached
+        raise Exception(
+            "WARNING : Function is expected to get timed out. Configured timeout duration = {} seconds".format(
```
untitaker, line 899: Can we also show `get_remaining_time_in_millis` here?

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -275,14 +276,5 @@ def test_handler(event, context):
...
-        log_result = (base64.b64decode(response["LogResult"])).decode("utf-8")
-        assert expected_text in log_result
-    else:
-        # In case of Python 3.6, 3.7 & 3.8 runtime environments
-        (event,) = events
-        assert event["level"] == "error"
-        (exception,) = event["exception"]["values"]
-        assert exception["type"] == "Exception"
-        assert exception["value"] == expected_text
+    log_result = (base64.b64decode(response["LogResult"])).decode("utf-8")
```
untitaker, line 279: I guess I am still confused as to what we're asserting here. It seems we are not expecting any timeout event at all?

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -234,16 +233,29 @@ def _make_request_event_processor(aws_event, aws_context, configured_timeout):
 
     def event_processor(event, hint, start_time=start_time):
         # type: (Event, Hint, datetime) -> Optional[Event]
-        extra = event.setdefault("extra", {})
+        total_memory, used_memory, free_memory = map(
```
shantanu73, line 236: @rhcarvalho @untitaker This calculation of 'used_memory' is incorrect, but it is the only source I could find up until now.
I've done some manual testing on this, it seems to be in alignment with memory usage displayed in AWS Lambda console, but I couldn't see an exact formula which calculates the desired result.

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -234,16 +233,29 @@ def _make_request_event_processor(aws_event, aws_context, configured_timeout):
...
+        )
         remaining_time_in_milis = aws_context.get_remaining_time_in_millis()
         exec_duration = configured_timeout - remaining_time_in_milis
+
+        contexts = event.setdefault("contexts", {})
+        if (
+            isinstance(contexts, dict)
+            and "memory usage and execution duration" not in contexts
+        ):
+            contexts["memory usage and execution time"] = {
```
rhcarvalho, line 247: I didn't find docs suggesting against spaces in the context key, but in practice we tend not to use it.

This information seems like a very good fit to the standard `runtime` context.
https://develop.sentry.dev/sdk/event-payloads/contexts/#runtime-context

sentry_sdk/integrations/aws_lambda.py
rhcarvalho: Not a big fan of this. It looks at the memory consumption for the whole system, not a single process.
Please consider alternatives like https://stackoverflow.com/a/7669482/963287

Would be good to have some error handling here to avoid blowing everything up if we can't assess memory usage (graceful degradation, progressive enhancement).

Please update the PR title/description if you're changing the scope of this PR. IIRC memory usage was not part of the original patch.

sentry_sdk/integrations/aws_lambda.py
shantanu73: @rhcarvalho 
As per a suggestion given by AJ, I figured that a separate section for execution time and memory usage should be a good implementation. Also, the "Additional data" section which contains all data related to AWS Lambda contains space in the context key, and considering the visual aspect for UI, I kept spaces. 
So, as a solution, should I keep the execution time and memory sage inside lambda details or keep the current implementation (I'll share the screenshot of current implementation on slack)?

untitaker: >Also, the "Additional data" section which contains all data related to AWS Lambda contains space in the context key, and considering the visual aspect for UI, I kept spaces.

Let's stay consistent within the integration for now, but just as a side note rodolfo is right that most other integrations don't do this. It also leads to poor (or at least worse) UX when attempting to configure some other features such as server-side data scrubbing.

tests/integrations/aws_lambda/test_aws.py
shantanu73: @untitaker 
Here, we are asserting with the exception message that we are sending when the ServerlessTimeoutWarning custom exception is raise in case of timeout error. If the exception is raised it'll be there in the log result. 
Also, timeout event is not coming even after the solution that you suggested. I've applied it is this manner:-

`def run(thread_hub, configured_timeout):
    with thread_hub:
        try:
            raise ServerlessTimeoutWarning(
                    "WARNING : Function is expected to get timed out. Configured timeout duration = {} seconds.".format(
                        configured_timeout
                    )
                )
        except Exception:
            client = thread_hub.client
            exc_info = sys.exc_info()
            event, hint = event_from_exception(
                exc_info,
                client_options=client.options,
                mechanism={"type": "threads", "handled": False},
            )
            thread_hub.capture_event(event, hint=hint)
            reraise(*exc_info)`

And, while starting a new thread:

`thread_hub = Hub(Hub.current)

tr = threading.Thread(target=run, args=[thread_hub, configured_timeout])
tr.start()`

This does generate timeout warning and the mechanism I provided in above code appears on Sentry dashboard, but it still doesn't show the stacktrace which was required and neither does the events data come in case f the automation test cases.

sentry_sdk/integrations/aws_lambda.py
shantanu73: So, I'll put the Execution time & memory usage data in the "Additional Data" section inside the "lambda" context.

untitaker: yes, just put it into `extra['lambda']`, you're right. Sorry about the confusion

tests/integrations/aws_lambda/test_aws.py
untitaker: `thread_hub = Hub.current` ought to be correct.

I can't really tell what would be wrong but as-is the testcase is wrong. Y'all have to debug this. If it makes things easier you can also try `hub.capture_message` without any exception.

It's fine if there's no stacktrace. The only useful stacktrace comes from the sigalrm approach we discussed in Slack afaik.

shantanu73: So you mean that event should come irrespective of whatever be the test case? Maybe, it is related to the HttpTransport we are using and we might need to modify it to send event in case of timeout error.

untitaker: I mean the test case right here is supposed to assert that there is an event, right? but right now it's not asserting that.

I don't think the transport is the issue tbh

shantanu73: Okay, I'll debug through this issue and see why event data is not coming and post the updates.

sentry_sdk/integrations/aws_lambda.py
shantanu73: @rhcarvalho I've tried with the above mentioned method, it gave result as 22 MB in case where max memory usage was 61 MB. So, still wrong results. But, I've checked 'resource' library and it has some more parameters for different memory usage. I'll check those once and see if they add up to the exact value which is returned in the AWS lambda result console.

rhcarvalho: 👍 

Note that some memory is allocated after this code runs, so it wouldn't be a big surprise to see different numbers reported here and by the lambda runtime, though the difference shouldn't be significant (otherwise we have a performance improvement to go after in the SDK).

shantanu73: I had multiple runs of the code and the difference is actually this significant only.
I went through the link : https://docs.python.org/3/library/resource.html, I saw other parameters related to memory usage like ru_ixrss (shared memory size), ru_idrss (unshared memory size) & ru_isrss (unshared stack size), these parameters showed memory usage as 0.

The memory usage which comes as output is 24680 (in kilobytes) or a value around this. And the value which is displayed on AWS lambda console is 62 MB.

So, shall we keep using this value itself ? Also, wanted to know abut the graceful degradation suggestion on the previous comment.

rhcarvalho: By graceful degradation I mean we should have code written with error handling in mind, and isolate the errors as much as possible. E.g. if there is an exception when trying to figure out the memory usage, the outcome should be that we do not report on memory usage, while reporting on other aspects is unaffected. And particularly, an exception in any part of collecting extra event data should never prevent the SDK from sending the event.

shantanu73: @rhcarvalho will be removing memory usage data, so resolving this issue.

tests/integrations/aws_lambda/test_aws.py
shantanu73: @untitaker I've debugged this issue. As I suspected earlier as well, the events are not coming because of the _send_event() method defined inside LAMBDA_PRELUDE of test_aws.py. The events =basically are not coming because of the delay of 1 second that is kept there.
Also, the reason why events are not coming after 1 second delay is because of the same reason timeout warning needs to be generated a litter before the actual timeout.

Is it very necessary to keep that delay ? can we remove that delay ?

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -248,25 +247,27 @@ def test_init_error(run_lambda_function):
...
         def test_handler(event, context):
             return 0
         """
         ),
         b'{"foo": "bar"}',
+        syntax_check=False,
     )
     log_result = (base64.b64decode(response["LogResult"])).decode("utf-8")
     expected_text = "name 'func' is not defined"
     assert expected_text in log_result
```
shantanu73, line 260: @rhcarvalho @untitaker In this particular test case as well, there is no events data which was coming and thus only comparison was with log_result that we get from response.
Now, I've figured out why this event data is not coming as well. It is because in the initialization error wrapper that we have currently written, it takes the execution info directly from the system (sys.exc_info) and not from an exact exception which is generated. So, we actually need to generate an exception as we only receive data from (sys.exc_info) in the post_init_error method and not the actual exception. A simple solution is reraise the exception from the (sys.exc_info) and then capture that exception information, which solves all the problems. I've tested and verified it. I'll tag you guys at the exact method I'm talking about.

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -28,8 +29,8 @@
     F = TypeVar("F", bound=Callable[..., Any])
 
 # Constants
-TIMEOUT_THRESHOLD_MILLIS = 1500  # Minimum time required to capture TimeoutError
-SECONDS_CONVERSION_FACTOR = 1000.0
+TIMEOUT_WARNING_BUFFER = 1500  # Buffer time required to send timeout warning to Sentry
+MILLIS_TO_SECONDS = 1000.0
 
 
 def _wrap_init_error(init_error):
```
shantanu73, line 36: @rhcarvalho @untitaker in this method see line no. 57, the event is not captured from an actual exception.

shantanu73: I'm suggesting we make this change below (this change works):-

            `.if error.get("errorType"):
            .    try:
            .        exc_info = sys.exc_info()
            .        reraise(*exc_info)
            .    except Exception:
            .        exc_info = sys.exc_info()
            .        event, hint = event_from_exception(
            .            exc_info,
            .            client_options=client.options,
            .            mechanism={"type": "aws_lambda", "handled": False},
            .        )
            .    hub.capture_event(event, hint=hint)`

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -22,13 +22,16 @@
 import json
 from sentry_sdk.transport import HttpTransport
 
+FLUSH_EVENT = True
```
shantanu73, line 25: @rhcarvalho @untitaker defined a global constant to parameterize timeout error to capture events data.

untitaker: yeah that works for me

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -25,6 +27,45 @@
...
+    def sentry_init_error(*args, **kwargs):
+        # type: (*Any, **Any) -> Any
+
+        hub = Hub.current
+        integration = hub.get_integration(AwsLambdaIntegration)
+        if integration is None:
+            return init_error(*args, **kwargs)
+
+        # Fetch Initialization error details from arguments
+        error = json.loads(args[1])
```
rhcarvalho, line 46: I'd be more defensive here. This will blow up if called with less than two positional arguments or if the arg is not valid JSON.

Can be done in a follow up.

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -158,19 +226,23 @@ def inner(*args, **kwargs):
             )
 
 
-def _make_request_event_processor(aws_event, aws_context):
-    # type: (Any, Any) -> EventProcessor
+def _make_request_event_processor(aws_event, aws_context, configured_timeout):
+    # type: (Any, Any, Any) -> EventProcessor
```
rhcarvalho, line 230: The type here is well-known, `float`.

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -22,20 +22,23 @@
...
+
 class TestTransport(HttpTransport):
     def _send_event(self, event):
         # Delay event output like this to test proper shutdown
         # Note that AWS Lambda truncates the log output to 4kb, so you better
         # pray that your events are smaller than that or else tests start
         # failing.
-        time.sleep(1)
+        if FLUSH_EVENT:
+            time.sleep(1)
```
rhcarvalho, line 34: Hard to understand for maintenance why `FLUSH_EVENT` toggles calling `time.sleep(1)`.

I'd think it would control calling `sentry.flush()` 😕 

Might be a matter of naming?

PR #759, opened 2020-07-10
Use sentry_init fixture in tests instead of using Hub directly

alexmojaki:
@untitaker how's this?

Based on https://github.com/getsentry/sentry-python/pull/749#issuecomment-656827764

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -7,7 +7,14 @@ import sys
 import time
 
 from textwrap import dedent
-from sentry_sdk import Hub, Client, configure_scope, capture_message, capture_exception
+from sentry_sdk import (
+    Hub,
+    Client,
+    configure_scope,
+    capture_message,
+    capture_exception,
+    capture_event,
+)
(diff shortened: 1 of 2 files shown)


tests/test_client.py
```diff
@@ -149,41 +156,41 @@ def test_proxy_httpsselect_bothenv_http(monkeypatch):
     assert client.transport._pool.proxy.scheme == "http"
 
 
-def test_simple_transport():
+def test_simple_transport(sentry_init):
     events = []
```
untitaker, line 160: You do not need to pass in a transport if you're using sentry_init, just use the `capture_events` fixture instead which does the same thing.

alexmojaki: I did that at first but then decided that wasn't in the spirit of the original test and the name `test_simple_transport`.

untitaker: ah well you're right

untitaker:
Thanks! Not entirely sure why project coverage is still there. I think I disabled it.

PR #761, opened 2020-07-14
Add range of executing node to frames

alexmojaki:
Part of #748 

When ExecutingIntegration is enabled, each frame in an exception traceback now contains new data that looks like this:

```python
'executing_node': {'start': {'line': 19, 'column': 12}, 'end': {'line': 19, 'column': 17}}
```

This can be used to highlight exactly what the frame was doing, rather than just a line. Actually doing this will also require work on the server.

The line numbers may actually go beyond the amount of context that's been included, I don't know if that's something we want to address at this stage.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -15,7 +15,6 @@ from sentry_sdk import (
     capture_exception,
     capture_event,
 )
-from sentry_sdk.integrations.executing import ExecutingIntegration
 from sentry_sdk.transport import Transport
 from sentry_sdk._compat import reraise, text_type, PY2
 from sentry_sdk.utils import HAS_CHAINED_EXCEPTIONS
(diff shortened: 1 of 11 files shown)


tests/test_client.py
```diff
@@ -217,35 +216,6 @@ def test_with_locals_disabled(sentry_init, capture_events):
     )
 
 
-@pytest.mark.parametrize("integrations", [[], [ExecutingIntegration()]])
```
untitaker, line 220: Oh damn, thanks! Seems like I overlooked this.

alexmojaki:
I think this and https://github.com/getsentry/sentry-python/issues/748 might as well be left closed unless https://github.com/getsentry/sentry/pull/19924 is going to be reopened.

sl0thentr0py:
thx @alexmojaki, I'm just bringing this repo back to life and had to do some bulk changes, sorry for the confusion/notification spam!

PR #766, opened 2020-07-23
fix: Fix AWS Lambda under Python 3.8 and refactor test setup code

### LINKED ISSUE #764: AWS Lambda under Python 3.8 is currently broken
It seems they have done some updates on the runtime that broke our integration.

untitaker:
This reverts commit 1737ba8cc7fb3461bbe2ccab22532186f812e328.

Fix #764


sentry_sdk/integrations/aws_lambda.py
```diff
@@ -140,25 +135,39 @@ def __init__(self, timeout_warning=False):
...
-        pre_37 = True  # Python 3.6 or 2.7
-
-        if not hasattr(lambda_bootstrap, "handle_http_request"):
-            try:
-                import bootstrap as lambda_bootstrap  # type: ignore
 
-                pre_37 = False  # Python 3.7
-            except ImportError:
-                pass
+        # Python 2.7: Everything is in `__main__`.
```
untitaker, line 139: This is the crucial bit, where we do not attempt to import but rather see what's imported

PR #767, opened 2020-07-24
Enable CodeQL

dcramer:


dcramer:
welp no idea why is nots running, dont care enough to figure out... will try out another time

PR #779, opened 2020-08-10
feat: Integration for Chalice

Gleekzone:
This is a proposal for the integration of Chalice in Sentry


sentry_sdk/integrations/chalice.py
```diff
@@ -0,0 +1,182 @@
...
+
+class ChaliceIntegration(Integration):
+    identifier = "chalice"
+
+    @staticmethod
+    def setup_once():
+        # for @app.route()
+        Chalice._get_view_function_response = _get_view_function_response
+        # for everything else (like events)
+        AwsLambdaIntegration.setup_once()
```
untitaker, line 86: Would just document that one needs to enable this. Unfortunately we have no way to enforce this programmatically right now. What you're doing here will enable the integration twice, meaning double the instrumentation overhead.

sentry_sdk/integrations/chalice.py
```diff
@@ -0,0 +1,182 @@
...
+        # type: (Event, Hint, datetime) -> Optional[Event]
+        extra = event.setdefault("extra", {})
+        extra["lambda"] = {
+            "function_name": lambda_context.function_name,
+            "function_version": lambda_context.function_version,
+            "invoked_function_arn": lambda_context.invoked_function_arn,
+            "aws_request_id": lambda_context.aws_request_id,
+        }
+
+        extra["cloudwatch logs"] = {
```
untitaker, line 103: This should not be necessary if the aws lambda integration is enabled

test-requirements.txt
```diff
@@ -13,3 +13,5 @@ dnspython<2.0
 newrelic
 executing
 asttokens
+chalice
```
untitaker, line 16: Please put this stuff into tox.ini, i.e. create a new job for your integration (+ testing different versions you want to support)

sentry_sdk/integrations/chalice.py
Gleekzone: i actually found a better way to catch exceptions in events running in chalice, which doesn't require adding the aws lambda integration https://github.com/getsentry/sentry-python/blob/f1443129021cd1d4a61425b293eda49ec712200f/sentry_sdk/integrations/chalice.py#L113

sentry_sdk/integrations/chalice.py
Gleekzone: this is now addressed by the change mentioned in the comment above https://github.com/getsentry/sentry-python/pull/779#discussion_r470299614

test-requirements.txt
Gleekzone: done!

sentry_sdk/integrations/chalice.py
untitaker: Can we still deduplicate the logic of adding cloudwatch logs in both chalice and aws lambda? I.e. make sure that we write the same key in extra (so we don't accidentally send the same info twice)

Ideally you'd call the event processor from aws lambda in here.

sentry_sdk/integrations/chalice.py
```diff
@@ -0,0 +1,157 @@
...
+            exc_info = sys.exc_info()
+            event, hint = event_from_exception(
+                exc_info,
+                client_options=client.options,
+                mechanism={"type": "chalice", "handled": False},
+            )
+            hub.capture_event(event, hint=hint)
+            hub.flush()
+            headers = {}
+            app.log.error("Caught exception for %s", view_function, exc_info=True)
```
untitaker, line 81: I really don't think the SDK is supposed to do that. Just re-raise the exception.

tox.ini
```diff
@@ -69,8 +69,9 @@ envlist =
     {py2.7,py3.7,py3.8}-sqlalchemy-{1.2,1.3}
 
     py3.7-spark
-
+
```
untitaker, line 72: Please undo this change

sentry_sdk/integrations/__init__.py
```diff
@@ -62,7 +62,7 @@ def iter_default_integrations(with_auto_enabling_integrations):
     "sentry_sdk.integrations.rq.RqIntegration",
     "sentry_sdk.integrations.aiohttp.AioHttpIntegration",
     "sentry_sdk.integrations.tornado.TornadoIntegration",
-    "sentry_sdk.integrations.sqlalchemy.SqlalchemyIntegration",
```
untitaker, line 65: Please undo this change

tests/integrations/chalice/test_chalice.py
```diff
@@ -0,0 +1,53 @@
...
+
+from pytest_chalice.handlers import RequestHandler
+
+
+SENTRY_DSN = "https://111@sentry.io/111"
+
+
+@pytest.fixture
+def app():
+    sentry_sdk.init(dsn=SENTRY_DSN, integrations=[chalice_sentry.ChaliceIntegration()])
```
untitaker, line 18: Use the `sentry_init` pytest fixture instead, then you dont need to pass a bogus DSN

sentry_sdk/integrations/chalice.py
Gleekzone: i made a monkeypatch of the '_get_view_function_response' function from the chalice, in order to keep the uniformity of the code (a good practice of chalice), I keep this line https://github.com/aws/chalice/blob/2e44fe3b63378b488af88bd92254c58708cb9ac2/chalice/app.py#L1123

tox.ini
Gleekzone: done!

sentry_sdk/integrations/__init__.py
Gleekzone: done!

tests/integrations/chalice/test_chalice.py
Gleekzone: done!

sentry_sdk/integrations/chalice.py
Gleekzone: This is resolved adding the event processor from aws lambda as you suggested me

sentry_sdk/integrations/chalice.py
untitaker: Oh that's not great, copypasting the code. What about wrapping view_function, and passing the wrapped function into the original implementation of `_get_view_function_response`?

untitaker (changes requested):
Can you resolve the last issue about copypasted code from chalice, and add a `__init__` in `tests/integrations/chalice` that calls `importorskip` (like in other integrations), otherwise the base testsuite does not pass (see travis CI)

sentry_sdk/integrations/chalice.py
untitaker: Why don't you have access to the app context?

```python
def _get_view_function_response(app, view_function, ...):
    def new_view_function(**kw):
        # you can access app.lambda_context here
        try:
            return view_function(**kw)
        except Exception:
            ...

    return old_get_view_function_response(app, new_view_function, ...)

File: sentry_sdk/integrations/chalice.py [hunk 8]
```diff
@@ -0,0 +1,105 @@
...
+                scope.transaction = app.lambda_context.function_name
+                scope.add_event_processor(
+                    _make_request_event_processor(
+                        app.current_request.to_dict(),
+                        app.lambda_context,
+                        configured_time,
+                    )
+                )
+            exc_info = sys.exc_info()
+            event, hint = event_from_exception(
```
untitaker, line 72: Please add the event processor unconditionally, not only when an exception is fired. If the user calls `capture_message` we also want data to be attached

Gleekzone: Done!

sentry_sdk/integrations/chalice.py
Gleekzone: I solved the last problem about the code copied from chalice and this is my suggestion https://github.com/getsentry/sentry-python/blob/59f092d760da9ef888616e3c9cb7a53ece382f73/sentry_sdk/integrations/chalice.py#L48 and https://github.com/getsentry/sentry-python/blob/59f092d760da9ef888616e3c9cb7a53ece382f73/sentry_sdk/integrations/chalice.py#L69, could be another way to wrap the original function _get_view_function_response.

sentry_sdk/integrations/chalice.py
```diff
@@ -0,0 +1,93 @@
...
+                    exc_info,
+                    client_options=client.options,
+                    mechanism={"type": "chalice", "handled": False},
+                )
+                hub.capture_event(event, hint=hint)
+                hub.flush()
+                reraise(*exc_info)
+
+
+old_get_view_function_response = Chalice._get_view_function_response
```
untitaker, line 44: Please don't have a global backup of the view function response, this will break if multiple libraries monkeypatch chalice

Gleekzone: add this in setup_once().

Gleekzone: Regarding the tests, these fail because the parameters of the context '_max_runtime' arrive null, so this line is not executed, the response if it contains the status_code '500' and the json, I don't understand why it fails in travis.

untitaker:
Tried to fix conflicts but was denied pushing. Continuing at https://github.com/getsentry/sentry-docs/pull/2220

untitaker (approved):
almost there, please fix the linting errors (run `make lint` locally)

matin:
@untitaker you now have permissions to push to our fork.

We're already using [`sentry-chalice`](https://github.com/cuenca-mx/sentry-chalice) in multiple projects. It'll be great switch to the official `sentry_sdk` library!

matin:
fwiw, Chalice _does_ have a stub file for type hinting, [`app.pyi`](https://github.com/aws/chalice/blob/master/chalice/app.pyi), and includes [`py.typed`](https://github.com/aws/chalice/blob/master/chalice/py.typed).

It shouldn't be necessary to ignore the import to resolve this error: https://travis-ci.com/github/getsentry/sentry-python/jobs/382832642#L249-L251

untitaker:
Yes but we can't install all optional dependencies at once when type
checking. Also we should not rely on those stubs as they are wrong (eg
flask, celery) or change between versions.

Technically we would have to typecheck for each entry in the cross product
of all versions of frameworks we support.

PR #782, opened 2020-08-11
feat: Avoid truncating span descriptions

rhcarvalho:
For database auto-instrumented spans, the description contains potentially long SQL queries that are most useful when not truncated.

Because arbitrarily large events may be discarded by the server as a protection mechanism, we dynamically limit the description length, preserving the most important descriptions/queries.

**Performance impact**

Preliminary CPU profiling using [1] suggests that uuid4() dominates the execution time for code sending many transactions sequentially.

Preliminary memory profiling using [2] and looking at the max RSS of a benchmark script suggests that the max RSS has no significant change (JSON encoding in CPython is implemented in C).

In any case, we mitigate any increase in memory usage and run time for the majority of cases by avoiding any extra work when the total number of bytes consumed by descriptions do not exceed ~512 KB, which is equivalent to having the standard string truncation applied.

Integrating profiling to the SDK is left for a future PR.

[1]\: https://pypi.org/project/zprofile/
[2]\: `/usr/bin/time -l` (macOS)

---

Related to https://github.com/getsentry/relay/pull/674

### DIFF
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -6,28 +6,36 @@ from sentry_sdk.utils import (
     AnnotatedValue,
     capture_internal_exception,
     disable_capture_event,
+    format_timestamp,
+    json_dumps,
     safe_repr,
     strip_string,
-    format_timestamp,
 )
+
+import sentry_sdk.utils
 
 from sentry_sdk._compat import text_type, PY2, string_types, number_types, iteritems
 
 from sentry_sdk._types import MYPY
 
 if MYPY:
+    from datetime import timedelta
+
     from types import TracebackType
 
     from typing import Any
+    from typing import Callable
+    from typing import ContextManager
     from typing import Dict
     from typing import List
     from typing import Optional
-    from typing import Callable
-    from typing import Union
-    from typing import ContextManager
+    from typing import Tuple
     from typing import Type
+    from typing import Union
 
     from sentry_sdk._types import NotImplementedType, Event
+
+    Span = Dict[str, Any]
 
     ReprProcessor = Callable[[Any, Dict[str, Any]], Union[NotImplementedType, str]]
     Segment = Union[str, int]
@@ -47,6 +55,17 @@ else:
 
     # Bytes are technically not strings in Python 3, but we can serialize them
     serializable_str_types = (str, bytes)
+
+
+# Maximum length of JSON-serialized event payloads that can be safely sent
+# before the server may reject the event due to its size. This is not intended
+# to reflect actual values defined server-side, but rather only be an upper
+# bound for events sent by the SDK.
+#
+# Can be overwritten if wanting to send more bytes, e.g. with a custom server.
+# When changing this, keep in mind that events may be a little bit larger than
+# this value due to attached metadata, so keep the number conservative.
+MAX_EVENT_BYTES = 10 ** 6
 
 MAX_DATABAG_DEPTH = 5
 MAX_DATABAG_BREADTH = 10
@@ -98,6 +117,7 @@ def serialize(event, **kwargs):
     memo = Memo()
     path = []  # type: List[Segment]
     meta_stack = []  # type: List[Dict[str, Any]]
+    span_description_bytes = []
 
     def _annotate(**meta):
         # type: (**Any) -> None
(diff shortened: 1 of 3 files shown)


sentry_sdk/serializer.py
```diff
@@ -48,6 +56,17 @@
...
+
+# Maximum length of JSON-serialized event payloads that can be safely sent
+# before the server may reject the event due to its size. This is not intended
+# to reflect actual values defined server-side, but rather only be an upper
+# bound for events sent by the SDK.
+#
+# Can be overwritten if wanting to send more bytes, e.g. with a custom server.
+# When changing this, keep in mind that events may be a little bit larger than
+# this value due to attached metadata, so keep the number conservative.
+MAX_EVENT_BYTES = 10 ** 6
```
untitaker, line 68: Can we clarify what actually happens on the server: That the 200kb limit is applied on teh compressed payload etc. (I get that this is not supposed to reflect that)

sentry_sdk/serializer.py
```diff
@@ -98,6 +117,7 @@ def serialize(event, **kwargs):
     memo = Memo()
     path = []  # type: List[Segment]
     meta_stack = []  # type: List[Dict[str, Any]]
+    span_description_bytes = []
```
untitaker, line 120: I think this can be a simple integer, that would probably lower memory usage a tiny little bit. Since we don't have `nonlocal` you can still use a fixed-length list like this:

```
span_description_bytes[0] += value
```

instead of:

```
span_description_bytes.append(value)
```

rhcarvalho: I've considered both; `nonlocal` is not an option in Python 2.7 and among list with single int or adhoc class with mutable field I thought list of lengths is the least surprising option. This piece of code is already super hard to read and understand, a list of ~999 ints vs list of one int makes no practical difference.

sentry_sdk/serializer.py
rhcarvalho: https://develop.sentry.dev/sdk/envelopes/#size-limits

The 200KB limit is for `/store/`, IIUC `/envelope/` is limited to 20MB compressed, 1MB decompressed (for error events and transactions). So the bottleneck is 1MB decompressed.

I'd be fine lowering the number here if you think we need to be even more conservative.

Would you have a more specific suggestion, what to add here?

rhcarvalho:
> if possible I would make this an _experiment flag for now

Could you help me with that?

I guess two if's around `serializer.py` would do it.

sentry_sdk/serializer.py
```diff
@@ -323,14 +343,108 @@ def _serialize_node_impl(
...
+        # When truncating a description, preserve a small prefix.
+        min_length = 10
+
+        def shortest_duration_longest_description_first(args):
+            # type: (Tuple[int, Span]) -> Tuple[timedelta, int]
+            i, serialized_span = args
+            span = event["spans"][i]
+            now = datetime.utcnow()
+            start = span.get("start_timestamp") or now
+            end = span.get("timestamp") or now
```
AbhiPrasad, line 384: Is `now` the most correct here? How is it even possible these can be nil by the time we hit the serializer?

I assume if we get a span without a valid start_timestamp or timestamp we should disregard it right?

sentry_sdk/serializer.py
untitaker: I get the argument about readability but this is an extremely hot code path. Still you're right that this is unlikely to regress perf

sentry_sdk/serializer.py
untitaker: forgot about envelopes, ignore

sentry_sdk/serializer.py
rhcarvalho: It is there for the sake of defensive coding. Is it not the role of the serializer to validate spans.

Note that `now` is not used to update the span, only to compute a `timedelta` for sorting.

sentry_sdk/serializer.py
```diff
@@ -323,14 +343,113 @@ def _serialize_node_impl(
...
+        del meta_stack[len(path) + 1 :]
+
     disable_capture_event.set(True)
     try:
         rv = _serialize_node(event, **kwargs)
         if meta_stack and isinstance(rv, dict):
             rv["_meta"] = meta_stack[0]
 
+        sum_span_description_bytes = sum(span_description_bytes)
+        if sum_span_description_bytes > 0:
```
rhcarvalho, line 424: It would be clearer if we also checked for `smart_transaction_trimming` here. It is just a side-effect of `smart_transaction_trimming` being false that `span_description_bytes` is never changed and thus `sum_span_description_bytes` is 0

rhcarvalho: Updated.

sentry_sdk/serializer.py
AbhiPrasad: For sure, I see that. Maybe this is out of scope for the PR, but I wonder if we can make the guarantee that the spans are valid before hit hits the serializer. That way we don't need the extra func call.

Also, if `start` gets now, and `end` uses it's timestamp, wouldn't you get a negative timedelta?

rhcarvalho: > Also, if `start` gets now, and `end` uses it's timestamp, wouldn't you get a negative timedelta?

The original start and end timestamps could also give a negative duration. That will only affect the sort order.

PR #785, opened 2020-08-12
Added a new integration for Google Cloud Functions and automation test case framework

shantanu73:
Changes:

1) Added new GCP Integration for Google Cloud Functions.
2) Added event processor for this integration.
3) Added code for sending timeout warning.
4) Added automation test cases for the integration.
5) Modified tox.ini file for running automation test cases for GCP Integration.
6) Added code to capture timeout warning as well.

### DIFF
--- a/sentry_sdk/integrations/gcp.py
+++ b/sentry_sdk/integrations/gcp.py
@@ -1,0 +1,68 @@
+import sys
+
+from sentry_sdk.hub import Hub
+from sentry_sdk._compat import reraise
+from sentry_sdk.utils import (
+    capture_internal_exceptions,
+    event_from_exception,
+)
+from sentry_sdk.integrations import Integration
+
+from sentry_sdk._types import MYPY
+
+
+if MYPY:
+    from typing import Any
+    from typing import TypeVar
+    from typing import Callable
+
+    F = TypeVar("F", bound=Callable[..., Any])
+
+
+def _wrap_func(func):
+    # type: (F) -> F
+    def sentry_func(*args, **kwargs):
+        # type: (*Any, **Any) -> Any
+
+        hub = Hub.current
+        integration = hub.get_integration(GcpIntegration)
+        if integration is None:
+            return func(*args, **kwargs)
+
+        # If an integration is there, a client has to be there.
+        client = hub.client  # type: Any
+
+        with hub.push_scope() as scope:
+            with capture_internal_exceptions():
+                scope.clear_breadcrumbs()
+            try:
+                return func(*args, **kwargs)
+            except Exception:
+                exc_info = sys.exc_info()
+                event, hint = event_from_exception(
+                    exc_info,
+                    client_options=client.options,
+                    mechanism={"type": "gcp", "handled": False},
+                )
+                hub.capture_event(event, hint=hint)
+                reraise(*exc_info)
+            finally:
+                # Flush out the event queue
+                hub.flush()
+
+    return sentry_func  # type: ignore
+
+
+class GcpIntegration(Integration):
+    identifier = "gcp"
+
+    @staticmethod
+    def setup_once():
+        # type: () -> None
+        import __main__ as gcp_functions  # type: ignore
+
+        worker1 = gcp_functions.worker_v1
+
+        worker1.FunctionHandler.invoke_user_function = _wrap_func(
+            worker1.FunctionHandler.invoke_user_function
+        )
--- a/sentry_sdk/integrations/gcp.py
+++ b/sentry_sdk/integrations/gcp.py
@@ -42,7 +42,13 @@ def _wrap_func(func):
         # If an integration is there, a client has to be there.
         client = hub.client  # type: Any
 
-        configured_time = int(environ.get("FUNCTION_TIMEOUT_SEC"))
+        configured_time = environ.get("FUNCTION_TIMEOUT_SEC")
+        if not configured_time:
+            logger.debug("The configured timeout could not be fetched from Cloud Functions configuration.")
+            return func(*args, **kwargs)
+
+        configured_time = int(configured_time)
+
         initial_time = datetime.now()
 
         with hub.push_scope() as scope:
--- a/sentry_sdk/integrations/gcp.py
+++ b/sentry_sdk/integrations/gcp.py
@@ -44,7 +44,9 @@ def _wrap_func(func):
 
         configured_time = environ.get("FUNCTION_TIMEOUT_SEC")
         if not configured_time:
-            logger.debug("The configured timeout could not be fetched from Cloud Functions configuration.")
+            logger.debug(
+                "The configured timeout could not be fetched from Cloud Functions configuration."
+            )
             return func(*args, **kwargs)
 
         configured_time = int(configured_time)
--- a/sentry_sdk/integrations/gcp.py
+++ b/sentry_sdk/integrations/gcp.py
@@ -1,3 +1,5 @@
+from datetime import datetime, timedelta
+from os import environ
 import sys
 
 from sentry_sdk.hub import Hub
@@ -5,16 +7,23 @@ from sentry_sdk._compat import reraise
 from sentry_sdk.utils import (
     capture_internal_exceptions,
     event_from_exception,
+    TimeoutThread,
 )
 from sentry_sdk.integrations import Integration
 
 from sentry_sdk._types import MYPY
 
+# Constants
+TIMEOUT_WARNING_BUFFER = 1.5  # Buffer time required to send timeout warning to Sentry
+MILLIS_TO_SECONDS = 1000.0
 
 if MYPY:
     from typing import Any
     from typing import TypeVar
     from typing import Callable
+    from typing import Optional
+
+    from sentry_sdk._types import EventProcessor, Event, Hint
 
     F = TypeVar("F", bound=Callable[..., Any])
 
@@ -32,10 +41,27 @@ def _wrap_func(func):
         # If an integration is there, a client has to be there.
         client = hub.client  # type: Any
 
+        configured_time = int(environ.get("FUNCTION_TIMEOUT_SEC"))
+        initial_time = datetime.now()
+
         with hub.push_scope() as scope:
             with capture_internal_exceptions():
                 scope.clear_breadcrumbs()
+                scope.transaction = environ.get("FUNCTION_NAME")
+                scope.add_event_processor(
+                    _make_request_event_processor(configured_time, initial_time)
+                )
             try:
+                if (
+                    integration.timeout_warning
+                    and configured_time > TIMEOUT_WARNING_BUFFER
+                ):
+                    waiting_time = configured_time - TIMEOUT_WARNING_BUFFER
+
+                    timeout_thread = TimeoutThread(waiting_time, configured_time)
+
+                    # Starting the thread to raise timeout warning exception
+                    timeout_thread.start()
                 return func(*args, **kwargs)
             except Exception:
                 exc_info = sys.exc_info()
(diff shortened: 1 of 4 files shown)


test-requirements.txt
```diff
@@ -4,6 +4,10 @@ tox==3.7.0
 Werkzeug==0.15.5
 pytest-localserver==0.5.0
 pytest-cov==2.8.1
+google-api-python-client==1.10.0
```
untitaker, line 7: Please put those in tox.ini like the boto3 dependency

tests/integrations/gcp/test_gcp.py
```diff
@@ -0,0 +1,320 @@
...
+    if "SENTRY_PYTHON_TEST_GCP_CREDENTIALS_JSON" not in os.environ:
+        pytest.skip("GCP environ vars not set")
+
+    # The file token.pickle stores the user's access and refresh tokens, and is
+    # created automatically when the authorization flow completes for the first
+    # time.
+    with open(
+        os.environ.get("SENTRY_PYTHON_TEST_GCP_CREDENTIALS_JSON"), "rb"
+    ) as creds_file:
+        for line in creds_file.readlines():
```
untitaker, line 79: We will not be able to store any files in CI. Can we store the entire JSON in an envvar?

Also do I read it correctly that you're expecting that file to contain multiple JSON objects where you're only interested in the last one? Because you're splitting by line separator and override `creds_json` in each iteration.

tests/integrations/gcp/test_gcp.py
```diff
@@ -0,0 +1,320 @@
...
+            creds_json = json.loads(line)
+    project_id = creds_json.get("installed", {}).get("project_id")
+    if not project_id:
+        pytest.skip("Credentials json file is not valid")
+
+    if os.path.exists("token.pickle"):
```
untitaker, line 85: Can we avoid using pickle? If possible I'd just stick to JSON.

tests/integrations/gcp/test_gcp.py
```diff
@@ -0,0 +1,320 @@
...
+            functions_service.projects()
+            .locations()
+            .functions()
+            .create(location=location, body=body)
+        )
+        api_request.execute()
+
+        # STEP : Invoke the cloud function
+        # Adding delay of 60 seconds for new created function to get deployed.
+        time.sleep(60)
```
untitaker, line 199: That is an awfully slow test for CI. Is there a way around this, such as polling the deployment status?

tests/integrations/gcp/test_gcp.py
```diff
@@ -0,0 +1,320 @@
...
+            entry_text_payload = entry.get("textPayload", "")
+            if (
+                entry_log_name == log_name
+                and entry_function_name == function_name
+                and "EVENTS: " in entry_text_payload
+            ):
+                event = entry_text_payload[len("EVENTS: ") :]
+                events.append(json.loads(event))
+
+        log_flag = True
```
untitaker, line 231: This flag never gets set to `False` so I think it cannot serve a purpose.

tests/integrations/gcp/test_gcp.py
shantanu73: Actually it is only a single json object which is there in the file I'm just reading the file line by line and fetching one single json. 
The thing is that, these oauth client credentials are downloaded from  as a .json file. I was just setting this environment variable to that json which is stored in the file.
This will be a minor overhead on user to copy the json from the file and set it to the environment variable. Do we want to do that? If so, I'll modify this implementation.

tests/integrations/gcp/test_gcp.py
shantanu73: The thing is that, the pickle basically is used so that user doesn't have to re-authenticate its credentials from the browser for every test case. 
So, do we want to do that or keep this usage?

tests/integrations/gcp/test_gcp.py
shantanu73: I need to check if polling deployment status is possible or not. Although, GCP deployments do take an awful lot of time. I've checked and figured that this is a stable minimum time which is needed for the deployment to finish, (the minimum time which I observed was 45 seconds nothing less than that).
So, I'll check if it is possible to poll and verify the status of deployment, but the test cases will still be slow.

tests/integrations/gcp/test_gcp.py
shantanu73: Sorry, this was a minor mistake, I'll correct it, On line no. 255 it actually needs to be set to false.

tests/integrations/gcp/test_gcp.py
untitaker: Is there a way we could reuse existing deployments then perhaps?

tests/integrations/gcp/test_gcp.py
untitaker: no I would just replace `pickle.dumps` with `json.dumps` or something like that

tests/integrations/gcp/test_gcp.py
untitaker: My main point was that it's easy to set secrets in Travis using environment variables but I may have misunderstood the auth flow.

Reading your other comment is there even a way to run this in CI if there's manual user interaction involved?

test-requirements.txt
shantanu73: Okay sure, I'll update this.

tests/integrations/gcp/test_gcp.py
shantanu73: I'm not completely sure about this, though what I can do is try to give all permissions from my google account to the APIs which are using the credentials. But I think for the first run atleast it still requires user interaction.

tests/integrations/gcp/test_gcp.py
shantanu73: This .pickle file contains actually the authentication token and not json. Since we are using that authentication token to authenticate our requests, I feel this can't be avoided. If we avoid this .pickle file, we'll have to verify each test case run through the browser.

tests/integrations/gcp/test_gcp.py
shantanu73: Well, that is something I haven't check yet and I'm not completely sure about this. 
Although, even if a small change is there in a particular file in cloud function we need to redeploy the code. That's the reason building this integration took so much more time as compared to AWS. I'm not completely aware if there is a separate process than this.

tests/integrations/gcp/test_gcp.py
untitaker: I understand the purpose of the file but I don't understand why we need to use pickle as a serialization format specifically. It seems to me that we could use any other format, like yaml or json.

shantanu73: okay, I got your point, I'll try to do it with json serialization.

tox.ini
```diff
@@ -44,6 +44,9 @@ envlist =
     # The aws_lambda tests deploy to the real AWS and have their own matrix of Python versions.
     py3.7-aws_lambda
 
+    # The gcp deploy to the real AWS and have their own matrix of Python versions.
+    py3.7-gcp
```
shantanu73, line 48: @untitaker I've pushed new changes regarding GCP test integration, does this look fine?

rhcarvalho (commented):
Summary from call with @shantanu73:

- [ ] Fix linting error
- [ ] Fix missing test dependency (tox env config)
- [ ] Document limitation that the current solution only works on Python 3.7 and not on 3.8 / disable the integration if the environment is unsupported to avoid breaking user functions

For now tests will be skipped in CI. As a follow up, convert the tests to use https://github.com/GoogleCloudPlatform/functions-framework-python to implement `run_cloud_function` so that we don't need to deal with credentials nor network calls.

sentry_sdk/integrations/gcp.py
```diff
@@ -0,0 +1,168 @@
...
+
+        hub = Hub.current
+        integration = hub.get_integration(GcpIntegration)
+        if integration is None:
+            return func(*args, **kwargs)
+
+        # If an integration is there, a client has to be there.
+        client = hub.client  # type: Any
+
+        configured_time = int(environ.get("FUNCTION_TIMEOUT_SEC"))
```
rhcarvalho, line 45: This is [failing in `mypy`](https://travis-ci.com/github/getsentry/sentry-python/jobs/375319774) because `environ.get` may return `None` if the env var is not set.

```py
In [1]: import os

In [2]: os.environ.get("FUNCTION_TIMEOUT_SEC")

In [3]: int(os.environ.get("FUNCTION_TIMEOUT_SEC"))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-d034f2d5d55a> in <module>
----> 1 int(os.environ.get("FUNCTION_TIMEOUT_SEC"))

TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

We need to handle the case in which the variable is not set.

If we don't know the configured timeout, we cannot start the `TimeoutThread`.

I suggest handling this by writing a debug log and then ignoring the rest of the wrapper, similar to when the integration is not in the current hub, as above.

shantanu73: Okay, sure. I'll make this change.

shantanu73: @rhcarvalho modified this.

rhcarvalho:
@shantanu73 now the linting job fails because of code formatting -- https://travis-ci.com/github/getsentry/sentry-python/jobs/375368281

You can run the lint job locally before submitting your commit with `make lint` and you can update the code formatting with `make format`. That's faster than waiting for Travis 😉 

There are still import errors in the test jobs: https://travis-ci.com/github/getsentry/sentry-python/jobs/375368274

You can run them locally / partially using `tox -e [ENVIRONMENT_NAME]`, or `scripts/runtox.sh` (this is what is used in Travis, see `.travis.yml` file) to test the fixes locally.

sentry_sdk/integrations/gcp.py
shantanu73: @rhcarvalho Maybe bcoz f the last check-in the linting got failed. I'll use make format and make lint. Though I do follow this as standard practice.

I'll check for the import errors in test jobs.

shantanu73: @rhcarvalho I'm a little doubtful on updating of tox.ini file. I've updated it as per the reference from aws_lambda. But, it still gave the import error.
Are my changes appropriate?
Also, I've tried running the commands you suggested, it took a lot of time to run but didn't complete.

untitaker: currently CI fails because you are not declaring the dependency on `requests`

untitaker: you are missing an `tests/integrations/aws_lambda/__init__.py` which would skip the tests if the dependency is not installed (in other jobs)

shantanu73: @untitaker okay, so I need a similar file for gcp tests right?

tests/integrations/gcp/test_gcp.py
```diff
@@ -5,21 +5,22 @@
...
 import pytest
-import pickle
 import os.path
 import os
 from googleapiclient.discovery import build
 from google_auth_oauthlib.flow import InstalledAppFlow
 from google.auth.transport.requests import Request
 
+requests = pytest.importorskip("requests")
+pickle = pytest.importorskip("pickle")
```
untitaker, line 19: pickle is stdlib, this should not be necessary

tox.ini
untitaker: yes

tests/integrations/gcp/test_gcp.py
```diff
@@ -0,0 +1,384 @@
...
+import tempfile
+import shutil
+import sys
+import subprocess
+
+import pytest
+import os.path
+import os
+
+requests = pytest.importorskip("requests")
```
untitaker, line 18: I would importorskip the google cloud sdk additionally

tests/integrations/gcp/test_gcp.py
shantanu73: okay, I'll change this.

tests/integrations/gcp/test_gcp.py
shantanu73: Sure I'll do that as well.

untitaker (approved):

PR #787, opened 2020-08-12
feat(django): Instrument views as spans

untitaker:
This adds spans around URL resolving and the view itself.

### DIFF
--- a/sentry_sdk/integrations/django/views.py
+++ b/sentry_sdk/integrations/django/views.py
@@ -1,9 +1,17 @@
 import functools
 
 from sentry_sdk.hub import Hub
+from sentry_sdk._types import MYPY
+
+if MYPY:
+    from typing import Any
+
+    from django.core.handlers.wsgi import WSGIRequest
+    from django.urls.resolvers import ResolverMatch
 
 
 def patch_resolve_request():
+    # type: () -> None
     from django.core.handlers.base import BaseHandler
     from sentry_sdk.integrations.django import DjangoIntegration
 
@@ -14,6 +22,7 @@ def patch_resolve_request():
         return
 
     def resolve_request(self, request):
+        # type: (BaseHandler, WSGIRequest) -> ResolverMatch
         hub = Hub.current
         integration = hub.get_integration(DjangoIntegration)
 
@@ -27,7 +36,7 @@ def patch_resolve_request():
 
 
 def _wrap_resolver_match(hub, resolver_match):
-    # type: (django.urls.resolvers.ResolverMatch) -> django.urls.resolvers.ResolverMatch
+    # type: (Hub, ResolverMatch) -> ResolverMatch
 
     # XXX: The wrapper function is created for every request. Find more
     # efficient way to wrap views (or build a cache?)
@@ -36,6 +45,7 @@ def _wrap_resolver_match(hub, resolver_match):
 
     @functools.wraps(old_callback)
     def callback(*args, **kwargs):
+        # type: (*Any, **Any) -> Any
         with hub.start_span(op="django.view", description=resolver_match.view_name):
             return old_callback(*args, **kwargs)
 
--- a/sentry_sdk/integrations/django/views.py
+++ b/sentry_sdk/integrations/django/views.py
@@ -28,8 +28,7 @@ def patch_resolver():
         if integration is None or not integration.middleware_spans:
             return old_resolve(self, path)
 
-        with hub.start_span(op="django.urls.resolve"):
-            return _wrap_resolver_match(hub, old_resolve(self, path))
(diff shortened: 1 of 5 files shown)


tests/integrations/sqlalchemy/test_sqlalchemy.py
```diff
@@ -118,7 +118,7 @@ class Address(Base):
     assert (
         render_span_tree(event)
         == """\
-- op=None: description=None
+- op=None: description='None'
```
rhcarvalho, line 121: Why op is `None` and description is `'None'`? Other ops below have quotes.

rhcarvalho: Hmm maybe `None` => JSON's `null`, while the description will always be a string?

tests/conftest.py
```diff
@@ -334,7 +334,7 @@ def inner(event):
             by_parent.setdefault(span["parent_span_id"], []).append(span)
 
         def render_span(span):
-            yield "- op={!r}: description={!r}".format(
+            yield "- op='{}': description='{}'".format(
```
rhcarvalho, line 337: What's the motivation to change this?

sentry_sdk/integrations/django/views.py
```diff
@@ -0,0 +1,53 @@
...
+            return _wrap_resolver_match(hub, old_resolve(self, path))
+
+    URLResolver.resolve = resolve
+
+
+def _wrap_resolver_match(hub, resolver_match):
+    # type: (Hub, ResolverMatch) -> ResolverMatch
+
+    # XXX: The wrapper function is created for every request. Find more
+    # efficient way to wrap views (or build a cache?)
```
rhcarvalho, line 41: Yes, it makes sense to compute the wrapped resolver only once. Looks like given a `path` we always return the same `ResolverMatch`.

sentry_sdk/integrations/django/views.py
```diff
@@ -0,0 +1,53 @@
...
+
+    old_callback = resolver_match.func
+
+    @functools.wraps(old_callback)
+    def callback(*args, **kwargs):
+        # type: (*Any, **Any) -> Any
+        with hub.start_span(op="django.view", description=resolver_match.view_name):
+            return old_callback(*args, **kwargs)
+
+    resolver_match.func = callback
```
rhcarvalho, line 51: Isn't this re-wrapping the same `func` over and over every time the `_wrap_resolver_match` is called?!

sentry_sdk/integrations/django/views.py
untitaker: resolvermatch is instantiated on every request so unfortunately we can't use it as cachekey. We could use `resolver_match.func`, which has stable IDs afaict

sentry_sdk/integrations/django/views.py
rhcarvalho: Clarified, a new `ResolverMatch` is instantiated for every request.

untitaker: no, resolvermatch is instantiated fresh every time as it contains request data, so `func` is always the same for the same request.

tests/conftest.py
untitaker: It caused incompatibilities between Python 2 and 3 when rendering unicode strings.

tests/integrations/sqlalchemy/test_sqlalchemy.py
untitaker: Ah yeah that's an unfortunate side-effect. Will switch to `json.dumps` as representation method

sentry_sdk/integrations/django/views.py
rhcarvalho: Disregard, the `ResolverMatch` also carries request-specific args and kwargs.
If we cache it would be in the `func` level.

sentry_sdk/integrations/django/views.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def resolve(self, path):
+        # type: (URLResolver, Any) -> ResolverMatch
+        hub = Hub.current
+        integration = hub.get_integration(DjangoIntegration)
+
+        if integration is None or not integration.middleware_spans:
+            return old_resolve(self, path)
+
+        with hub.start_span(op="django.urls.resolve"):
```
rhcarvalho, line 31: I'm not sure this is of much use. If all the resolver does is call the view, isn't it going to be very similar in duration to the `django.view` span?

untitaker: yeah let's remove it. resolver is probably fast enough, usually

PR #791, opened 2020-08-18
fix(django): Un-break csrf_exempt

untitaker:


untitaker:
django-dev is broken but that's a master bug

PR #797, opened 2020-08-25
fix: Use UTC time in AWS Lambda integration

rhcarvalho:
We use UTC throughout the SDK, the Lambda integration was the only
exception, now fixed.

Explicitly setting the timezone to UTC fixes a problem when loading the
AWS CloudWatch Logs console, where using local time is unreliable.


untitaker (approved):

rhcarvalho:
@untitaker can I ignore https://travis-ci.com/github/getsentry/sentry-python/jobs/377422532#L1164 ?

untitaker:
Yes. I think this may be a new release of Hypothesis causing the error. I will likely end up pinning hypothesis.

PR #804, opened 2020-08-30
fix: Refactor testsuite for GCP and fix some bugs

shantanu73:
1) Changed test_gcp.py file to include unit tests and removed automation tests.
2) Modified tox.ini as per above changes.
3) Modified Google cloud logs url to include utc time.

### DIFF
--- a/sentry_sdk/integrations/gcp.py
+++ b/sentry_sdk/integrations/gcp.py
@@ -160,21 +160,22 @@ def _get_google_cloud_logs_url(final_time):
         str -- Google Cloud Logs Console URL to logs.
     """
     hour_ago = final_time - timedelta(hours=1)
+    formatstring = "%Y-%m-%dT%H:%M:%SZ"
 
     url = (
         "https://console.cloud.google.com/logs/viewer?project={project}&resource=cloud_function"
         "%2Ffunction_name%2F{function_name}%2Fregion%2F{region}&minLogLevel=0&expandAll=false"
-        "&timestamp={initial_time}Z&customFacets=&limitCustomFacetWidth=true"
-        "&dateRangeStart={timestamp_start}Z&dateRangeEnd={timestamp_end}Z"
-        "&interval=PT1H&scrollTimestamp={timestamp_current}Z"
+        "&timestamp={initial_time}&customFacets=&limitCustomFacetWidth=true"
+        "&dateRangeStart={timestamp_start}&dateRangeEnd={timestamp_end}"
+        "&interval=PT1H&scrollTimestamp={timestamp_current}"
     ).format(
         project=environ.get("GCP_PROJECT"),
         function_name=environ.get("FUNCTION_NAME"),
         region=environ.get("FUNCTION_REGION"),
-        initial_time=final_time,
-        timestamp_start=hour_ago,
-        timestamp_end=final_time,
-        timestamp_current=final_time,
+        initial_time=final_time.strftime(formatstring),
+        timestamp_start=hour_ago.strftime(formatstring),
+        timestamp_end=final_time.strftime(formatstring),
+        timestamp_current=final_time.strftime(formatstring),
     )
 
     return url
--- a/sentry_sdk/integrations/gcp.py
+++ b/sentry_sdk/integrations/gcp.py
@@ -165,17 +165,15 @@ def _get_google_cloud_logs_url(final_time):
     url = (
         "https://console.cloud.google.com/logs/viewer?project={project}&resource=cloud_function"
         "%2Ffunction_name%2F{function_name}%2Fregion%2F{region}&minLogLevel=0&expandAll=false"
-        "&timestamp={final_time}&customFacets=&limitCustomFacetWidth=true"
+        "&timestamp={timestamp_end}&customFacets=&limitCustomFacetWidth=true"
         "&dateRangeStart={timestamp_start}&dateRangeEnd={timestamp_end}"
-        "&interval=PT1H&scrollTimestamp={timestamp_current}"
+        "&interval=PT1H&scrollTimestamp={timestamp_end}"
     ).format(
         project=environ.get("GCP_PROJECT"),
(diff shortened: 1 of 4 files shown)


sentry_sdk/integrations/gcp.py
```diff
@@ -159,22 +159,22 @@ def _get_google_cloud_logs_url(initial_time):
...
 
     url = (
         "https://console.cloud.google.com/logs/viewer?project={project}&resource=cloud_function"
         "%2Ffunction_name%2F{function_name}%2Fregion%2F{region}&minLogLevel=0&expandAll=false"
-        "&timestamp={initial_time}&customFacets=&limitCustomFacetWidth=true"
-        "&dateRangeStart={timestamp_start}&dateRangeEnd={timestamp_end}"
-        "&interval=PT1H&scrollTimestamp={timestamp_current}"
+        "&timestamp={initial_time}Z&customFacets=&limitCustomFacetWidth=true"
+        "&dateRangeStart={timestamp_start}Z&dateRangeEnd={timestamp_end}Z"
+        "&interval=PT1H&scrollTimestamp={timestamp_current}Z"
```
rhcarvalho, line 169: I'd expect the timestamps to be explicitly formatted to RFC3399 format as in https://github.com/getsentry/sentry-python/pull/797.

There should not be several `Z`s manually placed in the constructed URL string, that's very error prone and repetitive.

The other thing here is that we're implicitly relying on how a datetime object is converted to string... Unlike RFC3399, it is missing the `T` between date and time segments, and has instead a space. I wonder if that's intentional or just an accident.

rhcarvalho (changes requested):

sentry_sdk/integrations/gcp.py
rhcarvalho: Probably the Sentry UI will not handle links correctly if the URLs contain spaces. (If there had to be any actual spaces, they should be URL-encoded as `%20`.)

shantanu73: It was by accident as currently we are relying on the time format given by datetime object.
I'll make this change with appropriate string format.

sentry_sdk/integrations/gcp.py
```diff
@@ -159,22 +159,22 @@ def _get_google_cloud_logs_url(initial_time):
...
     ).format(
         project=environ.get("GCP_PROJECT"),
         function_name=environ.get("FUNCTION_NAME"),
         region=environ.get("FUNCTION_REGION"),
-        initial_time=initial_time,
+        initial_time=final_time,
```
rhcarvalho, line 174: Why is `initial_time` set to `final_time`?!

sentry_sdk/integrations/gcp.py
```diff
@@ -159,22 +159,22 @@ def _get_google_cloud_logs_url(initial_time):
...
-        "&timestamp={initial_time}&customFacets=&limitCustomFacetWidth=true"
-        "&dateRangeStart={timestamp_start}&dateRangeEnd={timestamp_end}"
-        "&interval=PT1H&scrollTimestamp={timestamp_current}"
+        "&timestamp={initial_time}Z&customFacets=&limitCustomFacetWidth=true"
+        "&dateRangeStart={timestamp_start}Z&dateRangeEnd={timestamp_end}Z"
+        "&interval=PT1H&scrollTimestamp={timestamp_current}Z"
     ).format(
         project=environ.get("GCP_PROJECT"),
         function_name=environ.get("FUNCTION_NAME"),
         region=environ.get("FUNCTION_REGION"),
```
rhcarvalho, line 173: This was added in a previous commit, but since we're fixing the link, I ask -- is it guaranteed that those variables are always set in the environment?

If they are not set, they will produce `None` in the middle of the URL, possibly not a working URL to the GCP console. In that case, it seems more reasonable to omit the URL altogether.

sentry_sdk/integrations/gcp.py
```diff
@@ -159,22 +159,22 @@ def _get_google_cloud_logs_url(initial_time):
...
-        initial_time=initial_time,
+        initial_time=final_time,
         timestamp_start=hour_ago,
-        timestamp_end=initial_time,
-        timestamp_current=initial_time,
+        timestamp_end=final_time,
+        timestamp_current=final_time,
```
rhcarvalho, line 177: One of the advantages of using named keyword arguments to `str.format` is that you can reference the same value by name multiple times. Here, I believe there is no need to define both `timestamp_end` and `timestamp_current` when they always point to the same value.

In fact, `timestamp_current` is misleading because it is not the current time and it is used to set the `scrollTimestamp` value in the query string.

Suggestion: there should only be two timestamps passed to `format`, one for the start and the other for the end of the range.

sentry_sdk/integrations/gcp.py
shantanu73: These environment variables are reserved for Python 3.7 runtime environment by Google Cloud platform as per the documentation.
Link : https://cloud.google.com/functions/docs/env-var

sentry_sdk/integrations/gcp.py
shantanu73: I'll make this change, was left out in the last changes that were made.

sentry_sdk/integrations/gcp.py
shantanu73: Made this change in the latest commit.

sentry_sdk/integrations/gcp.py
rhcarvalho: 👍

sentry_sdk/integrations/gcp.py
rhcarvalho: Thanks!

sentry_sdk/integrations/gcp.py
rhcarvalho: Thank you!

rhcarvalho (approved):

Issue #805, opened 2020-08-31
pure_eval misses the variable on the crash line


vmarkovtsev:
(Related to https://github.com/getsentry/sentry-python/issues/748)

I keep experiencing the same situation for various crash reports in my project: the new "interesting variables" logic includes everything but the entities on the line where the crash happens. For example,

```python
def foo(arg: Dict[str, str]):
    bar = arg[5]
    x = func()
    x.attr = "AttributeError"  # crash here
    ...
```

The visible variables in the stack trace are `arg`, `bar`, and even `Dict[str, str]` from `typing` (`pufe_eval` loves `typing` for some reason), but `x` - never. Yet the most important local stack information for me is the variables from the line where I raise an exception. I guess the variable suggestion logic can be hacked to prioritize the variables on that line.

untitaker:
@alexmojaki can you look into this?

vmarkovtsev:
Sorry, I've just remembered another concern: we've got the number of max variables reported per stack frame (10 by default AFAIK). It frequently happens that there are many random noisy names before `x`-es, so even if the priority is fixed, such variables can still be discarded after truncation of the alphabetically sorted keys sequence.

untitaker:
Yes. I think the solution here is that pure_eval does its own trimming, such that it does not run into limits later, as it knows better what to trim first.

alexmojaki:
Hi @vmarkovtsev, thanks for the report. I'm unable to reproduce this. Using your code:

```python
from typing import Dict

from sentry_sdk.integrations.pure_eval import PureEvalIntegration
import sentry_sdk

sentry_sdk.init(
    dsn="...",
    integrations=[PureEvalIntegration()],
    send_default_pii=True
)


class X:
    __slots__ = ["none"]


def func():
    return X()


def foo(arg: Dict[int, str]):
    bar = arg[5]
    x = func()
    x.attr = "AttributeError"  # crash here


foo({5: "asd"})
```

I see `x` as expected:

![Screenshot from 2020-08-31 14-04-36](https://user-images.githubusercontent.com/3627481/91718098-03852a00-eb93-11ea-9697-97562689fce3.png)

Of course `x.attr` doesn't exist yet, but it sounds like you're talking about `x` itself. Do you think you can put together a reproducible example?

[...]

PR #807, opened 2020-08-31
Order variables by closeness to executing statement in pure_eval

alexmojaki:
Part of #805 

The idea here is to prioritise values which appear closer to the executing statement. The dict of values is ordered so that trimming removes lower priority values later on.

However this doesn't work in python 3.5 because the ordered dict is actually converted to a regular dict which loses the order:

https://github.com/getsentry/sentry-python/blob/4d91fe0944009a6e02450214f663037dc1ce056c/sentry_sdk/serializer.py#L286-L288

It seems that my options are:

1. Remove the above line and fix the errors that arise.
2. Keep the dict ordered when making the copy.
3. Preemptively trim in the pure_eval integration so order no longer matters.

What do you think @untitaker ?

### DIFF
--- a/tests/integrations/pure_eval/test_pure_eval.py
+++ b/tests/integrations/pure_eval/test_pure_eval.py
@@ -1,3 +1,6 @@
+import sys
+from types import SimpleNamespace
+
 import pytest
 
 from sentry_sdk import capture_exception
@@ -10,8 +13,27 @@ def test_with_locals_enabled(sentry_init, capture_events, integrations):
     events = capture_events()
 
     def foo():
-        foo.d = {1: 2}
-        print(foo.d[1] / 0)
+        namespace = SimpleNamespace()
+        q = 1
+        w = 2
+        e = 3
+        r = 4
+        t = 5
+        y = 6
+        u = 7
+        i = 8
+        o = 9
+        p = 10
+        a = 11
+        s = 12
+        str((q, w, e, r, t, y, u, i, o, p, a, s))  # use variables for linter
+        namespace.d = {1: 2}
+        print(namespace.d[1] / 0)
+
+        # Appearances of variables after the main statement don't affect order
+        print(q)
+        print(s)
+        print(events)
 
     try:
         foo()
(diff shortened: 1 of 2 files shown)


vmarkovtsev:
Note: this will automatically include the function arguments, because, for all the lines except the lowest in the call stack, they lie on an executing statement 👍 I am super excited and eager to try this 🎉

tests/integrations/pure_eval/test_pure_eval.py
```diff
@@ -28,8 +50,43 @@ def foo():
...
+            "namespace.d[1]",
+            "s",
+            "a",
+            "p",
+            "o",
+            "i",
+            "u",
+            "y",
+        ]
+        if sys.version_info[:2] == (3, 5):
```
untitaker, line 67: Instead of this conditional can't we do `assert set(frame_vars.keys()) == set(expected_keys)`

untitaker (approved):
tests fail on `tox -e py3.5-pure_eval`, but this looks good

tests/integrations/pure_eval/test_pure_eval.py
alexmojaki: I think you misunderstood. `frame_vars.keys()` is set-like and so `assert frame_vars.keys() == set(expected_keys)` would work in all versions. `assert list(frame_vars.keys()) == expected_keys` is a stronger assertion to make extra sure (when possible) that values are prioritised as expected. The tests are fine as they are, the reason for the failure in 3.5 is that the implementation actually doesn't work and the values are actually wrong.

The problem is that I `return OrderedDict` above so that the trimming in `serialize` will keep the first 10 items of the dict, but just before the trimming the `OrderedDict` gets converted to a `dict` (see the PR description) which breaks the order in 3.5 and thus the implementation (the wrong values get trimmed). I'm asking if I should just trim the values within the pure_eval integration and forget the `OrderedDict`, or try to preserve order in `serialize`.

alexmojaki: To be clear both options should be equivalent right now, but if `serialize` changes its algorithm and starts trimming things more cleverly in the future, it will be better if it has the full dict of variables in order of priority rather than a dict that was eagerly/preemptively trimmed by pure_eval.

untitaker: Ah right, I recommend going for option 3 but avoiding hardcoding the number 10 (instead importing the var). I don't think there's a safe alternative to `dict` outside of pure_eval context, so I'd like to avoid changes to serializer.

untitaker: >but if serialize changes its algorithm and starts trimming things more cleverly in the future

I doubt this will happen in serialization, let's assume pure_eval always knows better (seems likely given the amount of extra analysis it does)

alexmojaki: Done

untitaker:
Excellent, thank you!

PR #809, opened 2020-09-02
.flake8: Don't set --max-complexity if you don't care about code complexity

cclauss:
…plexity


untitaker (approved):
Thanks, fair enough

cclauss:
(subtext: we _should_ care about code complexity and _should_ set --max-complexity to something vaguely reasonable)

PR #819, opened 2020-09-11
fix: Do not use locks in worker

### LINKED ISSUE #803: Deadlock (Hangs) during Garbage Collection because of logging in destructor and lock on `Queue`.

untitaker:
Fix #803 

doesn't work yet

### DIFF
--- a/sentry_sdk/worker.py
+++ b/sentry_sdk/worker.py
@@ -1,8 +1,9 @@
 import os
 
-from threading import Thread, Lock
+from collections import deque
+from threading import Thread
 from time import sleep, time
-from sentry_sdk._compat import queue, check_thread_support
+from sentry_sdk._compat import check_thread_support
 from sentry_sdk.utils import logger
 
 from sentry_sdk._types import MYPY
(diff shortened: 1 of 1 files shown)


untitaker:
thread restart was deemed too risky if the queue often hits zero. I think we should eventually try this out, but it's not a quick fix

PR #822, opened 2020-09-15
chore: Clean up Flask CI and test 0.10

untitaker:


sentry_sdk/integrations/flask.py
```diff
@@ -69,7 +69,7 @@ def setup_once():
         except (ValueError, TypeError):
             raise DidNotEnable("Unparsable Flask version: {}".format(FLASK_VERSION))
 
-        if version < (0, 11):
+        if version < (0, 10):
```
saifelse, line 72: Noting that the next line should read `0.10`

PR #826, opened 2020-09-18
fix(transport): Use correct data category for transaction events

untitaker:
While working on https://github.com/getsentry/sentry-python/pull/825 I found a bug that caused us to use the "default" item type for transaction events. The bug happened because there were mismatched expectations as to what Item.get_event should return if the content is a transaction event.

Will add an explicit test later.

### DIFF
--- a/sentry_sdk/envelope.py
+++ b/sentry_sdk/envelope.py
@@ -242,7 +242,7 @@ class Item(object):
     def get_event(self):
         # type: (...) -> Optional[Event]
         """
-        Returns a regular (non-transaction) event if there is one.
+        Returns an error event if there is one.
         """
         if self.headers.get("type") == "event" and self.payload.json is not None:
             return self.payload.json
--- a/sentry_sdk/envelope.py
+++ b/sentry_sdk/envelope.py
@@ -224,8 +224,12 @@ class Item(object):
     def data_category(self):
         # type: (...) -> EventDataCategory
         ty = self.headers.get("type")
-        if ty in ("session", "attachment", "transaction"):
-            return ty
+        if ty == "session":
+            return "session"
+        elif ty == "attachment":
+            return "attachment"
+        elif ty == "transaction":
+            return "transaction"
         elif ty == "event":
             return "error"
         else:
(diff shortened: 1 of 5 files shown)


sentry_sdk/envelope.py
```diff
@@ -230,23 +230,25 @@ def __repr__(self):
...
 
     def get_bytes(self):
         # type: (...) -> bytes
         return self.payload.get_bytes()
 
     def get_event(self):
         # type: (...) -> Optional[Event]
-        if self.headers.get("type") == "event" and self.payload.json is not None:
+        """
+        Returns a regular event if there is one. Transaction events are exempt
```
rhcarvalho, line 248: What do you mean by "transaction events are exempt from this"?

tests/test_transport.py
```diff
@@ -116,7 +116,7 @@ def test_simple_rate_limits(httpserver, capsys, caplog, make_client):
     client.capture_event({"type": "transaction"})
     client.flush()
 
-    assert len(httpserver.requests) == 1
+    assert list(r.url[-18:] for r in httpserver.requests) == ["/api/132/envelope/"]
```
rhcarvalho, line 119: `url[-18:]` :-(

I'd keep the original check for 1 request, then you can `assert httpserver.requests[0].endswith("/api/132/envelope/"]`

rhcarvalho (changes requested):
At least one bug found :)

tests/test_transport.py
untitaker: yeah fair

sentry_sdk/envelope.py
untitaker: Nevermind. I reverted the changes to this method as they were not necessary.

sentry_sdk/envelope.py
```diff
@@ -230,22 +223,27 @@ def __repr__(self):
...
+            return "default"
 
     def get_bytes(self):
         # type: (...) -> bytes
         return self.payload.get_bytes()
 
     def get_event(self):
         # type: (...) -> Optional[Event]
+        """
+        Returns a regular (non-transaction) event if there is one.
```
rhcarvalho, line 245: "regular non-transaction event" is weird. Does this mean

```suggestion
        Returns an error event if there is one.
```

?

rhcarvalho (approved):

PR #830, opened 2020-09-22
Capturing Performance monitoring transactions for AWS and GCP

shantanu73:
1) Added code to capture transactions related to performance monitoring for AWS & GCP integrations.
2) Added test cases as per above changes.

### DIFF
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -6,6 +6,7 @@ from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk.tracing import Transaction
 from sentry_sdk._compat import reraise
 from sentry_sdk.utils import (
+    AnnotatedValue,
     capture_internal_exceptions,
     event_from_exception,
     logger,
@@ -282,9 +283,6 @@ def _make_request_event_processor(aws_event, aws_context, configured_timeout):
         if "headers" in aws_event:
             request["headers"] = _filter_headers(aws_event["headers"])
 
-        if "body" in aws_event:
-            request["data"] = aws_event.get("body", "")
-
         if _should_send_default_pii():
             user_info = event.setdefault("user", {})
 
@@ -295,6 +293,14 @@ def _make_request_event_processor(aws_event, aws_context, configured_timeout):
             ip = aws_event.get("identity", {}).get("sourceIp")
             if ip is not None:
                 user_info.setdefault("ip_address", ip)
+
+            if "body" in aws_event:
+                request["data"] = aws_event.get("body", "")
+        else:
+            if aws_event.get("body", None):
+                # Unfortunately couldn't find a way to get structured body from AWS
+                # event. Meaning every body is unstructured to us.
+                request["data"] = AnnotatedValue("", {"rem": [["!raw", "x", 0, 0]]})
 
         event["request"] = request
 
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -3,6 +3,7 @@ from os import environ
 import sys
 
 from sentry_sdk.hub import Hub, _should_send_default_pii
+from sentry_sdk.tracing import Transaction
 from sentry_sdk._compat import reraise
 from sentry_sdk.utils import (
     AnnotatedValue,
@@ -100,7 +101,9 @@ def _wrap_handler(handler):
                     timeout_thread.start()
 
             try:
-                return handler(event, context, *args, **kwargs)
+                transaction = Transaction(op="aws_lambda", name=context.function_name)
+                with hub.start_transaction(transaction):
+                    return handler(event, context, *args, **kwargs)
             except Exception:
                 exc_info = sys.exc_info()
                 event, hint = event_from_exception(
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -81,6 +81,7 @@ def _wrap_handler(handler):
                 scope.add_event_processor(
                     _make_request_event_processor(event, context, configured_time)
                 )
+                scope.set_tag("region", context.invoked_function_arn.split(":")[3])
                 # Starting the Timeout thread only if the configured time is greater than Timeout warning
                 # buffer and timeout_warning parameter is set True.
                 if (
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -100,7 +100,7 @@ def _wrap_handler(handler):
                     # Starting the thread to raise timeout warning exception
                     timeout_thread.start()
 
-            headers = _filter_headers(event.get("headers", {}))
+            headers = event.get("headers", {})
             transaction = Transaction.continue_from_headers(
                 headers, op="serverless.function", name=context.function_name
             )
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -82,7 +82,7 @@ def _wrap_handler(handler):
                 scope.add_event_processor(
                     _make_request_event_processor(event, context, configured_time)
                 )
-                scope.set_tag("region", context.invoked_function_arn.split(":")[3])
+                scope.set_tag("aws_region", context.invoked_function_arn.split(":")[3])
                 # Starting the Timeout thread only if the configured time is greater than Timeout warning
                 # buffer and timeout_warning parameter is set True.
                 if (
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -6,7 +6,6 @@ from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk.tracing import Transaction
 from sentry_sdk._compat import reraise
 from sentry_sdk.utils import (
-    AnnotatedValue,
     capture_internal_exceptions,
     event_from_exception,
     logger,
@@ -79,7 +78,6 @@ def _wrap_handler(handler):
         with hub.push_scope() as scope:
             with capture_internal_exceptions():
                 scope.clear_breadcrumbs()
-                scope.transaction = context.function_name
                 scope.add_event_processor(
                     _make_request_event_processor(event, context, configured_time)
                 )
@@ -100,19 +98,22 @@ def _wrap_handler(handler):
                     # Starting the thread to raise timeout warning exception
                     timeout_thread.start()
 
-            try:
-                transaction = Transaction(op="aws_lambda", name=context.function_name)
-                with hub.start_transaction(transaction):
+            headers = _filter_headers(event.get("headers", {}))
+            transaction = Transaction.continue_from_headers(
+                headers, op="serverless.function", name=context.function_name
+            )
+            with hub.start_transaction(transaction):
+                try:
                     return handler(event, context, *args, **kwargs)
-            except Exception:
-                exc_info = sys.exc_info()
-                event, hint = event_from_exception(
-                    exc_info,
-                    client_options=client.options,
-                    mechanism={"type": "aws_lambda", "handled": False},
-                )
-                hub.capture_event(event, hint=hint)
-                reraise(*exc_info)
+                except Exception:
+                    exc_info = sys.exc_info()
+                    event, hint = event_from_exception(
+                        exc_info,
+                        client_options=client.options,
+                        mechanism={"type": "aws_lambda", "handled": False},
+                    )
+                    hub.capture_event(event, hint=hint)
+                    reraise(*exc_info)
 
     return sentry_handler  # type: ignore
 
@@ -280,10 +281,8 @@ def _make_request_event_processor(aws_event, aws_context, configured_timeout):
         if "headers" in aws_event:
             request["headers"] = _filter_headers(aws_event["headers"])
 
-        if aws_event.get("body", None):
-            # Unfortunately couldn't find a way to get structured body from AWS
-            # event. Meaning every body is unstructured to us.
-            request["data"] = AnnotatedValue("", {"rem": [["!raw", "x", 0, 0]]})
+        if "body" in aws_event:
+            request["data"] = aws_event.get("body", "")
 
         if _should_send_default_pii():
             user_info = event.setdefault("user", {})
--- a/sentry_sdk/integrations/gcp.py
+++ b/sentry_sdk/integrations/gcp.py
@@ -62,6 +62,7 @@ def _wrap_func(func):
                 scope.add_event_processor(
                     _make_request_event_processor(event, configured_time, initial_time)
                 )
+                scope.set_tag("region", environ.get("FUNCTION_REGION"))
                 if (
                     integration.timeout_warning
                     and configured_time > TIMEOUT_WARNING_BUFFER
--- a/sentry_sdk/integrations/gcp.py
+++ b/sentry_sdk/integrations/gcp.py
@@ -1,12 +1,12 @@
 from datetime import datetime, timedelta
 from os import environ
 import sys
-import json
 
-from sentry_sdk.hub import Hub
+from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk.tracing import Transaction
 from sentry_sdk._compat import reraise
 from sentry_sdk.utils import (
+    AnnotatedValue,
     capture_internal_exceptions,
     event_from_exception,
     logger,
@@ -163,8 +163,14 @@ def _make_request_event_processor(gcp_event, configured_timeout, initial_time):
         if hasattr(gcp_event, "headers"):
             request["headers"] = _filter_headers(gcp_event.headers)
 
-        if hasattr(gcp_event, "data"):
-            request["data"] = json.loads(gcp_event.data)
+        if _should_send_default_pii():
+            if hasattr(gcp_event, "data"):
+                request["data"] = gcp_event.data
+        else:
+            if hasattr(gcp_event, "data"):
+                # Unfortunately couldn't find a way to get structured body from GCP
+                # event. Meaning every body is unstructured to us.
+                request["data"] = AnnotatedValue("", {"rem": [["!raw", "x", 0, 0]]})
 
         event["request"] = request
 
--- a/sentry_sdk/integrations/gcp.py
+++ b/sentry_sdk/integrations/gcp.py
@@ -62,7 +62,7 @@ def _wrap_func(func):
                 scope.add_event_processor(
                     _make_request_event_processor(event, configured_time, initial_time)
                 )
-                scope.set_tag("region", environ.get("FUNCTION_REGION"))
+                scope.set_tag("gcp_region", environ.get("FUNCTION_REGION"))
                 if (
                     integration.timeout_warning
                     and configured_time > TIMEOUT_WARNING_BUFFER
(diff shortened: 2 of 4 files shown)


sentry_sdk/integrations/aws_lambda.py
```diff
@@ -100,7 +101,9 @@ def sentry_handler(event, context, *args, **kwargs):
                     timeout_thread.start()
 
             try:
-                return handler(event, context, *args, **kwargs)
+                transaction = Transaction(op="aws_lambda", name=context.function_name)
```
untitaker, line 104: Would prefer something more generic here such as `http.server` or `serverless.function`.

Also this is missing trace propagation entirely. Please use `continue_from_headers` such that we actually get joined traces, not individual transactions. Somewhere in `make_request_event_processor` there should be logic for accessing HTTP headers that we can duplicate. I think it would also be nice if we could instrument functions invoking other functions, but HTTP should be easiest for now

The same comments apply to GCP

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -100,7 +101,9 @@ def sentry_handler(event, context, *args, **kwargs):
...
+                with hub.start_transaction(transaction):
+                    return handler(event, context, *args, **kwargs)
             except Exception:
```
untitaker, line 107: Please put the exception block inside of start_transaction.

Also setting `scope.transaction` should no longer be necessary.

The same comments apply to GCP

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -49,12 +62,17 @@ def _send_event(self, event):
...
+        envelope = envelope_processor(envelope)
+        print("\\nENVELOPE: {}\\n".format(json.dumps(envelope)))
+
+def init_sdk(timeout_warning=False, traces_sample_rate=0, **extra_init_args):
     sentry_sdk.init(
         dsn="https://123abc@example.com/123",
         transport=TestTransport,
         integrations=[AwsLambdaIntegration(timeout_warning=timeout_warning)],
         shutdown_timeout=10,
+        traces_sample_rate=traces_sample_rate,
```
untitaker, line 75: It should not be necessary to pass this explicitly, right? we pass `extra_init_args`

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -40,6 +40,19 @@ def event_processor(event):
     # to print less to logs.
     return event
 
+def envelope_processor(envelope):
```
untitaker, line 43: 👍

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -99,13 +118,19 @@ def inner(code, payload, timeout=30, syntax_check=True):
             line = line[len(b"EVENT: ") :]
             events.append(json.loads(line.decode("utf-8")))
 
-        return events, response
+        for line in base64.b64decode(response["LogResult"]).splitlines():
```
untitaker, line 121: I would merge this with the previous loop such that events and envelopes are parsed in one go

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -123,8 +148,6 @@ def test_handler(event, context):
         b'{"foo": "bar"}',
     )
 
-    assert response["FunctionError"] == "Unhandled"
```
untitaker, line 126: Why did this change?

tests/integrations/gcp/test_gcp.py
```diff
@@ -102,16 +112,24 @@ def inner(code, subprocess_kwargs=()):
...
             stream = os.popen("python {}/main.py".format(tmpdir))
-            event = stream.read()
-            event = json.loads(event[len("EVENT: ") :])
-
-        return event
+            stream_data = stream.read()
+            delimeter = "EVENT: "
+            if "ENVELOPE:" in stream_data:
+                delimeter = delimeter + "|ENVELOPE: "
+            stream_list = re.split(delimeter, stream_data)
```
untitaker, line 119: This seems odd, can't we do the same as in AWS?

tests/integrations/gcp/test_gcp.py
```diff
@@ -48,21 +48,30 @@ def event_processor(event):
...
+        print("\\nENVELOPE: {}\\n".format(envelope.decode(\"utf-8\")))
 
-def init_sdk(timeout_warning=False, **extra_init_args):
+def init_sdk(timeout_warning=False, traces_sample_rate=0, **extra_init_args):
     sentry_sdk.init(
         dsn="https://123abc@example.com/123",
         transport=TestTransport,
         integrations=[GcpIntegration(timeout_warning=timeout_warning)],
         shutdown_timeout=10,
+        traces_sample_rate=traces_sample_rate,
```
untitaker, line 74: like in aws I don't think this is necessary because of `**extra_init_args`

untitaker (changes requested):

tests/integrations/gcp/test_gcp.py
shantanu73: I checked and I did it as per AWS, changes will be in next commit.

tests/integrations/gcp/test_gcp.py
shantanu73: Made this change, this parameter was redundant.

tests/integrations/aws_lambda/test_aws.py
shantanu73: This test-case was failing when I ran the test-suite locally, apparently "FunctionError" parameter wasn't coming as part of response. Although I rechecked it now and it is working fine. I'll revert this line.

tests/integrations/aws_lambda/test_aws.py
shantanu73: Parsed it in one go. Change will be in next commit.

tests/integrations/aws_lambda/test_aws.py
shantanu73: Yes, removed this parameter. Change will be in next commit.

sentry_sdk/integrations/aws_lambda.py
shantanu73: Sure, made this change.

sentry_sdk/integrations/aws_lambda.py
shantanu73: @untitaker 

I implemented the Transactions using "continue_from_headers" method. I got the same results as before, i.e. we can't capture the internal transactions for 2nd lambda function in case of nested lambda functions.
By "internal transactions" I mean that 2nd lambda function actually calls Dynamo DB & S3 bucket using boto3 library, so internal transactions would involve transactions of API calls(as boto3 functions) to these services.
Also, the trace ID is same for the instrumentation of API call to 2nd lambda function which is happening automatically currently.
I want to know your expectations when you mentioned about the usage of "continue_from_headers" method in the code, as the trace ID is anyways getting passed on, even if we don't use it and even the headers are also getting displayed.

Second thing is that, I want to know about your expectations on Lambda function calling another Lambda function or nested Lambda functions. There are 2 ways as of now to trigger/invoke lambda functions as of now are:-
1. Using boto3 library (POST https://lambda.us-east-2.amazonaws.com/2015-03-31/functions/sentry_dynamo_db/invocations)
2. Using API Gateway (GET https://g8xk68m1n6.execute-api.us-east-2.amazonaws.com/default/sentry_dynamo_db)

In both the above mentioned methods, I've tried to capture traces and both basically involve calling a REST API for 2nd lambda function to be invoked.
So, basically an API call is involved in any case to invoke a lambda function and the invocation traces of the API calls are getting displayed in Sentry dashboard.

Thus, wanted to know if we need to skip the usage for "continue_from_headers" method or still use it even though the results remain the same?
Also, is this change sufficient enough as per the expectations from the code, atleast related to instrumentation of lambda function calling another lambda function?

untitaker: >the invocation traces of the API calls are getting displayed in Sentry dashboard.

yes but if you go down to the trace details, do the two function calls have the same trace ID?

Screenshot 2020-09-24 at 18 22 35

Imagine a non-serverless (serverful?) stack, such as Django + Celery. There is a Django request handler that spawns a celery task. Those are two transactions, but they should also have the same trace ID. So you can go from the celery task back to the api call and figure out how this data made it into your system.

Imagine two Flask services calling each other via HTTP. All of the transactions of those two services are linked together, because we 1) look at Flask request headers to find sentry-trace header 2) monkeypatch httplib to add sentry-trace to outgoing HTTP requests

Right now this is not really surfaced in the product in an optimal way, but we're working on it.

I don't really understand how function-to-function calls work. If they are not using HTTP, it's probably too hard to implement. But perhaps we can make sure that if the incoming HTTP request contains the sentry-trace header, we honor it instead of creating a new trace ID.

Let me know if that makes sense. I don't think we have to do any of that, if we say it's too hard right now let's just create a new transaction in a new trace. But I just wanted to explain why `continue_from_headers` is different over pure init.

shantanu73: Currently function-to-function calls use HTTP and the trace ID is in fact same for both function calls. Look at the trace IDs for both lambda functions and span IDs and parent/child span IDs in the image below.

trace ID


The implementation that I did right now goes like this:-
API call to first function using Postman   ->   1st Lambda function getting triggered using API Gateway    ->   1st Lambda function (Sentry SDK initialized here) doing API call using requests library in the code to call 2nd Lambda function   ->   2nd Lambda function getting triggered using API Gateway   ->   2nd Lambda function using boto3 library to do operations on Dynamo DB & S3 bucket (which is nothing but API calls)

What I exactly wanted to say was that, since Sentry SDK is initialized in the Lambda function 1, the spans for Lambda function 1 and Lambda function 2 are appearing on the Sentry dashboard as per the the implementation I mentioned above. But, there is no way to capture trace for Dynamo DB and S3 bucket implemented in Lambda function 2 coz Lambda function 2 is not initialized with Sentry SDK. 

Now, If it were the case that we initialized both lambda functions with Sentry SDK and implemented the same way as I mentioned above, then we can capture traces for even Dynamo DB & S3 bucket, and also keeping the span of 2nd lambda function as child of span of 1st lambda function (which is already happening anyways).

Also, I got the idea behind the usage of continue_from_headers and it is easy to implement just wanted to understand reason behind it. Will make the change accordingly as I was able to achieve what you mentioned in the first comment.

The main intention was to match my current understanding of the whole tracing mechanism with yours and to validate if I'm doing the right thing or not. I think it is in alignment now and I can push the final changes soon.

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -277,10 +281,8 @@ def event_processor(event, hint, start_time=start_time):
         if "headers" in aws_event:
             request["headers"] = _filter_headers(aws_event["headers"])
 
-        if aws_event.get("body", None):
-            # Unfortunately couldn't find a way to get structured body from AWS
-            # event. Meaning every body is unstructured to us.
-            request["data"] = AnnotatedValue("", {"rem": [["!raw", "x", 0, 0]]})
+        if "body" in aws_event:
+            request["data"] = aws_event.get("body", "")
```
shantanu73, line 285: @untitaker I made this change to display the request body appear on Sentry Dashboard as shown below. This change was discussed with AJ.

req_body

sentry_sdk/integrations/gcp.py
```diff
@@ -143,6 +153,18 @@ def event_processor(event, hint):
 
         request["url"] = "gcp:///{}".format(environ.get("FUNCTION_NAME"))
 
+        if hasattr(gcp_event, "method"):
```
shantanu73, line 156: @untitaker I've modified event processor for GCP integration to get request related for Sentry dashboard.

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -78,10 +78,10 @@ def sentry_handler(event, context, *args, **kwargs):
         with hub.push_scope() as scope:
             with capture_internal_exceptions():
                 scope.clear_breadcrumbs()
-                scope.transaction = context.function_name
                 scope.add_event_processor(
                     _make_request_event_processor(event, context, configured_time)
                 )
+                scope.set_tag("region", context.invoked_function_arn.split(":")[3])
```
shantanu73, line 84: @untitaker I've added code to set "region" tag for AWS as well as for GCP integration.

sentry_sdk/integrations/aws_lambda.py
untitaker: Please restore the old behavior when the client option `send_default_pii` is in its default `False`. If the user set `send_default_pii=True` we can do what you propose.

Also please restore the code comment.

The reason for this is that we want to be safe by default: the request body can contain a lot of sensitive data and setting sendDefaultPii is our way of having the user say "I accept the risks".

sentry_sdk/integrations/gcp.py
```diff
@@ -143,6 +153,18 @@ def event_processor(event, hint):
...
+            request["method"] = gcp_event.method
+
+        if hasattr(gcp_event, "query_string"):
+            request["query_string"] = gcp_event.query_string.decode("utf-8")
+
+        if hasattr(gcp_event, "headers"):
+            request["headers"] = _filter_headers(gcp_event.headers)
+
+        if hasattr(gcp_event, "data"):
```
untitaker, line 166: Please apply the same logic here as I commented on aws lambda. Do not attempt to json.loads unless you are sure it is json 100% of the time, attach raw data instead. Note though this is a Flask request object so the json property may be available too.

sentry_sdk/integrations/aws_lambda.py
untitaker: Can we prefix this with integration name such as `gcp_region`/`aws_region`. Region is too generic and it's often already hard to figure out how data made it into the system.

shantanu73: okay, sure @untitaker

sentry_sdk/integrations/aws_lambda.py
shantanu73: @untitaker you mean that essentially, I move this section of code I've written to the if section for send_default_pii part and revert the changes here.

untitaker: sorry I mean

```
if send_default_pii:
    new_code
else:
    old_code
```

sentry_sdk/integrations/gcp.py
shantanu73: Sure.

sentry_sdk/integrations/aws_lambda.py
shantanu73: Okay, I'll make these changes asap.

sentry_sdk/integrations/gcp.py
```diff
@@ -56,11 +59,10 @@ def sentry_func(*args, **kwargs):
         with hub.push_scope() as scope:
             with capture_internal_exceptions():
                 scope.clear_breadcrumbs()
-                scope.transaction = environ.get("FUNCTION_NAME")
                 scope.add_event_processor(
-                    _make_request_event_processor(configured_time, initial_time)
+                    _make_request_event_processor(event, configured_time, initial_time)
                 )
-            try:
+                scope.set_tag("gcp_region", environ.get("FUNCTION_REGION"))
```
shantanu73, line 65: @untitaker Made all changes as per review comments, you may review them.

sentry_sdk/integrations/gcp.py
```diff
@@ -71,19 +73,28 @@ def sentry_func(*args, **kwargs):
...
-                )
-                hub.capture_event(event, hint=hint)
-                reraise(*exc_info)
-            finally:
-                # Flush out the event queue
-                hub.flush()
+
+            headers = {}
+            if hasattr(event, "headers"):
+                headers = _filter_headers(event.headers)
```
untitaker, line 79:
```suggestion
            if hasattr(event, "headers"):
                headers = event.headers
```

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -99,17 +100,22 @@ def sentry_handler(event, context, *args, **kwargs):
...
-            except Exception:
-                exc_info = sys.exc_info()
-                event, hint = event_from_exception(
-                    exc_info,
-                    client_options=client.options,
-                    mechanism={"type": "aws_lambda", "handled": False},
-                )
-                hub.capture_event(event, hint=hint)
-                reraise(*exc_info)
+            headers = _filter_headers(event.get("headers", {}))
```
untitaker, line 103:
```suggestion
            headers = event.get("headers", {})
```

`_filter_headers` has a different purpose, it is used to remove sensitive data before attaching it to the event. But we don't attach `headers`, we only read one header from it.

sentry_sdk/integrations/gcp.py
```diff
@@ -31,13 +34,13 @@
 
 def _wrap_func(func):
     # type: (F) -> F
-    def sentry_func(*args, **kwargs):
-        # type: (*Any, **Any) -> Any
+    def sentry_func(functionhandler, event, *args, **kwargs):
+        # type: (Any, Any, *Any, **Any) -> Any
```
untitaker, line 38:
```suggestion
    def sentry_func(*args, **kwargs):
        # type: (*Any, **Any) -> Any
```

Just to keep the diff small, this change was not necessary

sentry_sdk/integrations/gcp.py
```diff
@@ -31,13 +34,13 @@
...
 
         hub = Hub.current
         integration = hub.get_integration(GcpIntegration)
         if integration is None:
-            return func(*args, **kwargs)
+            return func(functionhandler, event, *args, **kwargs)
```
untitaker, line 43:
```suggestion
            return func(*args, **kwargs)
```

sentry_sdk/integrations/gcp.py
```diff
@@ -47,7 +50,7 @@ def sentry_func(*args, **kwargs):
             logger.debug(
                 "The configured timeout could not be fetched from Cloud Functions configuration."
             )
-            return func(*args, **kwargs)
+            return func(functionhandler, event, *args, **kwargs)
```
untitaker, line 53:
```suggestion
            return func(*args, **kwargs)
```

untitaker (approved):

sentry_sdk/integrations/gcp.py
```diff
@@ -50,7 +50,7 @@ def sentry_func(functionhandler, event, *args, **kwargs):
             logger.debug(
                 "The configured timeout could not be fetched from Cloud Functions configuration."
             )
-            return func(functionhandler, event, *args, **kwargs)
```
shantanu73: @untitaker wanted to know why was this changed?
If we are going to change this, then I might need to fix code to get event object, as well as fix unit tests for this coz they might break.

untitaker: Sorry, I thought this was an unnecessary change, will revert.

shantanu73: okay, alright. Actually this change was done considering all trigger types for GCP Cloud functions. In all the trigger types, there are 2 parameters which are passed on to the "invoke_user_function" (similar to "request_handler" in AWS) which are "functionhandler" & "event" (similar to "event" & "context" in AWS).
And, since we do need headers from event object and other request related data, this change was made.

untitaker:
Perfect, thanks. I will release later this week, out of bandwidth right now unfortunately.

shantanu73:
> Perfect, thanks. I will release later this week, out of bandwidth right now unfortunately.

Sure. Thanks.

PR #831, opened 2020-09-22
fix: Second attempt at fixing trace propagation in Celery 4.2+

untitaker:
It turns out the bug is perfectly reproducible with the Redis backend, I just messed up the test.

### DIFF
--- a/sentry_sdk/integrations/celery.py
+++ b/sentry_sdk/integrations/celery.py
@@ -61,7 +61,6 @@ class CeleryIntegration(Integration):
                 # short-circuits to task.run if it thinks it's safe.
                 task.__call__ = _wrap_task_call(task, task.__call__)
                 task.run = _wrap_task_call(task, task.run)
-                task.apply_async = _wrap_apply_async(task, task.apply_async)
 
                 # `build_tracer` is apparently called for every task
                 # invocation. Can't wrap every celery task for every invocation
@@ -71,6 +70,10 @@ class CeleryIntegration(Integration):
             return _wrap_tracer(task, old_build_tracer(name, task, *args, **kwargs))
 
         trace.build_tracer = sentry_build_tracer
+
+        from celery.app.task import Task
+
+        Task.apply_async = _wrap_apply_async(Task.apply_async)
 
         _patch_worker_exit()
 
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/celery.py
```diff
@@ -61,7 +61,6 @@ def sentry_build_tracer(name, task, *args, **kwargs):
                 # short-circuits to task.run if it thinks it's safe.
                 task.__call__ = _wrap_task_call(task, task.__call__)
                 task.run = _wrap_task_call(task, task.run)
-                task.apply_async = _wrap_apply_async(task, task.apply_async)
```
untitaker, line 64: I think there was a reason to patch the instance...

untitaker: nvm I might have cargo-culted based on the run/call situation

sentry_sdk/integrations/celery.py
```diff
@@ -85,30 +88,31 @@ def sentry_build_tracer(name, task, *args, **kwargs):
...
+            with hub.start_span(op="celery.submit", description=args[0].name) as span:
                 with capture_internal_exceptions():
                     headers = dict(hub.iter_trace_propagation_headers())
+
                     if headers:
-                        kwarg_headers = kwargs.setdefault("headers", {})
+                        # Note: kwargs can contain headers=None, so no setdefault!
+                        # Unsure which backend though.
+                        kwarg_headers = kwargs.get("headers") or {}
                         kwarg_headers.update(headers)
```
untitaker, line 107: This line is currently crashing all the time (in `capture_internal_exceptions`) because the old code could not deal with `kwargs == {"headers": None}`, only `kwargs == {}`.

sentry_sdk/integrations/celery.py
```diff
@@ -85,30 +88,31 @@ def sentry_build_tracer(name, task, *args, **kwargs):
...
 
                         # https://github.com/celery/celery/issues/4875
                         #
                         # Need to setdefault the inner headers too since other
                         # tracing tools (dd-trace-py) also employ this exact
                         # workaround and we don't want to break them.
-                        #
```
untitaker, line 108: Turns out it is perfectly reproducible. The bug we're working around lives in celery.app.amqp, but it seems that module is used in redis too??? tbh I no longer understand how celery separates concerns.

tests/integrations/celery/test_celery.py
```diff
@@ -42,6 +42,7 @@ def inner(propagate_traces=True, backend="always_eager", **kwargs):
 
             # this backend requires capture_events_forksafe
             celery.conf.worker_max_tasks_per_child = 1
+            celery.conf.worker_concurrency = 1
```
untitaker, line 45: This speeds up tests, otherwise celery forks 20 times.

tests/conftest.py
```diff
@@ -235,11 +236,7 @@ def append_envelope(envelope):
 @pytest.fixture
 def capture_events_forksafe(monkeypatch, capture_events, request):
     def inner():
-        in_process_events = capture_events()
-
-        @request.addfinalizer
-        def _():
-            assert not in_process_events
+        capture_events()
```
untitaker, line 239: Need to remove this assertion because we actually do have events in the same process now.

We need to call capture_events or otherwise our transport will raise an error.

tests/integrations/celery/test_celery.py
rhcarvalho: Sounds good on the perspective of speeding up tests -- won't this have an effect on bugs and code paths surfaced by tests, though?

untitaker: So we haven't had any bugs related to this *so far* and my gut feeling tells me no. There is no direct communication between the forked processes so I'd say it's unlikely, and the tests won't send in high event volumes ever anyway.

rhcarvalho (approved):

flub (approved):

PR #833, opened 2020-09-23
Replace PyPI page with README.md

cclauss:
https://pypi.org/project/sentry-sdk/ is currently a bit too spartan and requires the user to do extra clicks to determine if they want to install or upgrade this package.

This PR proposes to replace the current one-liner with the markdown content from `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,8 @@ raise ValueError()  # Will also create an event.
 
 # Contributing to the SDK
 
-Please refer to [CONTRIBUTING.md](./CONTRIBUTING.md).
+Please refer to [CONTRIBUTING.md](https://github.com/getsentry/sentry-python/blob/master/CONTRIBUTING.md).
 
 # License
 
-Licensed under the BSD license, see [`LICENSE`](./LICENSE)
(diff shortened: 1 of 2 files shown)


untitaker:
Is there a place where we can figure out which constructs are supported in pypi? We have some html in the readme

cclauss:
https://github.com/pypa/warehouse is the place to open/read issues on these topics.

You can always [push a trial version to the ___testpypi___ server](https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives) with `python3 -m twine upload --repository testpypi dist/*`.  This will go to a test server instead of the public PyPI so you can see if/how the HTML renders.  Also, check the relative links in markdown.  These might need to be converted to absolute links to work correctly on PyPI.

My first commit tried to add CHANGES.md to the end but that file is not available.

untitaker:
Thanks! 👍

cclauss:
Which repo contains the Sentry web user interface or is that not open source?

untitaker:
That's getsentry/sentry. I think you already contributed to it :)

PR #836, opened 2020-09-24
tests: parametrize proxy tests

xavfernandez:
While working on adding tests for an other feature, I stumbled on those tests and found it hard to check which cases were tested among the 16 present tests.

Hopefully it's easier to read in the new version.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -55,107 +55,42 @@ def test_transport_option(monkeypatch):
     assert str(Client(transport=transport).dsn) == dsn
 
 
-def test_proxy_http_use(monkeypatch):
-    client = Client("http://foo@sentry.io/123", http_proxy="http://localhost/123")
-    assert client.transport._pool.proxy.scheme == "http"
-
-
-def test_proxy_https_use(monkeypatch):
-    client = Client("https://foo@sentry.io/123", http_proxy="https://localhost/123")
-    assert client.transport._pool.proxy.scheme == "https"
-
-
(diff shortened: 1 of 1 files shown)


tests/test_client.py
```diff
@@ -55,107 +55,113 @@ def test_transport_option(monkeypatch):
...
-
-def test_proxy_httpsselect_bothenv_http(monkeypatch):
-    monkeypatch.setenv("HTTP_PROXY", "http://localhost/123")
-    monkeypatch.setenv("HTTPS_PROXY", "https://localhost/123")
-    client = Client("http://foo@sentry.io/123", http_proxy=None, https_proxy=None)
-    assert client.transport._pool.proxy.scheme == "http"
+@pytest.mark.parametrize(
+    "dsn,env_http_proxy,env_https_proxy,arg_http_proxy,arg_https_proxy,expected_proxy_scheme",
+    [
+        ("http://foo@sentry.io/123", None, None, "http://localhost/123", None, "http"),
```
untitaker, line 61: I wonder if instead of these tuples we should have one parameter called `testcase` and pass dictionaries into the testcase. I think this is hard to read because you need to memorize which tuple index maps to which param.

untitaker (changes requested):
thanks for working on this, I'd like one change though

tests/test_client.py
xavfernandez: Done :)

untitaker (approved):

untitaker:
excellent, thanks!

Issue #178, opened 2018-11-24
Send events without proxy  when http_proxy is configured as env variable


cichy767:
Hi,
I have problem with configuration of sentry_sdk. My code is behind corporate proxy and `http_proxy` is configured on machine as env variable (required when I want to install something via pip for example). Also I have env variable `no_proxy` where I added ip address for host where sentry is working (this ip is inside corporate network). When I try send event to sentry I got return code 403 (proxy refuse request). Probably sentry_sdk automacally fetch proxy from http_proxy env but ignore no_proxy. When I remove http_proxy from env everything is working. How can I configure python_sdk to ignore my http_proxy env variable? My code: 

```
import logging
import sentry_sdk
from sentry_sdk.integrations.logging import LoggingIntegration


sentry_logging = LoggingIntegration(
    level=logging.INFO,       
    event_level=logging.ERROR 
)
sentry_sdk.init(
    dsn="<dsn>",
    integrations=[sentry_logging],
    debug=True,
)
```

untitaker:
Hi, yes there's currently no way to disable `http_proxy` if you have an envvar. I am not sure if `no_proxy` is easy to implement.

gelmiro:
The same problem behid proxy. In this situation sentry_sdk is difficult to use. Maybe better solution will be use requests instead urllib3? Requests can handle no_proxy out of the box.

amiraliakbari:
I want to add that switching to requests library can have other benefits like supporting SOCKS proxies. Currently, providing SOCKS proxy to `http_proxy` parameter results in `ProxySchemeUnknown` exception. To fix that with urllib3, you should explicitly check for `socks://` schemes and use `SOCKSProxyManager` instead of `ProxyManager`, but requests handles it seamlessly.

untitaker:
To keep this issue updated: We discussed this internally and decided on a hard rule not to use requests. The reason for this is that we would like to explicitly control which behavior we support (since all of the new SDKs share a single piece of documentation describing that behavior, and the behavior across SDKs and across languages should be the same), and requests ideologically is the opposite of that: I for one was unaware that requests reads environment variables like that, even though I've been using it for quite a while.

I would propose that we keep the current, simple, stupid proxying behavior, but add a new flag that disables any access to the process environment. Is this an acceptable fix?

kennell:
> I would propose that we keep the current, simple, stupid proxying behavior, but add a new flag that disables any access to the process environment. Is this an acceptable fix?

[...]

PR #838, opened 2020-09-24
Add basic support for no_proxy environment variable

xavfernandez:
This would match `http_proxy`/`https_proxy` support and was actually supported in the old raven-python client.

Given the lack of a standard on `no_proxy` interpretation (with/without asterisk, with IP CIDR, etc), I went with a first basic implementation that should cover the most frequent usecase.

### DIFF
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -276,6 +276,17 @@ class HttpTransport(Transport):
             "ca_certs": ca_certs or certifi.where(),
         }
 
+    def _in_no_proxy(self, parsed_dsn):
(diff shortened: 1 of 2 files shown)


untitaker:
Is this for https://github.com/getsentry/sentry-python/issues/178? Implementation looks ok, will look tomorrow closer.

I overlooked this but I think we may want some testcases for the lowercase envvar as that's hte one used by curl

untitaker (approved):

untitaker:
Excellent, thanks!

xavfernandez:
Yes, it was for #178 . Thanks for merging :)

PR #842, opened 2020-09-29
fix(django): Do not patch resolver_match

### LINKED ISSUE #837: Django view functions replaced by wrapper
Django 2.2.16
Sentry 0.16.4+

When using sentry 0.16.4+, the resolver has been patched to return phony view functions that wrap calling the actual view function. This breaks anything that uses the resolver matched view function to compare to an imported function, eg we have middleware acting as a guardian with code like this:

```python
@custom_login_required_decorator
def the_view(request):
    request.session["skip"] = 1
    return redirect(reverse("homepage"))


def can_skip(request):
    return "skip" in request.session


class Guardian:
    def process_view(self, request, view_func, view_args, view_kwargs):
        if can_skip(request) or view_func == the_view:
            return
        return redirect(reverse("the_view"))
```

With sentry 0.16.3 (and every version of sentry-sdk before that, and with raven...), this worked fine. With sentry 0.16.4+, the view_func has been replaced. I'm pretty sure that 193f591 is responsible (and this is by design...). If this is the only way of achieving the functionality you need from the views, then at the very least this deficiency needs to be documented prominently - [here perhaps](https://docs.sentry.io/platforms/python/guides/django/#behavior)

With sentry 0.17.8 enabled:
```
(Pdb) p view_func                                                      
<function set_client_code at 0x7fc81f1c26a8>                           
(Pdb) p set_client_code                                                
<portal.utils.login_utils._ExtendedCheckLogin object at 0x7fc8210b54a8>
```

With sentry 0.17.8, but disabled:
```
(Pdb) p view_func                                                      
<portal.utils.login_utils._ExtendedCheckLogin object at 0x7fc63afbb128>
(Pdb) p set_client_code                                                
<portal.utils.login_utils._ExtendedCheckLogin object at 0x7fc63afbb128>
```

With sentry 0.16.3 enabled:
```
(Pdb) p view_func                                                      
<portal.utils.login_utils._ExtendedCheckLogin object at 0x7f3324e504a8>
(Pdb) p set_client_code                                                
<portal.utils.login_utils._ExtendedCheckLogin object at 0x7f3324e504a8>
```

untitaker:
Fix #837

### DIFF
--- a/sentry_sdk/integrations/django/views.py
+++ b/sentry_sdk/integrations/django/views.py
@@ -8,60 +8,33 @@ if MYPY:
     from django.urls.resolvers import ResolverMatch
 
 
-def patch_resolver():
+def patch_views():
     # type: () -> None
-    try:
-        from django.urls.resolvers import URLResolver
-    except ImportError:
-        try:
-            from django.urls.resolvers import RegexURLResolver as URLResolver
-        except ImportError:
-            from django.core.urlresolvers import RegexURLResolver as URLResolver
 
+    from django.core.handlers.base import BaseHandler
     from sentry_sdk.integrations.django import DjangoIntegration
 
-    old_resolve = URLResolver.resolve
+    old_make_view_atomic = BaseHandler.make_view_atomic
 
-    def resolve(self, path):
-        # type: (URLResolver, Any) -> ResolverMatch
+    @_functools.wraps(old_make_view_atomic)
+    def sentry_patched_make_view_atomic(self, *args, **kwargs):
+        callback = old_make_view_atomic(self, *args, **kwargs)
+
+        # XXX: The wrapper function is created for every request. Find more
+        # efficient way to wrap views (or build a cache?)
+
         hub = Hub.current
         integration = hub.get_integration(DjangoIntegration)
 
-        if integration is None or not integration.middleware_spans:
-            return old_resolve(self, path)
+        if integration is not None and integration.middleware_spans:
+            @_functools.wraps(callback)
+            def sentry_wrapped_callback(request, *args, **kwargs):
+                with hub.start_span(op="django.view", description=request.resolver_match.view_name):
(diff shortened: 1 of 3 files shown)


rhcarvalho:
How does this change affect the timings of views?

sentry_sdk/integrations/django/views.py
```diff
@@ -8,60 +8,36 @@
...
-        try:
-            from django.urls.resolvers import RegexURLResolver as URLResolver
-        except ImportError:
-            from django.core.urlresolvers import RegexURLResolver as URLResolver
 
+    from django.core.handlers.base import BaseHandler
     from sentry_sdk.integrations.django import DjangoIntegration
 
-    old_resolve = URLResolver.resolve
+    old_make_view_atomic = BaseHandler.make_view_atomic
```
rhcarvalho, line 17: Is `make_view_atomic` available in every version of Django we support?

Had quick look at
https://github.com/django/django/blob/c1442e1192057a3bf14aecbaa1b713eee139eaff/django/core/handlers/base.py#L320

It is not so obvious that this is called for every view once.

untitaker: `make_view_atomic` is called in `get_response_async` and `get_response` which is where all requests go through. It is the only thing called directly before accessing the resolver match, so it should be pretty close with regard to how timings are done.

For example called here: https://github.com/django/django/blob/c1442e1192057a3bf14aecbaa1b713eee139eaff/django/core/handlers/base.py#L174

The impact on timing I could see is that the function wrapper is created at a different point in time, so it could shift instrumentation overhead to a different point in time and therefore move a small gap in teh trace somewhere else.

rhcarvalho (approved):
It looks like it could possibly work 😄 

Good job @untitaker! As a side-effect we got rid of the `csrf_exempt` exceptional handling, which is a nice bonus.

PR #845, opened 2020-10-02
feat: Auto enable integrations=true

HazAT:
We want to make auto enabling integrations the new default.

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -128,7 +128,7 @@ class _Client(object):
             self.integrations = setup_integrations(
                 self.options["integrations"],
                 with_defaults=self.options["default_integrations"],
-                with_auto_enabling_integrations=self.options["_experiments"].get(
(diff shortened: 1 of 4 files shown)


sentry_sdk/client.py
```diff
@@ -128,7 +128,10 @@ def _send_sessions(sessions):
             self.integrations = setup_integrations(
                 self.options["integrations"],
                 with_defaults=self.options["default_integrations"],
-                with_auto_enabling_integrations=self.options["_experiments"].get(
+                with_auto_enabling_integrations=self.options[
+                    "auto_enabling_integrations"
+                ]
+                or self.options["_experiments"].get(
                     "auto_enabling_integrations", False
                 ),
```
untitaker, line 136: let's get rid of this flag, there is no backwards compat concern if auto enabling integrations are becoming the default

dcramer:
<3 thanks folks

HazAT:
We really need to improve CI build times
image

PR #852, opened 2020-10-09
feat(test): Add `only` pytest marker

lobsterkatie:
This adds a pytest marker similar to `it.only` in jest.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,6 @@
 [pytest]
 DJANGO_SETTINGS_MODULE = tests.integrations.django.myapp.settings
 addopts = --tb=short
-markers = tests_internal_exceptions
+markers =
+    tests_internal_exceptions
+    only: A temporary marker, to make pytest only run the tests with the mark, similar to jest's `it.only`. To use, run `pytest -v -m only`.


pytest.ini
```diff
@@ -1,4 +1,6 @@
 [pytest]
 DJANGO_SETTINGS_MODULE = tests.integrations.django.myapp.settings
 addopts = --tb=short
-markers = tests_internal_exceptions
+markers =
+    tests_internal_exceptions
+    only: A temporary marker, to make pytest only run the tests with the mark, similar to jest's `it.only`. To use, run `pytest -v -m only`.
```
flub, line 6: Is this needed?  There is no `--strict` default set so you should be able to use any un-registered marker already.

(also, do you know of `-k` which can be handy for some sub-set of this usecase)

lobsterkatie: Yup, I know about `-k` but if I want to run just `test_dog`, `test_cat`, and `test_bird`, I have to write an expression which captures those and only those - way easier to just mark them individually while I'm testing. 

And yes - technically no need to register the marker since we don't have strict, but is there any reason _not_ to register it, so that it comes up as an option when you run `pytest --markers`?

flub: Sure, sorry - should have approved this right away, there was no reason not to leave it up to you :)

flub (approved):

pytest.ini
lobsterkatie: No worries. I appreciate the pushback. Not every idea I have is a good one, LOL!

PR #853, opened 2020-10-09
feat(testing): Add pytest-watch

lobsterkatie:
This adds `pytest-watch` to the test requirements, in order to be able to run `pytest` in watch mode. It also fixes two places where we weren't closing resources at the end of a test. (Normally, they're closed by default because the testing process ends. But with a watcher it doesn't end, and causing `pytest-watch` to complain about the abandoned resources.)


tests/integrations/stdlib/test_httplib.py
```diff
@@ -106,3 +106,5 @@ def test_httplib_misuse(sentry_init, capture_events):
         "status_code": 200,
         "reason": "OK",
     }
+
+    conn.close()
```
flub, line 110: I think this is better to use the `request` fixture and `request.addfinalizer(conn.close)` right after the conn was created on line 84.  That way this gets cleaned up even when then test fails

flub (approved):

tests/integrations/stdlib/test_httplib.py
lobsterkatie: Oh, good idea, thanks. Done.

untitaker (approved):

PR #854, opened 2020-10-12
Fix incorrect timeout warnings in AWS Lambda and GCP integrations

### LINKED ISSUE #846: Incorrect timeout warnings when using AWS Lambda integration
sentry-sdk version: 0.17.8
AWS Lambda runtime: Python 3.8

My lambda function has a timeout of 29 seconds, but typically completes in <1s. When the lambda function is called multiple times in quick succession, the container for the lambda function is reused. When this happens, the threads that Sentry creates to manage the timeout warnings produce their warnings in subsequent function invocations (long after the function that they were created in has completed).

shantanu73:
1) Added code to stop thread in aws_lambda.py & gcp.py.
2) Modified logic of run() function of class TimeoutThread to stop the thread and raise ServerlessTimeoutWarning exception, conditionally.

Fixes #846.

### DIFF
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -116,6 +116,13 @@ def _wrap_handler(handler):
                     )
                     hub.capture_event(event, hint=hint)
                     reraise(*exc_info)
+                finally:
+                    if (
+                        integration.timeout_warning
+                        and configured_time > TIMEOUT_WARNING_BUFFER
+                    ):
+                        timeout_thread.stop_thread = True
+                        timeout_thread.join()
 
     return sentry_handler  # type: ignore
 
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -83,6 +83,8 @@ def _wrap_handler(handler):
                     _make_request_event_processor(event, context, configured_time)
                 )
                 scope.set_tag("aws_region", context.invoked_function_arn.split(":")[3])
+
+                timeout_thread = None
                 # Starting the Timeout thread only if the configured time is greater than Timeout warning
                 # buffer and timeout_warning parameter is set True.
                 if (
@@ -94,7 +96,8 @@ def _wrap_handler(handler):
                     ) / MILLIS_TO_SECONDS
 
                     timeout_thread = TimeoutThread(
-                        waiting_time, configured_time / MILLIS_TO_SECONDS
+                        waiting_time,
+                        configured_time / MILLIS_TO_SECONDS,
                     )
 
                     # Starting the thread to raise timeout warning exception
@@ -117,12 +120,8 @@ def _wrap_handler(handler):
                     hub.capture_event(event, hint=hint)
                     reraise(*exc_info)
                 finally:
-                    if (
-                        integration.timeout_warning
-                        and configured_time > TIMEOUT_WARNING_BUFFER
-                    ):
-                        timeout_thread.stop_thread = True
-                        timeout_thread.join()
+                    if timeout_thread:
+                        timeout_thread.stop()
 
     return sentry_handler  # type: ignore
 
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -903,7 +903,7 @@ class TimeoutThread(threading.Thread):
 
         if self._stop_event.is_set():
             return
-        
+
         integer_configured_timeout = int(self.configured_timeout)
 
         # Setting up the exact integer value of configured time(in seconds)
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -891,27 +891,22 @@ class TimeoutThread(threading.Thread):
         threading.Thread.__init__(self)
         self.waiting_time = waiting_time
         self.configured_timeout = configured_timeout
-        self.stop_thread = False
+        self._stop_event = threading.Event()
+
+    def stop(self):
+        # type: () -> None
+        self._stop_event.set()
+
+    def stopped(self):
+        # type: () -> Any
+        return self._stop_event.is_set()
 
     def run(self):
         # type: () -> None
 
-        raise_exception = True
-        threading_is_running = True
-        start_time = time.time()
+        time.sleep(self.waiting_time)
 
-        while threading_is_running:
-            if self.stop_thread:
-                raise_exception = False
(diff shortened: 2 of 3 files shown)


sentry_sdk/integrations/aws_lambda.py
```diff
@@ -116,6 +116,13 @@ def sentry_handler(event, context, *args, **kwargs):
                     )
                     hub.capture_event(event, hint=hint)
                     reraise(*exc_info)
+                finally:
```
shantanu73, line 119: @untitaker In this "finally" block, the code to stop the thread basically works in case of :-
1) A successful run of lambda function.
2) When an exception is there in the code.
But, it doesn't stop the thread in case of timeout, which is an expected behavior.
By "stop the thread" I mean explicitly stopping a thread by sending a stop signal.

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -116,6 +116,13 @@ def sentry_handler(event, context, *args, **kwargs):
...
+                    if (
+                        integration.timeout_warning
+                        and configured_time > TIMEOUT_WARNING_BUFFER
+                    ):
+                        timeout_thread.stop_thread = True
+                        timeout_thread.join()
```
rhcarvalho, line 125: This could potentially block the main thread of the lambda handler, which is undesirable.

sentry_sdk/utils.py
```diff
@@ -891,21 +891,32 @@ def __init__(self, waiting_time, configured_timeout):
...
+        threading_is_running = True
+        while threading_is_running and self.waiting_time > 1:
+            time.sleep(1)
+            self.waiting_time = self.waiting_time - 1
+            if self.stop_thread:
+                raise_exception = False
+                threading_is_running = False
 
-        integer_configured_timeout = int(self.configured_timeout)
+        if raise_exception:
```
rhcarvalho, line 912: This is racy -- there is an execution flow where `raise_exception` is `True` and `self.stop_thread` is also `True` (think the main thread may change `stop_thread` between the end of the `while` loop and the beginning of the `if` block).

rhcarvalho (commented):
Maybe using a [`Timer`](https://docs.python.org/3/library/threading.html#timer-objects) would make things cleaner.

- Every handler execution starts a new timer (unless we don't have enough "buffer"/threshold).
- Whenever the original handler finishes running (the new `finally` block in this PR) and a timer has been started, call `timer.cancel()` -- no additional condition.
- The Timer runs a function that raises the timeout exception, no conditions.

The above is my suggestion.

---

**Side note**
Because the runtime can abort/pause the handler at any time, I imagine there'd be room for the timer firing incorrectly into a future handler execution, but not sure if there's anything we could do about it. (I thought of an Event to mark the end of the handler and the timer could check it, but that doesn't change anything if the `finally` block of one execution runs into the next execution -- depends how the timeout is implemented by the runtime.)

sentry_sdk/utils.py
```diff
@@ -898,12 +898,16 @@ def run(self):
...
+        start_time = time.time()
+
+        while threading_is_running:
             if self.stop_thread:
                 raise_exception = False
                 threading_is_running = False
+            current_time = time.time()
+            elapsed_time = current_time - start_time
+            if elapsed_time >= self.waiting_time:
+                threading_is_running = False
```
rhcarvalho: This is worse than the original in that it became a CPU intensive loop.

rhcarvalho:
Alternative to a Timer, Thread + Event to communicate the stop like in this example https://stackoverflow.com/a/325528/963287
`Event.wait(timeout)` can be used instead of the loop that sleeps second by second.

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -116,6 +116,13 @@ def sentry_handler(event, context, *args, **kwargs):
                     )
                     hub.capture_event(event, hint=hint)
                     reraise(*exc_info)
+                finally:
+                    if (
+                        integration.timeout_warning
+                        and configured_time > TIMEOUT_WARNING_BUFFER
+                    ):
```
rhcarvalho, line 123: I see that `timeout_thread` is only conditionally defined. That's why you need to repeat this condition here.

It would be clearer to always define `timeout_thread`, initially setting it to `None`.
That way, here it would become:

```py
if timeout_thread is not None:
    timeout_thread.stop()
```

`.stop()` sets a "stop event" to true, unblocking and terminating the thread if still running, and doing nothing otherwise.

shantanu73: Made this change in latest commit.

sentry_sdk/utils.py
```diff
@@ -891,21 +891,31 @@ def __init__(self, waiting_time, configured_timeout):
...
+        self._stop_event.set()
+
+    def stopped(self):
+        # type: () -> Any
+        return self._stop_event.is_set()
 
     def run(self):
         # type: () -> None
 
         time.sleep(self.waiting_time)
```
rhcarvalho, line 907: The thread doesn't need to sleep for the full `waiting_time`. When the handler returns and we call `timeout_thread.stop()`, the thread should wake up and terminate.

That's why I suggested using `self._stop_event.wait(self.waiting_time)`.

https://docs.python.org/2.7/library/threading.html#threading.Event.wait

You can either use the return value or `wait` or `.is_set()` to decide whether to raise `ServerlessTimeoutWarning`.

sentry_sdk/utils.py
```diff
@@ -891,21 +890,27 @@ def __init__(self, waiting_time, configured_timeout):
...
+        self._stop_event.set()
 
     def run(self):
         # type: () -> None
 
-        time.sleep(self.waiting_time)
+        self._stop_event.wait(self.waiting_time)
 
-        integer_configured_timeout = int(self.configured_timeout)
+        if not self._stop_event.is_set():
```
shantanu73, line 904: @rhcarvalho I made final changes, if these look okay then PR can be merged.

sentry_sdk/utils.py
shantanu73: Yes I figured that while I did testing and found it was still waiting for complete waiting time. 
I've made final changes accordingly now.

rhcarvalho (approved):
Thanks @shantanu73 👍 

I've made a small change to avoid increasing indentation.
Would be nice to have a test case to cover this bug and prevent regression in a follow up.

Let's get this in first though. Thanks again!

PR #857, opened 2020-10-12
ref(tests): Split up tracing tests

lobsterkatie:
The tracing test file was already getting long, and I'm about to add a bunch of new tests. Zero behavior change here - literally just a copy-paste job.

(Enough of the tests from the old `test_tracing.py` ended up in the new `tracing/test_integration_tests.py` that git is calling it a rename-with-a-bunch-of-changes, rather than a deletion of the former and creation of the latter. Unclear how to force GH into `git diff --no-renames`.)


tests/tracing/test_integration_tests.py
```diff
@@ -1,6 +1,5 @@
 import weakref
 import gc
-
```
flub, line 3: I think in pep8 the recommendation is to keep a blank line between stdlib imports and 3rd party imports.  Not sure why this is removed?

tests/tracing/test_integration_tests.py
```diff
@@ -46,25 +45,9 @@ def test_basic(sentry_init, capture_events, sample_rate):
...
-        pass
-
-    assert len(events) == 2
-    assert events[0]["transaction"] == "/1/"
-    assert events[1]["transaction"] == "/2/"
-
-
 @pytest.mark.parametrize("sampled", [True, False, None])
 def test_continue_from_headers(sentry_init, capture_events, sampled):
+    """"""
```
flub, line 50: Forgot to remove this again?

flub (approved):

tests/tracing/test_integration_tests.py
lobsterkatie: Shoot - good catch. Thanks.

PR #863, opened 2020-10-16
Add `traces_sampler` option

lobsterkatie:
This adds a `traces_sampler` option to the SDK, allowing dynamic sampling and filtering of transactions, as outlined [here](https://develop.sentry.dev/sdk/unified-api/tracing).

The one piece of that spec that this doesn't yet implement is pulling in default sampling context values from various frameworks/integrations, which will happen in subsequent PRs.

### DIFF
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -207,8 +207,7 @@ def start_span(
 @hubmethod
 def start_transaction(
     transaction=None,  # type: Optional[Transaction]
-    custom_sampling_context=None,  # type: Optional[Dict[str, Any]]
     **kwargs  # type: Any
 ):
     # type: (...) -> Transaction
-    return Hub.current.start_transaction(transaction, custom_sampling_context, **kwargs)
+    return Hub.current.start_transaction(transaction, **kwargs)
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -207,7 +207,8 @@ def start_span(
 @hubmethod
 def start_transaction(
     transaction=None,  # type: Optional[Transaction]
+    custom_sampling_context=None,  # type: Optional[Dict[str, Any]]
     **kwargs  # type: Any
 ):
     # type: (...) -> Transaction
-    return Hub.current.start_transaction(transaction, **kwargs)
+    return Hub.current.start_transaction(transaction, custom_sampling_context, **kwargs)
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -631,7 +631,7 @@ class Transaction(Span):
         # 0, it's a sign the transaction should be dropped
         if not sample_rate:
             logger.debug(
-                "[Tracing] Discarding {transaction_description} because ${reason}".format(
+                "[Tracing] Discarding {transaction_description} because {reason}".format(
                     transaction_description=transaction_description,
                     reason=(
                         "traces_sampler returned 0 or False"
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -337,7 +337,7 @@ class Span(object):
         return Transaction(
             trace_id=trace_id,
             parent_span_id=parent_span_id,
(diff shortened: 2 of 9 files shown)


sentry_sdk/tracing.py
```diff
@@ -555,6 +565,112 @@ def to_json(self):
...
+        rate for the transaction.
+
+        3. If `traces_sampler` is not defined, but there's a parent sampling
+        decision, the parent sampling decision will be used.
+
+        4. If `traces_sampler` is not defined and there's no parent sampling
+        decision, `traces_sample_rate` will be used.
+        """
+
+        def _inherit_or_use_static_rate(parent_sampled, traces_sample_rate):
```
untitaker, line 590: Put this into the global namespace, otherwise there's a function that's constantly being defined and discarded, on the hot path

sentry_sdk/tracing.py
```diff
@@ -2,14 +2,21 @@
...
 from datetime import datetime, timedelta
+from inspect import isfunction
 from numbers import Real
 
 import sentry_sdk
 
-from sentry_sdk.utils import capture_internal_exceptions, logger, to_string
+from sentry_sdk.utils import (
+    capture_internal_exceptions,
+    has_tracing_enabled,
```
untitaker, line 16: I know this was part of a diff PR but can we move `has_tracing_enabled` into this module?

untitaker (approved):
just doing a perf review of this since i have no context. I'm just going to trust you with the impl since this since rodolfo is ooo and you already implemented the JS one anyway

sentry_sdk/tracing.py
untitaker: actually this is only used once, just inline it (max perf!!!!)

sentry_sdk/tracing.py
```diff
@@ -555,6 +565,112 @@ def to_json(self):
...
+                )
+            )
+            self.sampled = False
+            return
+
+        # if the function returned 0 (or false), or if `traces_sample_rate` is
+        # 0, it's a sign the transaction should be dropped
+        if not sample_rate:
+            logger.debug(
+                "[Tracing] Discarding {transaction_description} because ${reason}".format(
```
untitaker, line 643:
```suggestion
                "[Tracing] Discarding {transaction_description} because {reason}".format(
```

sentry_sdk/tracing.py
```diff
@@ -555,6 +565,112 @@ def to_json(self):
...
+        # value when starting the transaction, go with that
+        if self.sampled is not None:
+            return
+
+        # we would have bailed already if neither `traces_sampler` nor
+        # `traces_sample_rate` were defined, so one of these should work; prefer
+        # the hook if so
+        sample_rate = (
+            options["traces_sampler"](sampling_context)
+            if isfunction(options.get("traces_sampler"))
```
untitaker, line 621:
```suggestion
            if callable(options.get("traces_sampler"))
```

sentry_sdk/api.py
```diff
@@ -207,7 +207,8 @@ def start_span(
 @hubmethod
 def start_transaction(
     transaction=None,  # type: Optional[Transaction]
+    custom_sampling_context=None,  # type: Optional[Dict[str, Any]]
```
untitaker, line 210: Is there a need for this change? This would just end up in kwargs

sentry_sdk/tracing.py
lobsterkatie: TBH, it was more about giving a name than anything else, LOL. I can switch that to a comment and inline it.

sentry_sdk/api.py
lobsterkatie: So it shows up in the automatically-generated docs? But can demote it if you'd like.

untitaker: feel like we need a different solution, btw docstring already links to the wrapped method

lobsterkatie: Good to know. Anyway, I've already pulled it into kwargs.

PR #864, opened 2020-10-16
feat(tracing): Add types for `traces_sampler` implementation

lobsterkatie:
This adds (but doesn't yet use) the types and attributes necessary for adding the `traces_sampler` option which is specced [here](https://develop.sentry.dev/sdk/unified-api/tracing). Extracted from the larger PR on that topic, https://github.com/getsentry/sentry-python/pull/863.

- Types for the `traces_sampler` itself (the function and its input)
- A new attribute on the `Transaction` class tracking the parent sampling decision separately from the sampling decision of the transaction itself, since part of the `traces_sampler` spec is that there needs to be a difference between an inherited decision and an explicitly set decision.

### DIFF
--- a/sentry_sdk/_types.py
+++ b/sentry_sdk/_types.py
@@ -5,6 +5,7 @@ except ImportError:
 
 
 if MYPY:
+    from numbers import Real
     from types import TracebackType
     from typing import Any
     from typing import Callable
@@ -12,6 +13,7 @@ if MYPY:
     from typing import Optional
     from typing import Tuple
     from typing import Type
+    from typing import Union
     from typing_extensions import Literal
 
     ExcInfo = Tuple[
@@ -24,9 +26,13 @@ if MYPY:
     Breadcrumb = Dict[str, Any]
     BreadcrumbHint = Dict[str, Any]
 
+    SamplingContext = Dict[str, Any]
+
     EventProcessor = Callable[[Event, Hint], Optional[Event]]
     ErrorProcessor = Callable[[Event, ExcInfo], Optional[Event]]
     BreadcrumbProcessor = Callable[[Breadcrumb, BreadcrumbHint], Optional[Breadcrumb]]
+
+    TracesSampler = Callable[[SamplingContext], Union[Real, bool]]
 
     # https://github.com/python/mypy/issues/5710
     NotImplementedType = Any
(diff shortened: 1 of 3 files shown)


sentry_sdk/_types.py
```diff
@@ -24,10 +25,15 @@
...
     BreadcrumbHint = Dict[str, Any]
 
+    SamplingContext = Dict[str, Any]
+
     EventProcessor = Callable[[Event, Hint], Optional[Event]]
     ErrorProcessor = Callable[[Event, ExcInfo], Optional[Event]]
     BreadcrumbProcessor = Callable[[Breadcrumb, BreadcrumbHint], Optional[Breadcrumb]]
 
+    # booleans count as instances of Number, so this covers both return types
+    TracesSampler = Callable[[SamplingContext], Number]
```
rhcarvalho, line 35:
```python-console
>>> import this
...
Explicit is better than implicit.
...
```

```py
from typing import Union
```

```suggestion
    TracesSampler = Callable[[SamplingContext], Union[Number, bool]]
```

sentry_sdk/tracing.py
```diff
@@ -449,11 +449,12 @@ def get_trace_context(self):
 
 
 class Transaction(Span):
-    __slots__ = ("name",)
+    __slots__ = ("name", "parent_sampled")
```
rhcarvalho, line 452: Why do we need to store this value in every transaction?

sentry_sdk/_types.py
lobsterkatie: Fair. Thanks for the fix.

sentry_sdk/tracing.py
lobsterkatie: Because there's no other way (that I can see) to get it to the `traces_sampler`. The only link between `from_traceparent` and `start_transaction` is the `Transaction` object itself.

It's true that once it's been used, I could delete it off of the transaction object, but is that necessary/worth it?

untitaker (commented):
I have zero context on this PR, I don't know how to review this

lobsterkatie:
> I have zero context on this PR, I don't know how to review this

Yeah, sorry - Rodolfo just went through helping me with the JS version of this, so I probably didn't explain as much as I should have in the PR description since he was originally going to be the one reviewing. I've added more context above.

untitaker (approved):

PR #865, opened 2020-10-16
ref(tracing): Pre-`traces_sampler` documentation additions

lobsterkatie:
A bunch of comments and docstrings which got added while I was working on https://github.com/getsentry/sentry-python/pull/863. Also adds data to both the `Span` and `Transaction` reprs, and changes a handful of variable names to be more accurate. No behavior changes.

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -166,8 +166,10 @@ class Span(object):
 
     def __repr__(self):
         # type: () -> str
-        return "<%s(trace_id=%r, span_id=%r, parent_span_id=%r, sampled=%r)>" % (
+        return "<%s(op=%r, description:%r, trace_id=%r, span_id=%r, parent_span_id=%r, sampled=%r)>" % (
             self.__class__.__name__,
+            self.op,
+            self.description,
             self.trace_id,
             self.span_id,
             self.parent_span_id,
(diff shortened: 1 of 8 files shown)


sentry_sdk/tracing.py
```diff
@@ -226,6 +229,14 @@ def continue_from_environ(
...
     ):
+        """
+        Create a Transaction with the given params, then add in data pulled from
+        the 'sentry-trace' header in the environ (if any) before returning the
+        Transaction.
+
+        If the 'sentry-trace' header is malformed or missing, just create and
+        return a Transaction instance with the given params.
+        """
         # type: (...) -> Transaction
```
flub, line 240: The linter doesn't seem to like this order of type comment after the docstring?

tests/integrations/stdlib/test_httplib.py
```diff
@@ -4,13 +4,17 @@
 import pytest
 
 try:
+    # py3
```
flub, line 7: I've seem people argue for `if PY3K:` kind of variations for these to make this clearer.  Anyway, this is fine by me - i don't know if we have internal coding standards which have other opinions.

flub (approved):

sentry_sdk/tracing.py
lobsterkatie: Oh, _that's_ what it was mad about! The problematic ordering predates me, so I figured it was just the linter not being 100% on board with type hints. Fixed now, thanks.

tests/integrations/stdlib/test_httplib.py
lobsterkatie: Yeah, that might be better, but for now my goal was only to mark which was which.

sentry_sdk/tracing.py
lobsterkatie: Ugh. After a lot of poking, I figured out what the problem is: `mypy` wants the type annotations before the docstring, whereas `pylance` (the VSCode Python engine) very much wants them after. So now my editor is all kinds of mad, but at least CI is passing.

PR #867, opened 2020-10-16
feat(dev): Add fixtures for testing `traces_sampler`

lobsterkatie:
Extracted from https://github.com/getsentry/sentry-python/pull/863.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -6,6 +6,7 @@ pytest-localserver==0.5.0
 pytest-cov==2.8.1
 jsonschema==3.2.0
 pyrsistent==0.16.0 # TODO(py3): 0.17.0 requires python3, see https://github.com/tobgu/pyrsistent/issues/205
+mock # for testing under python < 3.3
 
 gevent
 eventlet
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 import os
 import json
+from types import FunctionType
 
 import pytest
 import jsonschema
@@ -35,6 +36,11 @@ except ImportError:
 
 else:
     del pytest_benchmark
+
+try:
+    from unittest import mock  # python 3.3 and above
+except ImportError:
+    import mock  # python < 3.3
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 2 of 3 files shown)


tests/conftest.py
```diff
@@ -325,3 +331,54 @@ def render_span(span):
...
+
+
+@pytest.fixture(name="Any")
+def make_any():
+    class Any(object):
+        """
+        An object which matches any instance of the class passed to __init__.
+
+        Useful for assert_called_with statements.
+        """
```
flub, line 343: Putting the docstring on fixture function will make it show up with `pytest --fixtures` which is friendlier

lobsterkatie: Good to know - thanks!

test-requirements.txt
```diff
@@ -6,6 +6,7 @@ pytest-localserver==0.5.0
 pytest-cov==2.8.1
 jsonschema==3.2.0
 pyrsistent==0.16.0 # TODO(py3): 0.17.0 requires python3, see https://github.com/tobgu/pyrsistent/issues/205
+mock # for testing under python < 3.3
```
untitaker, line 9: Instead of doing this please just use the pytest monkeypatch fixture

untitaker (approved):
Can't we just include this as part of the PR where the functions are actually used?

lobsterkatie:
> Can't we just include this as part of the PR where the functions are actually used? I think I am reviewing this PR without context

Sure, if you'd rather. I've been getting coaching to keep my PR's a manageable size, so I've tried to pull out from the main PR anything which could be thought of on its own and/or which was a non-behavior-changing prerequisite for the main changes. I recognize that for that to work there's got to be more cross-referencing and possibly more faith that such prereq changes are going to be used in a reasonable way. Still trying to find the balance, I guess. (And very open to advice in this arena!)

Perhaps rather than mushing these fixtures back into the main PR, I could pull the tests out of that and include them here, and just mark them all xfail for the moment?

test-requirements.txt
lobsterkatie: I'm not sure how you mean. I do see how you could use `monkeypatch` for `mock.patch.object` calls, but how would you handle getting `mock.Mock` instances when you need them?

untitaker:
>Perhaps rather than mushing these fixtures back into the main PR, I could pull the tests out of that and include them here, and just mark them all xfail for the moment?

That's also fine, though I think anything under 400loc is reviewable.

test-requirements.txt
untitaker: I don't know how you're going to use mock but I've just been creating my own stub classes. I find a lot of mock's behavior too magical wrt attribute access on the mock obj

tests/conftest.py
```diff
@@ -327,3 +333,56 @@ def render_span(span):
...
+    Just like a mock.Mock object, but one which always passes an isfunction
+    test.
+    """
+
+    class FunctionMock(mock.Mock):
+        def __init__(self, *args, **kwargs):
+            super(FunctionMock, self).__init__(*args, **kwargs)
+            self.__class__ = FunctionType
+
+    return FunctionMock
```
lobsterkatie, line 388: > I don't know how you're going to use mock but I've just been creating my own stub classes. I find a lot of mock's behavior too magical wrt attribute access on the mock obj

Here, for instance:

```python
def test_passes_custom_samling_context_from_start_transaction_to_traces_sampler(
    FunctionMock, sentry_init, DictionaryContaining
):
    traces_sampler = FunctionMock()
    sentry_init(traces_sampler=traces_sampler)

    start_transaction(custom_sampling_context={"dogs": "yes", "cats": "maybe"})

    traces_sampler.assert_any_call(
        DictionaryContaining({"dogs": "yes", "cats": "maybe"})
    )
```

Would rather not have to rewrite the machinery of tracking calls.

untitaker:
```python
def test_passes_custom_samling_context_from_start_transaction_to_traces_sampler(
    sentry_init
):
    sampler_calls = []
    sentry_init(traces_sampler=sampler_calls.append)

    start_transaction(custom_sampling_context={"dogs": "yes", "cats": "maybe"})

    assert sampler_calls == [{"dogs": "yes", "cats": "maybe"}]
```

Unless you need to account for additional keys or calls I would go with this...

untitaker: Feel free to add... I am not entirely sure why people keep using this stuff but I don't want to fight it.

lobsterkatie: Okay, in the interest of time I'm going to leave it for now, but that's an interesting pattern to know about, thanks!

PR #870, opened 2020-10-16
feat(tracing): Make spans point to their transactions

lobsterkatie:
This adds a pointer to the `Span` class which points the the `Transaction` instance from which the span is descended. (It doesn't do anything with it yet, though - that will come in a subsequent PR which allows unsampled transactions living on the scope to be found.)

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -107,6 +107,9 @@ class Span(object):
         "_span_recorder",
         "hub",
         "_context_manager_state",
+        # TODO: rename this "transaction" once we fully and truly deprecate the
+        # old "transaction" attribute (which was actually the transaction name)?
+        "_containing_transaction",
     )
 
     def __new__(cls, **kwargs):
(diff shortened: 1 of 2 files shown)


rhcarvalho:
For reference, the WIP Go implementation always records the span tree, no matter what the sampling decision is.
Would be curious if that has any negative implication, or otherwise we could do the same in Python and JS to eliminate having "yet-another-way" to store references to spans.

lobsterkatie:
> For reference, the WIP Go implementation always records the span tree, no matter what the sampling decision is.
> Would be curious if that has any negative implication, or otherwise we could do the same in Python and JS to eliminate having "yet-another-way" to store references to spans.

This seems better to me for two reasons:

1) Why take up memory storing a span tree if we're not going to keep its data? This does make span trees we keep a teeny bit heavier while we're building them, but on the balance it's still a memory savings unless the sample rate is super high.

2) `self._span_recorder[0]` is a pretty darn opaque way to say "the transaction to which this span belongs," and feels more on the level of implementation detail than ideally even an internal API would be (IMHO).

untitaker (approved):

PR #872, opened 2020-10-16
fix(tracing): Make unsampled transactions findable on the scope

lobsterkatie:
Currently, when trying to find the transaction on the scope (if any), we look for a span on the scope (which could be the transaction itself, or could be one of its descendants) and grab the first span in its span recorder. This works, but has the disadvantage that it can only find transactions with `sampled = True`, since transactions with `sampled = False` don't have a span recorder.

This changes the getter to use the current span's transaction pointer (added in https://github.com/getsentry/sentry-python/pull/870) to find the transaction instead.

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -144,13 +144,25 @@ class Scope(object):
     def transaction(self):
         # type: () -> Any
         # would be type: () -> Optional[Transaction], see https://github.com/python/mypy/issues/3004
-        """Return the transaction (root span) in the scope."""
-        if self._span is None or self._span._span_recorder is None:
+        """Return the transaction (root span) in the scope, if any."""
+
+        # there is no span/transaction on the scope
+        if self._span is None:
             return None
-        try:
-            return self._span._span_recorder.spans[0]
-        except (AttributeError, IndexError):
-            return None
+
+        # the span on the scope is itself a transaction
+        if isinstance(self._span, Transaction):
+            return self._span
+
+        # the span on the scope isn't a transaction but belongs to one
+        if self._span._containing_transaction:
+            return self._span._containing_transaction
+
+        # there's a span (not a transaction) on the scope, but it was started on
+        # its own, not as the descendant of a transaction (this is deprecated
+        # behavior, but as long as the start_span function exists, it can still
+        # happen)
+        return None
 
     @transaction.setter
     def transaction(self, value):
--- a/tests/tracing/test_sampling.py
+++ b/tests/tracing/test_sampling.py
@@ -1,6 +1,6 @@
 import pytest
 
-from sentry_sdk import start_span, start_transaction
+from sentry_sdk import Hub, start_span, start_transaction
 from sentry_sdk.tracing import _is_valid_sample_rate
 from sentry_sdk.utils import logger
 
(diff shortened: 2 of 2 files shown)


HazAT (approved):

sentry_sdk/tracing.py
```diff
@@ -469,6 +475,10 @@ def __init__(
         Span.__init__(self, **kwargs)
         self.name = name
 
+        # transactions (when considered as spans) belong to themselves (when
+        # considered as transactions)
+        self._containing_transaction = self
```
untitaker, line 480: would much rather store `None` or some other special value than create circular references

lobsterkatie: What kind of a special value were you thinking? I suppose it could be `None`, as long as the getting does a typecheck first. Does that seem like a better option?

lobsterkatie:
```python
    @property
    def transaction(self):
        # type: () -> Any
        # would be type: () -> Optional[Transaction], see https://github.com/python/mypy/issues/3004
        """Return the transaction (root span) in the scope, if any."""

        # there is no span/transaction on the scope
        if self._span is None:
            return None

        # the span on the scope is itself a transaction
        if isinstance(self._span, Transaction):
            return self._span

        # the span on the scope isn't a transaction but belongs to one
        if self._span._containing_transaction:
            return self._span._containing_transaction

        # there's a span (not a transaction) on the scope, but it was started on
        # its own, not as the descendant of a transaction (this is deprecated
        # behavior, but as long as the start_span function exists, it can still
        # happen)
        return None
```

lobsterkatie: (Out of curiosity, is there a drawback to the circular reference?)

untitaker: I am not sure if `None` is already used as a special value. If it is we need another one ("self" or something else). But yeah in either case the getter needs to special-case.

lobsterkatie: Hmmm - that also means `start_child` is going to have to special-case, because the attribute's value can't just be passed from parent to child.

Can you help me understand the disadvantages of the circular reference? It seems a lot cleaner.

untitaker: creating circular references is generally less easy on memory as they are collected whenever, as opposed to "linear" ones, and the gc is also slower if it needs to do more.

don't ask me by how much or if it matters, but if it's avoidable it's certainly better. I would create a new sentinel value like this:

```
_SELF = object()
```

and check for it like this:

```
if self._containing_transaction is _SELF:
```

lobsterkatie: Ah, okay. Makes sense, thanks.

Sentinel better than typechecking, you think?

untitaker: Ah no I don't think that matters

lobsterkatie: 👍🏻

lobsterkatie: Okay, fixed in https://github.com/getsentry/sentry-python/pull/870. (Rodolfo has been coaching me to push smaller PRs, but I might have gone overboard a little...)

Once that's approved and merged, I'll push the getter changes.

PR #875, opened 2020-10-17
fix(django): Fix complex either url patterns in Django

ChillarAnand:


untitaker (approved):

untitaker:
Thanks, I think we actually need some matching that cannot be expressed by regex, i.e. actually match open/closing parens, but I think this is good enough.

Can you rebase to fix CI?

untitaker:
Thanks!

PR #880, opened 2020-10-20
fix(dev): Pin `eventlet` version

lobsterkatie:
See https://github.com/eventlet/eventlet/issues/660.

Until that gets addressed, this pins us at a version we know doesn't break py2.7 tests.


untitaker (approved):

temoto:
@lobsterkatie thanks, WIP. May I suggest you'd use relaxed pin next time, like `<0.28.1` so it doesn't conflict with whatever pins other people may want.

temoto:
although i just only now realized it was pin in test requirements, so never mind

PR #882, opened 2020-10-20
fix: Fix celery integration with custom tag class

### LINKED ISSUE #844: Version 0.17.8 broke Celery tasks with custom Task class
I am using [tenant-schemas-celery](https://github.com/maciej-gol/tenant-schemas-celery) package, which generates it's own TenantTask (extends celery.app.task.Task) class in it's own CeleryApp (extends celery.Celery). After updating from sentry-sdk from 0.17.7 to 0.17.8 the TenantTask's tenant context switching stopped working. I suspect this is because of the following change in 0.17.8:

```
diff --git a/sentry_sdk/integrations/celery.py b/sentry_sdk/integrations/celery.py
index 1a11d4a..2b51fe1 100644
--- a/sentry_sdk/integrations/celery.py
+++ b/sentry_sdk/integrations/celery.py
@@ -60,9 +60,8 @@ class CeleryIntegration(Integration):
                 # Need to patch both methods because older celery sometimes

                 # `build_tracer` is apparently called for every task
                 # invocation. Can't wrap every celery task for every invocation
                 # or we will get infinitely nested wrapper functions.
@@ -71,8 +70,12 @@ class CeleryIntegration(Integration):
             return _wrap_tracer(task, old_build_tracer(name, task, *args, **kwargs))

+        from celery.app.task import Task  # type: ignore
+
+        Task.apply_async = _wrap_apply_async(Task.apply_async)
+
         _patch_worker_exit()

         # This logger logs every status of every task that ran on the worker.
         # Meaning that every task's breadcrumbs are full of stuff like "Task
```

I think this problem has to do with this new way of importing celery.app.task.Task. Even though TenantTask extends the celery.app.task.Task, for some reason this change broke the TenantTask logic.

I'm not sure which package this should be fixed in, but I'm bit skeptical about this new import in sentry-sdk, so I'm reporting it here.

Here is my celery.py:

```
from tenant_schemas_celery.app import CeleryApp as TenantAwareCeleryApp


app = TenantAwareCeleryApp()
app.config_from_object('django.conf:settings')
app.autodiscover_tasks()
```

untitaker:
Fix #844

### DIFF
--- a/tests/integrations/celery/test_celery.py
+++ b/tests/integrations/celery/test_celery.py
@@ -99,10 +99,12 @@ def celery_invocation(request):
     """
     return request.param
 
+
 @pytest.mark.parametrize("custom_celery_task_cls", (True, False))
 def test_simple(capture_events, celery, celery_invocation, custom_celery_task_cls):
 
     if custom_celery_task_cls:
+
         class CustomTask(celery.Task):
             def __call__(self, *args, **kwargs):
                 return self.run(*args, **kwargs)
--- a/tests/integrations/celery/test_celery.py
+++ b/tests/integrations/celery/test_celery.py
@@ -23,12 +23,12 @@ def connect_signal(request):
 
 @pytest.fixture
 def init_celery(sentry_init, request):
-    def inner(propagate_traces=True, backend="always_eager", **kwargs):
+    def inner(propagate_traces=True, backend="always_eager", celery_kwargs=None, **kwargs):
         sentry_init(
             integrations=[CeleryIntegration(propagate_traces=propagate_traces)],
             **kwargs
         )
-        celery = Celery(__name__)
+        celery = Celery(__name__, **(celery_kwargs or {}))
 
         if backend == "always_eager":
             if VERSION < (4,):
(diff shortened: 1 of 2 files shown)


untitaker:
Ok, test passes so it's useless. Don't merge.

tests/integrations/celery/test_celery.py
```diff
@@ -100,7 +100,17 @@ def celery_invocation(request):
...
 
 
-def test_simple(capture_events, celery, celery_invocation):
+@pytest.mark.parametrize("custom_celery_task_cls", (True, False))
+def test_simple(capture_events, celery, celery_invocation, custom_celery_task_cls):
+
+    if custom_celery_task_cls:
+
+        class CustomTask(celery.Task):
+            def __call__(self, *args, **kwargs):
```
untitaker, line 109: note: try assertion here next

tests/integrations/celery/test_celery.py
```diff
@@ -307,9 +309,24 @@ def dummy_task(self):
...
     celery = init_celery(traces_sample_rate=1.0, backend="redis", debug=True)
 
+    custom_calls = []
+
+    if custom_celery_task_cls:
+        import celery as celery_mod
+
+        class CustomTask(celery_mod.Task):
+            def __call__(self, *args, **kwargs):
+                custom_calls.append(1)
```
untitaker, line 325: this is in forked process so this will always be empty. Tried with set_tag and asserted over event payload, could not repro either

PR #884, opened 2020-10-20
fix(breadcrumbs): Make all auto-generated breadcrumbs follow spec

lobsterkatie:
~Right now, if a user calls `sentry_sdk.add_breadcrumb()` on a dictionary containing non-standard keys, those keys will be serialized and sent to Sentry, but Relay will drop them without registering a processing error. In other words, it fails silently.~

~This PR adds a warning if a user includes such a call in their code. The list of standard keys is taken from the SDK spec [here](https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/).~

UPDATE: This is getting handled in relay, so scoping this down to just the below.

It also corrects two places where we were sending the key `ty` rather than `type` (which works, but will now cause a warning).


untitaker:
We should do this on the server imo, and fix the UI if it's overwhelming. This just adds overhead to the SDK for the happy path where the breadcrumb is valid.

lobsterkatie:
> We should do this on the server imo, and fix the UI if it's overwhelming. This just adds overhead to the SDK for the happy path where the breadcrumb is valid.

Happy to have this supplanted by a server fix. Will hold off on merging this until we can talk again about potentially doing that (cc @jan-auer).

jan-auer:
@lobsterkatie What is it that you would like the server to emit exactly? If I'm reading this right, it looks like normalization errors for unknown breadcrumb keys, is that correct?

lobsterkatie:
We have other places where if you send bogus data, you get one of those pink boxes at the top of the event detail page, with an error along the lines of "Invalid value for xxxxxxx: _____".

untitaker:
do we still need this if the server now does some validation?

lobsterkatie:
> do we still need this if the server now does some validation?

At this point it's just fixing the SDK to follow spec - any reason _not_ to do that?

PR #885, opened 2020-10-22
unpin eventlet and dnspython

temoto:
eventlet recursion on py2.7 fixed in 0.29.1 https://github.com/eventlet/eventlet/issues/660
dnspython was pinned in eventlet since 0.26.1


untitaker:
Thanks!

PR #892, opened 2020-10-26
Fix mypy hinting of toplevel sentry_sdk module

### LINKED ISSUE #375: Add PEP 561 support
It would be useful to be able to use the type hints defined in this library with mypy. This library needs to add support for PEP 561 for this to work.

intgr:
Mypy does not support runtime-calculated `__all__`, so duplicate symbols
from `sentry_sdk.api.__all__` to top-level `__init__.py`.

Fixes #375.

Tested with mypy 0.790.

### DIFF
--- a/sentry_sdk/__init__.py
+++ b/sentry_sdk/__init__.py
@@ -4,11 +4,10 @@ from sentry_sdk.transport import Transport, HttpTransport
 from sentry_sdk.client import Client
 
 from sentry_sdk.api import *  # noqa
-from sentry_sdk.api import __all__ as api_all
 
 from sentry_sdk.consts import VERSION  # noqa
 
-__all__ = api_all + [  # noqa
+__all__ = [  # noqa
     "Hub",
     "Scope",
     "Client",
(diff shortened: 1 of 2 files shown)


sentry_sdk/__init__.pyi
```diff
@@ -0,0 +1,12 @@
...
+Since mypy cannot deal with the runtime calculated __all__ variable in __init__.py.
+"""
+from sentry_sdk.hub import Hub, init
+from sentry_sdk.scope import Scope
+from sentry_sdk.transport import Transport, HttpTransport
+from sentry_sdk.client import Client
+
+from sentry_sdk.api import *  # noqa
+
+from sentry_sdk.consts import VERSION  # noqa
```
intgr, line 12: As pointed out in #375, instead of `__init__.pyi`, an alternative to this would be to duplicate contents of the `api_all` list in `__all__` of `__init__.py`, instead of calculating the list at runtime. I'll leave this decision up to the reviewers.

untitaker: yeah please let's do that instead

intgr:
Updated per comments.

untitaker:
Thanks!

PR #896, opened 2020-10-27
Boto3 integration

marshall-lee:
This is the integration for boto3 library for recording requests to AWS services as spans.

Another suggestion is to enable it by default in `aws_lambda` integration since boto3 package is pre-installed on every lambda.


sentry_sdk/integrations/boto3.py
```diff
@@ -0,0 +1,110 @@
...
+
+            streaming_span = span.start_child(
+                op="aws.request.stream",
+                description=span.description,
+            )
+
+            orig_read = body.read
+            orig_close = body.close
+
+            def sentry_streaming_body_read(amt=None):
```
untitaker, line 69: unless you need to interpret the arguments please use `*args, **kwargs` (in both read and close) to protect against api changes

sentry_sdk/integrations/boto3.py
```diff
@@ -0,0 +1,110 @@
...
+            def sentry_streaming_body_close():
+                # type: () -> None
+                streaming_span.finish()
+                orig_close()
+
+            body.close = sentry_streaming_body_close
+
+        def sentry_after_call_error(context, exception, **kwargs):
+            # type: (Dict[str, Any], Type[BaseException], **Any) -> None
+            span = context.get("_sentrysdk_span")  # type: Optional[Span]
```
untitaker, line 91: can we use context.pop in each function

sentry_sdk/integrations/boto3.py
```diff
@@ -0,0 +1,110 @@
...
+        # type: () -> None
+        from botocore.client import BaseClient  # type: ignore
+        from botocore.response import StreamingBody  # type: ignore
+
+        if MYPY:
+            from botocore.awsrequest import AWSRequest  # type: ignore
+
+        def sentry_request_created(service_id, request, operation_name, **kwargs):
+            # type: (str, AWSRequest, str, **Any) -> None
+            hub = Hub.current
```
untitaker, line 31: Please add a `Hub.current.get_integration` check here like in other integrations to allow the integration to be disabled at a later point in time

tox.ini
```diff
@@ -81,6 +81,8 @@ envlist =
 
     {py3.6,py3.7,py3.8}-chalice-{1.16,1.17,1.18,1.19,1.20}
 
+    {py2.7,py3.7,py3.8}-boto3-{1.16}
+
 [testenv]
 deps =
```
untitaker, line 87: I think you will have to add `boto3: boto3` (with a version constraint pls) here to make CI work

untitaker (changes requested):
hi, this looks good! I added a review but after that I think we can merge

sentry_sdk/integrations/boto3.py
```diff
@@ -0,0 +1,117 @@
...
+    span.__enter__()
+
+    # request.context is an open-ended data-structure
+    # where we can add anything useful in request life cycle.
+    request.context["_sentrysdk_span"] = span
+
+
+def _sentry_after_call(context, parsed, **kwargs):
+    # type: (Dict[str, Any], Dict[str, Any], **Any) -> None
+    span = context.pop("_sentrysdk_span")  # type: Optional[Span]
```
untitaker, line 74: Ah please use `pop("...", None)` here, this should not throw

untitaker (approved):
one minor change then we can merge

marshall-lee:
@untitaker CI is green now!

untitaker:
Thanks! I think we should add integration docs like this one: https://docs.sentry.io/platforms/python/configuration/integrations/redis/

PR #898, opened 2020-10-27
Feature/add scope to sampling context asgi integration

Lynaj:
Inside of the `traces_sampler` method, I need an access to the `path`, as well as to the `method`. These are stored inside of the `scope`.
It helps me to filter out URLs, that I don't want to include in the performance report.

### DIFF
--- a/tests/integrations/asgi/test_asgi.py
+++ b/tests/integrations/asgi/test_asgi.py
@@ -13,6 +13,7 @@ try:
 except ImportError:
     import mock  # python < 3.3
 
+
 @pytest.fixture
 def app():
     app = Starlette()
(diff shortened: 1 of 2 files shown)


untitaker:
I don't know the context of this so I will have to defer.

Lynaj:
@rhcarvalho @lobsterkatie can I somehow help with the review?

tests/integrations/asgi/test_asgi.py
```diff
@@ -202,3 +206,42 @@ def handler(*args, **kwargs):
...
+):
+    test_url = "/sync-message"
+
+    traces_sampler = mock.Mock()
+    sentry_init(
+        send_default_pii=True,
+        traces_sampler=traces_sampler
+    )
+
+    events = capture_events()
```
untitaker, line 224: can you do something with this var, such as asserting the event did/did not get sent?

lobsterkatie:
@Lynaj - sorry for the delay. This is actually part of a larger change I was working on, which I just merged: https://github.com/getsentry/sentry-python/pull/906.

Please let me know if that doesn't serve your needs.

PR #901, opened 2020-10-29
feat(dev): Add `type_name` option for `ObjectDescribedBy` test matcher

lobsterkatie:
Sometimes, it's easier to pass the name of the type you're checking against, rather than the type/class object itself, so this lets you do that. Using this option has the disadvantage of not being able to match on subclasses, so I added it rather than replacing the `type` check.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -413,8 +413,10 @@ def object_described_by_matcher():
     """
     An object which matches any other object with the given properties.
 
-    Available properties currently are "type" (a type object) and "attrs" (a
-    dictionary).
+    Available properties currently are "type" (a type object), "type_name" (a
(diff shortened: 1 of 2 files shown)


untitaker:
I would again have to see the context here before telling whether this is a good idea. I can imagine that due to lack of module name included this can lead to weak assertions. On the flipside I don't see where it would be a great inconvenience to pass the actual type.

lobsterkatie:
This came up when writing a test for the AWS integration, where I was finding the inner workings to be a little more of a black box than normal, because we can only control things from afar. That said, I think I may be able to work around it a different way, so if you think this is too gross a hack, I'm happy to try to make that work first and maybe it'll be nicer.

lobsterkatie:
I found a different workaround. Closing this.

Issue #911, opened 2020-11-04
Exeption using Django+channels ASGI


hubertokf:
When I use Django + Channels and ASGI application I got this exception.

```
Exception inside application: sentry_patched_asgi_handler() takes 3 positional arguments but 4 were given
Traceback (most recent call last):
  File "env/lib/python3.8/site-packages/channels/routing.py", line 71, in __call__
    return await application(scope, receive, send)
TypeError: sentry_patched_asgi_handler() takes 3 positional arguments but 4 were given
HTTP POST /api/v2/auth/login 500 [0.67, 127.0.0.1:37750]
```

This is my sdk init:

```
sentry_sdk.init(
    dsn="...",
    integrations=[DjangoIntegration()],
    environment=os.getenv('ENVIRONMENT', "development"),

    # If you wish to associate users to errors (assuming you are using
    # django.contrib.auth) you may enable sending PII data.
    send_default_pii=True
)
```

untitaker:
Can you tell me which versions of channels, django etc you have?

hubertokf:
Sure! This is my init log:

```
System check identified no issues (0 silenced).
November 04, 2020 - 13:55:24
Django version 2.2.11, using settings 'prendaApi.settings'
Starting ASGI/Channels version 3.0.0 development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
```

hubertokf:
I have just updated to `sentry-sdk==0.19.2` and the Exception persists.

lpomfrey:
We're experiencing the same problem, downgrading to channels < 3.0 fixes the issue. Looks like https://github.com/django/channels/commit/4fe2b81df0fa8b54df19054d78a17fd2d28da2a9 is the relevant change.

untitaker:
Thanks @lpomfrey this helps a lot!

lpomfrey:
With a little further investigation it seems that for channels 3 `patch_channels_asgi_handler_impl` from https://github.com/getsentry/sentry-python/blob/549b7df3707cb41edf88390a75132434d3ed8c01/sentry_sdk/integrations/django/asgi.py#L65 should do the same as `patch_django_asgi_handler_impl` from https://github.com/getsentry/sentry-python/blob/549b7df3707cb41edf88390a75132434d3ed8c01/sentry_sdk/integrations/django/asgi.py#L33

untitaker:
Yes exactly. Since you're investigating, do you have time to make this patch?

[...]

PR #912, opened 2020-11-05
Fix patching of AsgiHandler in Django Channels >= 3.0

lpomfrey:
Fixes https://github.com/getsentry/sentry-python/issues/911


untitaker:
only flask-dev broke, merging

untitaker:
Thank you very much!

Issue #891, opened 2020-10-26
v0.18.0+ regression with batched lambda events


stickystyle:
[This line](https://github.com/getsentry/sentry-python/blob/52830558bb535d7ff8e09b27703c99425262067f/sentry_sdk/integrations/aws_lambda.py#L106) got introduced in v0.18.0 that assumes that the `event` object will always be a `dict`, but there are cases where the lambda `event` object can be a `List[dict]`. 

`BatchInvoke` events with AppSync [1] are the ones that I'm running up against right now after trying to upgrade the sdk. Anytime I invoke the lambda with a resolver configured with `BatchInvoke`, the lambda function fails right away (and without a sentry error, which is understandable...but made finding this a real pain)

```
[ERROR] AttributeError: 'list' object has no attribute 'get'
Traceback (most recent call last):
  File "/var/task/sentry_sdk/integrations/aws_lambda.py", line 106, in sentry_handler
    headers = event.get("headers", 
{}
)
```

[1] https://docs.aws.amazon.com/appsync/latest/devguide/resolver-mapping-template-reference-lambda.html#request-mapping-template

tomazberisa:
We have also observed this unexpected behavior.

As per AWS Lambda docs [1], `event` is _usually_ `dict`, but can also be `list`, `str`, `int`, `float`, or `NoneType`, therefore the existence of `.get()` should not be assumed to exist in the line @stickystyle pointed out.

[1] https://docs.aws.amazon.com/lambda/latest/dg/python-handler.html

untitaker:
We are working on a fix.

lobsterkatie:
@stickystyle @tomazberisa - I am working on this, but don't have an AppSync set up, and am striking out in my googling today. Can you tell me:

If you are batch-invoking, and `event` is actually a list of events, which of the bits of data we use would you expect to be the same across events within the list? The entries we look for and use are `"headers"`, `"httpMethod"`, `"queryStringParameters"`, `"identity"`, `""body"`, and `"path"`.

Thanks!

tomazberisa:
FWIW, we are not batch processing, rather using AWS Lambda directly.

We've been able to replicate this issue by creating a Lambda function with the Sentry integration enabled and specifying a non-dictionary (e.g., string, number, or list) test event using the AWS Console.

Not sure what input shape from AppSync would be expected (defer to @stickystyle for that), but think there shouldn't be a hard expectation on event shape for Lambda in the general case.

[...]

PR #915, opened 2020-11-09
fix(aws): Don't crash if `event` isn't a single dict

lobsterkatie:
Per https://docs.aws.amazon.com/lambda/latest/dg/python-handler.html, the `event` argument passed to the lambda function handler can be any jsonifiable type - string, int, list, etc - rather than just the dictionary we've previously assumed it to be. (This is particularly relevant for batch requests, which come in as a list of event dictionaries.)

When faced with such an `event`, our current integration crashes, because it tries to run `.get()` on it. This fixes that, by introducing the following behavior:

- If `event` is a list, tag the transaction as a batch and with the batch size.
- If `event` is a list, take the first entry as representative for the purposes of grabbing request data.
- If `event` (or the representative) isn't a dictionary, handle it gracefully and move on without request data.

Fixes https://github.com/getsentry/sentry-python/issues/891.

### DIFF
--- a/tests/integrations/aws_lambda/test_aws.py
+++ b/tests/integrations/aws_lambda/test_aws.py
@@ -392,10 +392,6 @@ def test_performance_error(run_lambda_function):
         (b"11.21", False, 1),
         (b'"Good dog!"', False, 1),
         (b"true", False, 1),
-        (b"[12, 31, 2012]", False, 3),
-        (b"[11.21, 20.12]", False, 2),
-        (b'["good", "dog"]', False, 2),
-        (b"[true, false]", False, 2),
         (
             b"""
             [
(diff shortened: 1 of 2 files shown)


tests/integrations/aws_lambda/test_aws.py
```diff
@@ -362,3 +383,132 @@ def test_handler(event, context):
     assert envelope["contexts"]["trace"]["op"] == "serverless.function"
     assert envelope["transaction"].startswith("test_function_")
     assert envelope["transaction"] in envelope["request"]["url"]
+
+
+@pytest.mark.parametrize(
+    "aws_event, has_request_data, batch_size",
+    [
```
untitaker, line 390: Let's trim down the amount of parametrizations here unless this truly tests edgecases. Each of this will add 5 seconds to overall test time.

untitaker (approved):

tests/integrations/aws_lambda/test_aws.py
lobsterkatie: Okay, pulled ~80 sec worth of options.

PR #916, opened 2020-11-10
feat(tracing): Add sampling context from AWS and GCP

lobsterkatie:
Another follow-up to https://github.com/getsentry/sentry-python/pull/863. This adds sampling context from the AWS and GCP integrations.

In addition, this PR includes two changes to support testing of these changes - fixes to the `DictionaryContaining` and `ObjectDescribedBy` matchers, and closing of resources opened by tests to suppress unnecessary warnings.

### DIFF
--- a/tests/integrations/gcp/test_gcp.py
+++ b/tests/integrations/gcp/test_gcp.py
@@ -112,6 +112,8 @@ def run_cloud_function():
             stream = os.popen("python {}/main.py".format(tmpdir))
             stream_data = stream.read()
 
+            stream.close()
+
             for line in stream_data.splitlines():
                 print("GCP:", line)
                 if line.startswith("EVENT: "):
--- a/tests/integrations/gcp/test_gcp.py
+++ b/tests/integrations/gcp/test_gcp.py
@@ -64,6 +64,7 @@ class TestTransport(HttpTransport):
         envelope = envelope_processor(envelope)
         print("\\nENVELOPE: {}\\n".format(envelope.decode(\"utf-8\")))
 
+
 def init_sdk(timeout_warning=False, **extra_init_args):
     sentry_sdk.init(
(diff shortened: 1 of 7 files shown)


tests/integrations/gcp/test_gcp.py
```diff
@@ -265,3 +268,79 @@ def cloud_function(functionhandler, event):
...
+                except AssertionError as e:
+                    # catch the error and print it because the error itself will
+                    # get swallowed by the SDK as an "internal exception"
+                    print("\\nAssertionError: {}\\n".format(e))
+
+                return "dogs are great"
+            """
+        )
+        + FUNCTIONS_PRELUDE
+        + dedent(inspect.getsource(DictionaryContaining))
```
untitaker, line 314: can we not return the traces sampler calls as response body and do assertions on teh response body? `assert x not in` can easily pass if it shouldn't, as the log buffer has a limited size the assertionerrors can be flushed out.


```
calls = []
def traces_sampler(...):
    calls.append(...)
    return ...

init(traces_sampler=traces_sampler)


def cloud_function(...):
    return json.dumps(calls)

lobsterkatie: Yeah, I thought about the flushing problem, and wondered if maybe I should add an envelope processor which deletes all of the envelope's data, so that there'd be very little printed after the assertion error, which would (more or less) guarantee the error wouldn’t scroll off the top of the buffer.

For GCS the above should work fine (though in GCS we're running it locally, so there isn't a limited buffer). For AWS, the one tricky part is that I'm comparing to a class (`LambdaContext`) which only exists in the AWS runtime AFAICT. But your point about printing is valid. I didn't originally have these assertions in the function handler itself, but now that I do, it occurs to me I can just return the error (if it happens).

untitaker (approved):

PR #919, opened 2020-11-11
fix: Remove duplicate data from sampling context

untitaker:


lobsterkatie (changes requested):
Actually, I'd rather do this: https://github.com/getsentry/sentry-python/compare/kmclb-no-duplicate-parent-sampled-in-sampling-context - pop `parent_sampled` out of `transaction_context` and leave it in `sampling_context`, because that follows [the spec](https://develop.sentry.dev/sdk/unified-api/tracing#sampling-context).

untitaker:
Can we then remove `parent_sampled` from within the `to_json` implementation entirely? I don't think we need to send this to the server at all?

lobsterkatie:
Is it being sent to the server, though? The only place I see us calling `to_json()` (other than where I call it in `start_transaction()`) is here: https://github.com/getsentry/sentry-python/blob/cc08a6bed116e09db41c712c20ab63eb0a839e41/sentry_sdk/tracing.py#L537-L541 where it's being called on spans, _not_ including `self`, meaning it wouldn't get called on a transaction.

**EDIT:** Yeah, I checked. When I threw a `pdb` into my transport's `_send_envelope` method, here's what I got:

```
(Pdb) pp envelope.items[0].payload.json
{'_meta': {'...'},
 'contexts': {'runtime': {'...'},
              'trace': {'description': None,
                        'op': None,
                        'parent_span_id': None,
                        'span_id': '979d39004e00edcc',
                        'trace_id': '5a4fc60521604bfea8d89162ce497dc4'}},
 'environment': 'dev',
 'event_id': 'fab597c687f74d9d9b47dffc57dc237b',
 'extra': {'secret_word': None,
           'sys.argv': ['/Users/Katie/Documents/Sentry/test-apps/flask/server.py'],
           'things': 'stuff'},
 'platform': 'python',
 'request': {'...'},
 'sdk': {...'},
 'server_name': 'test-server',
 'spans': [],
 'start_timestamp': '2020-11-12T17:50:33.731546Z',
 'tags': {'dogs': 'rule', 'dogtag': 'yes'},
 'timestamp': '2020-11-12T17:50:33.731875Z',
 'transaction': 'dogpark',
 'type': 'transaction',
 'user': {'id': '122112', 'is_silly': True, 'username': 'maiseythedog'}}
```

So I think we're safe.

untitaker:
Ok right. Do you agree though that we can remove the field from the to_json return value?

lobsterkatie:
Ah, yeah, that would do it, too, at which point the entire PR becomes removing this line: https://github.com/getsentry/sentry-python/blob/c9120691c360336e7ef4767cafa14c7d7a7d8226/sentry_sdk/tracing.py#L561

Works for me.

(Looking at it now, I'm not actually sure why I felt like it needed to be there in the first place... Oh - I know. I was trying too hard to copy the JS implementation, which I now actually wish I could go back and redo more like the way I ended up doing it in Python... Anyway, yeah, let's just take out that line.)

sentry_sdk/tracing.py
```diff
@@ -609,8 +608,8 @@ def _set_initial_sampling_decision(self, sampling_context):
             if callable(options.get("traces_sampler"))
             else (
                 # default inheritance behavior
-                sampling_context["parent_sampled"]
-                if sampling_context["parent_sampled"] is not None
+                sampling_context["transaction_context"]["parent_sampled"]
+                if sampling_context["transaction_context"]["parent_sampled"] is not None
```
lobsterkatie, line 612:
```suggestion
                sampling_context["parent_sampled"]
                if sampling_context["parent_sampled"] is not None
```

It's no longer in the transaction context, so this shouldn't be changed.

lobsterkatie (changes requested):

PR #921, opened 2020-11-13
Check botocore version when activating integration

### LINKED ISSUE #920: Not working with older boto version
Hello, we use it in Django==2.1.7 app and this row breaks the app.

https://github.com/getsentry/sentry-python/blob/cc08a6bed116e09db41c712c20ab63eb0a839e41/sentry_sdk/integrations/boto3.py#L36

For versions
boto3==1.7.45
botocore==1.10.84

this throws
`
AttributeError: 'str' object has no attribute 'hyphenize'`

I'm not sure the base of the integrations but I thought they must be enabled in settings, but this part of Boto3Integration is triggered even if we have not enabled it in django settings.

marshall-lee:
Fixes #920.


untitaker:
You added a test matrix for 1.9 which appears to pass, including all assertions. I would've expected that those raise DidNotEnable. I suspect this is because pkg_resources is not actually installed.

I recommend two things:

* I'd parse the version manually with `tuple(map(str, ` like in other integrations to remove the pkg_resources dependency. At the moment a user does not have pkg_resources installed, the SDK will effectively crash unless they also have boto3 installed.
* Decide on the lowest boto3 version we want to support. It seems the tests run fine with 1.9 while we say we support 1.12+.

untitaker:
@marshall-lee can you look into this again?

marshall-lee:
@untitaker 

I changed the version check code.

> Decide on the lowest boto3 version we want to support. It seems the tests run fine with 1.9 while we say we support 1.12+.

There's some confusion with boto3 library versions.

There're two libraries — user-facing `boto3` and its "core" package `botocore`. We run tests against different `boto3` versions but the integration itself does its job on `botocore` level. Perhaps it was better to name it `BotocoreIntegration` instead of `Boto3Integration`.

So, the minimum `boto3` version on test matrix is `1.9` but it uses `botocore` version `1.12`.

untitaker:
CI randomly locks up. Probably something else.

PR #927, opened 2020-11-19
Add build scripts for building and publishing AWS Lambda layers.

marshall-lee:
These are my scripts that I wrote to package a Sentry SDK for AWS Lambda.

Lambda layers are being uploaded as zip archives, then being extracted to `/opt` directory in Lambda runtime.
And if you put your package to somewhere like `/opt/python/lib/python3.8/site-packages` then it will be available to import from lambda function code.

Having such layers that users can attach to their lambda function is nice because it's easier to integrate, especially in cases when lambda function does not have any required packages except Sentry SDK.

So i wrote a couple of scripts.

One that builds the zip bundle with a specific file structure for each Python version, e.g. `python/lib/python3.8/site-packages/sentry_sdk/`, 
`python/lib/python3.7/site-packages/sentry_sdk` and so on.

And the one that publishes a layer to a set of AWS regions. It's important to publish layer to every region the user might want because layer region and lambda function region must match.

To run the publish script, it's required to set `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables that `boto3` library understands.

For each Python version, publish-script prints a list of obtained ARNs. Having an ARN for the region, user could just paste it in the settings and the package will be available to the runtime.

So it's necessary to put the list with latest ARNs somewhere in documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,3 +73,49 @@ travis-upload-dist: dist install-zeus-cli
 	zeus upload -t "application/zip+wheel" dist/* \
 		|| [[ ! "$(TRAVIS_BRANCH)" =~ ^release/ ]]
 .PHONY: travis-upload-dist
+
+awslambda-layer-build-pre: dist
+	@$(VENV_PATH)/bin/pip install .
+.PHONY: awslambda-layer-build-pre
+
+awslambda-layer-build-py27: awslambda-layer-build-pre
+	$(VENV_PATH)/bin/python scripts/build-awslambda-layer.py --python 2.7
+.PHONY: awslambda-layer-build-py27
+
+awslambda-layer-build-py36: awslambda-layer-build-pre
+	$(VENV_PATH)/bin/python scripts/build-awslambda-layer.py --python 3.6
+.PHONY: awslambda-layer-build-py36
+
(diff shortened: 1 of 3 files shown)


untitaker:
@ajjindal I can't review this like that. Please coordinate with @HazAT about 1) how we are going to publish those layers in the future (I assume we want this in craft like anything else) 2) what our timeline is for delivering aws lambda layers

untitaker:
(rewrote a deleted comment)

@ajjindal I am not sure what the background on this is, but this doesn't integrate very well into our regular release process.

We use [craft](https://github.com/getsentry/craft) to publish new versions, which has a bunch of different features for publishing artifacts to various sources. Publishing a lambda layer clearly falls into its domain.

Also we need to figure out who owns that AWS account, where the secrets are and if we want all SDK maintainers to have access to this.

This is a good poc but in order to make this more maintainable/allow maintainers to publish new versions as easily as they can right now, we need to put more work into it. I'm not sure what exactly we need to change in Craft to make this work. Please talk to @HazAT.

ahmedetefy:
Closing this since something similar exists using Craft now to publish new versions

PR #938, opened 2020-12-01
Django 3.1 async views support

Ivo-Donchev:


sentry_sdk/integrations/django/middleware.py
```diff
@@ -128,11 +134,27 @@ def __getattr__(self, method_name):
...
             # type: (*Any, **Any) -> Any
+            if self.async_capable:
+                return self.__acall__(*args, **kwargs)
+
             f = self._call_method
             if f is None:
                 self._call_method = f = _get_wrapped_method(self._inner.__call__)
             return f(*args, **kwargs)
 
+        async def __acall__(self, *args, **kwargs):
```
untitaker, line 145: There are two problems with this approach:

* we need to support Python 2, so this will be invalid syntax
* the wrapped method will not have correct span timings (which might be fine just to fix your issue)

We usually fix this by splitting out everything with async syntax into separate files that we import conditionally. let me know if that's something you want to work on, otherwise I can take over this PR

Ivo-Donchev: @untitaker Ahh, I didn't think about python 2 support here :sweat_smile: 

Okay, I can make the files splitting and conditional import later today :slightly_smiling_face:  

> the wrapped method will not have correct span timings (which might be fine just to fix your issue)

I'm not really sure what you mean by this :thinking: Is that an issue with the error reporting ?

untitaker: the reason we wrap middlewares at all is to report middleware execution times for our [performance](https://sentry.io/for/performance/) product. if you wrap an async function with `_get_wrapped_method` you will get a wrapper that starts the timer, calls the async function but then stops the timer. Instead it should await the async function before stopping the timer.

Basically if we have async middlewares with this patch we will report 0ms timings, because the wrapped middleware immediately returns a future/coroutine.

Ivo-Donchev: Ahh, I see what you mean. 
Okay, I'll jump into the code later and will try to find some solution to this :thinking: 
Not really sure how to test it though

Ivo-Donchev: @untitaker I just make another iteration here.  
I haven't changed anything related to the span timings. I'm not sure it's actually possible (or if worth) to do this here since the code inside is supposed to run asynchronously :thinking:

Ivo-Donchev: Just pushed a quick commit to (hopefully) fix the build

Ivo-Donchev: @untitaker One more thing: I found that part of the tests are being skipped in pytest conditionaly.  (the ones that are failing). 
How can I run them? I can't find the info in https://github.com/getsentry/sentry-python/blob/master/CONTRIBUTING.md :slightly_smiling_face:

untitaker: The text buries the lede but you can control it through the makefile. `make help` is the actual entrypoint for test running.

Ivo-Donchev: image
something is not quite right with the makefile I think :(

untitaker: I see, it's broken outside of macos. I fixed the bug in master. Can you try again?

Ivo-Donchev: Still broker:/
```
./scripts/runtox.sh: 26: ./scripts/runtox.sh: Bad substitution
Makefile:36: recipe for target 'test-all' failed
make: *** [test-all] Error 2
```
I rebase to your last commit - `7dad958e`

untitaker: Ok, as a workaround: To run django tests you need to install pytest-django and django in your virtualenv manually.

Or you use `tox -e py3.7-django-2.2` as one example of many environments we test.

Ivo-Donchev: Yes, it's working now, thanks

However I'm still struggling to make it working both for WSGI and ASGI. The patch inside `integrations/django/middleware.py` that dynamically changes BaseHandler's behaviour, makes it dificult to add asgi or wsgi specific behaviour://

I'll try again when I have more time

untitaker: I wonder why you need to do WSGI specific things? Wouldn't the middleware behave the same? I mean that's the sort of abstraction Django is supposed to provide, at least.

Ivo-Donchev: @untitaker I don't need to. But every time I change something with the middleware, one of the handlers - wsgi or asgi breaks. If you look at the current state of branch, it's working fine with the ASGIHandler, but if you run `$python manage.py runserver --noasgi`, you'll see the error  "corouting object does not have attribute ..... " (or sth like that).

So I was trying to find a solution that won't introduce this regression. But the problem is that I cannot easily plug into and patch only the **ASGIHandler** since the sentry sdk is already patching the `load_middleware` which is part of the **BaseHandler**, so I cannot change the behaviour asgi-specifically. I tried different thing, but I couldn't manage to make them both working correctly (and I didn't have the time to continue debugging:/)

untitaker:
@Ivo-Donchev I am looking into this myself now and it is indeed quite a bit tricky.

Primarily because what we actually need to fix is `_get_wrapped_method`, as the context manager in there should exit until after the coroutine has finished.

I don't quite get how we could test this too, but I would like to start with this first now since there's going to be a lot of trial and error.

Does the above code work *for you* (ignoring py2), and how did you verify this?

Ivo-Donchev:
@untitaker Yes, the code above work for me :+1: 

> how did you verify this?
My team is using the sentry_sdk in a production django project. I just forked the repo and installed it locally via `pip install -e .` inside the virtual environment. Then I created a simple async view like this:
```

async def my_async_view(request):
    from django.http import HttpResponse
    import asyncio

    for i in range(3):
        print(i)
        await asyncio.sleep(1)

    return HttpResponse('asd')
```
And then I opened 2 shells and ran with `httpie` - `$ http GET http://localhost:8000/api/async-view/` and looked at the prints inside the for loop. If it's async, they look like this:
image

It's a trial and error indeed. But I think that case can hardly be caught in a single unit test :thinking: Unless we assert the "async capability" of the request at the end of the middleware change (..somehow.. :( ) . I'm not sure if I'll have enough time to finish this this week. But I'm open for discussion or chat so we fix the problem.

PS: It's now the last thing that stops us to use the async views in production. It's quite anoying that there's no easy way to "switch off" a certain middleware "per view" in django - that'd make it much more flexible :sweat_smile:

Ivo-Donchev:
Just thinking out loud: Do you think it'd be a good idea to introduce `Django3AsyncIntegration` (or sth like that)?

untitaker:
>Do you think it'd be a good idea to introduce Django3AsyncIntegration (or sth like that)?

I think middleware instrumentation is supposed to not care about sync/async at all. I wonder if I can abstract the method wrapping away in utils.

Ivo-Donchev:
> I think middleware instrumentation is supposed to not care about sync/async at all. I wonder if I can abstract the method wrapping away in utils.

I agree. But since sentry_sdk is patching the code in the middle of the http handler (which is taking care of the middlewares depending of asgi/wsgi functionality), I can't see how can this be done easily :/

Ivo-Donchev:
@untitaker About this one. Will you have time to look at it any time soon ? 
Or I will try dig into it again when I have some free time

untitaker:
I hope to get to it next week

Ivo-Donchev:
okay :slightly_smiling_face:

patrick91:
just tried this PR on a project of mine using an async view, and it seems to work properly :)

PR #945, opened 2020-12-08
fix: Fix header extraction for AWS Lambda/ApiGateway

### LINKED ISSUE #944: Lambda via ApiGateway Console has no headers
When testing a Lambda via ApiGateway (HTTP Call) everything's fine.

But when I call the same endpoint using [AWS' ApiGateway Console](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-test-method.html#how-to-test-method-console), I get this error.
Since headers are not passed in this specific context, `headers = None`.

```
{
   "errorMessage":"'NoneType' object has no attribute 'get'",
   "errorType":"AttributeError",
   "stackTrace":[
      "  File \"/var/task/sentry_sdk/integrations/aws_lambda.py\", line 138, in sentry_handler\n    transaction = Transaction.continue_from_headers(\n",
      "  File \"/var/task/sentry_sdk/tracing.py\", line 291, in continue_from_headers\n    headers.get(\"sentry-trace\"), **kwargs\n"
   ]
}
```

raphael-riel:
Fixes #944 

Since the `headers` property is present and set to `None`, the `.get` default value is ignored.
We need an explicit check on `None` for this one.

~I would advise checking other places in `sentry_sdk/integrations/aws_lambda.py` where the default value is expected to be an empty dictionary. This kind of error is frequent when working with Lambda/ApiGateway.~ Fixed.

### DIFF
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -134,7 +134,7 @@ def _wrap_handler(handler):
                     # Starting the thread to raise timeout warning exception
                     timeout_thread.start()
 
-            headers = request_data.get("headers", None)
+            headers = request_data.get("headers")
             # AWS Service may set an explicit `{headers: None}`, we can't rely on `.get()`'s default.
             if headers is None:
                 headers = {}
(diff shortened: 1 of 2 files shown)


untitaker:
Thanks!

Issue #956, opened 2020-12-23
Django performance integration suggestions


Christophe31:
Hello,

To have better view in performance with django I added myself a few extra tools in my project.

I feel like they answer common needs, here I propose them as is.

Not sure any of these tools would be of interest to you add to this lib. 

```python
#!/usr/bin/env python

import sentry_sdk

################### A template tag to see the cost of a block before thinking about caching.
from django.template import Node
from django.template.base import token_kwargs
from django.template.loader_tags import register

class SentrySpanNode(Node):
    def __init__(self, var, name, nodelist,  description, data, op):
        self.nodelist = nodelist
        self.description = description
        self.op = op
        self.data = data

    def __repr__(self):
        return '<%s>' % self.__class__.__name__

    def render(self, context):
        with sentry_sdk.start_span(
                op=self.op,
                description=self.description
        ) as span:
            span.set_data("data", self.data)
            return self.nodelist.render(context)


@register.tag('sentryspan')
def do_start_span(parser, token):
    bits = token.split_contents()
    remaining_bits = bits[1:]
    kwargs = token_kwargs(remaining_bits, parser)
    nodelist = parser.parse(('endsentryspan',))
    description = str(kwargs.get("description", "unnamed"))
    data = kwargs.get("data", {})
    op = str(kwargs.get("op", "render_node"))
    parser.delete_first_token()
    return SentrySpanNode(None, None, nodelist, description=description, data=data, op=op)
############################################################################
############## A simple decorator
from functools import wraps


def sentry_timeit(op="function", description=None):
    def decorator(method):
        @wraps(method)
        def timed(*args, **kw):
            with sentry_sdk.start_span(
                op=op,
                description=description or ".".join(
                    method.__code__.co_filename[:-3].split("/")[-1:] +
                    [method.__name__])
            ) as span:
                kwargs = kw.copy()
                for i, arg in enumerate(args):
                    kwargs[i] = arg
                span.set_tag("func_name", method.__name__)
                span.set_data("args", kwargs)
                return method(*args, **kw)
        return timed
    return decorator

###################### Monkeypatch django template processing most common entries
###### edit: This feature as been implemented more expensively in PR #957 ######
# note: monkeypatching the Template.render method seems overkill as the call often cascade.
import django.shortcuts
from django.template.response import SimpleTemplateResponse
django.shortcuts.render = sentry_timeit("render")(django.shortcuts.render)
_original = SimpleTemplateResponse.rendered_content


@property
[...]

PR #957, opened 2020-12-24
feat: Django rendering monkey patching

Christophe31:
Monkey patch `django.shortcuts.render` and `TemplateResponse` to have a django.render span in the sentry performance view.
One of the features suggested in #956 that @untitaker saw as useful and not requiring any django configuration update.


Christophe31:
Hum, the error seems to be a pip install issue. It may have impacted the coverage detection. How to re run Travis for confirmation?

(I have 2 of my 70 lines which are not tested, the use of default behavior if integration is not enabled which is took from elsewhere in the same file and not tested there either.)

Christophe31:
@untitaker happy new year, I'm allowing myself to ping you so you can tell me if this is ok or how I should improve this.
Note: it only monkey patch main django rendering entries. it should work with any django template engine.
Note 2: it only provide the auto rendering span, not the template tag as I'm not sure I can do it in the current philosophy (no settings) while django discourage to add these keywords

- here the issue where django says they don't want lib to add template tags to builtins https://code.djangoproject.com/ticket/17085 
- I see how we can monkey-patch it as loadable without being and installed django app but if we do it and a user load it from a template, this template would break if you remove sentry's Django integration at render time.

tests/integrations/django/test_basic.py
```diff
@@ -518,6 +518,22 @@ def test_does_not_capture_403(sentry_init, client, capture_events, endpoint):
...
+        integrations=[DjangoIntegration()],
+        traces_sample_rate=1.0,
+    )
+
+    for url in (reverse("template_test"), reverse("template_test2")):
+        events = capture_events()
+        _content, status, _headers = client.get(url)
+        transaction = events[0]
+        assert(
+            '- op="django.render": description="user_name.html"'
```
untitaker, line 532: I would do a strict equality check here. You can acquire the value by first doing `assert x == ''` and copying the error message

sentry_sdk/integrations/django/__init__.py
```diff
@@ -357,6 +359,38 @@ def sentry_patched_get_response(self, request):
...
+        hub = Hub.current
+        if hub.get_integration(DjangoIntegration) is None:
+            return real_render(request, template_name, context, *args, **kwargs)
+
+        with hub.start_span(op="django.render", description=template_name) as span:
+            span.set_data("context", context)
+            return real_render(request, template_name, context, *args, **kwargs)
+
+    @property
+    def rendered_content(self):
```
untitaker, line 381: Are invocations of `rendered_content` and `render` mutually exclusive? I have concerns that we will have two `django.render` spans in one transaction without a way to distinguish them.

sentry_sdk/integrations/django/__init__.py
```diff
@@ -357,6 +359,38 @@ def sentry_patched_get_response(self, request):
...
+    real_render = django.shortcuts.render
+    real_rendered_content = SimpleTemplateResponse.rendered_content
+
+    @_functools.wraps(real_render)
+    def render(request, template_name, context=None, *args, **kwargs):
+        hub = Hub.current
+        if hub.get_integration(DjangoIntegration) is None:
+            return real_render(request, template_name, context, *args, **kwargs)
+
+        with hub.start_span(op="django.render", description=template_name) as span:
```
untitaker, line 376: I think the span name is too nondescriptive, would go with django.template.render

untitaker (changes requested):

untitaker:
@Christophe31 You raise good points about the stability of the injected tag and I think I agree

I reviewed the PR

with regard to CI, we switched to GitHub actions in master. I'm going to close and reopen the PR to trigger a fresh build

sentry_sdk/integrations/django/__init__.py
Christophe31: yes they are exclusive, one render in the view span, the template response is lazy and rendered out of the view automatic span.

tests/integrations/django/test_basic.py
Christophe31: it was my first implementation but the message is a bit heavy and have not the same indent so it would require two big strings and it would break on new automatic integrations. This implementation is more permissive which can be good or bad. I'll follow your suggestion.

sentry_sdk/integrations/django/__init__.py
Christophe31: The mapping is the rendering of an HTTP response specifically, not rendering in generic context so it works with any template backend, and avoid flooding as django do recursive rendering.

I'm updating according to your suggestion.

Christophe31:
should I pull master to have the best tests configurations?

tests/integrations/django/test_basic.py
Christophe31: I did not implemented this one yet as I gave my point of view, just tell me I did not convince you and I'll comply.

Christophe31: https://github.com/getsentry/sentry-python/pull/957/commits/3403adfeffdb49888da131a53fb02dac07e4c471#diff-18f1bd8241119d62eddf36e6bf69d0f29bd8968bf2aca21d9d6238040c60189bR522

Christophe31:
@untitaker I disabled dj1.6 monkey patch on ``django.shortcut.render`` to pass the tests as I had an import error (it required loaded settings)

untitaker:
Excellent, thanks!

PR #967, opened 2021-01-08
Fix multiple **kwargs type hints

intgr:
A **kwargs argument should be hinted as `T`, instead of `Dict[str, T]`.
The dict wrapping is already implied by the type system.

See: https://mypy.readthedocs.io/en/stable/getting_started.html?highlight=kwargs#more-function-signatures


untitaker:
Thanks!

Issue #972, opened 2021-01-13
AWS Lambda integration cannot detect awslambdaric bootstrap correctly


adam-olema:
Environment:

* python 3.8.5
* awslambdarie (v1.0)
* awslambdaric (v1.0.0)
* sentry-sdk==0.19.5
* handler in a submodule, ie run with:

      aws-lambda-rie python -m awslambdaric my_module.app_module

Error:

     Not running in AWS Lambda environment, AwsLambdaIntegration disabled (could not find handle_event_request)

I have not tried this in production.

The problem appears to be in "get_lambda_bootstrap" which is choosing sys.modules["\_\_main\_\_"] when in fact, the correct module is sys.modules["\_\_main\_\_"].bootstrap

adam-olema:
Have a PR for this, but it doesn't look like the CI runs the tests at all, and thus it is failing the coverage check and I can't figure out how to get the tests to run in my local env at all.  They just skip.  Stuck on that.

untitaker:
I think this is fixed as well

PR #976, opened 2021-01-14
AWS Lambda integration fails to detect the aws-lambda-ric 1.0 bootstrap

adam-olema:
Fix for https://github.com/getsentry/sentry-python/issues/972

### DIFF
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -294,6 +294,8 @@ def get_lambda_bootstrap():
     if "bootstrap" in sys.modules:
         return sys.modules["bootstrap"]
     elif "__main__" in sys.modules:
+        if hasattr(sys.modules["__main__"], "bootstrap"):
+            return sys.modules["__main__"].bootstrap  # type: ignore
         return sys.modules["__main__"]
     else:
         return None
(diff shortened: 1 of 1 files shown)


adam-olema:
Don't know how to get the aws lambda tests to actually run.  Until then, the coverage test will not pass.

untitaker:
@adam-olema the AWS lambda tests run against a real instance and as such require credentials that are not available for PRs of forks.

Can you add a comment indicating why this is necessary? A hint at aws-lambda-ric suffices.

adam-olema:
Under normal documented usage, with the current aws-lambda-ric 1.0, without this change, you only get the following warning, and no errors sent to sentry:

    Not running in AWS Lambda environment, AwsLambdaIntegration disabled (could not find handle_event_request)

So, not sure what changed, but in my setup at least, the sentry library does not detect the AWS lambda runtime at all.  It may have to do with the structure of my application which places the main handler not at the top level module but in a submodule (ie: not the usual my_app.handler, but instead my_app.sub_module.handler

untitaker:
@adam-olema I understand, I am asking that you put this as code comment, to be clear

adam-olema:
> @adam-olema I understand, I am asking that you put this as code comment, to be clear

Gotcha.  Comment added.

untitaker:
Thanks!

PR #980, opened 2021-01-15
fix(environment): Remove release condition on default

lobsterkatie:
The SDK spec ([here](https://develop.sentry.dev/sdk/overview/)) doesn't list a release condition on the default environment value.

Specifically:

> The following items are expected of production-ready SDKs:
> [...]
> Additionally, the following features are highly encouraged:
> - [...]
> - Send an `environment` on each event. If none was detected or set by the user, `production` should be used.


rhcarvalho (approved):
@mitsuhiko was there any special reason to check the release in the Python implementation?

PR #991, opened 2021-01-29
ref(tracing): Restore ability to have tracing disabled

lobsterkatie:
This PR partially reverts https://github.com/getsentry/sentry-python/pull/948 and https://github.com/getsentry/sentry-python/commit/6fc2287c6f5280e5adf76bb7a66f05f7c8d18882, to restore the ability to disable tracing, which allows it to truly be opt-in [as per the spec](https://develop.sentry.dev/sdk/performance/#sdk-configuration).

Note that this does not change the behavior that PR was made to reinstate - the model wherein the front end makes sampling decisions, the backend has `traces_sample_rate` set to `0`, and the result is that the backend samples according to the front end decision when there is one, but otherwise does not send transactions. (That PR modified a test to test for that exact scenario, and it still passes.)

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -72,7 +72,7 @@ class ClientConstructor(object):
         attach_stacktrace=False,  # type: bool
         ca_certs=None,  # type: Optional[str]
         propagate_traces=True,  # type: bool
-        traces_sample_rate=0.0,  # type: float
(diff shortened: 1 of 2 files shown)


ahmedetefy (approved):
Looks good to me!

RaduW (approved):

PR #992, opened 2021-02-02
fix(django) - Fix Django async views not behaving asyncronuously 

### LINKED ISSUE #786: Django 3.1 async middleware not properly instrumented as spans in Performance
The new async middlewares in Django 3.1 are not properly instrumented, the spans have zero-sized durations, if not worse.

(please vote with 👍 on this issue if you need this, or escalate through support for bumping priority)

ahmedetefy:
fixes #786 
fixes #938

### DIFF
--- a/sentry_sdk/integrations/django/asgi.py
+++ b/sentry_sdk/integrations/django/asgi.py
@@ -105,6 +105,7 @@ def _asgi_middleware_mixin_factory(_check_middleware_span):
     Mixin class factory that generates a middleware mixin for handling requests
     in async mode.
     """
+
     class SentryASGIMixin:
         def __init__(self, get_response):
             # type: (F) -> None
--- a/sentry_sdk/integrations/django/asgi.py
+++ b/sentry_sdk/integrations/django/asgi.py
@@ -101,6 +101,7 @@ def _asgi_mixin_factory(_check_middleware_span):
     Class factory that generates a class that deals with handling requests
     in async mode.
     """
+
     class SentryASGIMixin:
         def __init__(self, get_response):
             # type: (F) -> None
--- a/sentry_sdk/integrations/django/asgi.py
+++ b/sentry_sdk/integrations/django/asgi.py
@@ -102,7 +102,6 @@ def _asgi_mixin_factory(_check_middleware_span):
     in async mode.
     """
     class SentryASGIMixin:
-
         def __init__(self, get_response):
             # type: (F) -> None
             self.get_response = get_response
--- a/sentry_sdk/integrations/django/asgi.py
+++ b/sentry_sdk/integrations/django/asgi.py
@@ -132,4 +132,5 @@ def _wrap_asgi_code(middleware, _check_middleware_span):
 
             with middleware_span:
                 return await f(*args, **kwargs)
+
     return SentryASGIMixin
--- a/sentry_sdk/integrations/django/asgi.py
+++ b/sentry_sdk/integrations/django/asgi.py
@@ -16,8 +16,12 @@ from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 if MYPY:
     from typing import Any
     from typing import Union
+    from typing import TypeVar
+    from typing import Callable
 
     from django.http.response import HttpResponse
+
+    F = TypeVar("F", bound=Callable[..., Any])
 
 
 def patch_django_asgi_handler_impl(cls):
@@ -120,7 +124,7 @@ def _asgi_mixin_factory(_check_middleware_span):
                 self._is_coroutine = asyncio.coroutines._is_coroutine
 
         def async_route_check(self):
-            # type: () -> Boolean
+            # type: () -> bool
             """
             Function that checks if we are in async mode,
             and if we are forwards the handling of requests to __acall__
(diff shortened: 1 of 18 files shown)


sentry_sdk/integrations/django/asgi.py
```diff
@@ -91,3 +97,52 @@ async def sentry_wrapped_callback(request, *args, **kwargs):
...
 
     return sentry_wrapped_callback
+
+
+def _asgi_mixin_factory(_check_middleware_span):
+    # type: (F) -> Any
+    """
+    Class factory that generates a class that deals with handling requests
+    in async mode.
+    """
```
untitaker, line 107: Please clarify in both method name and docstring that this is a mixin class for middlewares specifically

sentry_sdk/integrations/django/middleware.py
```diff
@@ -102,11 +121,20 @@ def sentry_wrapped_method(*args, **kwargs):
...
-        def __init__(self, *args, **kwargs):
-            # type: (*Any, **Any) -> None
-            self._inner = middleware(*args, **kwargs)
+    class SentryWrappingMiddleware(_asgi_mixin_factory(_check_middleware_span)):
+
+        async_capable = getattr(middleware, "async_capable", False)
+
+        def __init__(self, get_response=None, *args, **kwargs):
+            # type: (F, *Any, **Any) -> None
+            if get_response:
```
untitaker, line 130: When would this be `None`?

tests/integrations/django/asgi/test_asgi.py
```diff
@@ -5,11 +5,19 @@
...
+    from django.core.urlresolvers import reverse
+
 APPS = [channels_application]
+ASGI_APP = []
+
 if django.VERSION >= (3, 0):
     from tests.integrations.django.myapp.asgi import asgi_application
 
-    APPS += [asgi_application]
+    ASGI_APP = [asgi_application]
```
untitaker, line 19: You don't need this. Instead of using pytest.mark.parametrize you can just refer to `asgi_application` directly from within the test (instead of `application`). The test will only run on 3.1 anyway.

untitaker (changes requested):
Good stuff, some minor comments but I think this is good to go.

tests/integrations/django/asgi/test_asgi.py
```diff
@@ -68,3 +76,83 @@ async def test_async_views(sentry_init, capture_events, application):
...
+    (resp1, resp2), _ = await asyncio.wait(
+        [comm.get_response(timeout=5), comm2.get_response(timeout=5)]
+    )
+
+    end = time.time()
+
+    assert resp1.result()["status"] == 200
+    assert resp2.result()["status"] == 200
+
+    assert end - start < 1.5
```
untitaker, line 112: Appears to still fail for channels, not sure why.

sentry_sdk/integrations/django/middleware.py
ahmedetefy: This is for backwards compatibility. Older versions of Django only, should be deprecated in 4.0

Ref from django.utils.deprecation:
```python
# RemovedInDjango40Warning: when the deprecation ends, replace with:
    #   def __init__(self, get_response):
    def __init__(self, get_response=None):
```

tests/integrations/django/asgi/test_asgi.py
ahmedetefy: Where do you mean? This last CI run shows that the tests are failing for Python 3.6, 3.7, 3.8 for django 3.1 and dev, am I missing something?

untitaker: Right but within those environments only hte channels variant is failing afaict.

sentry_sdk/integrations/django/asgi.py
```diff
@@ -91,3 +97,52 @@ async def sentry_wrapped_callback(request, *args, **kwargs):
             return await callback(request, *args, **kwargs)
 
     return sentry_wrapped_callback
+
+
+def _asgi_middleware_mixin_factory(_check_middleware_span):
+    # type: (F) -> Any
```
untitaker, line 103:
```suggestion
    # type: (Callable[..., Any]) -> Any
```

We use typevar if we want to make the return type dependent on the input type

sentry_sdk/integrations/django/asgi.py
```diff
@@ -91,3 +97,52 @@ async def sentry_wrapped_callback(request, *args, **kwargs):
...
+    """
+    Mixin class factory that generates a middleware mixin for handling requests
+    in async mode.
+    """
+
+    class SentryASGIMixin:
+        def __init__(self, get_response):
+            # type: (F) -> None
```
untitaker, line 111:
```suggestion
            # type: (Callable[..., Any]) -> None
```

sentry_sdk/integrations/django/middleware.py
```diff
@@ -64,29 +72,40 @@ def _wrap_middleware(middleware, middleware_name):
     # type: (Any, str) -> Any
     from sentry_sdk.integrations.django import DjangoIntegration
 
+    def _check_middleware_span(old_method):
+        # type: (F) -> Span
```
untitaker, line 76:
```suggestion
        # type: (Callable[..., Any]) -> Optional[Span]
```

tests/integrations/django/asgi/test_asgi.py
untitaker: decided to skip test for channels

untitaker (approved):

untitaker (changes requested):
wrong button

untitaker (approved):

PR #993, opened 2021-02-04
release: 1.0.0

ahmedetefy:


CHANGELOG.md
```diff
@@ -1,24 +1,26 @@
...
+- Fix that corrects hypothesis tests #978
+- Fix that corrects AWS lambda integration failure to detect the aws-lambda-ric 1.0 bootstrap #976
+- Fix unbound local crash on handling aws lambda exception - Timeout #977
+- Fix that removes release condition on default #980
+- Added Make pytest run instruction clear in contribution guide #981
+- Fix headers not parsed correctly in ASGI middleware, Decode headers before creating transaction #984
+- CI Fix `py3.5-celery` and `*-django-dev`
+- Ref: Restore ability to have tracing disabled #991
+- Fix Django async views not behaving asyncronuously
+- Feat: Supported pre-aggregated sessions
```
HazAT, line 23:
```suggestion
We decided to use [semver](https://semver.org/) versioning from now on for the SDK. The bump to 1.0 doesn't contain any breaking changes.

- Added power support(ppc64le) with ci and testing to the project for architecture-independent
- Fix for header extraction for AWS lambda/API extraction
- Fix that unpins pytest, stop testing eventlet #965
- Remove Travis VI #962
- Use full git sha as release name #960
- Fix multiple **kwargs type hints
- Added feature Django rendering monkey patching #957
- Fix that corrects hypothesis tests #978
- Fix that corrects AWS lambda integration failure to detect the aws-lambda-ric 1.0 bootstrap #976
- Fix unbound local crash on handling aws lambda exception - Timeout #977
- Fix that removes release condition on default #980
- Added Make pytest run instruction clear in contribution guide #981
- Fix headers not parsed correctly in ASGI middleware, Decode headers before creating transaction #984
- CI Fix `py3.5-celery` and `*-django-dev`
- Ref: Restore ability to have tracing disabled #991
- Fix Django async views not behaving asynchronously
- Feat: Supported pre-aggregated sessions
```

docs/conf.py
```diff
@@ -22,7 +22,7 @@
 copyright = u"2019, Sentry Team and Contributors"
 author = u"Sentry Team and Contributors"
 
-release = "0.19.5"
+release = "1.0.0"
```
HazAT, line 25:
```suggestion
release = "0.19.5"
```
The prepare script does this for us

sentry_sdk/consts.py
```diff
@@ -99,7 +99,7 @@ def _get_default_options():
 del _get_default_options
 
 
-VERSION = "0.19.5"
+VERSION = "1.0.0"
```
HazAT, line 102:
```suggestion
VERSION = "0.19.5"
```

setup.py
```diff
@@ -21,7 +21,7 @@ def get_file_text(file_name):
 
 setup(
     name="sentry-sdk",
-    version="0.19.5",
+    version="1.0.0",
```
HazAT, line 24:
```suggestion
    version="0.19.5",
```

HazAT (changes requested):

CHANGELOG.md
```diff
@@ -1,24 +1,26 @@
...
-
-A major release `N` implies the previous release `N-1` will no longer receive updates. We generally do not backport bugfixes to older versions unless they are security relevant. However, feel free to ask for backports of specific commits on the bugtracker.
+## 1.0.0
+
+- This product follows [semver](https://semver.org/) from now on
+- No breaking changes, SDK is stable
+- Added power support(ppc64le) with ci and testing to the project for architecture independent
+- Fix for header extraction for AWS lambda/API extraction
+- Fix that unpins pytest, stop testing eventlet #965
+- Remove Travis VI #962
```
rhcarvalho, line 10:
```suggestion
- Remove Travis CI #962
```

CHANGELOG.md
lobsterkatie: > Added power support(ppc64le) with ci and testing to the project for architecture-independent

If we're going to make it an adjective (by adding the dash), then we need something for that adjective to modify. Otherwise, I think the fix here is "architecture independence."

CHANGELOG.md
```diff
@@ -1,24 +1,26 @@
...
+- Fix for header extraction for AWS lambda/API extraction
+- Fix that unpins pytest, stop testing eventlet #965
+- Remove Travis CI #962
+- Use full git sha as release name #960
+- Fix multiple **kwargs type hints
+- Added feature Django rendering monkey patching #957
+- Fix that corrects hypothesis tests #978
+- Fix that corrects AWS lambda integration failure to detect the aws-lambda-ric 1.0 bootstrap #976
+- Fix unbound local crash on handling aws lambda exception - Timeout #977
+- Fix that removes release condition on default #980
```
untitaker, line 17: This is not describing anything, I would have to check the PR to see what's going on

EDIT: I would say "the default environment is now production, not based on release"

**BTW this *is* a breaking change.**

CHANGELOG.md
```diff
@@ -1,24 +1,26 @@
...
+
+We decided to use [semver](https://semver.org/) versioning from now on for the SDK. The bump to 1.0 doesn't contain any breaking changes.
+
+- Added power support(ppc64le) with ci and testing to the project for architecture independence
+- Fix for header extraction for AWS lambda/API extraction
+- Fix that unpins pytest, stop testing eventlet #965
+- Remove Travis CI #962
+- Use full git sha as release name #960
+- Fix multiple **kwargs type hints
+- Added feature Django rendering monkey patching #957
```
untitaker, line 13: I would specifically say that this is about template rendering. Also "monkey patching" is not the feature. Suggestion:

>Django integration now creates transaction spans for template rendering.

untitaker (changes requested):
I have two main requests here:

* Remove all mentions of CI changes, they are not relevant to the user and are not even contained in the package one can download from PyPI. That especially goes for ppc64le support as there were zero changes required in SDK code to make this fix.
* Go into more detail for each line as to what it fixes. What is the user-visible behavior before and after the fix? Pre-aggregated sessions can probably be omitted or mentioned as performance improvement.

I have left some *example* suggestions on the PR, but I think more cases exist where it's not clear to the user what the change in behavior is.

I admit that the PR titles could be better. With more people working on the Python SDK we will have to rethink how we do changelogs.

untitaker:
I forgot about this but I'd like to keep the clause that says that experimental integrations are exempt from versioning -- for example: https://docs.sentry.io/platforms/python/gnu_backtrace/

CHANGELOG.md
ahmedetefy: yeah, you are right so should I remove the "no breaking changes " part in the then?

ahmedetefy: and perhaps put a reference to this breaking change, like explicitly state that this a breaking change

CHANGELOG.md
```diff
@@ -1,24 +1,27 @@
...
-```
-sentry-sdk>=0.10.0,<0.11.0
-sentry-sdk==0.10.1
-```
+## 1.0.0
+
+The bump to 1.0 contains a breaking change.
 
-A major release `N` implies the previous release `N-1` will no longer receive updates. We generally do not backport bugfixes to older versions unless they are security relevant. However, feel free to ask for backports of specific commits on the bugtracker.
+- Fix for header extraction for AWS lambda/API extractio
```
HazAT, line 14: Is this the breaking change?

Also, is this a typo `extractio`?

ahmedetefy: Yep it's a typo should be extraction*

And nope this is not the breaking change but you are right its not clear which is the breaking change.
So I changed it and added a prefix to breaking changes

So I am just stating underneath release 1.0.0 
That this bump  to 1.0 will have a breaking change
and I prefixed the breaking change with the "BREAKING CHANGE" prefix

Is it not clear?

HazAT (approved):

CHANGELOG.md
HazAT: 👍

HazAT (changes requested):
Wait, cc @mitsuhiko What about we frame 1.0 with Session on by default?

PR #1018, opened 2021-02-12
fix(release): excluded sentry python serverless zip package from pypi

ahmedetefy:


untitaker (approved):
this intuitively makes sense to me but I haven't verified whether this option exists. I added anton for review since he wrote the pypi target for craft

iker-barriocanal (approved):

PR #1024, opened 2021-02-16
feat(rq): Add enqueued_at and started_at to rq job extra

kruvasyan:
Fixes: #144

Add enqueued_at and started_at to extra["rq-job"] on RQ integration

The test was a bit tricky because the job was missing the started_at attribute after the worker finished. I added information about this to the commit with test.

Also another question. Are we satisfied that the time in the fields will be in the format '2021-02-16 18:55:17.220349', or do we need some special format?


vasiliykovalev:
Hm, the same tests fail in the master. I guess I have to wait for the green master

Rada8732 (approved):

sl0thentr0py (approved):
very late but thx for adding this @kruvasyan !

PR #1034, opened 2021-02-26
bug(flask): Transactions missing body

ahmedetefy:
This PR fixes the bug that transactions were missing data/body because weakref to the request object was garbage collected before the request object made into the event_processor
- Removed weakref to request object when passed to the event_processor_factory

Ref: https://app.asana.com/0/1142928854421177/1199663160066590

### DIFF
--- a/sentry_sdk/integrations/flask.py
+++ b/sentry_sdk/integrations/flask.py
@@ -111,9 +111,7 @@ def _request_started(sender, **kwargs):
         except Exception:
             pass
 
-        evt_processor = _make_request_event_processor(
-            app, request, integration  # type: ignore
-        )
+        evt_processor = _make_request_event_processor(app, request, integration)
         scope.add_event_processor(evt_processor)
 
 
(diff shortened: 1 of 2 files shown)


untitaker (commented):
please verify manually that the test fails if the fix is not applied

ahmedetefy:
> please verify manually that the test fails if the fix is not applied

Already did .. fails without the code changes

ahmedetefy:
There is no particular need for the `request` to be passed on to the event processor factory 
https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/flask.py#L117-L119

A discussed solution would be to just create an instance of `FlaskRequestExtractor` when the request is started but that poses two problems 
1- It adds overhead for users that do not have tracing enabled 
Possible solution: Have a different code path if you do not have tracing enabled
2- (More serious) Calling `request.data` eagerly loads to memory before the user does in his code is not ideal

@untitaker is that everything we discussed or would you like to add something?

untitaker:
nah sounds good to me

PR #1035, opened 2021-02-26
fix(tracing): Get HTTP headers from span rather than transaction if possible

lobsterkatie:
Currently, there are situations in which a trace ends up looking not like this (as it should):

image

but instead looks like this:

image

This happens because we currently create the outgoing tracing headers based not on the span representing the outgoing request but on whatever span is on the scope, which is most often the transaction itself. There are situations in which this is necessary (if there is no child span to tie to*), but in situations where it's not, we should be correctly associating child transactions with their parent spans, not their ancestor transactions.

\*It's a separate question whether, in the one place there isn't a span ([here](https://github.com/getsentry/sentry-python/blob/1279eeca6763e119d97da5da8318f48a04d3adef/sentry_sdk/integrations/rq.py#L99-L107), the `rq` integration), there should be.

### DIFF
--- a/sentry_sdk/hub.py
+++ b/sentry_sdk/hub.py
@@ -682,14 +682,14 @@ class Hub(with_metaclass(HubMeta)):  # type: ignore
         if client is not None:
             return client.flush(timeout=timeout, callback=callback)
 
-    def iter_trace_propagation_headers(self):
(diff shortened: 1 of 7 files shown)


untitaker (approved):
Please add a docstring to `iter_trace_propagation_headers` explaining in which situations one would pass in a span object explicitly.

PR #1037, opened 2021-03-01
fix(django):  Added SDK logic that honors the `X-Forwarded-For` header

ahmedetefy:
This PR, 
- Changes the `SentryWsgiMiddleware` to accept a `use_x_forwarded_for` arg that defaults to False. It acts as a flag that if enabled  allows for checking if the `HTTP_X_FORWARDED_HOST` and if it is present in the `environ`, it is used to generate the request URL rather than using the `HTTP_HOST` for it
- Modified the Django integration's initialization of `SentryWsgiMiddleware` to pass the value of `settings.USE_X_FORWARDED_HOST`

### DIFF
--- a/tests/integrations/django/test_basic.py
+++ b/tests/integrations/django/test_basic.py
@@ -40,6 +40,49 @@ def test_view_exceptions(sentry_init, client, capture_exceptions, capture_events
     assert event["exception"]["values"][0]["mechanism"]["type"] == "django"
 
 
+def test_ensures_x_forwarded_header_is_honored_in_sdk_when_enabled_in_django(
+        sentry_init, client, capture_exceptions, capture_events):
+    """
+    Test that ensures if django settings.USE_X_FORWARDED_HOST is set to True
+    then the SDK sets the request url to the `HTTP_X_FORWARDED_FOR`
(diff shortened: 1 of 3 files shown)


tests/integrations/django/test_basic.py
```diff
@@ -40,6 +40,50 @@ def test_view_exceptions(sentry_init, client, capture_exceptions, capture_events
...
+def test_ensures_x_forwarded_header_is_honored_in_sdk_when_enabled_in_django(
+    sentry_init, client, capture_exceptions, capture_events
+):
+    """
+    Test that ensures if django settings.USE_X_FORWARDED_HOST is set to True
+    then the SDK sets the request url to the `HTTP_X_FORWARDED_FOR`
+    """
+    from django.conf import settings
+
+    settings.USE_X_FORWARDED_HOST = True
```
untitaker, line 52: Instead of importing settings I suggest using the settings pytest fixture, then your change is automatically reverted at the end of the test

```
def test_foo(settings):
    settings.USE_X_FORWARDED_HOST = True
    ...
```

untitaker (approved):

PR #1040, opened 2021-03-04
Get rid of setup.cfg by moving the only option to setup.py

hynek:
🎉

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,0 @@
-[bdist_wheel]
-universal = 1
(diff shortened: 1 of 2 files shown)


untitaker (approved):
Thanks, I didn't know you can do that!

hynek:
Python packaging is a land full of mysteries. ;)

I've been told it once by the maintainers in connection with my PyPI blog post.

untitaker:
Do you know by chance when this feature was introduced? I mean, seems to run fine on last pip for 2.7, but we have people running on 2.6.

hynek:
I don't think this is a pip feature but a setuptools/wheel one.

Assuming it's a recent-ish features (which I don't think), it would only affect people who insist on building their sdists with ancient build chains into wheels.

untitaker:
Fair enough.

PR #1050, opened 2021-03-10
support SOCKS proxies in sentry_sdk

### LINKED ISSUE #1049: Support SOCKS proxies
If pysocks is available, urllib3 supports SOCKS proxies, but you can't use the `urllib3.ProxyManager` class and must instead use `urllib3.contrib.socks.SOCKSProxyManager`. It would be very nice if `sentry_sdk.transport.HttpTransport` could detect a proxy set to `socks://` (or one of the many variants) and Do The Right Thing.

Roguelazer:
Fixes #1049

Not sure that this is the cleanest solution, but it sure seems to work.


kenoss:
This PR is very useful.  I'm not a maintainer of this repo but I'll note some points I found.

1. About `PySocks` dependency.

[`urllib3` does not install `PySocks` by default.](https://github.com/urllib3/urllib3/blob/1.26.5/setup.cfg#L21)
In other words, after this PR, the users need to write dependencies both `sentry_sdk` and `PySocks` explicitly and separately, and `urllib3` can't control what version of `PySocks` is installed.
[`setry_sdk` uses `extra_require` already](https://github.com/getsentry/sentry-python/blob/1.1.0/setup.py#L41), and (AFAIK) it is better to do it.
(But, a little tricky. `sentry_sdk[socks]` should trigger `urllib3[socks]` rather than `PySocks`.)
Users will be able to install by `pip install "sentry_sdk[socks]"`.

2. About test.

This PR would be more great if it has tests, like https://github.com/getsentry/sentry-python/pull/838/files I guess.

Roguelazer:
reset the clock

antonpirker (approved):

sentrivana (approved):

PR #1054, opened 2021-03-12
fix(django): Deal with template_name being a list

### LINKED ISSUE #1048: Django template span has invalid description
### Environment

How do you use Sentry?
Self-hosted, version 21.2.0

Which SDK and version?
Python 1.0.0

### Steps to Reproduce

1. Set up performance tracing for a Django application
2. Send performance data for a view which is a subclass of `TemplateView`
3. View the performance information for the view in Sentry

### Expected Result

Sentry accepts all of the submitted performance data.

### Actual Result

The following alert appears above the performance information:

Sentry Error

This is caused by #957.

At https://github.com/Christophe31/sentry-python/blob/db2d87e503b0178a7d56bdedb00db19ac6ddc9d8/sentry_sdk/integrations/django/templates.py#L60, a span is created with the description `self.template_name`. If the `TemplateResponse` was created by a Django `TemplateView`, `template_name` will be a list, not a string.

A similar error can occur with `django.shortcuts.render`, but I don't use it, so I haven't experienced the issue there.

Christophe31:
Here my pull request for issue #1048 .

Fix #1048


Christophe31:
Indeed, I saw this in tests results, it's fixed and tested now. thx for the report.

untitaker:
Thanks!

PR #1060, opened 2021-03-19
feat: Support tracing on Tornado

### LINKED ISSUE #1057: Add performance monitoring to Tornado integration
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) 

Which SDK and version?
sentry-sdk==1.0.0

### Steps to Reproduce

Create a python Tornado/Celery/Redis application and integrate sentry using:

    sentry_sdk.init(
        "https://xxxxxxxxxxxxxxxxx",
        integrations=[TornadoIntegration(), CeleryIntegration(), RedisIntegration()],
        traces_sample_rate=1.0
    )

### Expected Result

Capture results and monitor performance.

### Actual Result

Sentry is only capturing errors but the performance monitoring didn't work.

untitaker:
Fix #1057

### DIFF
--- a/sentry_sdk/integrations/tornado.py
+++ b/sentry_sdk/integrations/tornado.py
@@ -1,7 +1,9 @@
 import weakref
+import contextlib
 from inspect import iscoroutinefunction
 
 from sentry_sdk.hub import Hub, _should_send_default_pii
+from sentry_sdk.tracing import Transaction
 from sentry_sdk.utils import (
     HAS_REAL_CONTEXTVARS,
     CONTEXTVARS_ERROR_MESSAGE,
@@ -32,6 +34,7 @@ if MYPY:
     from typing import Optional
     from typing import Dict
     from typing import Callable
+    from typing import Generator
 
     from sentry_sdk._types import EventProcessor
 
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/tornado.py
```diff
@@ -110,6 +91,39 @@ def sentry_log_exception(self, ty, value, tb, *args, **kwargs):
...
+
+    if integration is None:
+        yield
+
+    weak_handler = weakref.ref(self)
+
+    with Hub(hub) as hub:
+        with hub.configure_scope() as scope:
+            scope.clear_breadcrumbs()
+            processor = _make_event_processor(weak_handler)  # type: ignore
```
ahmedetefy, line 108: Does this not have the same issue as #1034 , the `weak_handler` is not accessible within the `tornado_processor` https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/tornado.py#L135 ?

untitaker: Good catch, but I believe `self` will stick around for the duration of the request handler invocation, which sticks around for the duration of the response streaming, so it should be fine afaict

let me add a test

ahmedetefy: I don't think the test you added covers that case since that request raises an exception and so triggers the error handler.
I would create another test to check if the request body exists when the request doesn't raise an exception and returns a success status code.

untitaker: Conceptually the exception won't make the request stay around for longer, but you're right it's not the same testcase so I added another one.

ahmedetefy (approved):

lobsterkatie (approved):
LGTM. Can you please also add a tornado version of this?

https://github.com/getsentry/sentry-docs/blob/master/src/includes/performance/default-sampling-context/python.rq.mdx

untitaker:
Yup will do, thanks.

PR #1066, opened 2021-03-28
Fix (Sanic 21.3): TypeError: sentry_router_get() takes 2 positional arguments but 4 were given

SaidBySolo:
Fixed an issue with processing requests in 21.3.

### DIFF
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -99,7 +99,7 @@ class SanicIntegration(Integration):
         def sentry_router_get(self, request, *args):
             # type: (Any, Request, Any) -> Any
             if version >= (21, 3):
-                rv = old_router_get(self, request, args[0], args[1])
+                rv = old_router_get(self, request, *args)
             else:
                 rv = old_router_get(self, request)
             hub = Hub.current
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -98,7 +98,7 @@ class SanicIntegration(Integration):
 
         def sentry_router_get(self, request, *args):
             # type: (Any, Request, Any) -> Any
-            if version < (21, 3):
+            if version >= (21, 3):
                 rv = old_router_get(self, request, args[0], args[1])
             else:
                 rv = old_router_get(self, request)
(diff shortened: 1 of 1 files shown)


cakemanny (commented):
In 21.3 and later the first non-self argument is no longer a `Request`, it's the path, so this looks a bit confusing.
Since `request` is not being used, is there a reason to not just write it as follows?
```python
def sentry_router_get(self, *args):
    # type: (Router, *Any) -> Any
    rv = old_router_get(self, *args)
    hub = Hub.current
    ...
```

SaidBySolo:
> In 21.3 and later the first non-self argument is no longer a `Request`, it's the path, so this looks a bit confusing.
> Since `request` is not being used, is there a reason to not just write it as follows?
> 
> ```python
> def sentry_router_get(self, *args):
>     # type: (Router, *Any) -> Any
>     rv = old_router_get(self, *args)
>     hub = Hub.current
>     ...
> ```
it was burdensome to modify the code that worked fine.

cakemanny:
Hey @untitaker or @ahmedetefy, what's needed for this PR to make it in? Hopefully you are the right people to ask.
We are currently stuck on an ageing version of Sanic and can't upgrade without this

HazAT:
We will handle it this week

ahmedetefy:
Hey @cakemanny did you manage to run the tests for `sanic > 21.0`?
I have tried running tests for version 21.3 and they seem to be broken

cakemanny:
Hey @ahmedetefy, well, I have now 😅 . Looks like they broke when going from 20.9.x to 20.12.x :(

This is what you were seeing, yes?
```
_____________________________________________________________________ ERROR at setup of test_errors ______________________________________________________________________
tests/integrations/sanic/test_sanic.py:19: in app
    app = Sanic(__name__)
venv/lib/python3.7/site-packages/sanic/app.py:97: in __init__
    self.__class__.register_app(self)
venv/lib/python3.7/site-packages/sanic/app.py:1444: in register_app
    raise SanicException(f'Sanic app name "{name}" already in use.')
E   sanic.exceptions.SanicException: Sanic app name "tests.integrations.sanic.test_sanic" already in use.
...
```

Basically it looks like they've added a check to stop people accidentally re-declaring the same app.
I'll try to have a look into this but I'm shortly about to go on holiday for a week and a half. So if anyone else reading this wants to look in the meantime, feel free.

ahmedetefy:
No need, I have already picked this up.. will push a PR soon

PR #1068, opened 2021-03-30
Minor tweak of recommended version to pin in changelog

hoylemd:
Since we're on major version 1, I think the docs should recommend that version.


untitaker:
Thanks!

Rada8732 (approved):

Rada8732:
Merge complete

Issue #1074, opened 2021-04-08
Suppress errors while RQ is retrying


BobReid:
I am using RQ as my job queue. I have some use cases where retries and optimistic locking are used to smooth out contentious writes to the database and out of order execution. 

This results in a lot of false positives in Sentry. These issues fix themselves after retry so I do not want them reported to sentry until all the retries are exhausted. Currently there isn't a way to suppress the exception due to the way the sentry integration is monkey patching RQ. You need to inspect the RQ job instance status in order to determine whether or not retries have been exhausted. 

I am happy to contribute the change if it has a chance at being accepted. 

I see three possible solutions:

1. Add a settings to the RQ integration along the lines of failed_jobs_only. This setting would default to false to maintain backwards compatibility. 
2. Provide a hook mechanism that passes the job in and let users customize whether or not to capture the event
3. Stuff some job information like the status into the hint so it can be inspected and filtered via the standard before_send hook. 

IMO 1 or 3 is the way to go. 1 is the simplest and 3 is more flexible. I am not sure if stuffing random data into hint like this is desirable from Sentry's standpoint.

untitaker:
celery integration does not report those kinds of exceptions so for sake of consistency I also wouldn't do that for RQ (regardless of whether the behavior is controllable by an option)

BobReid:
@untitaker 

I have thrown together a quick PR. It can be as simple as checking if the job has failed. Let me know your thoughts and I can write a test for it. 

https://github.com/getsentry/sentry-python/pull/1076

BobReid:
I have added a test to my PR. I had to add an ignore_logger call to the RQ integration or else the exception was still captured by the logging integration.

BobReid:
@untitaker any thoughts on the resolution I have proposed. 

This is causing unnecessary disruptions to my team. I would prefer to fix this upstream in sentry but if now we will have to look at a way to patch it in our system.

untitaker:
@BobReid CI is not green so I have restarted it

I think the fix is fine (and I really apprechiate it and will make sure to get it merged this week, sorry for the delay!) but in general I would always find a way to unblock yourself from getting OSS contributions merged. If you need to escalate an issue I would go through Sentry support to get things looked at, because to support tickets we actually have SLOs and resources attached.

BobReid:
[...]

PR #1076, opened 2021-04-08
fix(rq): Only capture exception if RQ job has failed (ignore retries)

BobReid:
Fixes https://github.com/getsentry/sentry-python/issues/1074

Sentry reports all errors from RQ even if it is still retrying the job. 

This changes checks to ensure that retries have been exhausted, ie. the job has failed and will be moved to the failed jobs queue, before capturing the exception. 

Changes:

integrations/rq.py
 - only call _capture exception if job.is_failed is true, indicating it is moving to the failed jobs registry
 - add ignore_logger('rq.worker') so the exception does not get captured via the logging integration 
 
tests/integrations/rq/test_rq.py
 - Add a failing test case for redis version 1.5 and above (where Retry was introduced) to ensure only one exception is captured in the event of a retry

### DIFF
--- a/sentry_sdk/integrations/rq.py
+++ b/sentry_sdk/integrations/rq.py
@@ -108,7 +108,6 @@ class RqIntegration(Integration):
 
         Queue.enqueue_job = sentry_patched_enqueue_job
 
-
         ignore_logger("rq.worker")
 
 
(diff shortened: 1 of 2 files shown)


BobReid:
POC for addressing https://github.com/getsentry/sentry-python/issues/1074

untitaker (approved):

untitaker:
I cannot merge master into your branch but I've manually verified that this works.

PR #1080, opened 2021-04-10
Permit the usage of a custom Logger class in the logging module

collinbanko:
The logging module allows one to replace the class used when instantiating a logger using `logging.setLoggerClass`. At the moment, Sentry's logger integration patches the callHandlers on only the standard Logger class.

This fix accounts for the possibility the logger class was customized and retrieves the class that will be used for logger instantiation via the `logging.getLoggerClass` method.


collinbanko:
Hi there, would it possible to have the checks rerun on this pull request (and https://github.com/getsentry/sentry-python/pull/1085)? The test failures for `{py3.8,py3.9}-django-dev` appear to have been resolved upstream (as seen in more recent PR checks).

PR #1085, opened 2021-04-13
fix(integration): Discard `-dev` when parsing required versions for bottle

OndrejIT:
This PR bring support dev branch integration.
Simply strip -dev suffix.

https://github.com/bottlepy/bottle/blob/master/bottle.py#L20

`>>> tuple(map(int, "0.13-dev".split(".")))
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '13-dev'
`


ahmedetefy (approved):
Thanks for the PR!

Issue #760, opened 2020-07-13
Would be helpful to have function decorators for transactions and spans


markstory:
The current context managers for starting transactions and marking spans are great. However, there have been a few cases where having decorators for transactions and spans would make code simpler to understand. For example celery tasks could be written like

```python
@celery.task
@sentry_sdk.transaction(transaction="celery.update.user" op="job")
def update_user(user_id):
    # Do things
```

Has less visual noise than the decorator does, and when adding/removing instrumentation the entire function doesn't change. The same idea applies to spans: 

```python
@sentry_sdk.span(description="github.fetch_users", op="github.api")
def fetch_github_users(...):
    # do the api request and data transformations.
```

The names of the decorators in this issue are likely rubbish, but I think supporting decorators will help make adding instrumentation simpler for end users as it generates less noisy differences and reduces the amount of indentation in their code.

untitaker:
Yup, has been brought up a couple times. I feel like we may only want this
for spans for now.

For transactions I feel like the decorator could push a scope and set an
error boundary at the same time as well

ynouri:
My team is currently adopting Sentry for our Python code base and have been feeling the need for a decorator too. The main goal is to make it easier to start tracing any arbitrary function, and make the business logic more readable by removing instrumentation code.

We came up with the decorator implementation attached in draft PR https://github.com/getsentry/sentry-python/pull/1089.

Does this match with what you had in mind by any chance? 

PS: I'm new to Sentry and to the `sentry-python` code base, apologies in advance if this proposal misses obvious features or edge cases!

ynouri:
Also curious if there are any existing alternatives or workarounds for this - we didn't mean to re-invent the wheel.

ynouri:
We have been using those decorators in production on a Python 3.8 FastAPI app for a few months now and they seem to get the job done.

I would be willing to push the PR to the finish line (linting, unit tests) but would prefer to get feedback first. @untitaker any thoughts?

brettc:
I'm starting to add some custom spans to our FastAPI app, and my immediate thought was: "where are the decorators for this?" What is the status of the PR #1089 ? (and thanks @ynouri, I'll cut and paste you code for now).

alecgerona:
Hey all, since this is merged, any chance we can update the docs as well to show how to use these decorators?

antonpirker:
Hey @alecgerona 

[...]

PR #1089, opened 2021-04-15
Add decorator for Sentry tracing

ynouri:
Draft proposal for https://github.com/getsentry/sentry-python/issues/760


ynouri:
> This pull request has gone three weeks without activity. In another week, I will close it.
> 
> But! If you comment or otherwise update it, I will reset the clock, and if you label it `Status: Backlog` or `Status: In Progress`, I will leave it alone ... forever!
> 
> ----
> 
> "A weed is but an unloved flower." ― _Ella Wheeler Wilcox_ 🥀

I would be happy to bring this to the finish line but probably best to get feedback from a maintainer first. Thank you

sentry_sdk/decorators.py
```diff
@@ -0,0 +1,89 @@
...
+    Usage:
+    @sentry_traced
+    def my_function():
+        ...
+
+    @sentry_traced(transaction_name="new_tx")
+    async def my_async_function():
+        ...
+
+    @sentry_trace(op="child_name")
```
janfabry, line 40: Typo: missing `d` at the end of `sentry_trace`.

janfabry:
Would it be better to have a separate decorators for transactions and spans? They have different concerns.

The transaction decorator:
 - Would ensure a transaction exists
 - Would overwrite the name (if passed)
 - Would ensure a passed `sampled` argument is adhered to (if `True` is  passed, but the existing tx is not sampled, it should restart the transaction - because if you start with sampling disabled I think you can't enable it later on, the `SpanRecorder` is not enabled) (calling the decorator `sentry_traced` when it does not ensure it actually is traced is confusing)

The span decorator:
 - Would start a new span, with the passed args
 - Would only work if a transaction exists, but would not care if it doesn't

Maybe you could also have a decorator that combines both of these, but then it needs a clear name.

(I'm struggling with these questions right now as we are designing the decorators for our use case, so maybe it's a useful brain dump, but if not, feel free to ignore it)

sentry_sdk/decorators.py
ynouri: Done

ynouri:
I would be happy to bring this to the finish line but probably best to get feedback from a maintainer first. I moved it from Draft to Ready for Review. Maybe it'll get picked up by someone with this new status. cc @sl0thentr0py

antonpirker:
Hey @ynouri 
I will change the API, ab bit, because we have some things that should be working the same over different programming languages.

Also the span.op should not be set to an arbitrary value because we index spans based on this so we have a fixed set we can choose: https://develop.sentry.dev/sdk/performance/span-operations/

(I will change your code to set it to `function`)

antonpirker:
I have now removed the creation of the transaction in your decorator and also set the span `op` always to `function`. 
I know this might now frustrate you because you did a lot of good work, and now I come along and delete half your code. 
Just be assured, that we really appreciate your contribution, but as Sentry has to support a lot of platforms and versions we have some constraints. Sorry for that.

One thing that we need to fix now: Python 2.7 does not understand the `async def something()` definition of a function. 
I am trying to figure out now, how the decorator can support sync functions in python 2 and 3 and async functions only in python 3.

sentry_sdk/tracing.py
```diff
@@ -803,6 +805,36 @@ def finish(self, hub=None, end_timestamp=None):
...
+    If there is no current transaction, than nothing will be traced.
+
+    Usage:
+        import sentry_sdk
+
+        @sentry_sdk.trace
+        def my_function():
+            ...
+
+        @senrty_sdk.trace
```
HazAT, line 821:
```suggestion
        @sentry_sdk.trace
```

antonpirker:
This is the updated documentation: https://github.com/getsentry/sentry-docs/pull/6439

antonpirker:
If you could have a look @sl0thentr0py , would be great. 

There is some duplicated code but this is necessary so the decorator works in Python 2 and in Python 3 with sync and async functions.

sentry_sdk/tracing.py
```diff
@@ -803,6 +805,36 @@ def finish(self, hub=None, end_timestamp=None):
...
+
+        @sentry_sdk.trace
+        def my_function():
+            ...
+
+        @sentry_sdk.trace
+        async def my_async_function():
+            ...
+    """
+    if PY2:
```
sl0thentr0py, line 825: style/convention nit but instead of different files can we just switch on the `def` inside `tracing_utils` based on `PY2`, this is the pattern we follow everywhere else.

sentry_sdk/tracing_utils.py
```diff
@@ -395,6 +395,16 @@ def should_propagate_trace(hub, url):
     return False
 
 
+def get_running_span_or_transaction(hub):
```
sl0thentr0py, line 398: @antonpirker  we've long wanted to just expose `current_span` and `current_transaction` on the top level api, so maybe split this out into a separate PR and just do it as well.

sl0thentr0py (changes requested):
some convention nits, I really think we don't wanna start splitting files for python 2/3

sentry_sdk/tracing.py
antonpirker: This does not work because it is a `async def`. As soon as the Py2 interpreter sees a `async def` in a file it stops parsing it, thas whey the separate file. 
This is the first case we have where we have a async function in code that is also used in python 2. (the `@trace` decorator should work in py2 and py3. 
All the other cases of async code is code that only works in py3 (because those frameworks only support py3)

sentry_sdk/tracing_utils.py
antonpirker: Ok, sounds like a useful thing. Have put it here: https://github.com/getsentry/sentry-python/pull/1954

sl0thentr0py:
@antonpirker why are there so many test changes in this PR now? can you summarize them?

antonpirker (approved):

antonpirker:
Because I now install `pytest_asyncio` for running the common tests some tests (like test_asyncio) have been run for the first time (they where not run at all before) and this lead to some broken tests.

Now I changed the test setup to have a "common" test environment that is kind of like an integration, so I can install `pytest_asyncio` to only this "common" tests without having it also there in all the integration tests.

sentry_sdk/tracing_utils_py3.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    # Asynchronous case
+    if inspect.iscoroutinefunction(func):
+
+        @wraps(func)
+        async def func_with_tracing(*args, **kwargs):
+            # type: (*Any, **Any) -> Any
+
+            span_or_trx = get_current_span_or_transaction(sentry_sdk.Hub.current)
```
sl0thentr0py, line 33: also as discussed I think `get_current_span` here is sufficient

sl0thentr0py (changes requested):

sl0thentr0py (approved):

antonpirker:
Hey @ynouri !

Thanks for the great contribution and the patience! Finally we have merged your contribution!

If you want to have a small gift for your work, please send me your shipping address and your t-shirt size to anton.pirker@sentry.io and I will send you something! Thanks!

ynouri:
Thanks @antonpirker for the heavy lifting!

PR #1093, opened 2021-04-17
Set daemon attribute instead of using setDaemon method that was deprecated in Python 3.10

### LINKED ISSUE #1092: threading.setDaemon has been deprecated in favor of setting daemon attribute directly in Python 3.10
Ref : https://github.com/python/cpython/pull/25174

https://github.com/getsentry/sentry-python/blob/927903e3b354a42e427d91129c399d64d480a6b9/sentry_sdk/worker.py#L69

tirkarthi:
Fixes #1092


ahmedetefy:
Thanks for the PR!

PR #1101, opened 2021-04-28
chore: Fix CI failures

untitaker:
Fix a bunch of issues with CI that prevent me from releasing

* django main branch has changed how classbased views work, slightly, view_func is now some sort of closure
* spark is unmaintained so we can do nothing but remove it from CI for now
* Flask has a RC-version released so it generally feels safer to allow those weird versions to be used with the SDK


ahmedetefy:
Please add a description to the PR

ahmedetefy (approved):

Rada8732 (approved):

PR #1119, opened 2021-05-19
feat(integration): Add Httpx Integration

### LINKED ISSUE #1084: integration for httpx
About [httpx](https://github.com/encode/httpx/):

> HTTPX is a fully featured HTTP client for Python 3, which provides sync and async APIs, and support for both HTTP/1.1 and HTTP/2.

About integration:

Currently, sentry has integration for `httplib` but httpx has its own transport layer and [we can customize the transport layer if we needed](https://www.python-httpx.org/advanced/#custom-transports). 

About implementation:

We can patch `Client.send` and `AsyncClient.send` to support both synchronous and async interfaces.

I already write this integration for internal use, I will make a pr if you want to accept this feature.

elonzh:
close #1084

### DIFF
--- a/sentry_sdk/integrations/httpx.py
+++ b/sentry_sdk/integrations/httpx.py
@@ -8,7 +8,7 @@ if MYPY:
 
 
 try:
-    from httpx import AsyncClient, Client, Request, Response
+    from httpx import AsyncClient, Client, Request, Response # type: ignore
 except ImportError:
     raise DidNotEnable("httpx is not installed")
 
--- a/sentry_sdk/integrations/httpx.py
+++ b/sentry_sdk/integrations/httpx.py
@@ -21,6 +21,10 @@ class HttpxIntegration(Integration):
     @staticmethod
     def setup_once():
         # type: () -> None
+        """
+        httpx has its own transport layer and can be customized when needed,
+        so patch Client.send and AsyncClient.send to support both synchronous and async interfaces.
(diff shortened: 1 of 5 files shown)


ahmedetefy:
Hey Thanks for the PR! It looks good!

Just a couple of minor points:
1- Can you add a test for outgoing trace headers?
For reference something like -> https://github.com/getsentry/sentry-python/blob/4b4ffc05795130c8a95577074a29462c2a512d66/tests/integrations/stdlib/test_httplib.py#L120

2- Please document your code and tests

elonzh:
> Hey Thanks for the PR! It looks good!
> 
> Just a couple of minor points:
> 1- Can you add a test for outgoing trace headers?
> For reference something like ->
> 
> https://github.com/getsentry/sentry-python/blob/4b4ffc05795130c8a95577074a29462c2a512d66/tests/integrations/stdlib/test_httplib.py#L120
> 
> 2- Please document your code and tests

check out the new commit

ahmedetefy:
Thanks for looking into my comments!

Please add a `type: ignore` comment to the `httpx` imports line to pass the linter tests

elonzh:
> Thanks for looking into my comments!
> 
> Please add a `# type: ignore` comment to the `httpx` imports line to pass the linter tests

Done.

ahmedetefy (approved):

PR #1124, opened 2021-05-31
fix(ci): Fix failing CI dependencies due to Werkzeug and pytest_django

ahmedetefy:
This PR fixes failing CI due to : 
- `trytond` failing dependency because it uses deprecated methods in Wekzeug and so pinning `werkzeug` to version `<2.0`
- Introduction of explicit database access requirement due to an experimental feature in `pytest-django` version `4.3`

### DIFF
--- a/tests/integrations/django/test_basic.py
+++ b/tests/integrations/django/test_basic.py
@@ -245,7 +245,7 @@ def test_sql_queries(sentry_init, capture_events, with_integration):
 
 
 @pytest.mark.forked
-@pytest.mark.django_db
+@pytest.mark.django_db(databases="__all__")
 def test_sql_dict_query_params(sentry_init, capture_events):
     sentry_init(
         integrations=[DjangoIntegration()],
@@ -257,8 +257,6 @@ def test_sql_dict_query_params(sentry_init, capture_events):
 
     if "postgres" not in connections:
         pytest.skip("postgres tests disabled")
-
-    pytest.mark.django_db.databases.append("postgres")
 
     sql = connections["postgres"].cursor()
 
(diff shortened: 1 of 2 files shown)


tests/integrations/django/test_basic.py
```diff
@@ -21,6 +22,19 @@
...
+pytest_mark_django_db_decorator = pytest.mark.django_db
+try:
+    pytest_version = tuple(map(int, pytest_django.__version__.split(".")))
+    if pytest_version > (4, 2, 0):
+        pytest_mark_django_db_decorator = pytest.mark.django_db(databases="__all__")
+except ValueError:
+    if "dev" in pytest_django.__version__:
+        pytest_mark_django_db_decorator = pytest.mark.django_db(databases="__all__")
+except AttributeError:
+    pass
```
untitaker, line 36: I think this is simpler:

```suggestion
pytest_mark_django_db_decorator = pytest.mark.django_db(databases="__all__")
try:
    pytest_version = tuple(map(int, pytest_django.__version__.split(".")))
    if pytest_version <= (4, 2, 0):
        pytest_mark_django_db_decorator = pytest.mark.django_db
except (ValueError, AttributeError):
    pass
```

you'd have to test if this works though

untitaker (approved):
works for me

tests/integrations/django/test_basic.py
ahmedetefy: Nah that would not work because `pytest_django.__version__` was introduced in `4.0` I think

ahmedetefy: https://pytest-django.readthedocs.io/en/latest/changelog.html#v4-0-0-2020-10-16

Issue #1128, opened 2021-06-02
Support Sanic 21.03


robd003:
Could you guys support the latest version of Sanic?

SaidBySolo:
#1066 Maybe can support

SaidBySolo:
I recently found it in the documentation.

It seems that the latest version of sanic is no longer supported.

https://docs.sentry.io/platforms/python/guides/sanic/

codesync-admin:
@SaidBySolo , Based on that page, that's due to a two-year old issue about Sanic not having a consistent deprecation policy for LTS releases. I'm not sure if that's still the case now (July 2021).

https://github.com/sanic-org/sanic/issues/1532 

Hopefully Sentry can revisit this! (And merge in your PR!)

codesync-admin:
@ahmedetefy Thanks -- can you update the docs as well please? :D

https://docs.sentry.io/platforms/python/guides/sanic/

(Assume that the integration's been tested and all of course)

ahmedetefy:
Yes we can. Thanks for the reminder :)

PR #1146, opened 2021-07-07
feat(integration): Add support for Sanic >=21.3

ahmedetefy:
This PR:
- Adds support for latest Sanic versions >= 21.3 because it was initially broken due to breaking changes after LTS of 20.12

Note/Todo:
- There **MIGHT** be missing support for new added functionality in version 21.3 that should be integrated in future work
- This PR is merely there to not break support for what we already have

Fixes: #1066, #1128

### DIFF
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -108,7 +108,9 @@ class SanicIntegration(Integration):
                             # route name, and so we need to remove it from Route name so the
                             # transaction name is consistent across all versions
                             sanic_app_name = self.ctx.app.name
-                            scope.transaction = rv[0].name.split("%s." % sanic_app_name)[1]
+                            scope.transaction = rv[0].name.split(
+                                "%s." % sanic_app_name
+                            )[1]
                         else:
                             scope.transaction = rv[0].__name__
             return rv
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -108,9 +108,15 @@ class SanicIntegration(Integration):
                             # route name, and so we need to remove it from Route name so the
                             # transaction name is consistent across all versions
                             sanic_app_name = self.ctx.app.name
-                            scope.transaction = rv[0].name.split(
-                                "%s." % sanic_app_name
-                            )[1]
+                            sanic_route = rv[0].name
+
+                            if sanic_route.startswith("%s." % sanic_app_name):
+                                # We add a 1 to the len of the sanic_app_name because there is a dot
+                                # that joins app name and the route name
+                                # Format: app_name.route_name
+                                sanic_route = sanic_route[len(sanic_app_name) + 1 :]
+
+                            scope.transaction = sanic_route
                         else:
                             scope.transaction = rv[0].__name__
             return rv
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -96,14 +96,21 @@ class SanicIntegration(Integration):
 
         old_router_get = Router.get
 
-        def sentry_router_get(self, request):
-            # type: (Any, Request) -> Any
-            rv = old_router_get(self, request)
+        def sentry_router_get(self, *args):
+            # type: (Any, Union[Any, Request]) -> Any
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/sanic.py
```diff
@@ -96,14 +96,21 @@ async def sentry_handle_request(self, request, *args, **kwargs):
...
-        def sentry_router_get(self, request):
-            # type: (Any, Request) -> Any
-            rv = old_router_get(self, request)
+        def sentry_router_get(self, *args):
+            # type: (Any, Union[Any, Request]) -> Any
+            if version >= (21, 3):
+                rv = old_router_get(self, *args)
+            else:
+                request = args[0]
+                rv = old_router_get(self, request)
```
untitaker, line 105: not sure why this branch is necessary, why not always do `*args`?

tox.ini
```diff
@@ -139,6 +139,9 @@ deps =
     sanic-0.8: sanic>=0.8,<0.9
     sanic-18: sanic>=18.0,<19.0
     sanic-19: sanic>=19.0,<20.0
+    sanic-20: sanic>=20.0,<21.0
+    {py3.7,py3.8,py3.9}-sanic-21: sanic>=21.0
```
untitaker, line 143: You haven't declared those jobs to run in line 40. If you run `tox -l` there will not be an entry for sanic 20 and sanic 21.

untitaker (changes requested):

sentry_sdk/integrations/sanic.py
ahmedetefy: Yeah I agree.. added it initially to make the distinction but I guess no need

tests/integrations/sanic/test_sanic.py
```diff
@@ -34,7 +40,11 @@ def test_request_data(sentry_init, app, capture_events):
     assert response.status == 200
 
     (event,) = events
-    assert event["transaction"] == "hi"
+    # Sanic versions above and including 21.3 append the app name to the route name
```
AbhiPrasad, line 43: Should this be the default? I think we should strip the app name so transactions stay consistent in between updates to the Sanic integration.

tox.ini
```diff
@@ -139,6 +141,9 @@ deps =
     sanic-0.8: sanic>=0.8,<0.9
     sanic-18: sanic>=18.0,<19.0
     sanic-19: sanic>=19.0,<20.0
+    sanic-20: sanic>=20.0,<21.0
+    sanic-21: sanic>=21.0
```
AbhiPrasad, line 145: This might break CI with sanic 22, is that fine?

tests/integrations/sanic/test_sanic.py
ahmedetefy: you are right! resolved

tox.ini
ahmedetefy: oops

sentry_sdk/integrations/sanic.py
```diff
@@ -96,14 +96,23 @@ async def sentry_handle_request(self, request, *args, **kwargs):
...
                     with hub.configure_scope() as scope:
-                        scope.transaction = rv[0].__name__
+                        if version >= (21, 3):
+                            # Sanic versions above and including 21.3 append the app name to the
+                            # route name, and so we need to remove it from Route name so the
+                            # transaction name is consistent across all versions
+                            sanic_app_name = self.ctx.app.name
+                            scope.transaction = rv[0].name.split(
+                                "%s." % sanic_app_name
+                            )[1]
```
AbhiPrasad, line 113: Could the name.split list length ever be 0 or 1?

ahmedetefy: I do not think so https://github.com/sanic-org/sanic/blob/108a4a99c7e12561b35b62e26ecfc33c77db802d/sanic/mixins/routes.py#L666

ahmedetefy: `self.name` cannot be null

ahmedetefy: I will add safe guards regardless

AbhiPrasad (approved):
Let's ship to unblock users.

PR #1151, opened 2021-07-16
Keeping compatibility with celery 4.x.x BaseTask subclass like PeriodicTask

### LINKED ISSUE #844: Version 0.17.8 broke Celery tasks with custom Task class

kalmik:
fix #844

I'm using sentry_sdk 1.3.0 on python 2.7 with celery 4.4.6 and I'm having an issue with apply_async from PeriodicTask task instances. Following the @ahmedetefy leads I decided to open this PullRequest I'm not sure if it is the best way to fix that.

```
[2021-07-16 18:15:01,512: ERROR/Beat] Message Error: Couldn't apply scheduled task cron_jobs.periodic_jobs.schedule_report: unbound method apply_async() must be called with Task instance as first argument (got TaskType instance instead)
  File "/usr/local/pythonenv/myproject/local/lib/python2.7/site-packages/redbeat/schedulers.py", line 437, in maybe_due
    result = self.apply_async(entry, **kwargs)
  File "/usr/local/pythonenv/myproject/local/lib/python2.7/site-packages/celery/beat.py", line 400, in apply_async
    entry, exc=exc)), sys.exc_info()[2])
  File "/usr/local/pythonenv/myproject/local/lib/python2.7/site-packages/celery/beat.py", line 392, in apply_async
    **entry.options)
  File "/usr/local/pythonenv/myproject/local/lib/python2.7/site-packages/sentry_sdk/integrations/celery.py", line 117, in apply_async
    return f(*args, **kwargs)
SchedulingError: Couldn't apply scheduled task cron_jobs.periodic_jobs.schedule_report: unbound method apply_async() must be called with Task instance as first argument (got TaskType instance instead)```

### DIFF
--- a/sentry_sdk/integrations/celery.py
+++ b/sentry_sdk/integrations/celery.py
@@ -70,10 +70,13 @@ class CeleryIntegration(Integration):
             return _wrap_tracer(task, old_build_tracer(name, task, *args, **kwargs))
 
         trace.build_tracer = sentry_build_tracer
+        if CELERY_VERSION < (5,):
+            # Keeping compatibility with celery 4.x.x BaseTask subclass like PeriodicTasks
+            from celery.task.base import BaseTask  # type: ignore
+        else:
+            from celery.app.task import BaseTask # type: ignore
 
-        from celery.app.task import Task  # type: ignore
-
-        Task.apply_async = _wrap_apply_async(Task.apply_async)
+        BaseTask.apply_async = _wrap_apply_async(BaseTask.apply_async)
 
         _patch_worker_exit()
 


kalmik:
Hi @ahmedetefy What are your thoughts about this change? I would like to know if I am on the right path.

Thanks

ahmedetefy:
Hey @kalmik 
Thanks for opening the PR.

My approach was similar.. but I decided not to pursue it because it doesn't capture spans properly (for performance). 
Did you manage to see a span of your celery task?

This is probably because Celery was setting the functions on this legacy Class dynamically on runtime so I was not able to monkey patch it, if I remember correctly

Anyways, considering this again maybe its not a bad idea to add this to not break backwards compatibility if we document that performance might not work properly with this hack...

Can you please document that somewhere and add a test?

kalmik:
Sure thing, I can provide it shortly.

kalmik:
I didn't forget about it, I've created the test and I think I could get the way to get it patched correctly (I HOPE) I think by this week I came up with the final patch.

imposeren:
Or at lease add information in migration guid, that old celery is not supported (I have celery version 3.1.27 in project and it also broken by newest sentry-sdk)

Issue #1063, opened 2021-03-22
Gevent, SentryAsgiMiddleware and ContextVars are not working together


wearebasti:
It seems to me that #630 re-opened, when using most up to date gevent with current sentry-sdk. The culprit seems to be that the sdk calls `_is_contextvars_broken` in `utils.py` which itself uses gevent's `is_object_patched` function to determine if contextvars was indeed patched.

If I understand the gevent sources correctly they indeed do no longer patch contextvars on python versions >= 3.7 (https://github.com/gevent/gevent/blob/83c9e2ae5b0834b8f84233760aabe82c3ba065b4/src/gevent/monkey.py#L604)

### Environment

How do you use Sentry?
Sentry SaaS (sentry.io), but error happens (already) on local installation

Which SDK and version?
python 3.9.0
sentry-sdk 1.0.0
gevent 21.1.2

### Steps to Reproduce

1. Set up gunicorn with gevent workers, use SentryAsgiMiddleware
2. Try to start gunicorn with gevent (monkey patching activated)
3. Error about missing contextvar support is thrown

### Expected Result

Sentry should work just fine as stated in https://docs.sentry.io/platforms/python/troubleshooting/ for gevent since version 20.5

### Actual Result

Gunicorn workers die with the message

> RuntimeError: The ASGI middleware for Sentry requires Python 3.7+ or the aiocontextvars package.
>
> With asyncio/ASGI applications, the Sentry SDK requires a functional
> installation of `contextvars` to avoid leaking scope/context data across
> requests.

> Please refer to https://docs.sentry.io/platforms/python/contextvars/ for more information.

indiejohnny:
I've got the same issue.

tiangolo:
I implemented a potential fix for this in https://github.com/getsentry/sentry-python/pull/1157 :rocket:

tiangolo:
This was fixed (at least for me) in Sentry Python SDK version `1.3.1` released today. :tada:

PR #1157, opened 2021-07-26
🐛 Fix detection of contextvars compatibility with Gevent 20.9.0+

tiangolo:
🐛 Fix detection of contextvars compatibility with Gevent 20.9.0+

This is described in https://github.com/getsentry/sentry-python/issues/1063, and this PR would fix it (it should at least).

## Description

The current implementation checks if Gevent patched contextvars, but:

> Gevent 20.9.0 depends on Greenlet 0.4.17 which natively handles switching context vars when greenlets are switched.
> Older versions of Python [older than 3.7] that have the backport installed will still be patched.

Ref: https://github.com/gevent/gevent/blob/83c9e2ae5b0834b8f84233760aabe82c3ba065b4/src/gevent/monkey.py#L604-L609

This PR updates the implementation to check the version of Gevent and Python to account for that.

## Use Case

I'm migrating a code base with a couple of apps from Flask to FastAPI. During the migration both Flask and FastAPI are available at the same time, sharing a lot of the code and tests.

The Flask app is currently deployed with Gevent, so the tests use Gevent's monkey patching.

The FastAPI app uses the `SentryAsgiMiddleware`, which internally uses `contextvars` to store and handle all the Sentry stuff per request. And that in turn uses this utility function to check if the current environment has working contextvars available.

So, with Python 3.7+ and a recent installation of Gevent monkey patching stuff, Sentry errors out complaining about broken contextvars, but it seems it's actually just a false positive.

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -785,7 +785,7 @@ def _is_contextvars_broken():
     Returns whether gevent/eventlet have patched the stdlib in a way where thread locals are now more "correct" than contextvars.
     """
     try:
-        import gevent
+        import gevent  # type: ignore
         from gevent.monkey import is_object_patched  # type: ignore
 
         # Get the MAJOR and MINOR version numbers of Gevent
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -5,6 +5,7 @@ import os
 import sys
 import threading
 import subprocess
+import sys
 
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


untitaker (approved):
Thanks, I don't think this can have tests unless you want to play around with the version matrix in tox, barely worth it

ahmedetefy (approved):

tiangolo:
Thanks for the reviews @untitaker and @ahmedetefy! :rocket: 

Dang it, sorry for the linter warnings, I just ran them locally and fixed them (hopefully :crossed_fingers:).

I agree that all the extra complexity for testing different versions of Gevent with different versions of Python doesn't seem worth it as it's probably more of an edge case.

Nevertheless, let me know if they are needed and I can give that a try.

ahmedetefy:
Thanks @tiangolo !

tiangolo:
Awesome! Thanks for the quick response/review and merge! :rocket:

Do you have any estimate of when this could be included in a release?

ahmedetefy:
Anytime! Yep will cut a patch today

tiangolo:
Woohoo! Awesome, thanks a lot! :rocket:

PR #1177, opened 2021-08-30
feat(tracing): Add user data/transaction name to tracestate value

lobsterkatie:
This adds user data (specifically `id` and `segment`) and transaction name to the `tracetate` value. 

Doing this has two known limitations, which, though we've discussed them, I'm adding here for posterity:

1) Adding this data puts us over the character limit for tracestate values [listed in the W3C spec](https://www.w3.org/TR/trace-context/#value) (256 characters). For reference:

    Tracestate data:

    ```
    {
        "trace_id": "12312012123120121231201212312012",
        "environment": "dogpark",
        "release": "off.leash.park",
        "public_key": "dogsarebadatkeepingsecrets",
        "user": {"id": 12312013, "segment": "bigs"},
        "transaction": "/interactions/other-dogs/new-dog/"
    }
    ```

    `tracestate`  with without either: 196 characters
    `tracestate` with user data: 256 characters
    `tracestate` with transaction name: 264 characters
    `tracestate` with both: 324 characters

2) This data may change and/or get added to the scope after the tracestate value has been calculated, making it impossible to sample based on those attributes. This is especially a problem for transaction name, which in some frameworks isn't set to its final value until the transaction ends. This poses the added problem that the transaction name in its raw, un-finalized form may contain PII, because it is often the raw URL as opposed to the parameterized one (so, `/users/maisey/tricks/` rather than `/users/:username/tricks/`). More work needs to be done to investigate whether the final transaction name can be set earlier in any/all of the frameworks where this poses a problem. (For instance, it is a known problem in our Express integration, but not yet clear if it is a problem in any Python frameworks.)

### DIFF
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -309,16 +309,32 @@ def compute_tracestate_entry(span):
     """
     data = {}
 
-    client = (span.hub or sentry_sdk.Hub.current).client
+    hub = span.hub or sentry_sdk.Hub.current
+
+    client = hub.client
+    scope = hub.scope
 
     if client and client.options.get("dsn"):
         options = client.options
+        user = scope._user
+
         data = {
             "trace_id": span.trace_id,
             "environment": options["environment"],
             "release": options.get("release"),
             "public_key": Dsn(options["dsn"]).public_key,
         }
+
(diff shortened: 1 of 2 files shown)


untitaker (approved):
SHIP IT

PR #1179, opened 2021-09-01
feat(tracing): Add `tracestate` header handling

lobsterkatie:
This PR introduces handling of the `tracestate` header, as described in the [W3C Trace Context spec](https://www.w3.org/TR/trace-context/#tracestate-header) and our own [corresponding spec](https://develop.sentry.dev/sdk/performance/trace-context/).

 Key features:

- Deprecation of `from_traceparent` in favor of `continue_from_headers`, which now propagates both incoming `sentry-trace` and incoming `tracestate` headers.
- Propagation of `tracestate` value as a header on outgoing HTTP requests when they're made during a transaction.
- Addition of `tracestate` data to transaction envelope headers.

Supporting changes:

- New utility methods for converting strings to and from base64.
- Some refactoring vis-à-vis the links between transactions, span recorders, and spans. See https://github.com/getsentry/sentry-python/pull/1173 and https://github.com/getsentry/sentry-python/pull/1184.
- Moving of some tracing code to a separate `tracing_utils` file.

Note: `tracestate` handling  is currently feature-gated by the flag `propagate_tracestate` in the `_experiments` SDK option. 

More details can be found in the main PR on this branch, https://github.com/getsentry/sentry-python/pull/971.

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -22,7 +22,7 @@ from sentry_sdk.integrations import setup_integrations
 from sentry_sdk.utils import ContextVar
 from sentry_sdk.sessions import SessionFlusher
 from sentry_sdk.envelope import Envelope
-from sentry_sdk.tracing_utils import reinflate_tracestate
+from sentry_sdk.tracing_utils import has_tracestate_enabled, reinflate_tracestate
 
 from sentry_sdk._types import MYPY
 
@@ -349,7 +349,7 @@ class _Client(object):
             tracestate_data = raw_tracestate and reinflate_tracestate(
                 raw_tracestate.replace("sentry=", "")
             )
-            if tracestate_data:
+            if tracestate_data and has_tracestate_enabled():
                 headers["trace"] = tracestate_data
 
             envelope = Envelope(headers=headers)
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -346,7 +346,7 @@ class _Client(object):
                 "sent_at": format_timestamp(datetime.utcnow()),
             }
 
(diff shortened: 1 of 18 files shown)


untitaker (commented):
feedback: this PR is very hard to review. it appears to consist of already-reviewed PRs exclusively which have been reviewed by a different set of people sometimes. and now I can re-review 1.3k added lines at once

If we're going to release all of those changes at once I think it could be rather risky. It seems to me that we have not ever deployed any part of this into our own SaaS usage. I would suggest to split this PR back up into many PRs, merge those, run a (stable) release for each, deploy each release separately to SaaS. Otherwise I think the risk is too great.

lobsterkatie:
> feedback: this PR is very hard to review. it appears to consist of already-reviewed PRs exclusively which have been reviewed by a different set of people sometimes. and now I can re-review 1.3k added lines at once
> 
> If we're going to release all of those changes at once I think it could be rather risky. It seems to me that we have not ever deployed any part of this into our own SaaS usage. I would suggest to split this PR back up into many PRs, merge those, run a (stable) release for each, deploy each release separately to SaaS. Otherwise I think the risk is too great.

> rubberstamping. apparently this featurebranch has been deployed to saas already at one point, and there's not many ways to reduce risk afaict, unfortunately

We talked about this offline, but for posterity:

This was a feature branch, so yes, it's a bunch of already-reviewed PRs. But we did actually already deploy it, in our back and front ends (see https://github.com/getsentry/sentry/pull/24177 and https://github.com/getsentry/getsentry/pull/5192), at a point when 80-90 percent of the work had already been done.)

We can talk process for future work, but in the meantime, I'm going to merge this as agreed.

PR #1184, opened 2021-09-07
fix(tracing): Break `transaction` / `span` circular references before garbage collection

lobsterkatie:
_Background for anyone not familiar with Python's internal workings (skip ahead if you that's not you):_

_Python handles memory deallocation through a combination of reference counting and cyclic garbage collection, the former taking way fewer resources than the latter. Having circular references forces the cyclic garbage collector to run, since anything involved in a reference cycle will never have a refcount of 0, even once everything _outside_ of the cycle is done with all of its members. The `gc` module used in the test mentioned below is a window specifically into the cyclic garbage collector part of the memory deallocation system, and its `collect` method returns the number of objects it was forced to deal with. See https://devguide.python.org/garbage_collector/ and https://docs.python.org/3/library/gc.html?highlight=gc#module-gc for more info._

-------

There are currently a few places in the SDK where we have circular references:

1) Transaction -> span recorder -> spans including transaction itself
2) Child span -> span recorder -> spans including child span itself
3) Transaction -> span recorder -> spans -> containing transaction
4) In `serializer.py`, `_serialize_node()` -> `_serialize_node_impl()` -> `_serialize_node()`, making each a closure for the other.

This PR addresses points 1-3*, by making the following changes:

1) Transaction -> span recorder -> spans ~including transaction itself~
    Transactions are no longer added to their own span recorders. (The SDK doesn't ever use the fact that they're there, and they're stripped out before the event is sent to Sentry.)

2) ~Child span ->~ span recorder -> spans including child span itself
    Child spans no longer have their own span recorder pointer, and instead access the recorder through their containing transaction.

3) Transaction ~-> span recorder -> spans -> containing transaction~
    When a transaction ends, after it harvests any completed spans, it now jettisons its link to its span recorder before it (the transaction) goes out of scope.

It also adds to/modifies tests covering all three scenarios

\*Point 4, which we only discovered in the process of fixing 1-3, concerns a different system than the rest, and therefore will need to be fixed in a separate PR. (h/t @untitaker for tracking this down to the serializer)

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -133,11 +133,12 @@ class Span(object):
 
         self._span_recorder = None  # type: Optional[_SpanRecorder]
 
+    # TODO this should really live on the Transaction class rather than the Span
+    # class
     def init_span_recorder(self, maxlen):
         # type: (int) -> None
         if self._span_recorder is None:
             self._span_recorder = _SpanRecorder(maxlen)
-        self._span_recorder.add(self)
 
     def __repr__(self):
         # type: () -> str
@@ -577,7 +578,7 @@ class Transaction(Span):
         finished_spans = [
             span.to_json()
             for span in self._span_recorder.spans
-            if span is not self and span.timestamp is not None
+            if span.timestamp is not None
         ]
 
         return hub.capture_event(
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -200,9 +200,11 @@ class Span(object):
             **kwargs
         )
 
-        child._span_recorder = recorder = self._span_recorder
-        if recorder:
-            recorder.add(child)
+        span_recorder = (
+            self.containing_transaction and self.containing_transaction._span_recorder
+        )
+        if span_recorder:
+            span_recorder.add(child)
         return child
 
     def new_span(self, **kwargs):
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -583,6 +583,12 @@ class Transaction(Span):
             if span.timestamp is not None
         ]
 
+        # we do this to break the circular reference of transaction -> span
+        # recorder -> span -> containing transaction (which is where we started)
+        # before either the spans or the transaction goes out of scope and has
+        # to be garbage collected
+        del self._span_recorder
+
         return hub.capture_event(
             {
                 "type": "transaction",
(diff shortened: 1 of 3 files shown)


tests/integrations/sqlalchemy/test_sqlalchemy.py
```diff
@@ -189,15 +189,15 @@ def processor(event, hint):
...
-    assert len(event["spans"]) == 999
+    assert len(event["spans"]) == 1000
 
     # Some spans have their descriptions truncated. Because the test always
     # generates the same amount of descriptions and truncation is deterministic,
     # the number here should never change across test runs.
     #
     # Which exact span descriptions are truncated depends on the span durations
     # of each SQL query and is non-deterministic.
-    assert len(event["_meta"]["spans"]) == 536
```
untitaker, line 200: Why does this change? If this report changed because of spanrecorder changes I wonder if we should subtract 1 just to not change reporting format

tests/tracing/test_misc.py
```diff
@@ -152,6 +157,65 @@ def test_finds_non_orphan_span_on_scope(sentry_init):
...
+    # can mock out `serialize`. In the long run, though, we should probably fix
+    # that. (Whenever we do work on fixing it, it may be useful to add
+    #
+    #     gc.set_debug(gc.DEBUG_LEAK)
+    #     request.addfinalizer(lambda: gc.set_debug(~gc.DEBUG_LEAK))
+    #
+    # immediately after the initial collection below, so we can see what new
+    # objects the garbage collecter has to clean up once `transaction.finish` is
+    # called and the serializer runs.)
+    monkeypatch.setattr(
```
untitaker, line 176: interesting way to deal with this! like it

sentry_sdk/tracing.py
```diff
@@ -199,9 +200,11 @@ def start_child(self, **kwargs):
             **kwargs
         )
 
-        child._span_recorder = recorder = self._span_recorder
-        if recorder:
-            recorder.add(child)
+        span_recorder = (
```
untitaker, line 203: Does this mean the entire `Span._span_recorder` field should be moved to Transaction?

untitaker (approved):

tests/integrations/sqlalchemy/test_sqlalchemy.py
lobsterkatie: I'm not sure why testing for `x - 1 = y` is better than testing `x = y + 1`. We're making a change either way...

sentry_sdk/tracing.py
lobsterkatie: In the long run, yes, I think so.

PR #1196, opened 2021-09-20
misc(test): Dont run tests on -dev branches and add latest versions of Django and Flask

kamilogorek:
I'm not fully convinced that we should run everything on the latest branches.
It can be noisy and block our release process, while providing is with nothing else than a warning that upstream might be broken.


kamilogorek:
`apidocs` task is also failing due to this import - https://github.com/getsentry/sentry-python/commit/54bc81cfb68d4c1df752d2358b8caf1969f1490d#diff-7c64294459f5053c93d44e0e33e4e73ffcef0adefcd77ba91f4031aa461a8c42R35
Which i don't know how to resolve.

untitaker (approved):

Issue #1145, opened 2021-07-07
SentryWrappingMiddleware does not expose details of wrapped middleware


MattFisher:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
Python `sentry-sdk==1.1.0`

### Steps to Reproduce

Use the python sentry sdk alongside the python NewRelic agent (newrelic-python-agent v6.4.4.161).
Wait for slow transaction traces to appear in NewRelic UI.

### Expected Result

The middleware entries in transaction traces should show the names of the middlewares being wrapped.

### Actual Result

All middlewares in transaction traces appear listed as `sentry_sdk.integrations.django_middleware:_wrap_middleware.<locals>.SentryWrappingMiddleware`, as shown in the screenshot below.
<img width="1467" alt="Screen Shot 2021-07-07 at 11 29 21 am" src="https://user-images.githubusercontent.com/70594/124687254-e5592a80-df17-11eb-91cc-b227b108bb95.png">

I've done some investigation, and the NewRelic agent determines the strings to be displayed 
via a `callable_name` function, that bottoms out in a function called [_object_context_py3](https://github.com/newrelic/newrelic-python-agent/blob/v6.4.4.161/newrelic/common/object_names.py#L232). This function returns the module name (via `object.__module__`) and `object.__qualname__` attributes of the passed object.

In the [`sentry_sdk.integrations.django.middleware._wrap_middleware` function](https://github.com/getsentry/sentry-python/blob/1.2.0/sentry_sdk/integrations/django/middleware.py#L177), the `SentryWrappingMiddleware` is given the wrapped middleware's `__name__` attribute.
```python
	if hasattr(middleware, "__name__"):
        SentryWrappingMiddleware.__name__ = middleware.__name__
```
In the standard library's `functools.wraps()`, the wrapping entity is given `('__module__', '__name__', '__qualname__', '__doc__', '__annotations__')` of the wrapped object. This implies it should be pretty safe to do the same with `SentryWrappingMiddleware`, like so:
```python
    if hasattr(middleware, "__name__"):
        SentryWrappingMiddleware.__name__ = middleware.__name__
    if hasattr(middleware, "__module__"):
        SentryWrappingMiddleware.__module__ = middleware.__module__
    if hasattr(middleware, "__qualname__"):
        SentryWrappingMiddleware.__qualname__ = middleware.__qualname__
```
or potentially
```python
[...]

PR #1202, opened 2021-09-22
Make middleware expose more wrapped attributes

MattFisher:
Include __name__, __module__, __qualname__, __doc__, __annotations__
Addresses #1145


sl0thentr0py (approved):

sl0thentr0py:
As far as I understand from [this table](https://docs.python.org/3/library/inspect.html) annotations don't make sense for a class anyway and `__doc__` is nice to have I guess, but I'd prefer not to add another version check for it, sorry!

Issue #1226, opened 2021-10-24
How to scrub/sanitise transactions


i-am-david-fernandez:
I have only just begun evaluating Sentry (in a `python`/`django` backend context, to be expanded to cover a `javascript`/`angular` frontend/client), and so far I am very impressed. One thing stopping me from using it fully, however, and as CISO recommending its use in our products, is the seeming inability to scrub/sanitise performance data (transactions, I believe). The `before_send` and `before_breadcrumb` callbacks appear to be more than flexible enough to scrub events and breadcrumbs before they leave the application, but there appears to be no equivalent for scrubbing performance data. As a result, the data available via sentry.io contains things I do not wish it to (e.g., sensitive data within request bodies). The result is that I simply cannot use performance monitoring outside, perhaps, development environments with toy/non-sensitive data.

The documentation (https://develop.sentry.dev/sdk/performance/#interaction-with-beforesend-and-event-processors) states this:
> Transactions should not go through beforeSend.

which makes me confident I haven't simply missed something (though I'd be more than happy to be corrected here).

My request, then, is for the SDK to provide an equivalent to `before_send` that allows complete control over what data gets sent via performance monitoring. More broadly, it should be possible to scrub _any_ and _all_ data that is being sent (which at this stage appears to be only `events`, `breadcrumbs` and `transactions`, but perhaps the list is longer). I can't consider Sentry to be a security-compliant tool without such facilities, and quite frankly, given the obvious awareness of the importance of managing PII, I'm surprised that this problem exists.

I'd love to make full use of Sentry, as I think it's both a brilliant idea and, above issue not-withstanding, well executed, so I hope this request is thoughtfully considered.

Thank-you.

rhcarvalho:
@i-am-david-fernandez 

> This line is particularly problematic:
> 
> https://github.com/getsentry/sentry-python/blob/cad2f65316bab4ee5792b1b788c32c57293eea5e/sentry_sdk/scope.py#L392

As you noted, that line means global event processors run before scope event processors (as designed/intended).

For the Python SDK, I see that the web framework integrations like WSGI (used by Django) are setting an event processor in the scope:

https://github.com/getsentry/sentry-python/blob/cad2f65316bab4ee5792b1b788c32c57293eea5e/sentry_sdk/integrations/wsgi.py#L119-L123

That could be revisited, as it might be possible to make them global, but that doesn't change the solution for your use case.

The outgoing flow is:

```

[...]

PR #1236, opened 2021-11-02
Treat x-api-key header as sensitive

simonschmidt:
Putting credentials in `x-api-key` header seems to be popular, it should probably be treated as sensitive by default.

This PR adds it to the default list of filtered headers used by e.g. `SentryAsgiMiddleware`

I ran into this when wanting to filter the header client-side for asgi traces, where `before_send` and `add_global_event_processor` are a bit problematic (See https://github.com/getsentry/sentry-python/issues/1226 )


antonpirker (approved):
lgtm

PR #1239, opened 2021-11-03
feat(aws-lambda): Add Python 3.9 support

tomscytale:
- Added AWS Lambda Python 3.9 runtime support
- Ignore mypy module check
- Fixed check bug and added python3.9 runtime to tests
- Revert comment
- run black on aws_lambda.py and test
- add python3.9 as compatible runtime in .craft.yml

### DIFF
--- a/sentry_sdk/integrations/aws_lambda.py
+++ b/sentry_sdk/integrations/aws_lambda.py
@@ -284,11 +284,13 @@ def get_lambda_bootstrap():
     # Python 3.7: If the bootstrap module is *already imported*, it is the
     # one we actually want to use (no idea what's in __main__)
     #
-    # On Python 3.8 bootstrap is also importable, but will be the same file
+    # Python 3.8: bootstrap is also importable, but will be the same file
     # as __main__ imported under a different name:
     #
     #     sys.modules['__main__'].__file__ == sys.modules['bootstrap'].__file__
     #     sys.modules['__main__'] is not sys.modules['bootstrap']
+    #
+    # Python 3.9: bootstrap is in __main__.awslambdaricmain
     #
     # On container builds using the `aws-lambda-python-runtime-interface-client`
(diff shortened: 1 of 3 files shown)


AbhiPrasad (approved):
This LGTM, thanks for your contribution!

tomscytale:
@AbhiPrasad thanks .  What's with the test failures? They look spurious.

AbhiPrasad:
Yup they seem to be flaky tests for sanic, we’ll take care of that before merging in this PR.

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -297,10 +297,16 @@ def get_lambda_bootstrap():
     if "bootstrap" in sys.modules:
         return sys.modules["bootstrap"]
     elif "__main__" in sys.modules:
-        if hasattr(sys.modules["__main__"], "bootstrap"):
+        module = sys.modules["__main__"]
+        if hasattr(module, "awslambdaricmain") and hasattr(
```
iker-barriocanal, line 301: Can we add a comment to note this case is specific to Python 3.9? Also providing a link to docs/code where we can see this change will help a lot.

tomscytale: I added some comments.

I can't find any documentation - I suspect the original author repeatedly ran a lambda which logged information about the contents of `sys.modules['__main__']` until they found the location of `bootstrap`

aiguofer:
I was going to open an issue about breadcrumbs bleeding through multiple lambda invocations on 3.9. I'm guessing this PR would fix that?

iker-barriocanal (approved):
LGTM. Thanks for contributing, @tomscytale! I think we should fix master before merging this PR, what do you think @AbhiPrasad?

AbhiPrasad:
Master should be fixed, please rebase.

tomscytale:
@AbhiPrasad  done

PR #1241, opened 2021-11-04
build(craft): Remove Python 2.7 support for AWS Lambda layers

iker-barriocanal:
From [AWS docs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-python.html):

> End of support for the Python 2.7 runtime started on July 15, 2021.

Since Python 2.7 is no longer supported, there's no point in having it as a compatible runtime for the created layers. Removing this version means new layers created by Craft on releases won't have it as a compatible runtime; so it doesn't affect existing layers.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -18,7 +18,6 @@ targets:
         # On the other hand, AWS Lambda does not support every Python runtime.
         # The supported runtimes are available in the following link:
         # https://docs.aws.amazon.com/lambda/latest/dg/lambda-python.html
-          - python2.7
           - python3.6
           - python3.7
           - python3.8


rhcarvalho (approved):
Solid 👍 

Note that end of support Phase 2 is November 22nd, but if the timelines there are correct, since July one cannot create new functions with 2.7, but can change existing functions. I think we're safe.
https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html

Also note that existing Python 2.7 lambdas will continue to run indefinitely. But IIUC one would not be able to change it to use a new lambda layer for Sentry.
https://aws.amazon.com/blogs/compute/announcing-end-of-support-for-python-2-7-in-aws-lambda/

Need to check the failed CI tests.

iker-barriocanal:
Good context.

PR #1248, opened 2021-11-14
Add a Quart integration

pgjones:
This is based on the Flask integration but includes background and
websocket exceptions, and works with asgi. 

Note I think the background task exception monitoring and Quart-Auth integration make this more useful than the ASGI middleware alone.

If this is welcomed I'm happy to write the tests (based on the Flask ones).

### DIFF
--- a/sentry_sdk/integrations/__init__.py
+++ b/sentry_sdk/integrations/__init__.py
@@ -54,6 +54,7 @@ def _generate_default_integrations_iterator(integrations, auto_enabling_integrat
 _AUTO_ENABLING_INTEGRATIONS = (
     "sentry_sdk.integrations.django.DjangoIntegration",
     "sentry_sdk.integrations.flask.FlaskIntegration",
+    "sentry_sdk.integrations.quart.QuartIntegration",
     "sentry_sdk.integrations.bottle.BottleIntegration",
     "sentry_sdk.integrations.falcon.FalconIntegration",
     "sentry_sdk.integrations.sanic.SanicIntegration",
--- a/sentry_sdk/integrations/__init__.py
+++ b/sentry_sdk/integrations/__init__.py
@@ -54,7 +54,6 @@ def _generate_default_integrations_iterator(integrations, auto_enabling_integrat
 _AUTO_ENABLING_INTEGRATIONS = (
     "sentry_sdk.integrations.django.DjangoIntegration",
     "sentry_sdk.integrations.flask.FlaskIntegration",
-    "sentry_sdk.integrations.quart.QuartIntegration",
     "sentry_sdk.integrations.bottle.BottleIntegration",
     "sentry_sdk.integrations.falcon.FalconIntegration",
     "sentry_sdk.integrations.sanic.SanicIntegration",
--- a/sentry_sdk/integrations/quart.py
+++ b/sentry_sdk/integrations/quart.py
@@ -1,0 +1,171 @@
+from __future__ import absolute_import
+
+from sentry_sdk.hub import _should_send_default_pii, Hub
+from sentry_sdk.integrations import DidNotEnable, Integration
+from sentry_sdk.integrations._wsgi_common import _filter_headers
+from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
+from sentry_sdk.utils import capture_internal_exceptions, event_from_exception
+
+from sentry_sdk._types import MYPY
+
+if MYPY:
+    from typing import Any
+    from typing import Dict
+    from typing import Union
+
+    from sentry_sdk._types import EventProcessor
+
+try:
+    import quart_auth  # type: ignore
+except ImportError:
+    quart_auth = None
+
+try:
+    from quart import (  # type: ignore
+        Request,
+        Quart,
+        _request_ctx_stack,
+        _websocket_ctx_stack,
+        _app_ctx_stack,
+    )
+    from quart.signals import (  # type: ignore
+        got_background_exception,
+        got_request_exception,
+        got_websocket_exception,
+        request_started,
+        websocket_started,
+    )
+except ImportError:
+    raise DidNotEnable("Quart is not installed")
+
+TRANSACTION_STYLE_VALUES = ("endpoint", "url")
+
+
+class QuartIntegration(Integration):
+    identifier = "quart"
+
+    transaction_style = None
+
+    def __init__(self, transaction_style="endpoint"):
+        # type: (str) -> None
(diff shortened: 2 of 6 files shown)


sentry_sdk/integrations/quart.py
```diff
@@ -0,0 +1,180 @@
...
+            raise ValueError(
+                "Invalid value for transaction_style: %s (must be in %s)"
+                % (transaction_style, TRANSACTION_STYLE_VALUES)
+            )
+        self.transaction_style = transaction_style
+
+    @staticmethod
+    def setup_once():
+        # type: () -> None
+
```
pgjones, line 67: There is usually a version check here, but Quart only provides the version via the importlib machinery - which I think is avoided in this codebase. It is required as the `got_background_exception` only exists in the latest Quart. What is the preference.

AbhiPrasad:
Awesome, thank you for your contribution! You might have to rebase because we fixed some broken tests on master.

I think we'll need some tests if we want to get this merged, but more than happy to provide you with any assistance you need writing them. Once that's in a good place, we can also update the Sentry docs to link to this!

sentry_sdk/integrations/quart.py
AbhiPrasad: We avoid `importlib` because it's `3.1`+ (our internals are `2.7` compatible) - though we are less stringent with our integrations. We can enforce a minimum Python and Quart version that requires an integration to work. 

`importlib.metadata` requires `3.8`, should we enforce that you need [Python 3.8+ and Quart 0.11+](https://gitlab.com/pgjones/quart/-/blob/main/CHANGELOG.rst) then?

tox.ini
```diff
@@ -30,6 +30,8 @@ envlist =
     {pypy,py2.7,py3.5,py3.6,py3.7,py3.8,py3.9}-flask-1.1
     {py3.6,py3.8,py3.9}-flask-2.0
 
+    {py3.7,py3.8,py3.9}-quart
```
AbhiPrasad, line 33: Do we want to do a min quart version we support?

pgjones: Added as 0.16.0

sl0thentr0py (commented):
just one minor comment for now since @AbhiPrasad already reviewed

sentry_sdk/integrations/quart.py
```diff
@@ -0,0 +1,180 @@
...
+        got_request_exception,
+        got_websocket_exception,
+        request_started,
+        websocket_started,
+    )
+except ImportError:
+    raise DidNotEnable("Quart is not installed")
+
+try:
+    import blinker  # noqa
```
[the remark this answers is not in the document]
pgjones: Ta

sentry_sdk/integrations/quart.py
pgjones: I forgot it requires 3.8, I'd suggest we just use the setup.py markers for now then.

sentry_sdk/integrations/quart.py
```diff
@@ -0,0 +1,173 @@
...
+
+    hub.capture_event(event, hint=hint)
+
+
+def _add_user_to_event(event):
+    # type: (Dict[str, Any]) -> None
+    if quart_auth is None:
+        return
+
+    print("!!!!!!!!!!!!!!!!!!!!!", quart_auth.current_user._auth_id)
```
augustozanellato, line 164: This might be a leftover debug print statement

pgjones: Woops, thanks.

sentry_sdk/integrations/__init__.py
```diff
@@ -54,6 +54,7 @@ def iter_default_integrations(with_auto_enabling_integrations):
 _AUTO_ENABLING_INTEGRATIONS = (
     "sentry_sdk.integrations.django.DjangoIntegration",
     "sentry_sdk.integrations.flask.FlaskIntegration",
+    "sentry_sdk.integrations.quart.QuartIntegration",
```
AbhiPrasad, line 57: Not sure we want Quart to be auto enabling quite yet, we can turn it on later based on user feedback. For now let's leave it off?

AbhiPrasad (commented):
Awesome looking great. Do you also mind also adding docs for quart in https://github.com/getsentry/sentry-docs/tree/master/src/platforms/python/guides? We can then mention the `0.16.1` version requirement, as well as what SDK version it's available in.

As an example, see what we do for Sanic: https://github.com/getsentry/sentry-docs/blob/master/src/platforms/python/guides/sanic/index.mdx

pgjones:
I've added documentation, https://github.com/getsentry/sentry-docs/pull/4444.

sentry_sdk/integrations/__init__.py
pgjones: Should I remove all the tests for the auto integration as well?

AbhiPrasad:
Thank @pgjones. We haven't had the time to look at this recently, but it's on our radar and we really want to see it merged! Will do another review soon, thanks for your patience :)

sl0thentr0py:
Hey @pgjones, many apologies for taking so long on this. I'm in for merging this now.

* let's disable auto-enabling and remove the auto tests as discussed above
* not sure what to do about the request body [TODO](https://github.com/getsentry/sentry-python/pull/1248/files#diff-fbd2538914f0a6e7a097e492cbeb85303d3a662b94159433ab9a881bf73e0906R123), I'm fine with enhancing the request parsing in a later iteration

pgjones:
I've disabled the auto integration as an additional commit, which will make it easier to revert when desired.

tests/integrations/quart/test_quart.py
```diff
@@ -0,0 +1,507 @@
+import pytest
+
+quart = pytest.importorskip("quart")
```
sl0thentr0py, line 3: you'll have to move this to an `__init__.py` like [the others](https://github.com/pgjones/sentry-python/blob/master/tests/integrations/sanic/__init__.py) to fix the test failures. I think pytest tries to parse the whole file and in older versions, the `async def`'s throw syntax errors.

sl0thentr0py (changes requested):

tests/integrations/quart/test_quart.py
pgjones: Thanks, fixed.

sl0thentr0py:
Thanks a lot @pgjones for the contribution and for your patience! 
master is failing again due to an unrelated sanic issue, so I'll merge this and handle that separately.

sl0thentr0py (approved):

PR #1254, opened 2021-11-16
fix(tests): Pin more-itertools in tests for Python 3.5 compat

AbhiPrasad:
Version 8.11.0 of more-itertools drops Python 3.5 support. This pins
the library to <8.11.0 so that we still run tests.


AbhiPrasad:
Merging to unblock tests

PR #1256, opened 2021-11-16
Fix "shutdown_timeout" typing

seedofjoy:
Fix "shutdown_timeout" parameter typing


AbhiPrasad (approved):
Yup it seems like shutdown can be a float, thank you for your contribution!

https://github.com/getsentry/sentry-python/blob/5bd47750871a392be4ed2632b70c444990844b51/sentry_sdk/client.py#L404-L419

PR #1268, opened 2021-12-06
fix(client-reports): Record lost `sample_rate` events only if tracing is enabled

sl0thentr0py:
Spurious client report outcomes were recorded before even if `traces_sample_rate` and `traces_sampler` were both `None` in the config.

Fixes https://getsentry.atlassian.net/browse/WEB-279

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -543,6 +543,10 @@ class Transaction(Span):
         hub = hub or self.hub or sentry_sdk.Hub.current
         client = hub.client
 
+        if client is None:
+            # We have no client and therefore nowhere to send this transaction.
+            return None
+
         # This is a de facto proxy for checking if sampled = False
         if self._span_recorder is None:
             logger.debug("Discarding transaction because sampled = False")
@@ -550,15 +554,11 @@ class Transaction(Span):
             # This is not entirely accurate because discards here are not
             # exclusively based on sample rate but also traces sampler, but
             # we handle this the same here.
-            if client and client.transport:
+            if client.transport and has_tracing_enabled(client.options):
                 client.transport.record_lost_event(
                     "sample_rate", data_category="transaction"
                 )
 
-            return None
-
-        if client is None:
-            # We have no client and therefore nowhere to send this transaction.
             return None
 
         if not self.name:
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -109,7 +109,7 @@ def has_tracing_enabled(options):
     # type: (Dict[str, Any]) -> bool
     """
     Returns True if either traces_sample_rate or traces_sampler is
-    non-zero/defined, False otherwise.
+    defined, False otherwise.
     """
 
     return bool(
--- a/tests/tracing/test_sampling.py
+++ b/tests/tracing/test_sampling.py
@@ -286,13 +286,18 @@ def test_warns_and_sets_sampled_to_false_on_invalid_traces_sampler_return_value(
         assert transaction.sampled is False
 
 
-@pytest.mark.parametrize("traces_sample_rate", [None, 0.0, 1.0])
-def test_records_lost_event_only_if_tracing_enabled(
-    sentry_init, traces_sample_rate, monkeypatch
+@pytest.mark.parametrize(
+        "traces_sample_rate,sampled_output,reports_output",
+        [
+            (None, False, []),
+            (0.0, False, [("sample_rate", "transaction")]),
+            (1.0, True, []),
+        ],
+)
+def test_records_lost_event_only_if_traces_sample_rate_enabled(
+    sentry_init, traces_sample_rate, sampled_output, reports_output, monkeypatch
 ):
     reports = []
-    sampled = traces_sample_rate == 1.0
-    dropped = traces_sample_rate == 0.0
 
     def record_lost_event(reason, data_category=None, item=None):
         reports.append((reason, data_category))
@@ -304,10 +309,36 @@ def test_records_lost_event_only_if_tracing_enabled(
     )
 
     transaction = start_transaction(name="dogpark")
-    assert transaction.sampled == sampled
+    assert transaction.sampled == sampled_output
     transaction.finish()
 
-    if dropped:
-        assert reports == [("sample_rate", "transaction")]
-    else:
-        assert not reports
+    assert reports == reports_output
+
+
+@pytest.mark.parametrize(
+        "traces_sampler,sampled_output,reports_output",
+        [
+            (None, False, []),
+            (lambda _x: 0.0, False, [("sample_rate", "transaction")]),
+            (lambda _x: 1.0, True, []),
+        ],
+)
+def test_records_lost_event_only_if_traces_sampler_enabled(
+    sentry_init, traces_sampler, sampled_output, reports_output, monkeypatch
+):
+    reports = []
+
+    def record_lost_event(reason, data_category=None, item=None):
+        reports.append((reason, data_category))
+
+    sentry_init(traces_sampler=traces_sampler)
+
+    monkeypatch.setattr(
+        Hub.current.client.transport, "record_lost_event", record_lost_event
+    )
+
+    transaction = start_transaction(name="dogpark")
+    assert transaction.sampled == sampled_output
+    transaction.finish()
+
+    assert reports == reports_output
--- a/tests/tracing/test_sampling.py
+++ b/tests/tracing/test_sampling.py
@@ -287,12 +287,12 @@ def test_warns_and_sets_sampled_to_false_on_invalid_traces_sampler_return_value(
 
 
 @pytest.mark.parametrize(
-        "traces_sample_rate,sampled_output,reports_output",
-        [
-            (None, False, []),
-            (0.0, False, [("sample_rate", "transaction")]),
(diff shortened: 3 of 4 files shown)


sentry_sdk/tracing_utils.py
```diff
@@ -109,7 +109,7 @@ def has_tracing_enabled(options):
     # type: (Dict[str, Any]) -> bool
     """
     Returns True if either traces_sample_rate or traces_sampler is
-    non-zero/defined, False otherwise.
+    defined, False otherwise.
```
sl0thentr0py, line 112: this was changed originally here and was lost in later refactors so I brought it back
https://github.com/getsentry/sentry-python/pull/1065

rhcarvalho: Good catch.

tests/tracing/test_sampling.py
```diff
@@ -284,3 +284,30 @@ def test_warns_and_sets_sampled_to_false_on_invalid_traces_sampler_return_value(
         transaction = start_transaction(name="dogpark")
         logger.warning.assert_any_call(StringContaining("Given sample rate is invalid"))
         assert transaction.sampled is False
+
+
+@pytest.mark.parametrize("traces_sample_rate", [None, 0.0, 1.0])
```
rhcarvalho, line 289: Do we have tests that cover the `traces_sampler` equivalent side of this?

sl0thentr0py: will add another one, basically the same logic

tests/tracing/test_sampling.py
```diff
@@ -284,3 +284,30 @@ def test_warns_and_sets_sampled_to_false_on_invalid_traces_sampler_return_value(
...
+    )
+
+    transaction = start_transaction(name="dogpark")
+    assert transaction.sampled == sampled
+    transaction.finish()
+
+    if dropped:
+        assert reports == [("sample_rate", "transaction")]
+    else:
+        assert not reports
```
rhcarvalho, line 313: This requires quite some thinking to map input -> expected output.

I suggest we make the expected output part of the test parametrization (list of tuples).
This way we also replace lines 294-295 with something more obvious and telling what the expected behavior is.

sentry_sdk/tracing.py
```diff
@@ -543,24 +543,24 @@ def finish(self, hub=None):
         hub = hub or self.hub or sentry_sdk.Hub.current
         client = hub.client
 
+        if client is None:
+            # We have no client and therefore nowhere to send this transaction.
+            return None
```
rhcarvalho, line 548: 👍  makes sense to move this up

tests/tracing/test_sampling.py
lobsterkatie, line 313: Nit: Perhaps I'm suggesting a change which would make this less Pythonic, so take it with a grain of salt, but for super clarity, what do you think of this? (Just since empty arrays aren't falsy in every language, and we all have to be polyglots around here.) Up to you.

```suggestion
        assert len(reports) == 0
```

lobsterkatie (approved):
LGTM

tests/tracing/test_sampling.py
sl0thentr0py: I just followed the existing style from other client report tests, but will do. Explicit always nicer.

sl0thentr0py: I mean I can also split them up into separate tests for each case, not sure what best practice convention is.

tests/tracing/test_sampling.py
```diff
@@ -284,3 +284,61 @@ def test_warns_and_sets_sampled_to_false_on_invalid_traces_sampler_return_value(
...
+    reports = []
+
+    def record_lost_event(reason, data_category=None, item=None):
+        reports.append((reason, data_category))
+
+    sentry_init(traces_sample_rate=traces_sample_rate)
+
+    monkeypatch.setattr(
+        Hub.current.client.transport, "record_lost_event", record_lost_event
+    )
```
rhcarvalho, line 309: I'd probably use a custom transport instead of `monkeypatch` because doing so tends to be less brittle.

I was about to disregard this, but then noticed that all other tests in the same file use `mock` and `mock.patch` for the same effect -- any reason to do it differently here? If not, then let's at least keep local consistency.

tests/tracing/test_sampling.py
```diff
@@ -284,3 +284,61 @@ def test_warns_and_sets_sampled_to_false_on_invalid_traces_sampler_return_value(
...
+
+    transaction = start_transaction(name="dogpark")
+    assert transaction.sampled == sampled_output
```
rhcarvalho, line 312: Use `is` if we want the comparison to be stricter.

E.g.:

```py
>>> 1 == True
True
>>> 1 is True
False
```

```suggestion
    assert transaction.sampled is sampled_output
```

sentry_sdk/tracing.py
```diff
@@ -543,24 +543,24 @@ def finish(self, hub=None):
...
+
         # This is a de facto proxy for checking if sampled = False
         if self._span_recorder is None:
             logger.debug("Discarding transaction because sampled = False")
 
             # This is not entirely accurate because discards here are not
             # exclusively based on sample rate but also traces sampler, but
             # we handle this the same here.
```
rhcarvalho, line 556: Isn't this comment outdated now?

rhcarvalho (changes requested):
Noticed we only look at `traces_sampler` and `traces_sample_rate` in isolation, missing a test case for when both are set, but I'm okay to leave it out.

Only two possible points left:

- [ ] Potentially stale comment
- [ ] Potentially use `mock.patch` instead of `monkeypatch` (or just a good reason to be different than the rest of the file)

rhcarvalho:
Also need to investigate the failed tests 

image

tests/tracing/test_sampling.py
sl0thentr0py: I followed the same style as all other client report related tests from Armin.
Examples:
https://github.com/getsentry/sentry-python/blob/b681fcfbbdaf8cae55f069d97b1e310dbb657d60/tests/test_transport.py#L198
https://github.com/getsentry/sentry-python/blob/b681fcfbbdaf8cae55f069d97b1e310dbb657d60/tests/test_basics.py#L103-L105

sentry_sdk/tracing.py
sl0thentr0py: hmm I think it tries to convey that we are recording drops due to both `traces_sample_rate` and `traces_sampler` which still holds.

tests/tracing/test_sampling.py
rhcarvalho: The local consistency I brought up comes from ideas as in https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds.

Considering that the different test files are following different styles already, I'll let this be your call.

tests/tracing/test_sampling.py
```diff
@@ -284,3 +284,61 @@ def test_warns_and_sets_sampled_to_false_on_invalid_traces_sampler_return_value(
...
         logger.warning.assert_any_call(StringContaining("Given sample rate is invalid"))
         assert transaction.sampled is False
+
+
+@pytest.mark.parametrize(
+    "traces_sample_rate,sampled_output,reports_output",
+    [
+        (None, False, []),
+        (0.0, False, [("sample_rate", "transaction")]),
+        (1.0, True, []),
```
rhcarvalho, line 294: My curiosity, were there no tests covering this case before?

The key behavior this PR is changing is the first case, when `traces_sample_rate` is `None`.

I read the test names again, and while it is better now that each input and output is an explicit parameter, it is not so obvious why only the `0.0-False-...` case has a report (one needs to think/know what reports are for reporting "lost" events).

Maybe a clearer test would be one where there's always some lost event, and the expected output for the `None-...` is that only the transaction-related report should be skipped, but not one about errors. Is that covered somewhere else?

rhcarvalho (approved):
Thanks for the fix @sl0thentr0py!

tests/tracing/test_sampling.py
sl0thentr0py: no that's not covered afaik, will add one.

One high level comment I want to make here re: my personal testing philosophy is about the impossibility of testing all cases.
If we have 20 config parameters for `sentry.init` each taking 2 values (boolean), all possible combinations of those are `2^20 =1048576` total cases. Even more if those are not booleans. If any of those values is continuous, theoretically you need an infinite number of tests to test every possible state of your program/api surface.
Given this theoretical state of affairs, as _engineers_, we need to pick and choose where we write tests and not all tests add value/are worth adding.

I will add this one anyway but I just wanted to clarify how I think about testing for the future.

rhcarvalho: I see your extrapolation. But we're not trying to test all combinations of input, but rather behaviors, and there are not so many interesting behaviors to test here.

For testing arbitrary inputs then we can rely on other tools such as fuzzing, and not hand-written unit tests.

sl0thentr0py: > Maybe a clearer test would be one where there's always some lost event, and the expected output for the None-... is that only the transaction-related report should be skipped, 

this tbh is a fairly contrived case which involves `sample_rate` too so I don't believe this belongs to the scope that these tests are trying to tackle.

PR #1272, opened 2021-12-09
fix(django): Fix django legacy url resolver regex substitution

sl0thentr0py:
Upstream django [CVE fix](https://github.com/django/django/compare/2.2.24...2.2.25#diff-ecd72d5e5c6a5496735ace4b936d519f89699baff8d932b908de0b598c58f662L233) caused master tests to fail.
This patches our url resolver regex substition to account for \A and \Z metacharacters.

### DIFF
--- a/sentry_sdk/integrations/django/transactions.py
+++ b/sentry_sdk/integrations/django/transactions.py
@@ -76,6 +76,8 @@ class RavenResolver(object):
             result.replace("^", "")
             .replace("$", "")
             .replace("?", "")
+            .replace("\\A", "")
+            .replace("\\Z", "")
             .replace("//", "/")
             .replace("\\", "")
         )
--- a/tox.ini
+++ b/tox.ini
@@ -114,6 +114,7 @@ deps =
     django-2.2: Django>=2.2,<2.3
     django-3.0: Django>=3.0,<3.1
     django-3.1: Django>=3.1,<3.2
+    django-3.2: Django>=3.1,<3.3
 
     flask: flask-login
     flask-0.10: Flask>=0.10,<0.11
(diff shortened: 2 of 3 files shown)


AbhiPrasad (approved):
love regex

tox.ini
```diff
@@ -114,6 +114,7 @@ deps =
     django-2.2: Django>=2.2,<2.3
     django-3.0: Django>=3.0,<3.1
     django-3.1: Django>=3.1,<3.2
+    django-3.2: Django>=3.1,<3.3
```
iker-barriocanal, line 117: Why did we introduce this dependency?

sentry_sdk/integrations/django/transactions.py
```diff
@@ -76,6 +76,8 @@ def _simplify(self, pattern):
             result.replace("^", "")
             .replace("$", "")
             .replace("?", "")
+            .replace("\\A", "")
+            .replace("\\Z", "")
```
rhcarvalho, line 80: Since this is probably in the hot path for transactions, might be worth thinking about a more efficient way to carry out the replacements in the future, e.g. https://stackoverflow.com/questions/3367809/efficiently-carry-out-multiple-string-replacements-in-python

tox.ini
sl0thentr0py: because django 4 came out and this 3.2 would resolve to 4 instead

iker-barriocanal: Shouldn't it be `>=3.2` instead?
```suggestion
    django-3.2: Django>=3.2,<3.3
```
I'm not familiar with this, but how would django v4 be resolved without this change? Just asking.

sl0thentr0py: indeed, typo, thx for spotting

mlissner:
The best part is how this bug looks sort of like Sentry is supporting Russia against Ukraine by inserting the letter Z in a random place in the interface. Love it....

PR #1276, opened 2021-12-11
No threads HTTP transport

erkib:
Some hosting providers do not allow threads in python apps (like www.pythonanywhere.com) and this enhancement provides this.

### DIFF
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -8,6 +8,7 @@ import time
 
 from datetime import datetime, timedelta
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


antonpirker:
Hey @erkib !
Thanks for the contribution! Currently the Sentry SDK relies heavily on threads (not only for the transport, but also internal health checks, profiling and sessions.)

I will close this PR for now, when we are deciding to support single threaded environments (in this linked issue: https://github.com/getsentry/sentry-python/issues/125) we will pick up this PR again for sure!

PR #1277, opened 2021-12-12
fix(django): Fix 404 Handler handler being labeled as "generic ASGI request"

BeryJu:


sentry_sdk/integrations/django/__init__.py
```diff
@@ -325,6 +329,8 @@ def _before_get_response(request):
     integration = hub.get_integration(DjangoIntegration)
     if integration is None:
         return
+    if integration.urlconf is None:
+        integration.urlconf = import_module(settings.ROOT_URLCONF)
```
antonpirker, line 333: Can we move this `import_module` call into the `except Resolver404` block down below?
So the urlconf is only imported when there is an 404 error and not for every call to _before_get_response? This would make this more scallable.

antonpirker (changes requested):
I have one consideration to make the change use less resources.

antonpirker:
It seems that we changed a lot since this PR was submitted originally. @BeryJu I know it is a lot to ask, but could you move your check for the 404 into the new `_set_transaction_name_and_source`? (you probably have to rebase your branch on master or merge master into your branch

sentry_sdk/integrations/django/__init__.py
BeryJu: The import should only happen once in the first request since the instance of `DjangoIntegration()` stays the same, and we check for `integration.urlconf`. I think the reason I didn't put the import in the 404 handler was to prevent any other errors if importing the URL conf fails during the 404 handling

BeryJu: Just rebased it on main and your suggestion makes a lot more sense now, I've moved the import call to the 404 handler

sentry_sdk/integrations/django/__init__.py
```diff
@@ -370,6 +373,15 @@ def _set_transaction_name_and_source(scope, transaction_style, request):
...
         )
+    except Resolver404:
+        urlconf = import_module(settings.ROOT_URLCONF)
+        # This exception only gets thrown when transaction_style is `function_name`
+        # So we don't check here what style is configured
+        if hasattr(urlconf, "handler404"):
+            handler = urlconf.handler404
+            scope.transaction = transaction_from_function(
+                getattr(handler, "view_class", handler)
+            )
```
antonpirker, line 384:
```suggestion
            scope.transaction = urlconf.handler404
```

I guess it would suffice to just set the handler404 as the transaction name. As the handler404 is already the dotted path to the function. (`transaction_from_function` tries to get a dotted path from a given function object. In my demo project it returned none, because `getattr(handler, "view_class", handler)` just returned the handler404 string)

antonpirker (changes requested):
See comment

sentry_sdk/integrations/django/__init__.py
BeryJu: `urlconf.handler404` doesn't have to be a dotted path to a function, especially when using a class based view it can also be an actual function call of `ViewClass.as_view()`

antonpirker: ok, got it. I have changed now the code to take the handler as-is if it is a string type and otherwise try to get the `transaction_from_function`

antonpirker (approved):
lgtm

Issue #1210, opened 2021-09-30
[client-reports] Include event processors and before send


smeubank:
On Python and JS sdks there is now the added functionality to gather information from client side and send to sentry. The JS SDK is currently gathering more outcome types for discarded events than Python. They should gather as much as possibly the same metrics

* event_processors & before_send

https://develop.sentry.dev/sdk/client-reports/

original client report support PRs for both SDKs

https://github.com/getsentry/sentry-javascript/pull/3955
https://github.com/getsentry/sentry-python/pull/1181

PR #1281, opened 2021-12-16
feat(client-reports): Record event_processor client reports

sl0thentr0py:
Record client reports of `event_processor` type when dropped.

Closes https://github.com/getsentry/sentry-python/issues/1210
and https://getsentry.atlassian.net/browse/WEB-341

### DIFF
--- a/tests/integrations/gcp/test_gcp.py
+++ b/tests/integrations/gcp/test_gcp.py
@@ -81,6 +81,9 @@ def init_sdk(timeout_warning=False, **extra_init_args):
         transport=TestTransport,
         integrations=[GcpIntegration(timeout_warning=timeout_warning)],
         shutdown_timeout=10,
+        # excepthook -> dedupe -> event_processor client report gets added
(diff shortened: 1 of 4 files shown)


tests/integrations/gcp/test_gcp.py
```diff
@@ -81,6 +81,9 @@ def init_sdk(timeout_warning=False, **extra_init_args):
         transport=TestTransport,
         integrations=[GcpIntegration(timeout_warning=timeout_warning)],
         shutdown_timeout=10,
+        # excepthook -> dedupe -> event_processor client report gets added
+        # which we don't really care about for these tests
+        send_client_reports=False,
```
sl0thentr0py, line 86: I just turned client reports off for the GCP tests since the actual problem is unrelated and caused by the same exception being caught twice and dropped by the `DedupeIntegration` event processor.

I added a separate test for the case below.

rhcarvalho (approved):

PR #1287, opened 2021-12-22
Support Celery abstract tasks

johnzeringue:
I have a use case for Celery [abstract
tasks](https://docs.celeryproject.org/en/latest/userguide/application.html#abstract-tasks)
to handle exceptions for an entire Celery app in a common way. Handled
exceptions should be swallowed in the abstract task and never make it to
Sentry.

Prior to this change, the Celery integration always instruments
`task.run` and incorrectly instruments `task.__call__` (`task(...)` is
equivalent to `type(task).__call__(...)`, not `task.__call__(...)`).
After this change, we'll use the same logic as Celery to decide whether
to instrument `task.__call__` or `task.run`. That change allows abstract
tasks to catch/raise exceptions before the Sentry wrapper.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -59,3 +59,5 @@ ignore_missing_imports = True
 [mypy-sentry_sdk._queue]
 ignore_missing_imports = True
 disallow_untyped_defs = False
+[mypy-celery.app.trace]
+ignore_missing_imports = True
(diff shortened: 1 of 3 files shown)


johnzeringue:
Fixed the linting errors. Should be ready for another CI run

sl0thentr0py:
@johnzeringue thx for the contribution. Can you first rebase on master and push again?

sl0thentr0py (approved):

PR #1300, opened 2022-01-12
Do not serialize mappings with custom __repr__ as dict-s

### LINKED ISSUE #1296: Support custom repr() defined in classes inheriting from Mapping
I noticed that if I define a class like this:
```python
from typing import Mapping

class Foo(Mapping):
    def __repr__(self) -> str:
        return "custom repr() to save space in the stack trace parameters list"
```
then Sentry does not call my custom `repr()` and serializes the object like a regular dict. That's because how `_serialize_node_impl` in `serializer.py` branches:

```python
if obj is None or isinstance(obj, (bool, number_types)):
    ...
elif isinstance(obj, datetime):
    ...
elif isinstance(obj, Mapping):  # <- this is where all classes inherited from Mapping land
    ...
elif not isinstance(obj, serializable_str_types) and isinstance(obj, (Set, Sequence)):
    ...
obj = safe_repr(obj)  # <- I want to flow here
```

I need this feature much because I deal with huge dict-likes (custom dataclasses with tens of key-values), they don't fully fit in the stack trace parameters list and thus clobber the rest of the values standing after them. My handcrafted `__repr__` summarizes them instead of printing all the contents and significantly reduces the envelope size. My current workaround is:

```python
class OverriddenReprMeta(abc.ABCMeta):
    """Metaclass for OverriddenReprMapping to make isinstance() return False on custom repr()-s."""

    def __instancecheck__(self, instance):
        """Override for isinstance(instance, cls)."""
        if type(instance).__repr__ not in (object.__repr__, dict.__repr__):
            return False
        return super().__instancecheck__(instance)


class OverriddenReprMapping(Mapping, metaclass=OverriddenReprMeta):
    """Mapping that fails isinstance() check if the object contains a custom repr()."""


sentry_sdk.serializer.serialize.__globals__["Mapping"] = OverriddenReprMapping
```

It's a broken ugly hack, of course, but it works really well for my project.

Would you please add custom `__repr__` support for classes that inherit from Mapping?

vmarkovtsev:
Fixes #1296

### DIFF
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -288,7 +288,10 @@ def serialize(event, smart_transaction_trimming=False, **kwargs):
                 else safe_repr(obj)
             )
 
-        elif isinstance(obj, Mapping):
+        elif isinstance(obj, Mapping) and type(obj).__repr__ in (
+            object.__repr__,
+            dict.__repr__,
+        ):
             # Create temporary copy here to avoid calling too much code that
             # might mutate our dictionary while we're still iterating over it.
             obj = dict(iteritems(obj))
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -646,7 +646,8 @@ def test_chained_exceptions(sentry_init, capture_events):
 
 
 @pytest.mark.tests_internal_exceptions
-def test_broken_mapping(sentry_init, capture_events):
+@pytest.mark.parametrize("with_custom_repr", [False, True])
+def test_broken_mapping(sentry_init, capture_events, with_custom_repr):
     sentry_init()
     events = capture_events()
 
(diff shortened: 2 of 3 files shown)


vmarkovtsev:
@sl0thentr0py can you please review this :pray:

vmarkovtsev:
I have fixed some tests and the style, but the CI does not trigger again.

vmarkovtsev:
As I see, the only test that's failing is the serialization of `QueryDict` in Django that has a custom `__repr__` definition.
We can have a global list of class exclusions where the Django integration can append `QueryDict`, WDYT?

sl0thentr0py:
@vmarkovtsev yes I think we want to preserve older behavior for classes such as `QueryDict` here. I'm not sure what other things out there in the wild this will affect.

The other option here is to give you a custom `def __sentry_repr__` that if available on any class will automatically short circuit the serializer and use that instead. Would that work for you?

vmarkovtsev:
Yes, `__sentry_repr__` should work. Let's go that path then :+1:

sl0thentr0py:
ok I'll make another PR then

sl0thentr0py:
closing in favor of https://github.com/getsentry/sentry-python/pull/1322

PR #1308, opened 2022-01-17
feat(django): Pick custom urlconf up from request if any

sl0thentr0py:
## Premise
Django middlewares sometimes [can override `request.urlconf`](https://github.com/django-tenants/django-tenants/blob/eee77af68004745ed104b63ed69995789e3a6b32/django_tenants/middleware/main.py#L73) for the current request which it has a [contract](https://github.com/django/django/blob/main/django/core/handlers/base.py#L283-L288) for using subsequently as the source of truth for url related operations.

As a result, we also need to respect the overwritten `urlconf` in our transaction name resolving.

## Solution

I finally went with an `_after_get_response` solution. The resolution is repeated code but runs only if `urlconf` is overwritten. I wanted to touch as little of existing logic as possible.

For reference, I went back and forth several times between doing it in a couple of other places.
* [`_before_get_response`](https://github.com/getsentry/sentry-python/blob/f92e9707ea73765eb9fdcf6482dc46aed4221a7a/sentry_sdk/integrations/django/__init__.py#L340) - this doesn't work because it runs before all the django middlewares
* [middleware wrapping](https://github.com/getsentry/sentry-python/blob/f92e9707ea73765eb9fdcf6482dc46aed4221a7a/sentry_sdk/integrations/django/middleware.py#L172) - this feels clunky since the new code block would run on every middleware invocation

## Problem
One of our customers have a `django-tenants` [multi-tenant setup](https://django-tenants.readthedocs.io/en/latest/use.html#multi-types-tenants). They have a dynamically resolved `urlconf` and set `ROOT_URLCONF` to an empty string which [breaks this `get_resolver`](https://github.com/django/django/blob/main/django/urls/resolvers.py#L83) in Django.

This results in a silent exception when we try to [update the transaction name](https://github.com/getsentry/sentry-python/blob/2e18af218996cf9214c124457bc384976ed02921/sentry_sdk/integrations/django/__init__.py#L347) and hence they see `Generic WSGI request` for all their requests.

## References

* https://docs.djangoproject.com/en/4.0/topics/http/urls/#how-django-processes-a-request
* https://docs.djangoproject.com/en/4.0/topics/http/middleware/
* https://getsentry.atlassian.net/browse/WEB-530

### DIFF
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -338,8 +338,7 @@ def _before_get_response(request):
                 )
             elif integration.transaction_style == "url":
                 scope.transaction = LEGACY_RESOLVER.resolve(
-                    request.path_info,
-                    getattr(request, 'urlconf', None)
+                    request.path_info, getattr(request, "urlconf", None)
                 )
         except Exception:
             pass
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -338,7 +338,8 @@ def _before_get_response(request):
                 )
             elif integration.transaction_style == "url":
                 scope.transaction = LEGACY_RESOLVER.resolve(
-                    request.path_info, getattr(request, "urlconf", None)
+                    request.path_info,
+                    urlconf=getattr(request, "urlconf", None),
                 )
         except Exception:
             pass
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -337,7 +337,10 @@ def _before_get_response(request):
                     getattr(fn, "view_class", fn)
                 )
             elif integration.transaction_style == "url":
-                scope.transaction = LEGACY_RESOLVER.resolve(request.path_info)
+                scope.transaction = LEGACY_RESOLVER.resolve(
+                    request.path_info,
+                    getattr(request, 'urlconf', None)
+                )
         except Exception:
             pass
 
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -346,6 +346,31 @@ def _before_get_response(request):
         )
 
 
+def _after_get_response(request):
+    # type: (WSGIRequest) -> None
+    """
+    Some django middlewares overwrite request.urlconf
+    so we need to respect that contract,
+    so we try to resolve the url again.
+    """
+    if not hasattr(request, "urlconf"):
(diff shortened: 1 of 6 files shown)


sentry_sdk/integrations/django/__init__.py
```diff
@@ -346,6 +346,31 @@ def _before_get_response(request):
...
+        return
+
+    with hub.configure_scope() as scope:
+        try:
+            scope.transaction = LEGACY_RESOLVER.resolve(
+                request.path_info,
+                urlconf=request.urlconf,
+            )
+        except Exception:
+            pass
```
antonpirker, line 371: This is probably because integrations (and sdks in general) should fail as silent as possible right?

antonpirker (approved):
Looks nice. I have just one question.

tests/integrations/django/test_basic.py
```diff
@@ -755,3 +755,30 @@ def test_csrf(sentry_init, client):
...
+    settings.ROOT_URLCONF = ""
+    settings.MIDDLEWARE.insert(0, urlconf)
+    client.application.load_middleware()
+
+    sentry_init(integrations=[DjangoIntegration()], traces_sample_rate=1.0)
+    events = capture_events()
+
+    content, status, _headers = client.get("/custom/ok")
+    assert status.lower() == "200 ok"
+    assert b"".join(content) == b"custom ok"
```
AbhiPrasad, line 778: why do we need the bytes literal here?

tests/integrations/django/myapp/custom_urls.py
```diff
@@ -0,0 +1,31 @@
...
+    from django.conf.urls import url
+
+    def path(path, *args, **kwargs):
+        return url("^{}$".format(path), *args, **kwargs)
+
+
+from . import views
+
+urlpatterns = [
+    path("custom/ok", views.custom_ok, name="custom_ok"),
```
AbhiPrasad, line 30: > Django loads that Python module and looks for the variable urlpatterns

lol - thats a little bit too much magic for my taste but to each their own I guess

AbhiPrasad (approved):

tests/integrations/django/test_basic.py
sl0thentr0py: it's a byte stream or something from `werkzeug`, not a string. I just followed the other tests.

tests/integrations/django/myapp/custom_urls.py
sl0thentr0py: in other news, software is unnecessarily complicated

PR #1314, opened 2022-01-20
fix(python): Capture only 5xx HTTP errors in Falcon Integration

### LINKED ISSUE #753: Falcon sentry integration catch all http errors
Hi,
seems that the falcon integration catches all the http errors and not just the 500 http error code as the documentation says, https://docs.sentry.io/platforms/python/falcon/#behavior

https://github.com/getsentry/sentry-python/blob/7d482b5bfa1d4f58eb090818496eba8fee8e63aa/sentry_sdk/integrations/falcon.py#L156

antonpirker:
fixes #753

### DIFF
--- a/sentry_sdk/integrations/falcon.py
+++ b/sentry_sdk/integrations/falcon.py
@@ -187,8 +187,13 @@ def _patch_prepare_middleware():
 
 
 def _exception_leads_to_http_5xx(ex):
-    # type: (BaseException) -> bool
-    status = ex.status or ""
+    # type: (Exception) -> bool
+    status = ""
+    try:
+        status = ex.status
+    except:
+        pass
+
     is_server_error = isinstance(ex, falcon.HTTPError) and status.startswith('5')
     is_unhandled_error = not isinstance(ex, (falcon.HTTPError, falcon.http_status.HTTPStatus))
     return is_server_error or is_unhandled_error
--- a/sentry_sdk/integrations/falcon.py
+++ b/sentry_sdk/integrations/falcon.py
@@ -187,7 +187,6 @@ def _patch_prepare_middleware():
 
 
 def _exception_leads_to_http_5xx(ex):
-<<<<<<< HEAD
     # type: (Exception) -> bool
     status = ""
     try:
@@ -196,15 +195,9 @@ def _exception_leads_to_http_5xx(ex):
         pass
 
     is_server_error = isinstance(ex, falcon.HTTPError) and status.startswith('5')
-    is_unhandled_error = not isinstance(ex, (falcon.HTTPError, falcon.http_status.HTTPStatus))
-=======
-    # type: (BaseException) -> bool
-    status = ex.status or ""
-    is_server_error = isinstance(ex, falcon.HTTPError) and status.startswith("5")
     is_unhandled_error = not isinstance(
         ex, (falcon.HTTPError, falcon.http_status.HTTPStatus)
     )
->>>>>>> 1bfb78c9d19c3913f66a0aa957a5bdbfe2b41179
     return is_server_error or is_unhandled_error
 
 
--- a/sentry_sdk/integrations/falcon.py
+++ b/sentry_sdk/integrations/falcon.py
@@ -194,7 +194,7 @@ def _exception_leads_to_http_5xx(ex):
     except:
         pass
 
-    is_server_error = isinstance(ex, falcon.HTTPError) and status.startswith('5')
+    is_server_error = isinstance(ex, falcon.HTTPError) and status.startswith("5")
     is_unhandled_error = not isinstance(
         ex, (falcon.HTTPError, falcon.http_status.HTTPStatus)
     )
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/falcon.py
```diff
@@ -186,9 +187,15 @@ def sentry_patched_prepare_middleware(
     falcon.api_helpers.prepare_middleware = sentry_patched_prepare_middleware
 
 
-def _is_falcon_http_error(ex):
-    # type: (BaseException) -> bool
-    return isinstance(ex, (falcon.HTTPError, falcon.http_status.HTTPStatus))
+def _exception_leads_to_http_5xx(ex):
+    # type: (Union[Exception, falcon.HTTPError, falcon.http_status.HTTPStatus]) -> bool
```
rhcarvalho, line 191: I'd imagine the input (`ex`) is still of type `BaseException` -- how could we guarantee it is limited to this `Union...`? Also, note that both [`falcon.HTTPError`](https://falcon.readthedocs.io/en/stable/_modules/falcon/http_error.html) and [`falcon.http_status.HTTPStatus`](https://falcon.readthedocs.io/en/stable/_modules/falcon/http_status.html#HTTPStatus) are subtypes of `Exception`, so I guess the union is basically the same as saying `Exception`?

```suggestion
    # type: (BaseException) -> bool
```

sentry_sdk/integrations/falcon.py
```diff
@@ -186,9 +187,15 @@ def sentry_patched_prepare_middleware(
...
+    is_server_error = isinstance(ex, falcon.HTTPError) and (ex.status or "").startswith(
+        "5"
+    )
```
rhcarvalho, line 194: This is good, I noticed that even though there's an `HTTPError.code` field, that's not the HTTP status code but some custom user-defined code.

I'm missing here though looking at `HTTPStatus` with a custom code >= 500.
For example, https://falcon.readthedocs.io/en/stable/_modules/falcon/http_status.html#HTTPStatus says `'748 Confounded by Ponies'`, I believe any code that is not in the [0, 499] range should be treated as an error to report to Sentry, perhaps including misuses of the status field, like not starting with a number, what do you think?

tests/integrations/falcon/test_falcon.py
```diff
@@ -96,6 +96,75 @@ def on_get(self, req, resp):
...
+    client.simulate_get("/")
+
+    assert len(exceptions) == 0
+    assert len(events) == 0
+
+
+def test_http_status(sentry_init, capture_exceptions, capture_events):
+    """
+    This just demonstrates, that if Falcon raises a HTTPStatus with code 500
+    (instead of a HTTPError with code 500) Sentry will not capture it.
```
rhcarvalho, line 149: What's the reasoning behind this decision of not capturing the error?

rhcarvalho (commented):
Very spot on, thanks for the clean, easy to review, patch @antonpirker! 

I have one suggestion re: types, and one question about desired behavior.

tests/integrations/falcon/test_falcon.py
antonpirker: In Falcon there are two classes HTTPError and HTTPStatus. HTTPError is used for all the errors and HTTPStatus for "normal" HTTP responses. This test is more of a documentation on how the integration works. Currently Falcon is never raising HTTPStatus with an 5xx status code, but if they will do in the future, we will tread is a "not an error" because it is of class HTTPStatus. (if they decide to do something like this we have to refactor the integration anyways)

sentry_sdk/integrations/falcon.py
antonpirker: In the http standard there are only codes from 100-599. The 7xx codes are joke codes. So I think we should be save when we just everything that is in the 5xx class. See also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

sentry_sdk/integrations/falcon.py
antonpirker: You are right. I will change this.

sl0thentr0py (approved):
lgtm!

sentry_sdk/integrations/falcon.py
rhcarvalho: This is the same topic as in `test_http_status`. It is not clear why raising `HTTPStatus(status="500 Internal Server Error")` would, by design, not report to Sentry. We're fixing an "over reporting", I wonder if we're accidentally introducing a case of under-reporting.

sentry_sdk/integrations/falcon.py
```diff
@@ -12,6 +12,7 @@
     from typing import Any
     from typing import Dict
     from typing import Optional
+    from typing import Union
```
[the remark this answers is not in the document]
antonpirker: oh, will remove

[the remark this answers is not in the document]
antonpirker: done.

sentry_sdk/integrations/falcon.py
antonpirker: Falcon is not using HTTPStatus for reporting errors, so I think we are good in not reporting raised HTTPStatus to Sentry.

rhcarvalho (approved):

PR #1318, opened 2022-01-24
Provide default argument to json.dumps in compute_tracestate_value

tomchuk:
Granted, this is a bit of an edge-case, but I'll explain the rationale.

In our Django application, we define `MERCURIAL_TAG` in our settings. To compute this we subprocess out to `hg`. As you can imagine, this is quite expensive and has a significant impact on application startup time. So we've wrapped this in `django.utils.functional.lazy()` so that it's only evaluated when it's needed, not every time the app starts or every time a uwsgi worker is recycled.

The issue is, we primarily use this setting by passing it to `sentry_sdk.init()`. When the Sentry SDK goes to report an error, `compute_tracestate_value()` calls `json.dumps()` which blows up with a whole stack of `TypeError: Object of type __proxy__ is not JSON serializable` as the `_got_request_exception` signal fires over and over.

image

I've added a `default=safe_str` to `json.dumps()` to make sure that whatever weird thing some random idiot (read: me) passes to to `sentry_sdk.init()` is properly converted to a string.

I've left the type hint intact as `(typing.Mapping[str, str]) -> str` as it's probably worth maintaining the expectation that this is the case.

### DIFF
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -11,6 +11,7 @@ from sentry_sdk.utils import (
     capture_internal_exceptions,
     Dsn,
     logger,
+    safe_str,
     to_base64,
     to_string,
     from_base64,
@@ -288,7 +289,7 @@ def compute_tracestate_value(data):
     tracestate entry.
     """
 
-    tracestate_json = json.dumps(data)
+    tracestate_json = json.dumps(data, default=safe_str)
 
     # Base64-encoded strings always come out with a length which is a multiple
     # of 4. In order to achieve this, the end is padded with one or more `=`


sl0thentr0py:
@tomchuk thx for the PR! Looks fine in principle but I'm not sure I understand your usage.
I assume you pass that lazy function to `sentry_sdk.init(release=???)` somehow but I don't see how that would work in general. A lot of other things in the SDK also depend on that `release` config option being a string.

tomchuk:
@sl0thentr0py - You are correct, I'm using this lazy string for the `version` argument to `sentry_sdk.init()`. Remarkably, this seems to be the only place where this lazy string object affects things, though we don't use Sentry's performance product so there may be some code paths there I'm not exercising. It basically comes down to `json`'s `_iterencode` doing a bunch of `isinstance` checks while encoding. If you just blindly treat it as a string, it behaves just like a string. 

Eg:
```
In [1]: settings.MERCURIAL_REV
Out[1]: 'da15c89d169f'

In [2]: type(settings.MERCURIAL_REV)
Out[2]: django.utils.functional.lazy.<locals>.__proxy__

In [3]: "one" + settings.MERCURIAL_REV
Out[3]: 'oneda15c89d169f'

In [4]: settings.MERCURIAL_REV.encode("utf-8")
Out[4]: b'da15c89d169f'
```

sl0thentr0py (approved):

PR #1322, opened 2022-01-26
feat(serializer): Allow classes to short circuit serializer with `sentry_repr`

sl0thentr0py:
Based on the discussion in https://github.com/getsentry/sentry-python/pull/1300 and https://github.com/getsentry/sentry-python/issues/1296.

We do a lot of truncation magic in our serializer to not send huge payloads.
In the process, we ignore if a user/framework sets a custom `__repr__` on a class.
This PR lets users add a `sentry_repr` that will short circuit our truncation logic and pick the user defined repr.
We cannot just go with checking if `__repr__` is overridden because we still want to preserve the truncation behavior for classes out there in the wild like [`QueryDict`](https://docs.djangoproject.com/en/2.2/ref/request-response/#django.http.QueryDict).

Some notes from the TSC:
* for now this is python only because python is the only SDK with an elaborate serializer
* and in general, this only makes sense for dynamic languages

### DIFF
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -281,6 +281,9 @@ def serialize(event, smart_transaction_trimming=False, **kwargs):
             else:
                 return obj
 
+        elif callable(getattr(obj, "sentry_repr", None)):
+            return obj.sentry_repr()
+
         elif isinstance(obj, datetime):
             return (
                 text_type(format_timestamp(obj))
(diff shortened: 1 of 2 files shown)


antonpirker (approved):
lgtm

vmarkovtsev (commented):
Thanks!

sl0thentr0py:
@vmarkovtsev heads up, I had to change it to `__sentry_repr__` in the latest release because we had some mocking problems on our end due to this change.

PR #1324, opened 2022-01-29
Group captured warnings under separate issues

mnito:
Prior to https://bugs.python.org/issue46557 being addressed, warnings captured by logging.captureWarnings(True) were logged with logger.warning("%s", s) which caused them to be grouped under the sameissue. This change adds special handling for creating separate issues for captured warnings arriving with the %s format string by using args[0] as the message instead of the msg arg.

Refer to: https://docs.python.org/3/library/logging.html#logging.captureWarnings


sentry_sdk/integrations/logging.py
```diff
@@ -222,7 +222,20 @@ def _emit(self, record):
 
         event["level"] = _logging_to_event_level(record.levelname)
         event["logger"] = record.name
-        event["logentry"] = {"message": to_string(record.msg), "params": record.args}
+
+        # Ensure warnings captured by logging module arrive as separate issues
+        # Before https://bugs.python.org/issue46557 was addressed, captured
+        # warnings were logged with logger.warning("%s", s) which caused all
+        # captured warnings to be grouped under the same issue
+        # Refer to https://docs.python.org/3/library/logging.html#logging.captureWarnings
```
antonpirker, line 230: A little bit to specific. In a couple of months/years no one cares about the python issue. Please change to something like "Capture warnings from `warnings` module when logging.captureWarnings(True) was set."

tests/integrations/logging/test_logging.py
```diff
@@ -136,6 +137,30 @@ def filter(self, record):
...
+        integrations=[LoggingIntegration(event_level="WARNING")],
+        default_integrations=False,
+    )
+    events = capture_events()
+
+    logging.captureWarnings(True)
+    warnings.warn("first")
+    warnings.warn("second")
+    logging.captureWarnings(False)
+
```
antonpirker, line 151: Please add another `warning.warn()` and test below that this `warnings.warn()` was not captured in the events. Thanks

antonpirker (changes requested):

sentry_sdk/integrations/logging.py
mnito: Done ✅

tests/integrations/logging/test_logging.py
mnito: Done ✅

antonpirker: Nice.

sentry_sdk/integrations/logging.py
antonpirker: ok

antonpirker:
@mnito I have made some cosmetic changes to your branch to make it more readable

antonpirker:
I have one question @mnito can we check for messages from warnings module in another way other than `record.msg == "%s"`? This feels a little bit unstable to check for "%s" in the message

mnito:
> I have one question @mnito can we check for messages from warnings module in another way other than `record.msg == "%s"`? This feels a little bit unstable to check for "%s" in the message

Maybe the intent isn’t clear here, but the check for “%s” is important to this change. Even before this change, logged warnings were always captured by the Sentry SDK — however they are grouped under a single issue due to “%s” **always** being the logged string and the actual message being the first arg (this behavior is within CPython’s logging library). I changed that behavior in an approved change to CPython itself, but this change is a workaround for Python versions that won’t have that change since it likely won’t be backported.

antonpirker (approved):

antonpirker:
@mnito could you format the `logging.py` file so our CI won't complain? Thanks!

See:
https://github.com/getsentry/sentry-python/runs/5126865254?check_suite_focus=true

mnito:
> > @mnito could you format the `logging.py` file so our CI won't complain? Thanks!
> > See: https://github.com/getsentry/sentry-python/runs/5126865254?check_suite_focus=true
> 
> Done ✅

Fixed the `mypy` failure as well ✅

antonpirker:
Uh yeah! 👍 
Thanks a lot!

sl0thentr0py (approved):

antonpirker:
This is merged now @mnito Thanks for your contribution! It will be released with the next version of the SDK (probably in 1-2 weeks)

mnito:
Sounds good. Thanks!

PR #1331, opened 2022-02-11
AioHttpIntegration sentry_app_handle() now ignores ConnectionResetError

cmalek:
This just means that the upstream closed the connection.  aiohttp itself just debug logs this and discards it, so sentry_sdk should, too.


sentry_sdk/integrations/aiohttp.py
```diff
@@ -115,6 +115,9 @@ async def sentry_app_handle(self, request, *args, **kwargs):
                     except asyncio.CancelledError:
                         transaction.set_status("cancelled")
                         raise
+                    except ConnectionResetError:
+                        transaction.set_status("connection_reset")
```
sl0thentr0py, line 119: the span status needs to be one in the list defined [here](https://develop.sentry.dev/sdk/event-payloads/span/#attributes)

sl0thentr0py (changes requested):

sentry_sdk/integrations/aiohttp.py
cmalek: Thanks for reviewing my pull request!  I've updated the patch to consider `ConnectionResetError` to have "cancelled" status.

cmalek:
I did make the changes you requested, @sl0thentr0py.

sl0thentr0py (approved):

sl0thentr0py:
thx for the contribution @cmalek!

Issue #1171, opened 2021-08-26
Django integration: connect Javascript SDK


danielmanser:
I'm completely stuck with connecting the Python SDK (I'm using Django) and the Javascript SDK. I've installed `sentry-python` and it sends events to my onPrem Sentry installation. Now I want to connect it with the Frontend using Sentry SDK.

I followed the docs and I think I need to add the `trace_id` as a `<meta>` tag in my template. Either I'm too dumb or the docs could be improved, but I've spent 3+hrs trying to get it done with no success.

### Environment

How do you use Sentry?
Sentry onPrem, version 21.8.0, installed with the recommended Docker service using `install.sh`. Sentry seems to work, I can receive errors.

Which SDK and version?
Backend: Python 3.8.5, Django 3.2.6 using sentry-sdk, version 1.3.1
Frontend: Basic HTML, Bootstrap, nothing fancy

Python virtual environment:
```
asgiref==3.4.1
certifi==2021.5.30
Django==3.2.6
django-appconf==1.0.4
django-compressor==2.4.1
django-debug-toolbar==3.2.2
django-extensions==3.1.3
django-honeypot==1.0.1
psycopg2-binary==2.9.1
pytz==2021.1
rcssmin==1.0.6
rjsmin==1.1.0
sentry-sdk==1.3.1
six==1.16.0
sqlparse==0.4.1
urllib3==1.26.6
```

### Steps to Reproduce

1. Set up the Django integration following [the official docs](https://docs.sentry.io/platforms/python/guides/django/). Unhandled errors in Django are working and being sent to my Sentry instance, so I guess this part is not in question.
2. Following [Connecting Services](https://docs.sentry.io/platforms/python/performance/#connecting-services), I decided it would be easiest to add a `<meta>` tag in my Django template, so I followed [this](https://docs.sentry.io/platforms/javascript/performance/connect-services/#pageload):

```html
<meta name="sentry-trace" content="{{ span.toSentryTrace() }}" />
```

I understand that the docs are not Django-specific and I can't call a function within a template in Django. So I've tried a lot of different things, but I don't seem to have the `trace_id` available in my template, nor in the Python environment.

Do I need to add it using a custom template tag? Or is there a middleware I'm not aware of? I've googled for "Django Sentry trace_id" and a lot of other things, searched Github for examples, but didn't find anything that worked.

[...]

PR #1336, opened 2022-02-14
feat(flask): Add `sentry_trace()` template helper

BYK:
To setup distributed tracing links between a Flask app and a front-end app, one needs to figure out how to get the current hub, safely get the traceparent and then properly pass it into a template and then finally use that properly in a `meta` tag. [The guide](https://docs.sentry.io/platforms/javascript/performance/connect-services/) is woefully inadequate and error-prone so this PR adds a built-in helper `sentry_trace()` to the Flask integration to simplify this linking.

I'll follow up with documentation updates once the approach here is approved.

### DIFF
--- a/examples/tracing/templates/index.html
+++ b/examples/tracing/templates/index.html
@@ -1,6 +1,6 @@
 <meta name="sentry-trace" content="{{ sentry_trace }}" />
 <script src="https://browser.sentry-cdn.com/6.17.7/bundle.js" crossorigin="anonymous"></script>
-{{ sentry_trace() }}
+{{ sentry_trace }}
 <!-- TODO: Replace with real tracing integration once it's fixed -->
 <script src="/static/tracing.js" crossorigin="anonymous"></script>
 
--- a/examples/tracing/templates/index.html
+++ b/examples/tracing/templates/index.html
@@ -1,4 +1,6 @@
-<script src="https://browser.sentry-cdn.com/5.4.1/bundle.js" crossorigin="anonymous"></script>
+<meta name="sentry-trace" content="{{ sentry_trace }}" />
+<script src="https://browser.sentry-cdn.com/6.17.7/bundle.js" crossorigin="anonymous"></script>
+{{ sentry_trace() }}
 <!-- TODO: Replace with real tracing integration once it's fixed -->
 <script src="/static/tracing.js" crossorigin="anonymous"></script>
 
@@ -12,14 +14,6 @@ Sentry.init({
         new Sentry.Integrations.Tracing({ tracingOrigins: ['']})
     ],
     debug: true
-});
-
-window.setTimeout(function() {
-    const scope = Sentry.getCurrentHub().getScope();
-    // TODO: Wait for Daniel's traceparent API
-    scope.setSpan(scope.getSpan().constructor.fromTraceparent(
-        "00-{{ traceparent['sentry-trace'].strip("-") }}-00"
-    ));
 });
 
 async function compute() {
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/flask.py
```diff
@@ -87,10 +96,14 @@ def sentry_patched_wsgi_app(self, environ, start_response):
             if Hub.current.get_integration(FlaskIntegration) is None:
                 return old_app(self, environ, start_response)
 
-            return SentryWsgiMiddleware(lambda *a, **kw: old_app(self, *a, **kw))(
+            patched_app = SentryWsgiMiddleware(lambda *a, **kw: old_app(self, *a, **kw))(
                 environ, start_response
             )
 
+            patched_app.jinja_env.globals['sentry_trace'] = self._get_sentry_trace
```
untitaker, line 103: there's a flask signal like `got_first_request` or something like that which would probably be better for patching in the jinja globals. as-is, this config change is done on every request

BYK: Ugh, good point. Let me check.

BYK: So, there is `before_first_request` that said we'll be calling that on the new app instance anyway?

BYK: Okay, this should be fixed now so resolving.

sl0thentr0py (approved):

sl0thentr0py:
@BYK the tests on py3.4/old flask versions fail because there's no `add_template_global` apparently, maybe add a `hasattr` check and patch only if exists.

sl0thentr0py:
hmm no ignore that, `add_template_global` exists since version 0.10, so something else going on here.

BYK:
@sl0thentr0py I figured out the issue here. It's what @untitaker said. `Flask.__call__()` creates you a `wsgi` app from an already existing Flask instance. I got confused by that. I'll try to use `before_render_template` signal but not sure if I can modify the context at that point. Will try.

BYK:
@silent1mezzo finally, ready to merge. Sorry for all that noise!

BYK:
@untitaker @mitsuhiko @sl0thentr0py I don't think that `format` job will ever run

BYK:
@sl0thentr0py it won't because this is set to run on `push`: https://github.com/getsentry/sentry-python/blob/master/.github/workflows/black.yml#L3

But I can only push to my own repo.

sl0thentr0py:
yea we need to kill that `black` action and move to pre-commit, it's super annoying.
I can't merge without it passing because we removed admin overrides 2 weeks ago, sigh..

BYK:
>I can't merge without it passing because we removed admin overrides 2 weeks ago, sigh..

How do you release with Craft then?

AbhiPrasad:
We might want to do something similar for django as well. See: https://github.com/getsentry/sentry-python/issues/1171

BYK:
@AbhiPrasad if only Functional Software Inc. was still paying me... 😛

AbhiPrasad:
> if only Functional Software Inc. was still paying me

Well if ever get the itch to help out, opened https://github.com/getsentry/sentry-python/issues/1341 to track. Thanks for your help BYK, glad to see you're still around :)

examples/tracing/templates/index.html
```diff
@@ -1,4 +1,6 @@
-<script src="https://browser.sentry-cdn.com/5.4.1/bundle.js" crossorigin="anonymous"></script>
+<meta name="sentry-trace" content="{{ sentry_trace }}" />
```
Sija, line 1: This line shouldn't be here as it's being included via `{{ sentry_trace }}` interpolation itself.

BYK: #1434

Issue #1334, opened 2022-02-12
KeyError: 'auto_session_tracking' in version 1.5.5


tonal:
### Environment

Python 3.10.2
sentry-sdk=**1.5.5**
fastapi==0.73.0

<details><summary>All packages</summary><p>

aiojobs==1.0.0
aiomysql==0.0.22
anyio==3.5.0
asgiref==3.5.0
async-timeout==4.0.2
attrs==21.4.0
Brotli==1.0.9
cached-property==1.5.2
certifi==2021.10.8
charset-normalizer==2.0.11
click==8.0.3
coverage==6.3.1
dnspython==2.2.0
email-validator==1.1.3
envparse==0.2.0
execnet==1.9.0
fastapi==0.73.0
fastapi-utils==0.2.1
fastnumbers==3.2.1
fluent-logger==0.10.0
greenlet==1.1.2
h11==0.12.0
httpcore==0.14.7
httpx==0.22.0
idna==3.3
iniconfig==1.1.1
isodate==0.6.1
jmespath==0.10.0
lxml==4.7.1
motor==2.5.1
msgpack==1.0.3
mysqlclient==2.1.0
natsort==8.1.0
packaging==21.3
pep517==0.12.0
pip-tools==6.5.1
platformdirs==2.5.0
pluggy==1.0.0
py==1.11.0
pydantic==1.9.0
pymongo==3.12.3
PyMySQL==0.9.3
pyparsing==3.0.7
pytest==7.0.1
pytest-asyncio==0.18.1
pytest-cov==3.0.0
pytest-forked==1.4.0
pytest-xdist==2.5.0
python-snappy==0.6.0
pytz==2021.3
PyYAML==6.0
requests==2.27.1
requests-file==1.5.1
requests-toolbelt==0.9.1
rfc3986==1.5.0
sentry-sdk==1.5.5
six==1.16.0
sniffio==1.2.0
SQLAlchemy==1.4.31
starlette==0.17.1
tomli==2.0.1
typing_extensions==4.0.1
ujson==5.1.0
urllib3==1.26.8
uvicorn==0.17.4
uvloop==0.16.0
zeep==4.1.0
zstandard==0.17.0
</p>
</details>
### Steps to Reproduce

1. test_sentry.py
```python
# -*- codong: utf-8 -*-
from fastapi import FastAPI
from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
from starlette.testclient import TestClient


def test_sentry():
[...]

PR #1337, opened 2022-02-14
Added default value for auto_session_tracking ...

antonpirker:
… to make it work in fastapi

refs #1334


sl0thentr0py (approved):

sentry_sdk/sessions.py
```diff
@@ -10,23 +10,27 @@
...
-    from typing import List
     from typing import Generator
+    from typing import List
+    from typing import Optional
+    from typing import Union
 
 
 def is_auto_session_tracking_enabled(hub=None):
-    # type: (Optional[sentry_sdk.Hub]) -> bool
+    # type: (Optional[sentry_sdk.Hub]) -> Union[Any, bool, None]
```
wbolster, line 23: you sure this shouldn't have stayed `bool`?

perhaps a `return bool(should_track)` can avoid mypy errors

PR #1315, opened 2022-01-21
feat(tooling): Enabled local linting

antonpirker:
To speed up development and prevent a lot of front and back with the CI linter, I added a pre-commit hook, so before one can commit something the linting (black, flake8 and mypy) will run locally to format and lint the changed files.

PR #1339, opened 2022-02-14
meta: Remove black GH action

sl0thentr0py:
Since we moved to `pre-commit` in #1315 anyway, we don't need this flaky black stuff to run on push.
The `lint` step checks with `black --check` either way.


BYK (approved):

antonpirker (approved):

antonpirker:
I am also for removing it.

One thing we might keep in mind: people submitting PRs maybe do not have pre-commit installed so can submit horribly formatted code.

sl0thentr0py:
I wanted to bring this up in the pre-commit PR but anyway, let's do it now.
We already have `make format` [here](https://github.com/getsentry/sentry-python/blob/384cf8741c7d493dd84b6be939d1b068962adf6f/1#L27-L30), now that we added pre-commit, that's redundant.
I initially thought why add a new tool when we can just run `make format` in a simple git pre-commit hook?
But anyway, now that we added it, we can just change the `Makefile` and make users setup `pre-commit` in `CONTRIBUTING.md`.

PR #1340, opened 2022-02-14
build(changelogs): Use automated changelogs from Craft

BYK:
See https://github.com/getsentry/craft#changelog-policies for details

Moved 'versioning policy' to README so Craft does not push it down.


BYK:
@untitaker @mitsuhiko

Issue #1165, opened 2021-08-19
Add ability to configure transaction style to ASGI integration 


AbhiPrasad:
In the bottle integration, we can set a transaction style to either `endpoint` or `url` based on integration setting. We should add the same to ASGI.

See: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/bottle.py#L45

antonpirker:
A lot of our Integrations have this feature, so it makes sense to add it to the ASGI Integration.

antonpirker:
This is done.

PR #1349, opened 2022-02-21
✨ Add support for setting transaction name to path in FastAPI

tiangolo:
✨ Add support for setting transaction name to `path` in FastAPI

This is related to https://github.com/getsentry/sentry-python/issues/829 and https://github.com/getsentry/sentry-python/issues/1165

This would work with FastAPI `0.74.1` that I just released 5 minutes ago.

Most of the functionality is already tested by the Starlette tests. I added a FastAPI test only to test the new `transaction_style="url"`. But I'm not sure if that's desired. If it is, FastAPI would have to be added to the testing dependencies here, but I'm not even sure where, I suspect somewhere in the Tox configs, but I don't really know much about Tox :grimacing: , so I'll wait for feedback first.

### DIFF
--- a/sentry_sdk/integrations/asgi.py
+++ b/sentry_sdk/integrations/asgi.py
@@ -187,15 +187,16 @@ class SentryAsgiMiddleware:
             event.get("transaction", _DEFAULT_TRANSACTION_NAME)
             == _DEFAULT_TRANSACTION_NAME
         ):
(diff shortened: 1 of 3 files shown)


sl0thentr0py (approved):

sl0thentr0py:
thx for the contribution @tiangolo!

PR #1359, opened 2022-03-04
feat(measurements): Add experimental set_measurement api on transaction

sl0thentr0py:
I went with  `set_measurement(name, value, unit)` (singular) instead of `set_measurements` as in JS since I think this is more convenient for the end user. Otherwise they need to build the  name -> unit/value dict themselves.

This should be merged once [develop](https://github.com/getsentry/develop/pull/515#discussion_r819587128) is synced and the changes in relay have been made to support units for `measurements`.

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -132,17 +132,14 @@ class Span(object):
 
     def __repr__(self):
         # type: () -> str
-        return (
-            "<%s(op=%r, description:%r, trace_id=%r, span_id=%r, parent_span_id=%r, sampled=%r)>"
-            % (
-                self.__class__.__name__,
-                self.op,
-                self.description,
-                self.trace_id,
-                self.span_id,
-                self.parent_span_id,
-                self.sampled,
-            )
+        return "<%s(op=%r, description:%r, trace_id=%r, span_id=%r, parent_span_id=%r, sampled=%r)>" % (
+            self.__class__.__name__,
+            self.op,
+            self.description,
+            self.trace_id,
+            self.span_id,
+            self.parent_span_id,
+            self.sampled,
         )
 
     def __enter__(self):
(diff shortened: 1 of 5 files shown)


sentry_sdk/tracing.py
```diff
@@ -588,17 +590,34 @@ def finish(self, hub=None):
...
+            "spans": finished_spans,
+        }
+
+        if has_custom_measurements_enabled():
+            event["measurements"] = self._measurements
+
+        return hub.capture_event(event)
+
+    def set_measurement(self, name, value, unit="ms"):
+        # type: (str, float, MeasurementUnit) -> None
```
antonpirker, line 609: is the type hinting preventing from using None as a name. Or: should we test name here for invalid names?

antonpirker (commented):
I like this better then the plural version. Very clean interface. This way one can set and update values.

sentry_sdk/tracing.py
sl0thentr0py: I have no clue what types/mypy do here lol, I would kill them if you asked me 🤷

tests/tracing/test_misc.py
```diff
@@ -246,3 +246,18 @@ def test_has_tracestate_enabled(sentry_init, tracestate_enabled):
...
     else:
         assert has_tracestate_enabled() is False
+
+
+def test_set_meaurement(sentry_init, capture_events):
+    sentry_init(traces_sample_rate=1.0, _experiments={"custom_measurements": True})
+
+    events = capture_events()
+
+    transaction = start_transaction(name="measuring stuff")
```
marandaneto, line 256: if I do:

`child = transaction.start_child(x)`

can I do `child.set_measurement(x, y)` too?

Or `start_child` would return a different interface that does not provide `set_measurement`? since measurements are only available at the transaction level.

sl0thentr0py: Currently they are just on transaction and not span, `start_child` returns a span so you can't add measurements there. From what I understood of the requirements, we only want to expose this api on the transaction but this is a design decision and we can also change it.

Open decision is also if we want to expose this on the hub/top-level api as well where it'll just fetch the current hub/scope/transaction and set it there.

marandaneto: I agree that we should not leak this API to the Child level as well, but a few SDKs return the very same interface for `start_transaction` and `start_child`, so unless we change that, the public API will leak.
cc @bruno-garcia

sl0thentr0py:
tested end-end with relay/ingestion

image
https://sentry.io/api/0/projects/sentry-sdks/sentry-python/events/547beff9de5445d9953de3b802adf2d0/json/

sl0thentr0py:
tested graceful error handling (unit name too long)
image

antonpirker (approved):

PR #1363, opened 2022-03-07
fix(serializer): Implement abstract base class for customizing serialization

sl0thentr0py:
The addition of `sentry_repr` in #1322 breaks `MagicMock` since it creates attrs on the fly.
This adds a new ABC mixin class called `SentryReprMixin` which the user has to inherit and implement the `sentry_repr` method to customize serialization instead of directly implementing it on their class.


sl0thentr0py:
closing in favour of `__sentry_repr__`

PR #1364, opened 2022-03-07
fix(serializer): Make sentry_repr dunder method to avoid mock problems

sl0thentr0py:
The addition of `sentry_repr` in #1322 breaks `MagicMock` since it creates attrs on the fly.
This converts it to a dunder magic method `__sentry_repr__`.

See [here](https://github.com/python/cpython/blob/main/Lib/unittest/mock.py#L639-L640) that mock doesn't mess with custom magics and the following test code:

```bash
In [22]: from unittest import mock

In [23]: m = mock.Mock()

In [24]: m
Out[24]: <Mock id='140712474488560'>

In [25]: type(m)
Out[25]: unittest.mock.Mock

In [26]: getattr(m, "foo")
Out[26]: <Mock name='mock.foo' id='140712490825808'>

In [27]: getattr(m, "__foo__")
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [27], in <module>
----> 1 getattr(m, "__foo__")

File /usr/lib/python3.10/unittest/mock.py:636, in NonCallableMock.__getattr__(self, name)
    634         raise AttributeError("Mock object has no attribute %r" % name)
    635 elif _is_magic(name):
--> 636     raise AttributeError(name)
    637 if not self._mock_unsafe:
    638     if name.startswith(('assert', 'assret', 'asert', 'aseert', 'assrt')):

AttributeError: __foo__
```

### DIFF
--- a/tests/test_serializer.py
+++ b/tests/test_serializer.py
@@ -1,7 +1,5 @@
 import sys
-
 import pytest
-from unittest import mock
 
 from sentry_sdk.serializer import serialize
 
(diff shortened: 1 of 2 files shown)


tests/test_serializer.py
```diff
@@ -68,8 +67,25 @@ def test_serialize_sets(extra_normalizer):
...
 
     result = extra_normalizer(CustomReprDict(one=1, two=2))
     assert result == "custom!"
+
+
+try:
+    from unittest import mock
+except ImportError:
+    pass
+else:
```
untitaker, line 81: I would use pytest.importorskip here, or generally use pytest-skipping instead of letting the test be undefined.

untitaker (approved):

PR #1366, opened 2022-03-09
fix(perf): Fix transaction setter on scope to use containing_transaction

### LINKED ISSUE #1155: Setting transaction name doesn't work from child span.
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
Python/Django: django-sentry-1.13.5, sentry-sdk-1.3.0 (also tried sentry-sdk-1.1.0)

### Steps to Reproduce

We're using Sentry in Django and Graphene (for GraphQL support). We would like to set the transaction name based on GraphQL operation name.

```python

urlpatterns = [
    # ...snip
    path(
        'graphql',
        csrf_exempt(MyGraphQLView.as_view(graphiql=settings.GRAPHIQL_ENABLED)),
    ),
]

class MyGraphQLView(graphene_django.views.GraphQLView):
    def execute_graphql_request(
        self, request, data, query, variables, operation_name, show_graphiql=False
    ):
            transaction_name = f'graphql.{operation_name}' # graphql.OperationName
            scope.transaction = transaction_name
            # This works:
            scope._span._containing_transaction.name = transaction_name
            return super().execute_graphql_request(
                request, data, query, variables, operation_name, show_graphiql
            )
```

From the trace view in Sentry, the GraphQLView span is nested several spans down (in Django middlewares). From my understanding of the [sentry documentation: set transaction name](https://docs.sentry.io/platforms/python/enriching-events/transaction-name/) It seems like the `scope.transaction = 'name'` setter should be enough, but doesn't seem to work from child spans.

### Expected Result

Sentry transaction name updated to `graphql.OperationName`.

### Actual Result

Sentry transaction name stays as `/graphql`

sl0thentr0py:
fixes #1155, not sure why we have this discrepancy between getter and setter behaviour in the first place

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -173,9 +173,8 @@ class Scope(object):
         # transaction name or transaction (self._span) depending on the type of
         # the value argument.
         self._transaction = value
-        span = self._span
-        if span and isinstance(span, Transaction):
-            span.name = value
+        if self._span and self._span.containing_transaction:
+            self._span.containing_transaction.name = value
 
     @_attr_setter
     def user(self, value):


lobsterkatie:
I think it's fine to change this, but the entire business of getter and setter handling two different types is a headache which I'd love to see us fix. (It would take a new major, since it'd be a breaking change, but it's never led to anything but confusion, for us and users, so highly worth it, IMHO.)

So while it's true that this would make what the issue creator is doing work, IMHO it's kind of an antipattern which we should remove from the docs even if we don't remove from the code. `scope.transaction.name = whatever` will do it just fine, and doesn't rely on the weird legacy behavior.

vladanpaunovic (approved):

Issue #1132, opened 2021-06-17
SQLAlchemy Scoped Session - ValueError: generator already executing


bobbydams:
Hello,

In part of our code we use [scoped sessions](https://docs.sqlalchemy.org/en/14/orm/contextual.html) from SQLAlchemy. Frequently we run into the error `ValueError: generator already executing` (see screenshot). The only workaround we've found at the moment is to set `auto_enabling_integrations=False` in the `sentry.init` function.

Let me know if any additional information is needed. Thanks!

### Environment

How do you use Sentry?

on-premise v21.5.1

Which SDK and version?

Python v3.8, Sentry SDK 1.1.0

### Steps to Reproduce

Creating a scoped session

```
scoped_session = db.create_scoped_session()
```

This bug seems to be triggered by this bit of code:

```
    def get_files_from_blob(self, files, text=False):
        """get files content in thread pool in bytes or text format"""
        start = time.time()
        pool = ThreadPool(4)
        results = pool.map(partial(self._get_file, text=text), files)

        # close the pool and wait for the work to finish
        pool.close()
        pool.join()
        logger.debug("DEBUG time to get in pool: %d", time.time() - start)
        return results

    def _get_file(self, file, text):
        """..."""
        blob_file = BlobFile(file.filename)
        if text:
            data = blob_file.read_from_blob_to_text().replace("\r", "")
        else:
            data = blob_file.read_from_blob_to_bytes()
        return file, data
```

### Expected Result

There should be no errors thrown by the sentry-sdk

### Actual Result

```
Traceback (most recent call last):
  File "/src/helpers/processing.py", line 165, in instrumented_upload_processing
    processed_data = processor.run_processing()
  File "/src/helpers/processing.py", line 670, in run_processing
    outputs += self.load_hmd()
  File "/src/helpers/processing.py", line 785, in load_hmd
    data = self.get_files_from_blob(self.hmd_files, text=True)
  File "/src/helpers/processing.py", line 556, in get_files_from_blob
    results = pool.map(partial(self._get_file, text=text), files)
  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 364, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 771, in get
    raise self._value
  File "/usr/local/lib/python3.8/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
[...]

PR #1368, opened 2022-03-11
(fix) Change context manager type to avoid race in threads

Fofanko:
Fix issue from issue #1132 when we get race condition in multithreading with generator based context manager. The generator is not thread safe!

### DIFF
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -9,7 +9,7 @@ from sentry_sdk._types import MYPY
 from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk.scope import add_global_event_processor
 from sentry_sdk.serializer import add_global_repr_processor
-from sentry_sdk.tracing_utils import record_sql_queries
+from sentry_sdk.tracing_utils import RecordSqlQueries
 from sentry_sdk.utils import (
     HAS_REAL_CONTEXTVARS,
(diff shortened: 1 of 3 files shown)


sl0thentr0py:
@Fofanko if this really fixes the problem, happy to merge it in, but I'm trying to fully understand where the threading problem actually lies.

In the original stacktrace on the linked issue, the error occurs in `_after_cursor_execute`, so it seems the `ctx_mgr` variable on `conn` is being acted on by multiple threads, and so the `conn` itself is being used across threads.

From the [`Connection` documentation](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection)

> The Connection object is not thread-safe. While a Connection can be shared among threads using properly synchronized access, it is still possible that the underlying DBAPI connection may not support shared access between threads. Check the DBAPI documentation for details.

So I believe the actual fix for this would be to wrap the `_sentry_sql_span_manager` getter/setter in a lock?

sl0thentr0py (approved):
thx @Fofanko for the PR!
I'm still not convinced this is the actual problem, but it still does get rid of generator magic so I'm merging it in.

sl0thentr0py:
fyi @Fofanko I'm reverting this since it caused a regression and trying to fix the problem properly in https://github.com/getsentry/sentry-python/pull/1388.

sl0thentr0py:
@Forty-Bot this is reverted anyway in one of the newer releases.

PR #1379, opened 2022-03-30
Update black

antonpirker:
.. and reformat code with new black.

### DIFF
--- a/tests/utils/test_general.py
+++ b/tests/utils/test_general.py
@@ -31,19 +31,23 @@ else:
     def test_safe_repr_never_broken_for_strings(x):
         r = safe_repr(x)
         assert isinstance(r, text_type)
-        assert u"broken repr" not in r
+        assert "broken repr" not in r
 
 
 def test_safe_repr_regressions():
+    # fmt: off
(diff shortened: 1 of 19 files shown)


tests/utils/test_general.py
```diff
@@ -31,19 +31,23 @@
     def test_safe_repr_never_broken_for_strings(x):
         r = safe_repr(x)
         assert isinstance(r, text_type)
-        assert u"broken repr" not in r
+        assert "broken repr" not in r
 
 
 def test_safe_repr_regressions():
+    # fmt: off
```
sl0thentr0py, line 38: why are these needed?

antonpirker: So that black does not remove the u"" from the strings and breaks our tests.

antonpirker: You can "turn off" black formatting for parts of the code.

antonpirker: strings containing only ascii characters do not need u"" in python 2.7. Only strings containing unicode chars need the u"". (hence "u") 

I think Python (2) did a really bad job in adding unicode support to the language

PR #1414, opened 2022-04-28
feat(extension): Replace original DSN host/port with localhost:3000

sl0thentr0py:
https://getsentry.atlassian.net/browse/WEBBACKEND-143

### DIFF
--- a/scripts/init_serverless_sdk.py
+++ b/scripts/init_serverless_sdk.py
@@ -18,14 +18,13 @@ if MYPY:
     from typing import Any
 
 
-# TODO(neel) maybe conditional on extension presence/layer version?
-# not sure best way to check
 def extension_relay_dsn(original_dsn):
     dsn = Dsn(original_dsn)
     dsn.host = "localhost"
     dsn.port = 3000
     dsn.scheme = "http"
     return str(dsn)
+
 
 # Configure Sentry SDK
 sentry_sdk.init(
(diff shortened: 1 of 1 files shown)


scripts/init_serverless_sdk.py
```diff
@@ -11,15 +11,25 @@
...
 import sentry_sdk
 from sentry_sdk._types import MYPY
+from sentry_sdk.utils import Dsn
 from sentry_sdk.integrations.aws_lambda import AwsLambdaIntegration
 
 if MYPY:
     from typing import Any
 
 
+# TODO(neel) maybe conditional on extension presence/layer version?
```
antonpirker, line 21: We should add a check if the extension is present and only in this case change the dsn.

Ways to do this:
- checking for the /opt/extensions folder (hackish)
- introduce env var that defaults to "extension present". this way we do not need to set it in existing lambda functions, we need to set it only for cases where there is the layer but no extension.  (should this case ever happen in the future?)

scripts/init_serverless_sdk.py
```diff
@@ -11,15 +11,25 @@
...
+# not sure best way to check
+def extension_relay_dsn(original_dsn):
+    dsn = Dsn(original_dsn)
+    dsn.host = "localhost"
+    dsn.port = 3000
+    dsn.scheme = "http"
+    return str(dsn)
```
antonpirker, line 28: 👍 💯

antonpirker (approved):

PR #1436, opened 2022-05-05
chore: Update logo in readme (again)

mattjohnsonpint:
In #1426, I updated the logo to use white or black based on theme.  While this works well for GitHub, it was pointed out to me that we also render these readme files on package registries, and some (including NPM and also PyPi) do not set their theme correctly.  (See https://github.com/getsentry/sentry-electron/pull/479#issuecomment-1118755203)

Updating to use the purple Sentry logo instead, which renders well on both white and black backgrounds, and will work for both GitHub and PyPi.

Sorry for the friction.

#skip-changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,7 @@
 <p align="center">
-    <a href="https://sentry.io/?utm_source=github&utm_medium=logo" target="_blank" align="center">
-        <picture>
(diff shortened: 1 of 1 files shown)


mattjohnsonpint:
See also https://github.com/getsentry/sentry/pull/34229 and https://github.com/getsentry/sentry-electron/pull/479#issuecomment-1118755203 for more detail.

sl0thentr0py (approved):

PR #1441, opened 2022-05-10
Starlette Integration

sl0thentr0py:


sentry_sdk/integrations/starlette.py
```diff
@@ -0,0 +1,414 @@
...
+    """
+    old_call = middleware_class.__call__
+
+    async def _sentry_authenticationmiddleware_call(self, scope, receive, send):
+        # type: (Dict[str, Any], Dict[str, Any], Callable[[], Awaitable[Dict[str, Any]]], Callable[[Dict[str, Any]], Awaitable[None]]) -> None
+        try:
+            await old_call(self, scope, receive, send)
+        except Exception as exc:
+            _add_user_to_sentry_scope(scope)
+            raise exc
```
sl0thentr0py, line 185: do we still need this since you added `_add_user_to_sentry_scope` to `patch_exception_middleware` too?

sl0thentr0py (commented):
just two questions for now, amazing work!

sentry_sdk/integrations/starlette.py
antonpirker: Oh, yes. I wanted to clean this up. As ExceptionMiddleware is always present, we do not need this here. Good catch!

antonpirker (approved):

mypy.ini
```diff
@@ -63,3 +63,7 @@ disallow_untyped_defs = False
 ignore_missing_imports = True
 [mypy-flask.signals]
 ignore_missing_imports = True
+[mypy-starlette.*]
+ignore_missing_imports = True
+[mypy-fastapi.*]
+ignore_missing_imports = True
```
tiangolo, line 69: Are these really needed? In particular the FastAPI one. If there are mypy errors in FastAPI I haven't seen I would like to get them fixed!

sentry_sdk/integrations/fastapi.py
```diff
@@ -0,0 +1,122 @@
...
+except ImportError:
+    raise DidNotEnable("FastAPI is not installed")
+
+try:
+    from starlette.types import ASGIApp, Receive, Scope, Send
+except ImportError:
+    raise DidNotEnable("Starlette is not installed")
+
+
+_DEFAULT_TRANSACTION_NAME = "generic FastApi request"
```
tiangolo, line 28: Just a tiny nitpick, but it would be cool if this used the spelling `FastAPI`, with the "API" part in upper case.

sentry_sdk/integrations/fastapi.py
```diff
@@ -0,0 +1,122 @@
...
+    def setup_once():
+        # type: () -> None
+        StarletteIntegration.setup_once()
+        patch_middlewares()
+
+
+def patch_middlewares():
+    # type: () -> None
+
+    old_build_middleware_stack = FastAPI.build_middleware_stack
```
tiangolo, line 44: The Starlette base middleware will probably change to fix a couple of issues, and this logic to build the middleware stack will probably change in the future. But I would expect the new version should be easier to patch too.

sentry_sdk/integrations/starlette.py
```diff
@@ -0,0 +1,459 @@
...
+    # type: (Dict[str, Any]) -> None
+    """
+    Extracts user information from the ASGI scope and
+    adds it to Sentry's scope.
+    """
+    if "user" not in scope:
+        return
+
+    if not _should_send_default_pii():
+        return
```
tiangolo, line 173: This is great for Starlette!

For FastAPI, the `AuthenticationMiddleware` is probably not commonly used because that is handled with dependencies, and there's no predefined expected/standardized user data/model.

It would probably make sense to explain in the docs how to manually and optionally record user data for FastAPI users in dependencies.

antonpirker:
Good feedback! I will look at your comments tomorrow!

sentry_sdk/integrations/fastapi.py
antonpirker: done

antonpirker: (in some other branch)

mypy.ini
antonpirker: Those are there because the integrations import stuff (aka fastapi) that is outside of the sentry sdk in the users projects.

And i basically just was to lacy to set mypy up correctly so these ignores are not necessary. But I will do so know :-)

antonpirker: done.

sentry_sdk/integrations/starlette.py
antonpirker: Good to know. What is the most used dependency for user handling?

We have something in the documentation on how to set the user: https://docs.sentry.io/platforms/python/enriching-events/identify-user/
I will make sure it lands in the FastAPI docs too!

sentry_sdk/integrations/fastapi.py
antonpirker: Good to know. (And yes, it was a bit cumbersome to instrument the middlewares :-) )
Any plans on when this will change?

tiangolo: Not sure yet because it will probably have to be fixed in Starlette. There are several people trying to solve it, I still have to help review all that too. And then port/update it in FastAPI

sentry_sdk/integrations/starlette.py
tiangolo: That sounds perfect!

---

There's not an expected predefined user data model or dependency, so every application can use the data model they find more useful. It normally doesn't require adopting a predefined model or any type of dependency, because installing, learning, and using some package would require the same amount of effort and code as implementing it directly.

The documentation explaining how to handle users is here: https://fastapi.tiangolo.com/tutorial/security/get-current-user/

But again, it's just explaining the concepts, not enforcing a data model.

---

I think that documenting how to handle PII in the developers' code, as you suggest with the existing docs, would probably be the best approach here too. Instead of trying to solve it behind the scenes as there's nothing obvious to expect.

sentry_sdk/integrations/fastapi.py
tiangolo: Thanks!

sentry_sdk/integrations/fastapi.py
Kludex: > The Starlette base middleware will probably change to fix a couple of issues

Which issues?

Kludex:
I don't know how to say this without sounding pedantic, but... Is it possible to change the name of the integration to `FastAPIIntegration` instead of `FastApiIntegration`? The name of the framework is FastAPI.

PR #1450, opened 2022-05-24
Serverless V2

antonpirker:


antonpirker:
Hey @getsentry/open-source!

We use certifi to validate SSL certificates in this package. Probably since forever. But the license bot does not like it.
Here is the license in question: https://github.com/certifi/python-certifi/blob/master/LICENSE

How should we continue with this?

vladanpaunovic (approved):

PR #1486, opened 2022-07-08
Profiling new data format

szokeasaurusrex:
Changed the output data format for the profiler, and added ability for the profiler to send profiles to Sentry

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -402,7 +402,7 @@ class _Client(object):
 
             if is_transaction:
                 if "profile" in event_opt:
-                    envelope.add_profile(event_opt.pop("profile")) 
+                    envelope.add_profile(event_opt.pop("profile"))
                 envelope.add_transaction(event_opt)
             else:
                 envelope.add_event(event_opt)
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -402,8 +402,7 @@ class _Client(object):
 
             if is_transaction:
                 if "profile" in event_opt:
-                    envelope.add_profile(event_opt["profile"])
-                    del event_opt["profile"]
+                    envelope.add_profile(event_opt.pop("profile")) 
                 envelope.add_transaction(event_opt)
             else:
                 envelope.add_event(event_opt)
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -391,6 +391,9 @@ class _Client(object):
             envelope = Envelope(headers=headers)
 
             if is_transaction:
+                if "profile" in event_opt:
+                    envelope.add_profile(event_opt["profile"])
+                    del event_opt["profile"]
                 envelope.add_transaction(event_opt)
             else:
                 envelope.add_event(event_opt)
--- a/sentry_sdk/profiler.py
+++ b/sentry_sdk/profiler.py
@@ -26,7 +26,7 @@ def nanosecond_time():
 class FrameData:
     def __init__(self, frame):
         self.function_name = frame.f_code.co_name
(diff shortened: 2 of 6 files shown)


sl0thentr0py (approved):

sentry_sdk/tracing.py
```diff
@@ -606,6 +607,9 @@ def finish(self, hub=None):
             "spans": finished_spans,
         }
 
+        if hub.client.options["enable_profiling"]:
+            event["profile"] = self._profile.to_json()
```
Zylphrex, line 611: @szokeasaurusrex The profile should not be sent in the same envelope item as the transaction. The cocoa/android sdks currently create a separate envelope item containing just the profile and send that in the same envelope. One reason for this is in case we decide to separate the transaction and profile in separate envelopes.

sentry_sdk/tracing.py
```diff
@@ -488,6 +488,7 @@ class Transaction(Span):
         # tracestate data from other vendors, of the form `dogs=yes,cats=maybe`
         "_third_party_tracestate",
         "_measurements",
+        "_profile",
```
sl0thentr0py, line 491: pls add a 
```
self._profile = None  # type: Optional[Sampler]
``` 
inside `__init__` below for completeness.

sentry_sdk/tracing.py
```diff
@@ -606,6 +607,12 @@ def finish(self, hub=None):
             "spans": finished_spans,
         }
 
+        if hub.client.options["enable_profiling"]:
```
sl0thentr0py, line 610: remove the `except` below and just add a condition here
```suggestion
        if hub.client.options["enable_profiling"] and self._profile:
```

sentry_sdk/client.py
```diff
@@ -391,6 +391,9 @@ def capture_event(
             envelope = Envelope(headers=headers)
 
             if is_transaction:
+                if "profile" in event_opt:
+                    envelope.add_profile(event_opt["profile"])
```
sl0thentr0py, line 395:
```suggestion
                    envelope.add_profile(event_opt.pop("profile"))
```
and remove the `del`

sl0thentr0py (approved):
just some style changes and one question, otherwise looks good!

sl0thentr0py (changes requested):
sry picked the wrong option earlier, didn't mean to approve yet

sentry_sdk/profiler.py
```diff
@@ -120,9 +116,22 @@ def _sample(self, _, frame):
...
         """
-        return json.dumps(self, cls=self.JSONEncoder)
+        thread_id = threading.get_ident()
+        return {
+            'samples': [{
+                'frames': sample.stack,
+                'relative_timestamp_ns': sample.sample_time,
+                'thread_id': thread_id
+            } for sample in self.stack_samples],
+            'frames': [{
```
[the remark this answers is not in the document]
szokeasaurusrex: We will be adding a `frames` field to the relay since it is needed for Python

sentry_sdk/tracing.py
szokeasaurusrex: I changed this to `and self._profile is not None` because I thought that was more readable; is that OK?

sl0thentr0py: yes that's fine

sl0thentr0py (approved):
also lgtm, you can merge it into the feature branch if you want

sentry_sdk/profiler.py
```diff
@@ -114,9 +110,22 @@ def _sample(self, _, frame):
...
```
phacops, line 123: We don't need to have the frame list here since it's already captured in the samples.

Issue #1489, opened 2022-07-11
feat(transcations) Transaction source


smeubank:
### Problem Statement

Add the `source` of a transaction name to all events sent with the SDK. 
See https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations

### To do:

Integrations where we have to check/add transaction source:
- [x] asgi (on event, in event processor, has code for saving url style transaction names, but can this only if fastapi is used...)
- [x] flask (on scope, outside event processor)
- [x] django (on scope, outside event processor and multiple other places, TODO later: test for "/message/:id" names)
- [x] bottle (on event, in event processor)
- [x] falcon (on event, in event processor, TODO: there is no endpoint transaction_style (create ticket for later?))
- [x] pyramid (on scope, outside event processor)
- [x] quart (on scope, outside event processor)
- [x] sanic
- [x] celery (TODO: update TESTS! rest is done)
- [x] wsgi (does not set transaction names (or sources) what I think should also be that way in ASGI...)
- [x] spark (does not have transactions)
- [x] tornado (on event, in event processor)
- [x] trytond (does not have transactions)
- [x] aiohttp (on scope, outside event processor)
- [x] starlette (needs to use transaction_from_function() to have better endpoint names)
- [x] aws_lambda (need to check tests in CI because did not run locally because I am too lazy to setup env vars to actual AWS to run tests.)
- [x] gcp
- [x] chalice (on scope, outside event processor)

### Reference

[Develop Spec: Transaction Annotations](https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations)

https://github.com/getsentry/sentry-javascript/pull/5367

https://github.com/getsentry/sentry-javascript/issues/5345

https://github.com/getsentry/relay/pull/1330

https://github.com/getsentry/develop/pull/624

PR #1490, opened 2022-07-11
feat(transactions): Transaction Source

antonpirker:
This adds the `transaction_info.source` to events. See https://github.com/getsentry/sentry-python/issues/1489

### DIFF
--- a/sentry_sdk/integrations/asgi.py
+++ b/sentry_sdk/integrations/asgi.py
@@ -13,9 +13,8 @@ from sentry_sdk._types import MYPY
 from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk.integrations._wsgi_common import _filter_headers
 from sentry_sdk.sessions import auto_session_tracking
+from sentry_sdk.tracing import SOURCE_FOR_STYLE, TRANSACTION_SOURCE_UNKNOWN
 from sentry_sdk.utils import (
-    SOURCE_FOR_STYLE,
-    TRANSACTION_SOURCE_UNKNOWN,
     ContextVar,
     event_from_exception,
     transaction_from_function,
--- a/sentry_sdk/integrations/asgi.py
+++ b/sentry_sdk/integrations/asgi.py
@@ -149,7 +149,7 @@ class SentryAsgiMiddleware:
                         transaction = Transaction(op="asgi.server")
 
                     transaction.name = _DEFAULT_TRANSACTION_NAME
-                    transaction.source + TRANSACTION_SOURCE_UNKNOWN
+                    transaction.source = TRANSACTION_SOURCE_UNKNOWN
                     transaction.set_tag("asgi.type", ty)
 
                     with hub.start_transaction(
@@ -218,7 +218,7 @@ class SentryAsgiMiddleware:
             # done, which is sometime after the request has started. If we have
             # an endpoint, overwrite our generic transaction name.
             if endpoint:
-                name = transaction_from_function(endpoint)
+                name = transaction_from_function(endpoint) or ""
 
         elif transaction_style == "url":
             # FastAPI includes the route object in the scope to let Sentry extract the
--- a/sentry_sdk/integrations/asgi.py
+++ b/sentry_sdk/integrations/asgi.py
@@ -14,8 +14,7 @@ from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk.integrations._wsgi_common import _filter_headers
 from sentry_sdk.sessions import auto_session_tracking
 from sentry_sdk.utils import (
-    TRANSACTION_SOURCE_COMPONENT,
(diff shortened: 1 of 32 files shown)


sentry_sdk/integrations/asgi.py
```diff
@@ -147,6 +149,7 @@ async def _run_app(self, scope, callback):
                         transaction = Transaction(op="asgi.server")
 
                     transaction.name = _DEFAULT_TRANSACTION_NAME
+                    transaction.source = TRANSACTION_SOURCE_UNKNOWN
```
sl0thentr0py, line 152: so I don't think we should mark the default 'generic..' name as unknown because relay will then drop all those. The default name is low cardinality anyway since it's always the same value.

(following up in #client-infra)

sentry_sdk/integrations/bottle.py
```diff
@@ -176,24 +177,32 @@ def size_of_file(self, file):
         return file.content_length
 
 
+def _set_transaction_name_and_source(event, transaction_style, request):
+    # type: (Event, str, Any) -> Event
+    name_for_style = {
```
sl0thentr0py, line 182: why did you choose this hash style, you now call `transaction_from_function` in some cases even when it's not necessary.

sentry_sdk/integrations/django/__init__.py
```diff
@@ -319,6 +320,34 @@ def _patch_django_asgi_handler():
     patch_django_asgi_handler_impl(ASGIHandler)
 
 
+def _set_transaction_name_and_source(scope, transaction_style, request):
```
sl0thentr0py, line 323: same comment above about not returning scope since it's mutated anyway

sentry_sdk/integrations/asgi.py
sl0thentr0py: change to `route` as discussed on slack

sentry_sdk/integrations/django/__init__.py
```diff
@@ -438,7 +454,7 @@ def _got_request_exception(request=None, **kwargs):
 
         if request is not None and integration.transaction_style == "url":
             with hub.configure_scope() as scope:
-                _attempt_resolve_again(request, scope)
+                _attempt_resolve_again(request, scope, integration.transaction_style)
```
sl0thentr0py, line 457: this only runs in the `transaction_style="url"` case so not sure you need the extra argument here

sl0thentr0py (changes requested):
some stuff to start with

sentry_sdk/tracing.py
```diff
@@ -513,6 +514,7 @@ class Transaction(Span):
     def __init__(
         self,
         name="",  # type: str
+        source=TRANSACTION_SOURCE_UNKNOWN,  # type: str
```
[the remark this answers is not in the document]
antonpirker: right!

sentry_sdk/integrations/asgi.py
antonpirker: ok

sentry_sdk/integrations/asgi.py
```diff
@@ -213,6 +200,46 @@ def event_processor(self, event, hint, asgi_scope):
     # data to your liking it's recommended to use the `before_send` callback
     # for that.
 
+    def _set_transaction_name_and_source(self, event, transaction_style, asgi_scope):
```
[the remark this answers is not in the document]
antonpirker: I just wanted to make it extra clear that the event is manipulated. (and also make it a bit looking like functional programming) But I guess it is python, so you are right :-)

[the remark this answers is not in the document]
antonpirker: done

sentry_sdk/integrations/django/__init__.py
antonpirker: done

sentry_sdk/integrations/bottle.py
antonpirker: Just for readability. I like this hash style. But performance matters, so I changed it. (and it is still readable)

sentry_sdk/integrations/django/__init__.py
antonpirker: I rather keep the argument so `_attempt_resolve_again` is not assuming outside state.

sentry_sdk/integrations/asgi.py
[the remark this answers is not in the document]
sl0thentr0py: so as a thumbrule in dynamic languages, I abide by the following pattern.
If I take some input and am returning something, I try to make the function as immutable as possible, so if I'm returning the same object it needs to be copied and not mutated.
If the function returns nothing anyway, then it's clear it does something which has side effects and then it can mutate some of it's inputs.

sl0thentr0py (approved):
thanks for making the changes!
shippit

PR #1504, opened 2022-07-15
Fast tests

antonpirker:
The first iteration of making our test suite faster. This PR runs our test suite using `tox --parallel auto` to use all CPU cores for running the tests. This brings the duration down from almost 40 minutes to around 13 minutes.

One of our tests (`test_redis_backend_trace_propagation`) needed to be disabled because it hung when run in parallel. We will fix this test at a later stage in the process of optimizing our test suite.


tests/integrations/celery/test_celery.py
```diff
@@ -313,56 +313,57 @@ def dummy_task(self):
         assert e["type"] == "ZeroDivisionError"
 
 
-@pytest.mark.forked
-def test_redis_backend_trace_propagation(init_celery, capture_events_forksafe, tmpdir):
-    celery = init_celery(traces_sample_rate=1.0, backend="redis", debug=True)
+# TODO: This test is hanging when running test with `tox --parallel auto`. Find out why and fix it
```
sl0thentr0py, line 316: can you use 
https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-skip

sl0thentr0py (changes requested):

tests/integrations/celery/test_celery.py
antonpirker: Done

sl0thentr0py (approved):

PR #1507, opened 2022-07-18
fix: avoid sending empty Baggage header

intgr:
According to W3C Working Draft spec, the `Baggage` header must contain at least one value; header with empty value string is invalid.


sl0thentr0py (approved):
thx

PR #1508, opened 2022-07-18
fix: properly freeze Baggage object

intgr:
Clearly the intent was to call the function, but previously the `freeze` method was dead code.

I have not verified that this actually works correctly, however.


sl0thentr0py:
thx @intgr

Issue #1515, opened 2022-07-22
Starlette integration requires python-multipart


matyasrichter:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.8.0

### Steps to Reproduce

Using the Starlette/FastAPI integration while `python-multipart` isn't installed raises an exception. It's an optional dependency for Starlette, but not optional for sentry-sdk.
```python
sentry_sdk.init(
    dsn=settings.SENTRY_DSN,
    integrations=[StarletteIntegration(), FastApiIntegration()],
)
```

This can be fixed by gathering form data conditionally.

### Expected Result

No exception.

### Actual Result

Sending any request, even a GET, raises an exceptions, here's the relevant part of the stacktrace:
```
.venv/lib/python3.10/site-packages/fastapi/applications.py:269: in __call__
    await super().__call__(scope, receive, send)
.venv/lib/python3.10/site-packages/sentry_sdk/integrations/starlette.py:268: in _sentry_patched_asgi_app
    return await middleware(scope, receive, send)
.venv/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py:126: in _run_asgi3
    return await self._run_app(scope, lambda: self.app(scope, receive, send))
.venv/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py:176: in _run_app
    raise exc from None
.venv/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py:171: in _run_app
    return await callback()
.venv/lib/python3.10/site-packages/sentry_sdk/integrations/starlette.py:268: in _sentry_patched_asgi_app
    return await middleware(scope, receive, send)
.venv/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py:126: in _run_asgi3
    return await self._run_app(scope, lambda: self.app(scope, receive, send))
.venv/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py:137: in _run_app
    raise exc from None
.venv/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py:134: in _run_app
    return await callback()
.venv/lib/python3.10/site-packages/starlette/applications.py:124: in __call__
    await self.middleware_stack(scope, receive, send)
.venv/lib/python3.10/site-packages/sentry_sdk/integrations/fastapi.py:122: in __call__
    await self.app(scope, receive, send)
[...]

PR #1516, opened 2022-07-22
Fixed problem with broken response

antonpirker:


sl0thentr0py (approved):

antonpirker:
This should now fix #1515

sentry_sdk/integrations/starlette.py
```diff
@@ -339,7 +339,13 @@ async def form(self):
         curl -X POST http://localhost:8000/upload/somethign -H "Content-Type: application/x-www-form-urlencoded" -d "username=kevin&password=welcome123"
         curl -X POST http://localhost:8000/upload/somethign  -F username=Julian -F password=hello123
         """
-        return await self.request.form()
+        try:
```
sl0thentr0py, line 342: in general in lots of other places, we try to import stuff and fail there, but up to you however you prefer to do it
https://github.com/getsentry/sentry-python/blob/9e69fcf61a537ae30ea687f5b55957b6f3441935/sentry_sdk/integrations/flask.py#L23-L26

antonpirker: Ok. Done.

sentry_sdk/integrations/starlette.py
```diff
@@ -339,6 +338,12 @@ async def form(self):
         curl -X POST http://localhost:8000/upload/somethign -H "Content-Type: application/x-www-form-urlencoded" -d "username=kevin&password=welcome123"
         curl -X POST http://localhost:8000/upload/somethign  -F username=Julian -F password=hello123
         """
+        try:
+            # Optional dependency of Starlette to parse form data.
+            import multipart  # type: ignore # noqa: F401
```
sl0thentr0py, line 343: I'd do at the top of the file
```python
try:
    import multipart
except ImportError:
    multipart = None
```

and then
```
if multipart is None:
    return None
```
in the function

antonpirker: nice!

sentry_sdk/integrations/starlette.py
```diff
@@ -423,6 +423,7 @@ async def __call__(self, scope, receive, send):
         hub = Hub.current
         integration = hub.get_integration(StarletteIntegration)
         if integration is None:
+            await self.app(scope, receive, send)
```
axyjo, line 434: Thanks for this -- v1.8.0 is ~unusable with POST requests Starlette without this change

PR #1520, opened 2022-07-25
Update Flask and Quart integrations

pgjones:
Flask and Quart are deprecating and removing the ``_xxx_ctx_stack``s
and adopting a more direct usage of ContextVars. The previous code
will therefore break for the latest version of Quart and start to warn
for Flask and then break.

This fix should work with any version of Flask or Quart, and hence is
a more robust version. There is an extra indirection, however I don't
think this is on any hot path.


pgjones:
There is a test failure `test_cli_commands_raise` that isn't related to this change, but I think may be related to changes in click.

sentry_sdk/integrations/quart.py
```diff
@@ -27,11 +27,12 @@
 
 try:
     from quart import (  # type: ignore
+        has_request_context,
```
antonpirker, line 30: This does not work with Quart 0.16.1 (the first version we support). We need to be backwards compatible

Could you change the imports to something like this: 
https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/starlette.py#L43-L47

Where you then down on line 112 can check do something like:
```
if has_request_context is None:
    # old way of getting request
else:
    # new way of getting request
```

antonpirker (changes requested):

antonpirker:
Hey @pgjones !
Thanks for your PR, great work!

I have one request for making it backwards compatible. Otherwise it looks good!

sentry_sdk/integrations/quart.py
pgjones: Are you sure it doesn't work with 0.16.1, `has_request_context` is present in that release, https://github.com/pallets/quart/blob/0.16.1/src/quart/__init__.py#L15

antonpirker: But it can not be imported from quart as in master, because the "as xxx" is missing: https://github.com/pallets/quart/blob/main/src/quart/__init__.py#L15

pgjones: Sorry, I'm not following. In [0.16.1 an `__all__`](https://github.com/pallets/quart/blob/0.16.1/src/quart/__init__.py#L87) is used rather than `as xxx`. It is only the latest, 0.18.0 that uses the `as xxx` syntax.

I've just installed 0.16.1 and can import it.

antonpirker: Hey @pgjones !
I have now checked it again, and you are right. It works...
I don't know why it did not work when I was testing it the first time. 
So I guess this is fine!

antonpirker (approved):

antonpirker:
Hey again @pgjones! If you send me your shipping address to anton |dot| pirker \at\ sentry [dot] io, I can send you a couple of stickers as a small token of appreciation.

Issue #1514, opened 2022-07-21
FastAPI Integration - TypeError: 'coroutine' object is not callable


danilogco:
### How do you use Sentry?

Self-hosted/on-premise

### Version

1.8.0

### Steps to Reproduce

1. Install the lib v1.8.0;
2. Our Sentry server is self-hosted;
3. Configure the sentry inicialization as described in the release documentation:
```python
from fastapi import FastAPI
from sentry_sdk.integrations.starlette import StarletteIntegration
from sentry_sdk.integrations.fastapi import FastApiIntegration
sentry_sdk.init(
dsn="...",
integrations=[StarletteIntegration(), FastApiIntegration()],
)
app = FastAPI()
```
4. So I did purposefully a request that would give me a 404/500 to activate the sentry warning.

Thanks in advance.

### Expected Result

Sentry 404 log sent successfully.

### Actual Result

web-1        | /usr/local/lib/python3.10/site-packages/sentry_sdk/serializer.py:216: RuntimeWarning: coroutine 'request_validation_exception_handler' was never awaited
web-1        |   with memo.memoize(obj) as result:
web-1        | RuntimeWarning: Enable tracemalloc to get the object allocation traceback
web-1        | ERROR:    Exception in ASGI application
web-1        | Traceback (most recent call last):
web-1        |   File "/usr/local/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
web-1        |     result = await app(self.scope, self.receive, self.send)
web-1        |   File "/usr/local/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
web-1        |     return await self.app(scope, receive, send)
web-1        |   File "/usr/local/lib/python3.10/site-packages/fastapi/applications.py", line 269, in __call__
web-1        |     await super().__call__(scope, receive, send)
web-1        |   File "/usr/local/lib/python3.10/site-packages/sentry_sdk/integrations/starlette.py", line 268, in _sentry_patched_asgi_app
web-1        |     return await middleware(scope, receive, send)
web-1        |   File "/usr/local/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py", line 126, in _run_asgi3
web-1        |     return await self._run_app(scope, lambda: self.app(scope, receive, send))
web-1        |   File "/usr/local/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py", line 176, in _run_app
[...]

PR #1532, opened 2022-08-02
Fix FastAPI issues

antonpirker:
This issue is a collection of bug fixes for the few bugs that where discovered from the FastAPI alpha release.

* Patch FastAPI `get_request_handler` to extract request body data (fixes the infinite loop when doing form posts in FastAPI) #1514
* Fix patching exception handlers in `_sentry_patched_exception_handler`(fixes the error on non existent routes) #1514
* Made Starlette middleware patching idempotent to prevent patching functions twice (no GitHub issue for that)
* Patch `request_response` in Starlette to extract request body data  (fixes the infinite loop when doing form posts in Starlette) #1514
* Starlette has async and sync "views" (or route handlers) in the sync ones it is not easily possible to get the request body data (because those functions are async) so we just get the cookies form the sync route handlers and not the body data. (no GitHub issue for that)


sentry_sdk/integrations/fastapi.py
```diff
@@ -82,6 +83,60 @@ def _set_transaction_name_and_source(event, transaction_style, request):
     event["transaction_info"] = {"source": SOURCE_FOR_STYLE[transaction_style]}
 
 
+def patch_get_request_handler():
```
sl0thentr0py, line 86: if we patch the request handler, do we need `SentryFastApiMiddleware` anymore?

sl0thentr0py: same question for `SentryStarletteMiddleware`

sl0thentr0py (commented):
question before detailed review

sentry_sdk/integrations/fastapi.py
antonpirker: Right now the `SentryFastApiMiddleware` is only setting the transaction name and source. But I guess we can move this into the request handler and get rid of the middleware!

sl0thentr0py: yea I'd prefer only one place to set transaction name/source.

antonpirker: There were some differences in Starlette and FastAPI. Can't remember exactly what, but something with routing or handling the routes. So I think we need to set the transaction name and source in both integrations to make sure we have the correct one in FastAPI projects. 
But we can get rid of the `SentryFastApiMiddleware`.

sl0thentr0py: yes I meant one place per Starlette/Fastapi integration.
So I think we can get rid of both `SentryFastApiMiddleware` and `SentryStarletteMiddleware` in theory.

MrSalman333:
when will this be released? 

thanks for the great work

antonpirker:
We are working on the final touches. Should not take that long anymore.

sentry_sdk/integrations/fastapi.py
antonpirker: I have now cleaned up the code and the transaction name/source are only set once in fastapi and once in starlette integration.

sl0thentr0py (approved):

PR #1538, opened 2022-08-04
Replace Travis CI badge with GitHub Actions badge

153957:
Travis CI is no longer being used and has been replaced by GitHub Actions.
The ci.yml workflow contains the main tests, so show that status in the README.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,8 +95,6 @@ jobs:
           PGHOST: localhost
           PGPASSWORD: sentry
         run: |
-          psql -c 'create database travis_ci_test;' -U postgres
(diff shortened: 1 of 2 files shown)


sl0thentr0py (approved):
thx @153957

153957:
@sl0thentr0py there is still a 'reference' to travis in the actual ci.yml file, creating postgres databases, maybe those should also be renamed?

https://github.com/getsentry/sentry-python/blob/master/.github/workflows/ci.yml#L98-L99

sl0thentr0py:
hmm yea don't see where they're used, can you remove them? let's see if CI passes. Thanks again!

PR #1543, opened 2022-08-05
feat(redis): Add instrumentation for redis pipeline

jjbayer:
Add automatic instrumentation of [redis pipelining](https://redis.io/docs/manual/pipelining/) for both `redis` and `rediscluster`. 

Note: This does not add instrumentation for [`StrictRedisCluster`](https://github.com/Grokzen/redis-py-cluster/blob/master/docs/release-notes.rst#200-aug-12-2019).

### DIFF
--- a/sentry_sdk/integrations/redis.py
+++ b/sentry_sdk/integrations/redis.py
@@ -28,7 +28,11 @@ def patch_redis_pipeline(pipeline_cls, parse_command_fn):
 
         with hub.start_span(op="redis", description="redis.pipeline.execute") as span:
             with capture_internal_exceptions():
-                print(self.command_stack)
+                transaction = (
+                    self.transaction if pipeline_cls.__name__ == "Pipeline" else False
+                )
+                span.set_tag("transaction", transaction)
+
                 commands = [parse_command_fn(c) for c in self.command_stack[:10]]
                 if len(self.command_stack) > 10:
                     commands.append("...")
--- a/sentry_sdk/integrations/redis.py
+++ b/sentry_sdk/integrations/redis.py
@@ -109,6 +109,9 @@ class RedisIntegration(Integration):
 
         patch_redis_client(redis.StrictRedis, is_cluster=False)
         patch_redis_pipeline(redis.client.Pipeline, False, _get_redis_command_args)
+        patch_redis_pipeline(
+            redis.client.StrictPipeline, False, _get_redis_command_args
+        )
 
         try:
             import rb.clients  # type: ignore
--- a/sentry_sdk/integrations/redis.py
+++ b/sentry_sdk/integrations/redis.py
@@ -66,6 +66,14 @@ def _parse_rediscluster_command(command):
     return command.args
 
 
+def _get_pipeline_cls(rediscluster):
+    # type: (Any) -> Any
+    try:
+        return rediscluster.ClusterPipeline
+    except AttributeError:
+        return rediscluster.StrictClusterPipeline
+
+
 def _patch_rediscluster():
     # type: () -> None
     try:
@@ -75,7 +83,7 @@ def _patch_rediscluster():
 
     patch_redis_client(rediscluster.RedisCluster, is_cluster=True)
     patch_redis_pipeline(
-        rediscluster.ClusterPipeline, True, _parse_rediscluster_command
+        _get_pipeline_cls(rediscluster), True, _parse_rediscluster_command
     )
 
     # up to v1.3.6, __version__ attribute is a tuple
--- a/sentry_sdk/integrations/redis.py
+++ b/sentry_sdk/integrations/redis.py
@@ -66,7 +66,7 @@ def _parse_rediscluster_command(command):
     return command.args
 
 
-def _get_pipeline_cls(rediscluster):
+def _get_cluster_pipeline_cls(rediscluster):
     # type: (Any) -> Any
     try:
         return rediscluster.ClusterPipeline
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/redis.py
```diff
@@ -15,6 +15,43 @@
...
+        if hub.get_integration(RedisIntegration) is None:
+            return old_execute(self, *args, **kwargs)
+
+        with hub.start_span(op="redis", description="redis.pipeline.execute") as span:
+            with capture_internal_exceptions():
+                print(self.command_stack)
+                commands = [parse_command_fn(c) for c in self.command_stack[:10]]
+                if len(self.command_stack) > 10:
+                    commands.append("...")
+                span.set_data(
```
beezz, line 35: nit: Expose if the pipeline is transactional or not (the transaction arg ~ https://redis.readthedocs.io/en/stable/connections.html#redis.Redis.pipeline).

sentry_sdk/integrations/redis.py
```diff
@@ -15,6 +15,43 @@
...
+                )
+
+            return old_execute(self, *args, **kwargs)
+
+    pipeline_cls.execute = sentry_patched_execute
+
+
+def _parse_redis_command(command):
+    # type: (Any) -> str
+    return " ".join(map(str, command[0]))
```
untitaker, line 47: i would align this with the code at the bottom of this file. there we use `repr` instead of `str` and no list comprehension/slicing. Perhaps you find a way to factor this out.

sentry_sdk/integrations/redis.py
```diff
@@ -15,6 +15,43 @@
...
+            return old_execute(self, *args, **kwargs)
+
+        with hub.start_span(op="redis", description="redis.pipeline.execute") as span:
+            with capture_internal_exceptions():
+                print(self.command_stack)
+                commands = [parse_command_fn(c) for c in self.command_stack[:10]]
+                if len(self.command_stack) > 10:
+                    commands.append("...")
+                span.set_data(
+                    "commands",
```
untitaker, line 36: this should be `redis.commands` for consistency with how normal commands are parsed and stored

sentry_sdk/integrations/redis.py
jjbayer: Done!

sentry_sdk/integrations/redis.py
jjbayer: Done. I factored out the constant.

sentry_sdk/integrations/redis.py
jjbayer: 👍 Done.

sentry_sdk/integrations/redis.py
```diff
@@ -7,13 +7,64 @@
...
+                commands = []
+                for i, arg in enumerate(self.command_stack):
+                    if i > _MAX_NUM_ARGS:
+                        break
+                    command_args = []
+                    for j, command_arg in enumerate(get_command_args_fn(arg)):
+                        if j > 0:
+                            command_arg = repr(command_arg)
+                        command_args.append(command_arg)
+                    commands.append(" ".join(command_args))
```
jjbayer, line 47: SDK folks: What prevents us from using list comprehensions and/or slicing for this kind of code? Do we support Python < 2.0?

sentry_sdk/integrations/redis.py
```diff
@@ -45,25 +99,26 @@ def setup_once():
...
             import rb.clients  # type: ignore
         except ImportError:
             pass
         else:
-            patch_redis_client(rb.clients.FanoutClient)
-            patch_redis_client(rb.clients.MappingClient)
-            patch_redis_client(rb.clients.RoutingClient)
+            patch_redis_client(rb.clients.FanoutClient, is_cluster=False)
+            patch_redis_client(rb.clients.MappingClient, is_cluster=False)
+            patch_redis_client(rb.clients.RoutingClient, is_cluster=False)
```
jjbayer, line 112: Not sure whether Redis Blaster should be considered cluster or not.

sentry_sdk/integrations/redis.py
untitaker: I suspect this was my code: It's a performance thing. Don't want to allocate unnecessarily large intermediate lists

sentry_sdk/integrations/redis.py
untitaker: blaster is legacy, i wouldn't care

sentry_sdk/integrations/redis.py
```diff
@@ -7,13 +7,72 @@
...
+    # type: (Any) -> Sequence[Any]
+    return command[0]
+
+
+def _parse_rediscluster_command(command):
+    # type: (Any) -> Sequence[Any]
+    return command.args
+
+
+def _get_cluster_pipeline_cls(rediscluster):
```
sl0thentr0py, line 69: since the `Strict..` renaming is part of the same [2.0.0 change](https://github.com/Grokzen/redis-py-cluster/blob/8a8102a9d758d61a7ec1e2ac9050fcd34029ff3f/docs/release-notes.rst#200-aug-12-2019), can we move this under the check at line 95 too?

tests/integrations/rediscluster/test_rediscluster.py
```diff
@@ -12,6 +13,11 @@
 
 @pytest.fixture(scope="module", autouse=True)
 def monkeypatch_rediscluster_classes():
+    pipeline_cls = _get_cluster_pipeline_cls(rediscluster)
```
sl0thentr0py, line 16: just a nit, but why do we need to patch pipeline here? we can use the same `Strict` checking logic above to also decide which of the classes to patch and then just patch `execute/execute_command`?

sl0thentr0py (commented):
thx @jjbayer cool stuff!
just a question / nit about the `Strict..` renaming mess

sentry_sdk/integrations/redis.py
jjbayer: Done! The subtle difference between `..Cluster` and `..Pipeline` is that `RedisCluster` and `StrictRedisCluster` can exist simultaneously, while `ClusterPipeline` and `StrictClusterPipeline` are mutually exclusive.

sl0thentr0py (approved):
thx, lgtm!

PR #1548, opened 2022-08-05
Handle no release when uploading profiles

szokeasaurusrex:
We noticed that an error occurs when a profile is being uploaded and there is no release set (the error can only occur when profiling is enabled). This PR stops the error from occurring.

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -403,10 +403,7 @@ class _Client(object):
             if is_transaction:
                 if "profile" in event_opt:
                     event_opt["profile"]["transaction_id"] = event_opt["event_id"]
-                    try:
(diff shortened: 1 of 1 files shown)


sentry_sdk/client.py
```diff
@@ -403,7 +403,10 @@ def capture_event(
             if is_transaction:
                 if "profile" in event_opt:
                     event_opt["profile"]["transaction_id"] = event_opt["event_id"]
-                    event_opt["profile"]["version_name"] = event_opt["release"]
+                    try:
```
sl0thentr0py, line 406:
```suggestion
                        event_opt["profile"]["version_name"] = event_opt.get("release", "")
```

sl0thentr0py (changes requested):

sl0thentr0py (approved):

PR #1558, opened 2022-08-11
Remove TRANSACTION_SOURCE_UNKNOWN and default to CUSTOM

### LINKED ISSUE #1557: Transaction source custom semantics not correct
We should get rid of `unknown` as a source completely since this is never supposed to be set by an SDK, only by relay as an inference in absence of other sources.
Also default to `custom` as the default in `start_transaction` / set transaction name helpers, and make sure there are no other cases where it ends up in a null/undefined state.

sl0thentr0py:
Fixes #1557 
see https://github.com/getsentry/develop/pull/667

`unknown` is only supposed to be inferred by relay as a default and not
set by any SDKs.
Additionally, fix some of the other cases where `start_transaction` was
being called without a source in integrations.

### DIFF
--- a/sentry_sdk/integrations/aiohttp.py
+++ b/sentry_sdk/integrations/aiohttp.py
@@ -9,7 +9,7 @@ from sentry_sdk.integrations._wsgi_common import (
     _filter_headers,
     request_body_within_bounds,
 )
-from sentry_sdk.tracing import SOURCE_FOR_STYLE, Transaction
+from sentry_sdk.tracing import SOURCE_FOR_STYLE, Transaction, TRANSACTION_SOURCE_ROUTE
 from sentry_sdk.utils import (
     capture_internal_exceptions,
     event_from_exception,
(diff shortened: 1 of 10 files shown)


Lms24 (approved):
LGTM! Just a question: Are we sure that whenever we set `TRANSACTION_SOURCE_ROUTE`, the transaction name is actually a route and not some sort of unparameterized URL? Just asking because I'm probably not up to date w/ the latest status of high-cardinality transaction names in Pythonland.

sl0thentr0py:
@Lms24 yea I manually checked all the cases, `route` is here only for the `Generic WSGI..` stuff which is also not exactly a route but just the default name.
Later in other integrations, we do the actual resolving and set the correct source. Examples
* [ASGI](https://github.com/getsentry/sentry-python/blob/6d242c88e7baf253bca36753b576d58403141c23/sentry_sdk/integrations/asgi.py#L240-L246)
* [Django](https://github.com/getsentry/sentry-python/blob/6d242c88e7baf253bca36753b576d58403141c23/sentry_sdk/integrations/django/__init__.py#L339-L343)

Lms24:
@sl0thentr0py perfect, thanks!

PR #1570, opened 2022-08-16
feat(profiling): Support for multithreaded profiles

Zylphrex:
A signal handler can only be installed on the main thread, this was the reason
why we could not use signals to profile multithreaded programs. This change
installs the signal handler during sdk initialization which should happen on the
main thread. The timers are still started on the individual threads to allow for
profiles being recorded from different threads.


sentry_sdk/profiler.py
```diff
@@ -13,200 +13,333 @@
...
-    def frame_list(self):
-        # type: () -> typing.List[FrameData]
-        # Build frame array from the frame indices
-        frames = [None] * len(self._frame_indices)  # type: typing.List[typing.Any]
-        for frame, index in self._frame_indices.items():
-            frames[index] = frame
-        return frames
+                for frame in stack:
+                    if frame not in frames:
+                        frames[frame] = len(frames)
```
sl0thentr0py, line 286: I couldn't follow what you're doing with the `frames` dict here.
Is it memoizing something? And you update the `frames` dict with the current length of the dict itself, or it that a typo?

sl0thentr0py (approved):
thx a lot @Zylphrex, this is so much cleaner now!
Just one question because I could not follow the logic there, but otherwise lgtm.
:shipit:

sentry_sdk/profiler.py
JonasBa: @sl0thentr0py this is format size optimization. You index the stack frames so that you convert each sample's `frame[]` stack to a `int[]` and store each frame in the dict. The len(frames) is the sequential id of each frame which is then resolved when we reconstruct the stack.

PR #1574, opened 2022-08-19
Add more version constraints

isra17:
For some reason, `poetry` will run the solver at least twice if python version are above 3.6, each with a different constraint for urllib3. This add a significant slowdown on our end in some project.


sl0thentr0py (approved):
thx @isra17

PR #1590, opened 2022-08-31
feat(pymongo): add PyMongo integration

### LINKED ISSUE #1585: pymongo integration
### Discussed in https://github.com/getsentry/sentry-python/discussions/1580

<div type='discussions-op-text'>

<sup>Originally posted by **abhishek17569** August 24, 2022</sup>
Hello All,
I'm using pymongo and performing multiple aggregation pipelines. 
I want to get the DB process spans to be shown in event breakdown.

Agalin:
Adds breadcrumbs and performance traces for PyMongo queries using an
official monitoring API. Integration is similar to the one available in
OpenTelemetry, tags set to values recommended for attributes by OT as
specified in `Span Operations` guidelines.

PyMongo version selection explanation:
* 3.1 - introduction of monitoring API. Only Python 2.7 and 3.6
supported.
* 3.12 - latest 3.x release, support for 2.7, 3.6-3.9 (3.7-3.9 added in
various minor releases between 3.1 and 3.12).
* 4.0 - no support for 2.7, added support for 3.10.
* 4.1 - no support for 3.6.0-3.6.1.
* 4.2 - no support for any 3.6.

Uses `MockupDB` for tests (Python implementation of MongoDB protocol server). `mongomock` is a reimplementation of the `MongoClient` and doesn't support tracing. This way we get real world like results. Note that pymongo has changed queries sent for some requests between 3.1 and 3.12 (probably 3.7 or 3.11) so additional responder is necessary to handle tests with 3.1.

There is no support for GridFS nor server connect/disconnect events. Those can be added in the future although I don't think there is an API similar to `pymongo.monitoring` for GridFS (so monkeypatching would be necessary in this one).

closes #1585

### DIFF
--- a/sentry_sdk/integrations/pymongo.py
+++ b/sentry_sdk/integrations/pymongo.py
@@ -1,0 +1,120 @@
+from __future__ import absolute_import
+
+from sentry_sdk import Hub
+from sentry_sdk.integrations import DidNotEnable, Integration
+from sentry_sdk.tracing import Span
+from sentry_sdk.utils import capture_internal_exceptions
+
+from sentry_sdk._types import MYPY
+
+try:
+    from pymongo import monitoring
+except ImportError:
+    raise DidNotEnable("Pymongo not installed")
+
+if MYPY:
+    from typing import Dict
+
+    from pymongo.monitoring import (
+        CommandFailedEvent,
+        CommandStartedEvent,
+        CommandSucceededEvent,
+    )
+
+
+class CommandTracer(monitoring.CommandListener):
+    def __init__(self):
+        self._ongoing_operations = {}  # type: Dict[int, Span]
+
+    def _operation_key(self, event):
+        # type: (CommandFailedEvent | CommandStartedEvent | CommandSucceededEvent) -> int
+        return event.request_id
+
+    def started(self, event):
+        # type: (CommandStartedEvent) -> None
+        hub = Hub.current
+        if hub.get_integration(PyMongoIntegration) is None:
+            return
+        with capture_internal_exceptions():
+            command = dict(event.command)
+
+            command.pop("$db", None)
+            command.pop("$clusterTime", None)
+            command.pop("$signature", None)
+
+            if event.command_name:
+                op = "db." + event.command_name
+            else:
+                op = "db"
+
+            tags = {
+                "db.name": event.database_name,
+                "db.system": "mongodb",
+                "db.operation": event.command_name,
+            }
+
+            try:
+                tags["net.peer.name"] = event.connection_id[0]
+                tags["net.peer.port"] = str(event.connection_id[1])
+            except TypeError:
+                pass
+
+            data = {"operation_ids": {}}
+
+            data["operation_ids"]["operation"] = event.operation_id
+            data["operation_ids"]["request"] = event.request_id
+
+            try:
+                lsid = command.pop("lsid")["id"]
+                data["operation_ids"]["session"] = str(lsid)
+            except KeyError:
+                pass
+
+            query = "{} {}".format(event.command_name, command)
+            span = hub.start_span(op=op, description=query)
+
+            for tag, value in tags.items():
+                span.set_tag(tag, value)
+
+            for key, value in data.items():
+                span.set_data(key, value)
+
+            with capture_internal_exceptions():
+                hub.add_breadcrumb(message=query, category="query", type=op, data=tags)
+
+            self._ongoing_operations[self._operation_key(event)] = span.__enter__()
+
+    def failed(self, event):
+        # type: (CommandFailedEvent) -> None
+        hub = Hub.current
+        if hub.get_integration(PyMongoIntegration) is None:
+            return
+
+        try:
+            span = self._ongoing_operations.pop(self._operation_key(event))
+            span.set_status("internal_error")
+            span.__exit__(None, None, None)
+        except KeyError:
+            return
+
+    def succeeded(self, event):
+        # type: (CommandSucceededEvent) -> None
+        hub = Hub.current
+        if hub.get_integration(PyMongoIntegration) is None:
+            return
+
+        try:
+            span = self._ongoing_operations.pop(self._operation_key(event))
+            span.set_status("ok")
+            span.__exit__(None, None, None)
+        except KeyError:
+            pass
+
+
+class PyMongoIntegration(Integration):
+    identifier = "pymongo"
+
+    @staticmethod
+    def setup_once():
+        # type: () -> None
+        monitoring.register(CommandTracer())
--- a/sentry_sdk/integrations/pymongo.py
+++ b/sentry_sdk/integrations/pymongo.py
@@ -44,31 +44,42 @@ SAFE_COMMAND_ATTRIBUTES = [
 def _strip_pii(command):
     # type: (Dict[str, Any]) -> Dict[str, Any]
     for idx, key in enumerate(command):
-        if key in SAFE_COMMAND_ATTRIBUTES:
+        is_safe_field = key in SAFE_COMMAND_ATTRIBUTES
+        if is_safe_field:
             # Skip if safe key
             continue
 
-        if key == "update" and "findAndModify" not in command:
+        update_db_command = key == "update" and "findAndModify" not in command
+        if update_db_command:
             # Also skip "update" db command because it is save.
             # There is also an "update" key in the "findAndModify" command, which is NOT safe!
             continue
 
-        if key == "documents":
+        # Special stripping for documents
+        is_document = key == "documents"
+        if is_document:
             for doc in command[key]:
                 for doc_key in doc:
                     doc[doc_key] = "%s"
+            continue
 
-        elif key in ["filter", "query", "update"]:
+        # Special stripping for dict style fields
+        is_dict_field = key in ["filter", "query", "update"]
+        if is_dict_field:
             for item_key in command[key]:
                 command[key][item_key] = "%s"
+            continue
 
-        elif key == "pipeline":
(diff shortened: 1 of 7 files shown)


Agalin:
BTW I really hope spans like `db.find` are visible in DB spans panel and similar places, didn't have time to dive into implementation details of span retrieval endpoints.

Agalin:
Docs PR: https://github.com/getsentry/sentry-docs/pull/5469

sentry_sdk/integrations/pymongo.py
```diff
@@ -0,0 +1,120 @@
...
+from sentry_sdk.integrations import DidNotEnable, Integration
+from sentry_sdk.tracing import Span
+from sentry_sdk.utils import capture_internal_exceptions
+
+from sentry_sdk._types import MYPY
+
+try:
+    from pymongo import monitoring
+except ImportError:
+    raise DidNotEnable("Pymongo not installed")
```
polina-popova, line 13: Hi,
May be would be nice also to explicitly check the pymongo version and give the more specific alert as in SqlalchemyIntegration, for example, because pymongo could be installed but has the wrong version.

sentry_sdk/integrations/pymongo.py
```diff
@@ -0,0 +1,120 @@
...
+            data["operation_ids"]["operation"] = event.operation_id
+            data["operation_ids"]["request"] = event.request_id
+
+            try:
+                lsid = command.pop("lsid")["id"]
+                data["operation_ids"]["session"] = str(lsid)
+            except KeyError:
+                pass
+
+            query = "{} {}".format(event.command_name, command)
```
polina-popova, line 73: I checked this implementation with my project and found

`"find {'find': 'users', 'filter': {'_id': '1'}, 'limit': 1, 'singleBatch': True, '$readPreference': {'mode': 'primary'}}"` and
`'saslStart {}'` as a query values.
So there are double mentioning of the command name and empty curve brackets in case of no `command` value.

Are this results the same as you expected?

Agalin: It's the expected result. It's command + its arguments. As you can see in tests content of the arguments dict varies between mongo client (protocol) versions and command being repeated here is not always the case.

One place where there is an explicit duplication of command is op name - and I'm still not sure if it should be repeated there or only one of those should stay (and which one).

sentry_sdk/integrations/pymongo.py
Agalin: I can add it although barely any integration does it - even new ones, like FastAPI just check for the import itself.

polina-popova: I am not sure that FastAPI has any restrictions in its version to be able to use the integration. Anyway, It was just a proposal.

Agalin: It does - otherwise there would be no minimal version specified in `setup.py`. For example see https://github.com/getsentry/sentry-python/pull/1349 - FastAPI `0.74.1` had dedicated changes to enable transaction name retrieval. And I believe there were further changes in it to improve integration.

antonpirker:
Hey @Agalin and @polina-popova !
Sorry for the delay! And thank you soo much for the great work! 
I will have a look at this today, and I will also be the person to help bring this into a release sooner than later :-)

Could you maybe rebase your branch again on the newest master, so everything is up to date? Thanks!

Agalin:
Awesome. Rebased to master (clean rebase as the only existing file is tox.ini)

sentry_sdk/integrations/pymongo.py
antonpirker: Hey! I guess the import check as you implemented is totally fine.

sentry_sdk/integrations/pymongo.py
```diff
@@ -0,0 +1,120 @@
...
+            return
+        with capture_internal_exceptions():
+            command = dict(event.command)
+
+            command.pop("$db", None)
+            command.pop("$clusterTime", None)
+            command.pop("$signature", None)
+
+            if event.command_name:
+                op = "db." + event.command_name
```
antonpirker, line 46: As the span `op` please use "db.query" instead of "db.find". (I know it messes up this elegant code but it helps us figure out performance problems with queries)

sentry_sdk/integrations/pymongo.py
```diff
@@ -0,0 +1,120 @@
...
+        # type: (CommandFailedEvent | CommandStartedEvent | CommandSucceededEvent) -> int
+        return event.request_id
+
+    def started(self, event):
+        # type: (CommandStartedEvent) -> None
+        hub = Hub.current
+        if hub.get_integration(PyMongoIntegration) is None:
+            return
+        with capture_internal_exceptions():
+            command = dict(event.command)
```
antonpirker, line 39: There is a function called `sentry_sdk.hub._should_send_default_pii()` that should be used to guard personal identifiable information. (by default this function returns `False`, this behavior can be changed with this option: https://docs.sentry.io/platforms/python/configuration/options/#send-default-pii)

Right now if you do inserts or make queries, all the user data is sent to Sentry. This must not be the case unless `_should_send_default_pii()` returns `True`.

Can you please make sure that all the values in the command are replaced by `%s`? 
(Maybe there is a way to get the redacted/parameterized command from MongoDB?)

antonpirker (changes requested):
Hey! 

Extraordinary work! 
Just one change to make sure the personal data of users is save and one change to make it easier for Sentry to identify performance problems with queries.

antonpirker:
Oh, and please fix also the linting errors. Thanks!

sentry_sdk/integrations/pymongo.py
Agalin: Nah, it's fine and actually solves a discussion above if `command_name` makes sense being duplicated in `op` and `description`.

cleptric: We should try to stay in sync with the list of operations outlined at https://develop.sentry.dev/sdk/performance/span-operations/#database
We just spend a significant amount of time unifying span ops across our SDKs.

sentry_sdk/integrations/pymongo.py
Agalin: There is no such option. 😞 Those aren't events sent to the database but objects crafted specifically to be used by monitoring components and there is no way from there to ensure access the original query object - which doesn't solve the problem either as it's a wrapper around the raw BSON data. There is no SQL-like convention of parameter replacement in mongo.
I don't think there is any clean way to only substitute user-provided data either, especially when you consider protocol changes between pymongo versions.

Basically I can do one of the two things:
* Only add `command` part when PII can be sent.
* Replace values in this dict by `%s` (recursively?).

TBH I'd rather do the first one to avoid performance penalty.

antonpirker: I guess doing suggestion 2 would be cool. But I do not think we need recursion right now.
Just something like this should be ok:
```
>>> bla = {
...   "blub": "hello",
...   "foo": {
...     "bar": "hello2",
...   }
... }
>>> bla.keys()
dict_keys(['blub', 'foo'])
>>> for key in bla.keys():
...   bla[key] = "%s"
... 
>>> bla
{'blub': '%s', 'foo': '%s'}
```

If we find out that this is not enough data to debug queries (or just know what queries are run) we can change them in the future.

Agalin: I've implemented the first one for now.

sentry_sdk/integrations/pymongo.py
Agalin: Yeah, I've seen that change. I believe this doc looked a bit different when I was writing the pull request. 😄 Now it's a bit cleaner.

sentry_sdk/integrations/pymongo.py
Agalin: Ah. Missed your comment. Will do it like suggested then.

Agalin: Added top level replacement.

Agalin: Oh great. Missed tests update.

BTW should I add tests for that PII part?

antonpirker: Tests for PII stuff would be great, yes!

Agalin: Added.

antonpirker:
I have now added a new function for nicer PII stripping, could you review my PR @Agalin : https://github.com/operasoftware/sentry-python/pull/1 
Thank you very much!

antonpirker:
I have to stay home to take care of my daughter today (some childrens sickness) Could you fix the tests please @Agalin ? I think when the tests are green we are ready to merge.

Agalin:
@antonpirker tests now pass in tox on my machine (tested 2.7 with pymongo 3.1 and 3.8 with 4.2).

antonpirker:
Amazing @Agalin ! Thank you so much again for your great work!

antonpirker (approved):

antonpirker:
Hey @Agalin if you now rebase on the new master again, it will be merged automatically (after running the tests once again)

Agalin:
Ah. My rebase stopped automerge. 😓 @antonpirker can you reapprove?

antonpirker:
Done! (now for the documentation)

antonpirker:
Hey everyone!

This will be released today!

Again: thank you so much for your contribution, it is really great work!
@Agalin If you send me your shipping address to anton.pirker@sentry.io I can send you a little "thank you" gift.

Agalin:
Yay, it's happening!
Should I add an entry in the main Sentry codebase that adds integration recommendation? It couldn't be done earlier as version number is required.

PR #1610, opened 2022-09-12
ref: upgrade linters to flake8 5.x

asottile-sentry:
noqa removal was automated via [yesqa](https://github.com/asottile/yesqa)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,16 +1,17 @@
 [flake8]
-ignore = 
-  E203,  // Handled by black (Whitespace before ':' -- handled by black)
-  E266,  // Handled by black (Too many leading '#' for block comment)
-  E501,  // Handled by black (Line too long)
-  W503,  // Handled by black (Line break occured before a binary operator)
-  E402,  // Sometimes not possible due to execution order (Module level import is not at top of file)
-  E731,  // I don't care (Do not assign a lambda expression, use a def)
-  B950,  // Handled by black (Line too long by flake8-bugbear)
-  B011,  // I don't care (Do not call assert False)
-  B014,  // does not apply to Python 2 (redundant exception types by flake8-bugbear)
-  N812,  // I don't care (Lowercase imported as non-lowercase by pep8-naming)
-  N804   // is a worse version of and conflicts with B902 (first argument of a classmethod should be named cls)
-max-line-length = 80
-select = N,B,C,E,F,W,T4,B9
-exclude=checkouts,lol*,.tox
+extend-ignore =
+  # Handled by black (Whitespace before ':' -- handled by black)
+  E203,
+  # Handled by black (Line too long)
+  E501,
+  # Sometimes not possible due to execution order (Module level import is not at top of file)
+  E402,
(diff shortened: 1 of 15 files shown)


.flake8
```diff
@@ -1,16 +1,17 @@
 [flake8]
-ignore =
```
asottile-sentry, line 2: extend-ignore is better than ignore, it preserves the default ignore

.flake8
```diff
@@ -1,16 +1,17 @@
 [flake8]
-ignore = 
-  E203,  // Handled by black (Whitespace before ':' -- handled by black)
-  E266,  // Handled by black (Too many leading '#' for block comment)
-  E501,  // Handled by black (Line too long)
-  W503,  // Handled by black (Line break occured before a binary operator)
-  E402,  // Sometimes not possible due to execution order (Module level import is not at top of file)
-  E731,  // I don't care (Do not assign a lambda expression, use a def)
```
asottile-sentry, line 8: ini comments are either `#` or `;` not `//` -- these are actually being interpreted as ignore codes https://flake8.pycqa.org/en/latest/user/configuration.html

.flake8
```diff
@@ -1,16 +1,17 @@
...
-  B950,  // Handled by black (Line too long by flake8-bugbear)
-  B011,  // I don't care (Do not call assert False)
-  B014,  // does not apply to Python 2 (redundant exception types by flake8-bugbear)
-  N812,  // I don't care (Lowercase imported as non-lowercase by pep8-naming)
-  N804   // is a worse version of and conflicts with B902 (first argument of a classmethod should be named cls)
-max-line-length = 80
-select = N,B,C,E,F,W,T4,B9
```
asottile-sentry, line 15: this isn't necessary -- flake8 enables the codes of all extension by default

.flake8
```diff
@@ -1,16 +1,17 @@
...
-exclude=checkouts,lol*,.tox
```
asottile-sentry, line 16: extend-exclude is better as it preserves the default exclude list

sentry_sdk/api.py
```diff
@@ -130,7 +130,7 @@ def configure_scope(  # noqa: F811
 
 
 @overload
-def push_scope():  # noqa: F811
```
asottile-sentry, line 133: I fixed pyflakes a while back to understand `@overload`

linter-requirements.txt
```diff
@@ -1,10 +1,9 @@
 black==22.3.0
-flake8==3.9.2
-flake8-import-order==0.18.1
```
asottile-sentry, line 3: this wasn't actually being executed (since it wasn't part of the `select` list below) so I removed it

antonpirker:
Thanks for this update @asottile-sentry !
I will review this after DjangoCon

antonpirker (approved):

linter-requirements.txt
antonpirker: ok

sentry_sdk/api.py
antonpirker: 👍

PR #1611, opened 2022-09-13
feat(profiling): Convert profile output to the sample format

phacops:
We're looking to introduce a new profile format referred to as the Sample format (https://github.com/getsentry/relay/pull/1462). This PR implements this format.


sentry_sdk/profiler.py
```diff
@@ -291,11 +306,18 @@ def slice_profile(self, start_ns, stop_ns):
                                 "line": frame[2],
                             }
                         )
-                    sample["frames"].append(frames[frame])
+                    current_stack.append(frames[frame])
+
+                try:
+                    stack_idx = stacks.index(current_stack)
```
sl0thentr0py, line 312: I have no clue what the sizes of these are, but why not do the same optimization as with `frames` and `frames_list`, so have a dict `stacks_dict` and a list `stacks_list` instead of walking the current `stacks` list with `index`?

sl0thentr0py: oh ok @Zylphrex said the same, did not refresh

sl0thentr0py (approved):
lgtm except for that one comment

sentry_sdk/profiler.py
```diff
@@ -308,13 +309,11 @@ def slice_profile(self, start_ns, stop_ns):
                         )
                     current_stack.append(frames[frame])
 
-                try:
-                    stack_idx = stacks.index(current_stack)
-                except ValueError:
-                    stack_idx = len(stacks)
-                    stacks.append(current_stack)
+                if current_stack not in stacks:
+                    stacks[current_stack] = len(stacks)
```
Zylphrex, line 313: Isn't `current_stack` a list here? This would throw a `TypeError: unhashable type: 'list'`.

sentry_sdk/profiler.py
```diff
@@ -308,13 +309,11 @@ def slice_profile(self, start_ns, stop_ns):
...
+                    stacks_list.append(current_stack)
```
Zylphrex, line 314: What's this for? You just append to it and never use it.

phacops: I should have returned it instead of `stacks`.

sentry_sdk/profiler.py
phacops: I'm now making it into a `tuple` before using it as a key.

sentry_sdk/profiler.py
```diff
@@ -203,20 +195,42 @@ def to_json(self):
...
+                "version": platform.python_version(),
+            },
+            "transactions": [
+                {
+                    "id": None,  # Gets added in client.py
+                    "name": self.transaction.name,
+                    "relative_start_ns": str(relative_start_ns),
+                    "relative_end_ns": str(relative_end_ns),
+                    "trace_id": self.transaction.trace_id,
+                    "thread_id": str(self.transaction._thread_id),
```
Zylphrex, line 231: I don't think you need to get the `_thread_id` on the transaction. Let's move it to the profile's `__init__` and that should suffice. Just trying to keep profiling code separated from transactions where possible.

sentry_sdk/profiler.py
```diff
@@ -352,13 +374,25 @@ def _should_profile(hub):
...
+            with Profile(transaction, hub=hub):
+                yield
+        else:
             yield
-    else:
+
+else:
+    import warnings
+
+    warnings.warn("profiling is only supported on Python >= 3.3")
```
Zylphrex, line 393: I think this warning needs to be in the `setup` function instead so it doesn't appear from just importing this module

Zylphrex (approved):

PR #1616, opened 2022-09-14
feat(profiling): Introduce different profiler schedulers

Zylphrex:
Previously, the only scheduling mechanism was via `signals.SIGPROF`. This was limited to UNIX platforms and was not always consistent. This PR introduces more ways to schedule the sampling. They are the following:

- `_SigprofScheduler` uses `signals.SIGPROF` to schedule
- `_SigalrmScheduler` uses `signals.SIGALRM` to schedule
- `_SleepScheduler` uses threads and `time.sleep` to schedule
- `_EventScheduler` uses threads and `threading.Event().wait` to schedule

### DIFF
--- a/sentry_sdk/profiler.py
+++ b/sentry_sdk/profiler.py
@@ -76,7 +76,6 @@ def setup_profiler(buffer_secs=60, frequency=101):
     `buffer_secs` determines the max time a sample will be buffered for
     `frequency` determines the number of samples to take per second (Hz)
     """
-
     global _sample_buffer
     global _scheduler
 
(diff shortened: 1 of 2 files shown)


sentry_sdk/profiler.py
```diff
@@ -298,33 +289,240 @@ def slice_profile(self, start_ns, stop_ns):
...
 
-        if should_stop_timer:
-            signal.setitimer(signal.ITIMER_PROF, 0)
-        return should_stop_timer
+    def run(self):
+        # type: () -> None
+        while True:
+            if self.event.is_set():
+                break
+            self.event.wait(timeout=self._interval)
```
sl0thentr0py, line 401: the only `event.set`  is in `stop_profiling` so won't this always timeout?

Zylphrex: Yes, that's the intent. The idea with this scheduler is to use this timeout mechanism to schedule the next time the sampling function is called. One possible concern with this approach is that CPU usage may spike due to the wait but this will be tested more and verified before we make a final decision.

sl0thentr0py: ah gotcha, thx for explaining?

sl0thentr0py (approved):

PR #1624, opened 2022-09-16
fix(profiling): Check transaction sampled status before profiling

Zylphrex:
Should always check if the transaction is sampled before deciding to profile to avoid profiling when it's not necessary.


sl0thentr0py (approved):

phacops (approved):

sl0thentr0py:
@antonpirker another parallel side effect in sanic-0.8 again

Issue #1631, opened 2022-09-19
Upgrading from 1.9.5 to 1.9.8 triggers OOM error


euri10:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.9.8

### Steps to Reproduce

hello, fastapi app here with an upload endpoint, files uploaded are usually 15Mo in size, 

reverting to 1.9.5 fixes the issue, on 1.9.8 it's exhausting the container memory quite fast on an upload of several files (about 8 generally).

I'm using sentry this way, just reading the notes I see a deprecation, maybe the reason of the crash

```
    if app_settings_loaded.sentry_dsn is not None:
        # sentry middleware
        sentry_sdk.init(dsn=app_settings_loaded.sentry_dsn)
        app.add_middleware(SentryAsgiMiddleware)
```
that said I dont see the deprecation warning in logs.

the app crash and I dont get sentry logs, I think it's normal considering it appears to be an issue with it

### Expected Result

well to be able to upload my files :)

### Actual Result

```
backend_1          | INFO:     172.18.0.3:54200 - 2022-09-19 13:57:21,244 -'POST /uploads/x1 HTTP/1.1' 500 Internal Server Error
backend_1          | ERROR:    Exception in ASGI application
backend_1          | Traceback (most recent call last):
backend_1          |   File "/opt/venv/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 404, in run_asgi
backend_1          |     result = await app(  # type: ignore[func-returns-value]
backend_1          |   File "/opt/venv/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
backend_1          |     return await self.app(scope, receive, send)
backend_1          |   File "/opt/venv/lib/python3.10/site-packages/fastapi/applications.py", line 269, in __call__
backend_1          |     await super().__call__(scope, receive, send)
backend_1          |   File "/opt/venv/lib/python3.10/site-packages/sentry_sdk/integrations/starlette.py", line 293, in _sentry_patched_asgi_app
backend_1          |     return await middleware(scope, receive, send)
backend_1          |   File "/opt/venv/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py", line 138, in _run_asgi3
backend_1          |     return await self._run_app(scope, lambda: self.app(scope, receive, send))
[...]

Issue #1595, opened 2022-09-01
FastAPI file upload not working with sentry after 1.9.6


dearkafka:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.9.6

### Steps to Reproduce

```
from fastapi import FastAPI, File, UploadFile
from PIL import Image
import uvicorn
import io

import sentry_sdk
from sentry_sdk.integrations.starlette import StarletteIntegration
from sentry_sdk.integrations.fastapi import FastApiIntegration

sentry_sdk.init(
    dsn=SECRET,
    release="server",
    integrations=[
        StarletteIntegration(),
        FastApiIntegration(),
    ],
    attach_stacktrace=True,
    request_bodies="always",
    send_default_pii=True,
    traces_sample_rate=0.0,
)

app = FastAPI()


@app.post("/predict")
async def predict(file: UploadFile = File(default=None)):
    contents = await file.read()
    Image.open(io.BytesIO(contents)).convert("RGB")
    return {"result": "ok"}


uvicorn.run(app, host="0.0.0.0", port=8888, loop="asyncio", lifespan="on")
```

### Expected Result

```{"result": "ok"}```

### Actual Result

```RuntimeError: The Sentry Python SDK can now automatically support ASGI frameworks like Starlette and FastAPI.
Please remove 'SentryAsgiMiddleware' from your project.
See https://docs.sentry.io/platforms/python/guides/asgi/ for more information.
```

so, when disabling sentry - all works fine, when adding sentry - it breaks. apparently sentry messes up content of uploaded file.

vladanpaunovic:
@dearkafka, we released 1.9.7, https://github.com/getsentry/sentry-python/blob/master/CHANGELOG.md#197. You can give it a try and it may result in a success for you.

RockBomber:
with version 1.9.8 still got the error.

vladanpaunovic:
Thanks @RockBomber! We identified this a few days ago too. It's on our backlog and the team will be on it soon.

tristan:
We managed to work around this problem for now by adding `await file.seek(0)` before the `await file.read()` call.

antonpirker:
Hey @dearkafka ! I have now created you sample project like this:
![sample-code](https://user-images.githubusercontent.com/202325/193295396-ef2db48f-2f92-4e39-bc5f-3f6cdbf76f7a.png)

(note that I have NOT specified the integrations, because they are now detected automatically!)

[...]

PR #1632, opened 2022-09-20
Django4 support

antonpirker:
Support for Django 4.0 and 4.1


antonpirker:
The only really new parts in Django 4 (more specifically 4.1) are async class based view and async ORM queries. Both of them I have tested and everything shows up in "Performance". So I think we are good to go.

One problem is there if you run Django in a ASGI server, but this is the same underlying issue as https://github.com/getsentry/sentry-python/issues/1631 and https://github.com/getsentry/sentry-python/issues/1595 (this problem has not yet surfaced in Django because most people use Django in a sync way.)

This problem will be handled here: https://github.com/getsentry/sentry-python/issues/1646

vladanpaunovic (approved):

PR #1634, opened 2022-09-20
fix(profiling): Dynamically adjust profiler sleep time

Zylphrex:
Because more time may have elapsed between 2 samples due to us calling the sampling function and other threads executing, we need to account for it in the sleep or the time between samples will often be greater than the expected interval. This change ensures we account for this time elapsed and dynamically adjust the amount of time we sleep for between samples.

### DIFF
--- a/sentry_sdk/profiler.py
+++ b/sentry_sdk/profiler.py
@@ -377,10 +377,23 @@ class _SleepScheduler(_ThreadScheduler):
 
     def run(self):
         # type: () -> None
+        last = time.perf_counter()
+
         while True:
+            # some time may have elapsed since the last time
+            # we sampled, so we need to account for that and
+            # not sleep for too long
+            now = time.perf_counter()
+            elapsed = max(now - last, 0)
+
+            if elapsed < self._interval:
+                time.sleep(self._interval - elapsed)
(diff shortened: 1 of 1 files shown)


sentry_sdk/profiler.py
```diff
@@ -377,10 +377,23 @@ class _SleepScheduler(_ThreadScheduler):
...
+        last = time.perf_counter()
+
         while True:
+            # some time may have elapsed since the last time
+            # we sampled, so we need to account for that and
+            # not sleep for too long
+            now = time.perf_counter()
+            elapsed = max(now - last, 0)
+
+            if elapsed < self._interval:
```
phacops, line 389: I'm not sure why you test this before you sleep. Shouldn't we sleep all the time at least a little bit, just the remaining amount of time until `self._interval` passed?

Zylphrex: If the elapsed time is greater than `self._interval`, then we'd want to sample as soon as possible otherwise the time between the 2 samples will be much larger than expected. This is because other threads could be running without giving this thread a chance to execute.

sl0thentr0py (approved):

PR #1641, opened 2022-09-27
Fix for partial signals in old Django and old Python versions.

### LINKED ISSUE #1640: Django Signals integration breaks on partial objects for python <3.10
### How do you use Sentry?

Self-hosted/on-premise

### Version

1.9.9

### Steps to Reproduce

1. Use python older than 3.10.
2. Register a partial function as a signal handler.

### Expected Result

Signal is traced correctly.

### Actual Result

Exception is raised from `_get_receiver_name` function as `partial` objects don't have `__module__` before python 3.10 (and even there it's undocumented from what I can see).

It fails in our tests where we don't even register any signals so either Django itself or some kind of integration (Sentry?) registers such signals by default.

The whole signals integration is missing a `capture_internal_exceptions` context too I believe.

antonpirker:
Making sure signal name can also be set when partials are used as signal in  old Django versions in old Python versions.

Fixes #1640

### DIFF
--- a/sentry_sdk/integrations/django/signals_handlers.py
+++ b/sentry_sdk/integrations/django/signals_handlers.py
@@ -15,11 +15,17 @@ if MYPY:
 
 def _get_receiver_name(receiver):
     # type: (Callable[..., Any]) -> str
-    if hasattr(receiver, "__qualname__"):
-        return receiver.__qualname__
+    name = ""
+    if hasattr(receiver, "__module__"):
+        name += receiver.__module__ + "."
 
-    return str(receiver)
+    if hasattr(receiver, "__name__"):
+        name += receiver.__name__
 
+    if name == "":
+        return str(receiver)
+
+    return name
 
 def patch_signals():
     # type: () -> None
--- a/sentry_sdk/integrations/django/signals_handlers.py
+++ b/sentry_sdk/integrations/django/signals_handlers.py
@@ -21,9 +21,11 @@ def patch_signals():
 
     def _get_receiver_name(receiver):
         # type: (Callable[..., Any]) -> str
-        name = receiver.__module__ + "."
+        name = receiver.__module__ + "." if hasattr(receiver, "__name__") else ""
+
         if hasattr(receiver, "__name__"):
             return name + receiver.__name__
+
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/django/signals_handlers.py
```diff
@@ -21,9 +21,14 @@ def patch_signals():
 
     def _get_receiver_name(receiver):
         # type: (Callable[..., Any]) -> str
-        name = receiver.__module__ + "."
+        if hasattr(receiver, "__module__"):
```
sl0thentr0py, line 24: can't we just use [`__qualname__`](https://docs.python.org/3/glossary.html#term-qualified-name) here?

antonpirker: Good idea.

But I checked, the partials also do not have a `__qualname__`, so: no.

antonpirker: oh, but we can use `__qualname__` if there is one, and if not we use `__name__`

sl0thentr0py (changes requested):
also please add a test using a partial :)

sentry_sdk/integrations/django/signals_handlers.py
```diff
@@ -13,19 +13,32 @@
...
+        name += receiver.__qualname__
+    elif hasattr(receiver, "__name__"):  # Python 2.7 has no __qualname__
+        name += receiver.__name__
+
+    if (
+        name == ""
+    ):  # certain functions (like partials) dont have a name so return the string representation
+        return str(receiver)
+
+    if hasattr(receiver, "__module__"):  # prepend with module, if there is one
```
sl0thentr0py, line 30: qualname already has the module prefix right?

antonpirker: No. It is only the fully qualified name of the class itself (with dots if is a nested class) but it does not include the python module.

sl0thentr0py: ah gotcha, thx

sl0thentr0py (approved):

PR #1650, opened 2022-09-29
Fix CI tests

antonpirker:
The new version of Sanic that was released a couple of hours ago has some breaking changes and thus break our test suite. So I have now excluded the latest Sanic version from our test suite.

### DIFF
--- a/tests/integrations/sanic/test_sanic.py
+++ b/tests/integrations/sanic/test_sanic.py
@@ -32,8 +32,8 @@ def app():
 
         Sanic.test_client = property(new_test_client)
 
-    if SANIC_VERSION >= (20, 12):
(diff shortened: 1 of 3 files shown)


tests/integrations/sanic/test_sanic.py
```diff
@@ -229,7 +229,6 @@ def __init__(self, request_body):
                 def respond(self, response):
                     responses.append(response)
                     patched_response = HTTPResponse()
-                    patched_response.send = lambda end_stream: asyncio.sleep(0.001)
```
sl0thentr0py, line 232: what was this mega hack lol

sl0thentr0py (approved):

PR #1651, opened 2022-09-29
Cancel old CI runs when new one is started.

### LINKED ISSUE #1644: Save GitHub actions concurrency by cancelling old runs
### Problem Statement

On each push to a PR the whole test suite is running. This always starts a lot of concurrent Github action workflows. (And we as sentry only have 180 concurrent workflows for the whole organization. 

When there is a new commit pushed to a PR the still running workflows of the old commit should be canceled to free up workflows for other projects.

### Solution Brainstorm

One can define a concurrency group and auto cancel old runs.  Sentry and Getsentry project already did this.

See: https://github.com/getsentry/sentry/pull/38889/files

antonpirker:
This will save us some CI minutes when new commits are pushed to the same PR over and over again.

Fixes #1644

### DIFF
--- a/.github/workflows/test-integration-aiohttp.yml
+++ b/.github/workflows/test-integration-aiohttp.yml
@@ -8,6 +8,12 @@ on:
 
   pull_request:
 
+# Cancel in progress workflows on pull_requests.
+# https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value
+concurrency:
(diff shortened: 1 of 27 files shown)


antonpirker:
Look at the checks in the commit "Cancel old CI runs when new one is started". Some of them have been cancelled (grey exclamation mark) which means the setup works!

sl0thentr0py (approved):

PR #1662, opened 2022-10-05
Include framework in SDK name

### LINKED ISSUE #1497: SDK Name should include framework used
### Problem Statement

From many (most or all?) Sentry SDKs the sdk name in the event payload, query params, and auth headers follows a pattern of including the framework when being used. (like `sentry.ruby.rails` or `sentry.php.laravel`) The python SDK should follow the same pattern, this would make tracking usage per framework much easier.

### Solution Brainstorm

SDK name in the event body, header, query param, and auth header should all match, similar or same as other SDKs

### Where to include:

- [ ] event header [Dev docs headers](https://develop.sentry.dev/sdk/envelopes/#headers)
- [ ] event body [Develop docs Event payload Attributes](https://develop.sentry.dev/sdk/event-payloads/sdk/#attributes)
- [ ] auth header or query param: [Authentication dev docs](https://develop.sentry.dev/sdk/overview/#authentication)

antonpirker:
Currently the Python sdk reports always the name `sentry.python` as its name. For consistent sdk naming and better data analysis of events we want to have also include the framework in the SDK name. Like `sentry.python.django`.

Fixes #1497

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -19,7 +19,7 @@ from sentry_sdk.serializer import serialize
 from sentry_sdk.transport import make_transport
 from sentry_sdk.consts import (
     DEFAULT_OPTIONS,
-    SDK_INFO,
+    VERSION,
     ClientConstructor,
     _get_sdk_name,
 )
@@ -44,6 +44,13 @@ if MYPY:
 
 
 _client_init_debug = ContextVar("client_init_debug")
+
+
+SDK_INFO = {
+    "name": "sentry.python",  # SDK name be overridden after integrations have been loaded with sentry_sdk.integrations.setup_integrations()
+    "version": VERSION,
+    "packages": [{"name": "pypi:sentry-sdk", "version": VERSION}],
+}
 
 
 def _get_options(*args, **kwargs):
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -21,7 +21,6 @@ from sentry_sdk.consts import (
     DEFAULT_OPTIONS,
     VERSION,
     ClientConstructor,
-    _get_sdk_name,
 )
 from sentry_sdk.integrations import setup_integrations
 from sentry_sdk.utils import ContextVar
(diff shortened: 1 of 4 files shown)


sentry_sdk/client.py
```diff
@@ -128,6 +133,11 @@ def _capture_envelope(envelope):
                     "auto_enabling_integrations"
                 ],
             )
+
+            sdk_name = _get_sdk_name(list(self.integrations.keys()))
+            SDK_INFO["name"] = sdk_name
```
sl0thentr0py, line 138: if we're mutating `SDK_INFO` in another module, it's not a `const` anymore, so I'd just move it to client.py.

sl0thentr0py (changes requested):

sentry_sdk/client.py
antonpirker: yes, you are right!

antonpirker:
I did not make the var lower case although it is no real const anymore but having it uppercase just shows me that it is defined somewhere on the top (and not given as a parameter)

sentry_sdk/client.py
```diff
@@ -41,6 +46,13 @@
 _client_init_debug = ContextVar("client_init_debug")
 
 
+SDK_INFO = {
+    "name": "sentry.python",  # SDK name be overridden after integrations have been loaded with sentry_sdk.integrations.setup_integrations()
```
vladanpaunovic, line 50: nit: small typo

```suggestion
    "name": "sentry.python",  # SDK name will be overridden after integrations have been loaded with sentry_sdk.integrations.setup_integrations()
```

sentry_sdk/consts.py
```diff
@@ -104,9 +104,38 @@ def _get_default_options():
 del _get_default_options
 
 
+def _get_sdk_name(installed_integrations):
```
sl0thentr0py, line 107: this should also not be here, either in integrations or utils

sl0thentr0py (approved):

PR #1669, opened 2022-10-07
feat(profiling): Extract qualified name for each frame

Zylphrex:
Currently, we use `code.co_name` for the frame name. This does not include the name of the class if it was a method. This tries to extract the qualified name for each frame where possible.

- methods: *typically* have `self` as a positional argument and we can inspect it to extract the class name
- class methods: *typically* have `cls` as a positional argument and we can inspect it to extract the class name
- static methods: no obvious way of extract the class name

### DIFF
--- a/sentry_sdk/profiler.py
+++ b/sentry_sdk/profiler.py
@@ -135,7 +135,7 @@ def _extract_stack(frame):
     only the first `MAX_STACK_DEPTH` frames will be returned.
     """
 
-    stack = deque(maxlen=MAX_STACK_DEPTH)  # type: Deque[FrameData]
+    stack = deque(maxlen=MAX_STACK_DEPTH)  # type: Deque[FrameType]
 
     while frame is not None:
         stack.append(frame)
--- a/sentry_sdk/profiler.py
+++ b/sentry_sdk/profiler.py
@@ -16,21 +16,20 @@ import atexit
 import platform
 import random
 import signal
+import sys
 import threading
 import time
-import sys
 import uuid
-
-from collections import deque
+from collections import deque, namedtuple
 from contextlib import contextmanager
 
 import sentry_sdk
 from sentry_sdk._compat import PY33
-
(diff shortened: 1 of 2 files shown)


sentry_sdk/profiler.py
```diff
@@ -137,22 +135,52 @@ def _extract_stack(frame):
...
 
-    return stack
+    return [
+        FrameData(
+            name=get_frame_name(frame),
+            file=frame.f_code.co_filename,
+            line=frame.f_lineno,
+        )
+        for frame in stack
+    ]
```
Zylphrex, line 151: Instead of extracting the data for every frame while we step through the stack, we step through the stack first and only extract the data from the frames within the `MAX_STACK_DEPTH`.

Some quick benchmarks on 100,000 iterations
- stack depth of 100: 4.22s vs 4.55s (a little slower)
- stack depth of 500: 21.76s vs 8.32 (a lot faster)

Since the slowest part of this is the `get_frame_name` function and I figured it was worth the trade off here to optimize for the worst case and be a little slower on the common cases.

sentry_sdk/profiler.py
```diff
@@ -289,9 +317,9 @@ def slice_profile(self, start_ns, stop_ns):
                         frames[frame] = len(frames)
                         frames_list.append(
                             {
-                                "name": frame[0],
-                                "file": frame[1],
-                                "line": frame[2],
+                                "name": frame.name,
+                                "file": frame.file,
+                                "line": frame.line,
```
phacops, line 322: If you don't mind using `function`/`filename`/`lineno`/`colno`, that'd be great. We renamed them to align with the `Frame` object in the protocol but still provide compatibility though.

Also, we added `colno` in case you want/can report the column.

Zylphrex: I have this in a follow up 👍

sl0thentr0py (approved):

PR #1672, opened 2022-10-11
Fixes #1386 - is_valid_sample_rate-toostrict

### LINKED ISSUE #1386: is_valid_sample_rate may be too strict
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.5.6

### Steps to Reproduce

One day sentry stopped sending transactions to sentry.io.   After much digging I found the issue to be here:
https://github.com/getsentry/sentry-python/blob/0ba75fef404f877f3c7fc1afcc6013eb9c4b986c/sentry_sdk/tracing.py#L668

We have a custom traces_sampler function that reads in from a simple JSON file and sets the sample rate based on the endpoint.  An example of our JSON:
```
{
  "^$": 0.05,
  "default_sample_rate": 1.0
}
```

We limit our root path `/`, because it gets polled by our load balancer once a minute.  And the endpoint is blindingly simple and uninteresting, so we don't want to track all events and blow through our quota.

The problem came about when I switch from the `json` package to parse the json, to the `json_encoder` package, which generally is more robust json parser.   The issue is that the package json will create a type of `float` when it deserializes the json, json_encoder will create a type of `Decimal`

### Expected Result

* sample_rate with type Decimal should be sufficient for a return type of the traces_sampler function
* sentry documentation should state the exact allowable types for the traces_sampler function

### Actual Result

transactions will be skipped if the type returned by traces_sampler function is not of type float.

Arvind2222:
* Fix is_valid_sample_rate to allow check on `Decimal`

### DIFF
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -2,6 +2,7 @@ import re
 import contextlib
 import json
 import math
+from decimal import Decimal
 
 from numbers import Real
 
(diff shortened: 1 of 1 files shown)


Arvind2222:
@antonpirker @mitsuhiko Can you please review

PR #1673, opened 2022-10-11
Have instrumentation for ASGI middleware receive/send callbacks.

### LINKED ISSUE #1664: Set correct parent span in async environment
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.9.10

### Steps to Reproduce

* Have a FastAPI app
* Have a FastAPI middleware like this one: https://github.com/getsentry/demo-app-django-react/blob/main/backend/fastapi/main.py#L80-L101
* Do something in this middleware that creates a span (like the redis call in the span
* Reload an endpoint that uses this middleware
* Go to the "Performance" section of sentry.io and look at the event
* You will see this:
    Screenshot 2022-10-05 at 08 55 07

### Expected Result

The redis span should be under the `CounterMiddleware`.

### Actual Result

The redis span is under the `AsyncExitStackMiddleware`

antonpirker:
To better visualize what is going on when ASGI middlewares are executed we added instrumentation for the `receive` and `send` callbacks. 

FIxes #1664

### DIFF
--- a/sentry_sdk/integrations/starlette.py
+++ b/sentry_sdk/integrations/starlette.py
@@ -91,12 +91,40 @@ def _enable_span_for_middleware(middleware_class):
         integration = hub.get_integration(StarletteIntegration)
         if integration is not None:
             middleware_name = args[0].__class__.__name__
+
             with hub.start_span(
                 op=OP.MIDDLEWARE_STARLETTE, description=middleware_name
             ) as middleware_span:
                 middleware_span.set_tag("starlette.middleware_name", middleware_name)
 
-                await old_call(*args, **kwargs)
+                app, scope, receive, send = args
+
+                # Creating spans for the "receive" callback
+                async def _sentry_receive(*args, **kwargs):
+                    hub = Hub.current
+                    with hub.start_span(
+                        op=OP.MIDDLEWARE_STARLETTE_RECEIVE,
+                        description=receive.__qualname__,
+                    ) as span:
+                        span.set_tag("starlette.middleware_name", middleware_name)
+                        await receive(*args, **kwargs)
+
+                receive_patched = receive.__name__ == "_sentry_receive"
+                new_receive = _sentry_receive if not receive_patched else receive
+
+                # Creating spans for the "send" callback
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/starlette.py
```diff
@@ -91,12 +91,42 @@ async def _create_span_call(*args, **kwargs):
...
         if integration is not None:
             middleware_name = args[0].__class__.__name__
+
             with hub.start_span(
                 op=OP.MIDDLEWARE_STARLETTE, description=middleware_name
             ) as middleware_span:
                 middleware_span.set_tag("starlette.middleware_name", middleware_name)
 
-                await old_call(*args, **kwargs)
+                app, scope, receive, send = args
```
sl0thentr0py, line 100: can this unpacking ever fail?

sl0thentr0py (changes requested):
one question, and

in cases like these where we add some feature, can you attach a testing screenshot so that somebody who looks at the PR from the changelog can see clearly what changed?

antonpirker:
> one question, and
> 
> in cases like these where we add some feature, can you attach a testing screenshot so that somebody who looks at the PR from the changelog can see clearly what changed?

Good idea!

sentry_sdk/integrations/starlette.py
antonpirker: As we patch the `__call__` of a class, the first arg is always there. And when looking at the ASGI spec (and all the implemented middleware in starlette) the scope, receive and send args are NOT optional: https://asgi.readthedocs.io/en/latest/specs/main.html#

So: No, it most certainly can not fail. (but it is software and software is written by humans, it may fail anyways...)

sl0thentr0py: so I believe it is slightly better if we are unpacking anyway to change the `_create_span_call` to
```
def _create_span_call(app, scope, receive, send, **kwargs)
```
rather than unpacking later

antonpirker: 👍

sl0thentr0py (approved):

PR #1674, opened 2022-10-11
ref(profiling): Add types tests and update names in profiler

Zylphrex:
This change removes some of the `Any` types we've been using, introduces tests to some of the profiling functions and updates the payload to use the canonical names.


Zylphrex:
going to split this out into smaller PRs

PR #1676, opened 2022-10-12
fix(profiling): Race condition spawning multiple profiling threads

Zylphrex:
There is a race condition where multiple profiling threads may be spawned. Specifically, if `start_profiling` is called immediately after `stop_profiling`. This happens because `stop_profiling` does not immediately terminate the thread, instead the thread will check that the event was set and exit at the end of the current iteration. If `start_profiling` is called during the iteration, the event gets set again and the old thread will continue running. To fix this, a new event is created when a profiling thread starts so they can be terminated independently.

### DIFF
--- a/sentry_sdk/profiler.py
+++ b/sentry_sdk/profiler.py
@@ -16,22 +16,22 @@ import atexit
 import platform
 import random
 import signal
+import sys
 import threading
 import time
-import sys
 import uuid
-
 from collections import deque
 from contextlib import contextmanager
 
 import sentry_sdk
 from sentry_sdk._compat import PY33
-
+from sentry_sdk._queue import Queue
 from sentry_sdk._types import MYPY
 from sentry_sdk.utils import nanosecond_time
 
 if MYPY:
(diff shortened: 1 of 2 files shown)


sentry_sdk/profiler.py
```diff
@@ -374,34 +380,37 @@ def teardown(self):
 
     def start_profiling(self):
         # type: () -> bool
-        if super(_ThreadScheduler, self).start_profiling():
+        if super(ThreadScheduler, self).start_profiling():
             # make sure to clear the event as we reuse the same event
             # over the lifetime of the scheduler
-            self.event.clear()
+            event = threading.Event()
+            self.threads.put_nowait(event)
```
sl0thentr0py, line 387: just nitpicking but could we possible have a better name than `threads`?
Maybe signify purpose, like `termination_events/stop_events` or something since we use them to break the profiler loops?

sl0thentr0py (approved):
just a naming nitpick, feel free to keep/rename as you deem fit

PR #1689, opened 2022-10-18
Asyncio fixes

antonpirker:
* Fix task creation (give the coroutine object, not the function creating it)
* Make sure a user set task factory is not overridden by our factory.


antonpirker:
I will create a new PR for the exception handling. Please approve this @sl0thentr0py (if it is ok) so we can merge it.

sl0thentr0py (approved):

Issue #1678, opened 2022-10-13
Sentry's http putrequest override breaks if url is None, while native http does not


cpak:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.9.10

### Steps to Reproduce

Example code

```
import sentry_sdk
from http.client import HTTPConnection

print("sending reqeust with None url before sentry_sdk.init")
HTTPConnection("httpbin.org", port=443).putrequest("POST", None)
print("OK")

sentry_sdk.init(dsn="")
print("sending reqeust with None url after sentry_sdk.init")
HTTPConnection("httpbin.org", port=443).putrequest("POST", None)
print("OK")
```

Output:

```
sending reqeust with None url before sentry_sdk.init
OK
sending reqeust with None url after sentry_sdk.init
Traceback (most recent call last):
  File "/Users/cpak/projects/sentry-httplib-bug/main.py", line 10, in <module>
    HTTPConnection("httpbin.org", port=443).putrequest("POST", None)
  File "/Users/cpak/projects/sentry-httplib-bug/.venv/lib/python3.10/site-packages/sentry_sdk/integrations/stdlib.py", line 73, in putrequest
    if not real_url.startswith(("http://", "https://")):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

### Expected Result

I expect Sentry's http instrumentation to not alter the behavior of the underlying code.

I ran into this issue because a package I'm using ([azure-data-tables](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/tables/azure-data-tables)) apparently tries to send requests with a `None` url, i.e. I don't really have any control over this. It worked locally but not on our remote servers, which turned out to be because we have Sentry running on those servers 😅 

### Actual Result

An `AttributeError: 'NoneType' object has no attribute 'startswith'` is raised.

FWIW the native implementation has this line
```
url = url or '/'
```

PR #1693, opened 2022-10-19
fix(integrations): Fix http putrequest when url is None

MattFlower:
Modifies behavior of putrequest to check for None on real_url prior to using it.

Fixes GH-1678


antonpirker:
Hey @MattFlower, 
great work! Thanks for contributing.

antonpirker (approved):

PR #1694, opened 2022-10-19
perf(profiling): Tune the sample profile generation code for performance

Zylphrex:
We noticed that generating the sample format at the end of a profile can get rather slow and this aims to improve what we can here with minimal changes. A few things we took advantage of to accomplish this:

- Turning the extracted stack into a tuple so it is hashable so it can be used as a dictionary key. This let's us check if the stack is indexed first, and skip indexing the frames again. This is especially effective in profiles where it's blocking on a network request for example, since there will be many identical stacks.
- Using the hash of the stack as the dictionary key. Hashing the entire stack can be an expensive operation since a stack can have up to 128 frames. Using it as a dictionary key means it needs to be rehashed each time. To avoid this, we pre-hash the stack and use the hash as a dictionary key which is more efficient.
- Convert numbers to strings ahead of time if we know have to. Values like the tid and elapsed since start ns needs to be sent as a string. However, many samples share the same value for it, and we're doing the conversion each time. Instead, we convert them to a string upfront and reuse it as needed in order to minimize unnecessary calculations.

### DIFF
--- a/sentry_sdk/profiler.py
+++ b/sentry_sdk/profiler.py
@@ -380,7 +380,7 @@ class SampleBuffer(object):
         # will not have any metadata associated with it.
         thread_metadata = {
             str(thread.ident): {
-                "name": thread.name,
+                "name": str(thread.name),
             }
             for thread in threading.enumerate()
         }  # type: Dict[str, ProcessedThreadMetadata]
--- a/sentry_sdk/profiler.py
+++ b/sentry_sdk/profiler.py
@@ -45,7 +45,7 @@ if MYPY:
     from typing_extensions import TypedDict
     import sentry_sdk.tracing
 
-    RawSampleData = Tuple[int, Sequence[Tuple[int, Sequence[RawFrameData]]]]
+    RawSampleData = Tuple[int, Sequence[Tuple[str, Sequence[RawFrameData]]]]
 
     ProcessedStack = Tuple[int, ...]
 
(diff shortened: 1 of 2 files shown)


Zylphrex:
Some benchmarks I collected comparing the before and afters using a few arbitrary profiles.

```samples: 439		stacks: 4326
old	avg of 100 runs	   21.9256ms
new	avg of 100 runs	    3.5280ms
samples: 507		stacks: 4227
old	avg of 100 runs	   18.8779ms
new	avg of 100 runs	    4.2562ms
samples: 1688		stacks: 15167
old	avg of 100 runs	   61.5474ms
new	avg of 100 runs	   14.6774ms
samples: 418		stacks: 3347
old	avg of 100 runs	   14.5421ms
new	avg of 100 runs	    2.5589ms
samples: 513		stacks: 2568
old	avg of 100 runs	   14.5687ms
new	avg of 100 runs	    2.9569ms
samples: 379		stacks: 758
old	avg of 100 runs	    5.7061ms
new	avg of 100 runs	    1.5271ms
```

From this testing, this looks to be consistently faster than what it used to be, and I have no reason to believe it's slower in any case.

antonpirker:
Side note: The "daphne" errors above in the Django tests was fixed in master, to if you update your branches they should be gone.

sl0thentr0py (approved):
nice numbers :tada:

PR #1702, opened 2022-10-21
feat(profiling): Extract more frame info

Zylphrex:
This extracts a little more information around the frame that we'll use to improve the visualization/groupings including
- in_app
- module


sentry_sdk/profiler.py
```diff
@@ -243,18 +263,24 @@ def __exit__(self, ty, value, tb):
...
         assert self._start_ns is not None
         assert self._stop_ns is not None
 
+        profile = self.scheduler.sample_buffer.slice_profile(
+            self._start_ns, self._stop_ns
+        )
+
+        handle_in_app_impl(
+            profile["frames"], options["in_app_exclude"], options["in_app_include"]
+        )
```
Zylphrex, line 277: If I understood the helper correctly, this does the following
- if no `in_app_exclude` or `in_app_include` are specified, it'll add `"in_app": True` to every frame
- if it is specified
  - frames that match `in_app_include` will have `"in_app": True` added to it
  - frames that match `in_app_exclude` will have `"in_app": False` added to it
  - other frames will not specify, but we should treat them as if they're not in_app frames

sl0thentr0py (approved):

Issue #1712, opened 2022-10-27
Add `pytest_asyncio` in test-requirements.txt


mgaligniana:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.10.1

### Steps to Reproduce

I followed the steps defined in https://github.com/getsentry/sentry-python/blob/master/CONTRIBUTING.md#running-tests **WITHOUT USING** `tox`

1. Clone the repo
2. `cd sentry-python`
3. `python -m venv .env` (using python 3.7.13)
4. `source .env/bin/activate`
5. `pip install -e .`
6. `pip install -r test-requirements.txt`
7. `pytest tests/`

Note: The last week the `pytest_asyncio` import was added: https://github.com/getsentry/sentry-python/commit/973b2f6db7386aae50dd4279ffcead9a4c87d8c6#diff-eed9849977b164602f2509424aa79d672bfdf77cc7bcce9411a27fc8ad0dd3bfR5

### Expected Result

Test suite run successfully

### Actual Result

```
========================================================================================================================== ERRORS ==========================================================================================================================
_______________________________________________________________________________________________ ERROR collecting tests/integrations/asyncio/test_asyncio.py ________________________________________________________________________________________________
ImportError while importing test module '/.../sentry-python/tests/integrations/asyncio/test_asyncio.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../.pyenv/versions/3.7.13/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/integrations/asyncio/test_asyncio.py:5: in <module>
    import pytest_asyncio
E   ModuleNotFoundError: No module named 'pytest_asyncio'
```

antonpirker:
Will close. 
See comments in PR for details: https://github.com/getsentry/sentry-python/pull/1713

PR #1713, opened 2022-10-27
test(integrations): Add pytest_asyncio to test-requirements file

mgaligniana:
Fixes GH-1712

Hello!

When I was testing my first issue to the sentry-python repo I found this error when I tried to run the tests.

Hopefully this was the problem!

Thank you!

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -5,6 +5,7 @@ pytest-cov==2.8.1
 pytest-forked<=1.4.0
 pytest-localserver==0.5.0
 pytest-watch==4.2.0
+pytest-asyncio==0.20.1
 tox==3.7.0
 Werkzeug<2.1.0
 jsonschema==3.2.0


mgaligniana:
Oh it didn't work. If there is something I can read/change to fix it let me know! Thanks!

antonpirker:
Hey @mgaligniana 

Thanks for the PR! Really appreciate it. 
I have now fixed this in another way: https://github.com/getsentry/sentry-python/commit/914aa8ffc609efa230ed92dcaac35fb201bb8761

(Because just adding `pytest-asyncion` to test-requirements.txt will not work, because we also run our tests in Python 2.7 and this makes things a little bit more complicated :-)

I will close this PR and the related issue.

(But If you send me your mailing address to anton.pirker@sentry.io I will hook you up with a little "thank you")

PR #1716, opened 2022-10-31
Move relay to port 5333 to avoid collisions

sl0thentr0py:
goes with https://github.com/getsentry/action-build-aws-lambda-extension/pull/6
and https://github.com/getsentry/sentry-javascript/pull/6093

### DIFF
--- a/scripts/init_serverless_sdk.py
+++ b/scripts/init_serverless_sdk.py
@@ -21,7 +21,7 @@ if MYPY:
 def extension_relay_dsn(original_dsn):
     dsn = Dsn(original_dsn)
     dsn.host = "localhost"
-    dsn.port = 3000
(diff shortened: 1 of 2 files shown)


antonpirker (approved):

viperfx:
Hi there, I was wondering when would this change be included in a release?

viperfx:
@antonpirker I was wondering - when would this change be included in a release?

antonpirker:
This should already be working. You need to update your serverless functions.

You can do this by going to Sentry.io > Settings > Integrations > Select "AWS Lambda" from the list > Select "Configuration" Tab > Click "Configuration" next to your installation > Hit the "Update" button next to your function.

Does this solve your issue?

Issue #1675, opened 2022-10-11
RuntimeError: Stream consumed when reading body with startlette (bug since 1.9.10)


sevaho:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.9.10

### Steps to Reproduce

Server:

```python
import sentry_sdk

from fastapi import FastAPI, Request

sentry_sdk.init()

app = FastAPI()


@app.post("/")
async def index(request: Request):
    r = await request.body()
    return {"Hello": "World"}
```

Run with uvicorn:

```
uvicorn app:app --port 5000
```

As client do:

```python
#! /usr/bin/python
import httpx

import asyncio


async def main():
    image = b"\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01\x00\x00\x00\x01\x08\x06\x00\x00\x00\x1f\x15\xc4\x89\x00\x00\x00\rIDATx\xdac\xfc\xcf\xf0\xbf\x1e\x00\x06\x83\x02\x7f\x94\xad\xd0\xeb\x00\x00\x00\x00IEND\xaeB`\x82"
    files = {"file": ("test.png", image)}
    async with httpx.AsyncClient() as client:
        r = await client.post(
            "http://localhost:5000/",
            files=files,
        )
        print(r)


asyncio.run(main())
```

### Expected Result

To succeed.

### Actual Result

ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/sevaho/.cache/pypoetry/virtualenvs/centurion-X1KYOsH6-py3.10/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 404, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/sevaho/.cache/pypoetry/virtualenvs/centurion-X1KYOsH6-py3.10/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/home/sevaho/.cache/pypoetry/virtualenvs/centurion-X1KYOsH6-py3.10/lib/python3.10/site-packages/fastapi/applications.py", line 270, in __call__
    await super().__call__(scope, receive, send)
[...]

PR #1724, opened 2022-11-04
Fix reading FastAPI request body twice. 

antonpirker:
Starlette/FastAPI is internally caching the request body if read via `request.json()` or `request.body()` but NOT when using `request.form()`. This leads to a problem when our Sentry Starlette integration wants to read the body data and also the users code wants to read the same data. 

Solution:
Force caching of request body for `.form()` calls too, to prevent error when body is read twice.

The tests where mocking `.stream()` and thus hiding this problem. So the tests have been refactored to mock the underlying `._receive()` function instead.

### DIFF
--- a/sentry_sdk/integrations/starlette.py
+++ b/sentry_sdk/integrations/starlette.py
@@ -478,7 +478,7 @@ class StarletteRequestExtractor:
                 request_info["cookies"] = None
 
             # Add annotation if body is too big
-            content_length = await self.content_length()
+            content_length = self.content_length()
             if not request_body_within_bounds(client, content_length):
                 request_info["body"] = AnnotatedValue.removed_because_over_size_limit()
                 return request_info
--- a/sentry_sdk/integrations/starlette.py
+++ b/sentry_sdk/integrations/starlette.py
@@ -478,8 +478,10 @@ class StarletteRequestExtractor:
                 request_info["cookies"] = None
 
             # Add annotation if body is too big
-            content_length = self.content_length()
-            if not request_body_within_bounds(client, content_length):
+            content_length = await self.content_length()
+            if content_length and not request_body_within_bounds(
+                client, content_length
+            ):
                 request_info["body"] = AnnotatedValue.removed_because_over_size_limit()
                 return request_info
 
@@ -508,7 +510,7 @@ class StarletteRequestExtractor:
             request_info["body"] = AnnotatedValue.removed_because_raw_data()
             return request_info
 
-    def content_length(self):
+    async def content_length(self):
         # type: (StarletteRequestExtractor) -> Optional[int]
         if "content-length" in self.request.headers:
             return int(self.request.headers["content-length"])
--- a/sentry_sdk/integrations/starlette.py
+++ b/sentry_sdk/integrations/starlette.py
@@ -370,7 +370,8 @@ def patch_request_response():
                             # Add info from request to event
                             request_info = event.get("request", {})
                             if info:
-                                request_info["cookies"] = info["cookies"]
+                                if "cookies" in info:
+                                    request_info["cookies"] = info["cookies"]
                                 request_info["data"] = info["data"]
                             event["request"] = request_info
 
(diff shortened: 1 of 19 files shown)


sentry_sdk/integrations/starlette.py
```diff
@@ -516,12 +516,16 @@ async def raw_data(self):
...
-        curl -X POST http://localhost:8000/upload/somethign  -F username=Julian -F password=hello123
+        curl -X POST http://localhost:8000/upload/something -H "Content-Type: application/x-www-form-urlencoded" -d "username=kevin&password=welcome123"
+        curl -X POST http://localhost:8000/upload/something  -F username=Julian -F password=hello123
         """
         if multipart is None:
             return None
 
+        # Parse the body first to get it cached, as Starlette does not cache form() as it
+        # does with body() and json() https://github.com/encode/starlette/discussions/1933
+        await self.request.body()
```
sl0thentr0py, line 527: this is getting more and more hacky tbh.

I'd rather prefer we rewrite the whole `extract_request_info` thing to something that makes sense for starlette. The current structure was just copied from flask and might not make sense for another framework. The relevant goal is to extract as much helpful request metadata as possible, the actual logic depends on the framework and what it supports.

sl0thentr0py (changes requested):

sentry_sdk/integrations/starlette.py
antonpirker: I know this looks hacky. But for our integration to get the form data from the request calling .body() and then .form() is the only way to not break our users Starlette projects.

antonpirker: This reading of the `body()` is still there, but I have now refactored `extract_request_info` to make it clearer what is happening.

antonpirker: I guess this is now better understandable and enough "non hacky". What do you think @sl0thentr0py ?

tests/integrations/starlette/test_starlette.py
```diff
@@ -243,82 +247,77 @@ async def test_starlettrequestextractor_cookies(sentry_init):
...
+    side_effect = [_mock_receive(msg) for msg in FORM_RECEIVE_MESSAGES]
+    starlette_request._receive = mock.Mock(side_effect=side_effect)
+
+    extractor = StarletteRequestExtractor(starlette_request)
+
+    form_data = await extractor.form()
+    assert form_data.keys() == PARSED_FORM.keys()
+    assert form_data["username"] == PARSED_FORM["username"]
+    assert form_data["password"] == PARSED_FORM["password"]
+    assert form_data["photo"].filename == PARSED_FORM["photo"].filename
```
hasier, line 282: Maybe call `await request.stream()` or `body()` right at the end to ensure that there are no regressions? Maybe it'd be relevant to do the same in the rest of tests as well, just to make sure in all cases we still have access to the whole request stream/body?

tests/integrations/starlette/test_starlette.py
```diff
@@ -243,82 +247,77 @@ async def test_starlettrequestextractor_cookies(sentry_init):
...
-    with mock.patch(
-        "starlette.requests.Request.stream",
-        return_value=AsyncIterator(json.dumps(BODY_JSON)),
-    ):
-        starlette_request = starlette.requests.Request(SCOPE)
-        extractor = StarletteRequestExtractor(starlette_request)
+    await extractor.request.form()
 
-        assert await extractor.raw_data() == bytes(json.dumps(BODY_JSON), "utf-8")
+    with pytest.raises(RuntimeError):
```
sl0thentr0py, line 319: I don't understand this test, shouldn't we test that our stuff doesn't throw even if we try to access both form and body?

hasier: I think this was following [my suggestion](https://github.com/getsentry/sentry-python/issues/1675#issuecomment-1302200430). My idea was to have a test that crashes when you upgrade to a `starlette` version where they have fixed [the caching inconsistency](https://github.com/encode/starlette/discussions/1933), as that means you may need to double check your implementation and maybe simplify the code, right @antonpirker?

antonpirker: Yes. This test calls directly the `form()` from the Starlette request and not the `form()` from the extractor, that "fixes" the caching with internally calling `body()`.

sl0thentr0py: ahh gotcha, thx for the explanation both!

sl0thentr0py (approved):
alright lgtm

Issue #1721, opened 2022-11-03
Performance regression with Django


tevansuk:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.10.1

### Steps to Reproduce

Updating from 1.9.8 to 1.10.1 has introduced a huge performance regression which seems to be related to django template processing.

One test case (run with profiling using cProfile) went from 22 seconds to 19 minutes 18 seconds.

Python 3.9
Django 2.2.
sentry-sdk 1.9.8 / 1.10.1
pytest 7.2.0
pytest-django 4.5.2
coverage 6.5.0

The only library that changes version between the two tests is sentry-sdk.

Profiling reveals a huge number of additional function calls - approximately 7x as many as in sentry-sdk 1.9.8.

I've included snakeviz heatmaps of the profiling results drilled down to the test case to show where the time is being spent. I'd rather not directly put the `.prof` files up here.

### Expected Result

With the older `sentry-sdk==1.9.8`

```
pytest -m "not functional" --junitxml=qa_reports/junit/unit-tests.xml --reuse-db --profile --profile-svg tests/unit/portal/test_ip_blocked.py::LoginTestCase::test_login_from_NOT_BLOCKED_IP_SUCCEEDS
===================================== test session starts =====================================
platform linux -- Python 3.9.15, pytest-7.2.0, pluggy-1.0.0 -- /venv/bin/python
cachedir: .pytest_cache
django: settings: project.settings (from option)
rootdir: /app, configfile: pyproject.toml
plugins: celery-4.3.1, forked-1.4.0, mock-3.10.0, profiling-1.7.0, requests-mock-1.10.0, cov-4.0.0, xdist-2.5.0, django-4.5.2
collected 1 item                                                                              

tests/unit/portal/test_ip_blocked.py::LoginTestCase::test_login_from_NOT_BLOCKED_IP_SUCCEEDS PASSED [100%]

------------------ generated xml file: /app/qa_reports/junit/unit-tests.xml -------------------
Profiling (from /app/prof/combined.prof):
Thu Nov  3 17:43:40 2022    /app/prof/combined.prof

         5797777 function calls (5687652 primitive calls) in 7.768 seconds

   Ordered by: cumulative time
   List reduced from 5819 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    7.783    7.783 runner.py:109(pytest_runtest_protocol)
[...]

PR #1725, opened 2022-11-04
Performance optimizations

antonpirker:
Our Django signals integrations added a lot of overhead in some cases. 
See: https://github.com/getsentry/sentry-python/issues/1721

This code runs now over an order of magnitude faster for getting the name of a partials.

### DIFF
--- a/sentry_sdk/integrations/django/signals_handlers.py
+++ b/sentry_sdk/integrations/django/signals_handlers.py
@@ -25,7 +25,7 @@ def _get_receiver_name(receiver):
     elif hasattr(
         receiver, "func"
     ):  # certain functions (like partials) dont have a name
-        name = "functools.partial(<function " + receiver.func.__name__ + ">)"
+        name = "functools.partial(<function " + receiver.func.__name__ + ">)"  # type: ignore
 
     if (
         name == ""
--- a/sentry_sdk/integrations/django/signals_handlers.py
+++ b/sentry_sdk/integrations/django/signals_handlers.py
@@ -25,7 +25,7 @@ def _get_receiver_name(receiver):
     elif hasattr(
         receiver, "func"
     ):  # certain functions (like partials) dont have a name
-        name = "functools.partial(<function " + receiver.func.__name__ + ">)"  # type: ignore
+        name = "partial(<function " + receiver.func.__name__ + ">)"  # type: ignore
 
     if (
         name == ""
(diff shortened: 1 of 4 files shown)


sentry_sdk/integrations/django/signals_handlers.py
```diff
@@ -19,13 +19,17 @@ def _get_receiver_name(receiver):
...
     if hasattr(receiver, "__qualname__"):
-        name += receiver.__qualname__
+        name = receiver.__qualname__
     elif hasattr(receiver, "__name__"):  # Python 2.7 has no __qualname__
-        name += receiver.__name__
+        name = receiver.__name__
+    elif hasattr(
+        receiver, "func"
+    ):  # certain functions (like partials) dont have a name
+        name = "partial(<function " + receiver.func.__name__ + ">)"  # type: ignore
```
sl0thentr0py, line 28: can the func itself be a partial?

antonpirker: F*ck... Probably... I will check on Monday

antonpirker: I will check now.

antonpirker: Ok, `func` appears to always be the function:

```
Python 3.10.8 (main, Oct 13 2022, 09:48:40) [Clang 14.0.0 (clang-1400.0.29.102)] on darwin
>>> def f(a, b, c, x):
...     return 1000*a + 100*b + 10*c + x
... 
>>> f
<function f at 0x102752050>
>>> from functools import partial
>>> g = partial(f, 3)
>>> h = partial(g, 1)
>>> g
functools.partial(<function f at 0x102752050>, 3)
>>> h
functools.partial(<function f at 0x102752050>, 3, 1)
>>> h(3, 4)
3134
>>> h.func
<function f at 0x102752050>
>>> g.func
<function f at 0x102752050>
```

sl0thentr0py: cool, thx for checking

sl0thentr0py (approved):
change is fine, but why was it slow in the first place?

antonpirker:
Because `str(bla)` is very very slow compared to `bla.__name__`. 

I will see if the user can test this and confirm that it makes the situation way better before I merge.

PR #1729, opened 2022-11-08
ref(profiling): Use sleep scheduler by default

Zylphrex:
The sleep scheduler is the most reliable of the available schedulers, make it the default.


sl0thentr0py (approved):
yay science

PR #1297, opened 2022-01-11
The Falcon integration for Falcon 3.0.1

bartolootrit:
The Falcon integration passed the tests for Falcon 3.0.1, 2.0.0, 1.4.0 (Python 3.8) and Falcon 2.0.0, 1.4.0 (Python 2.7)

PR #1733, opened 2022-11-10
feat(falcon): Update of Falcon Integration

antonpirker:
Made Falcon integration support version 3.0.1.
Also added new Falcon versions to the test suite.

Used to be this PR: https://github.com/getsentry/sentry-python/pull/1297

Co-authored-by: bartolootrit <bartolootrit@users.noreply.github.com>


sentry_sdk/integrations/falcon.py
```diff
@@ -58,16 +73,27 @@ def raw_data(self):
...
-            return self.request.media
-        except falcon.errors.HTTPBadRequest:
-            # NOTE(jmagnusson): We return `falcon.Request._media` here because
-            # falcon 1.4 doesn't do proper type checking in
-            # `falcon.Request.media`. This has been fixed in 2.0.
-            # Relevant code: https://github.com/falconry/falcon/blob/1.4.1/falcon/request.py#L953
-            return self.request._media
+    if FALCON3:
+
+        def json(self):
```
sl0thentr0py, line 78: I'd just merge this into two `hasattr` checks for `media/_media` and pick whichever is available

sl0thentr0py (changes requested):

sentry_sdk/integrations/falcon.py
```diff
@@ -28,6 +28,21 @@
     raise DidNotEnable("Falcon not installed")
 
 
+# In Falcon 3.0 `falcon.api_helpers` is renamed to `falcon.app_helpers`
+# and `falcon.API` to `falcon.App`
+try:
+    import falcon.app  # type: ignore
+    import falcon.app_helpers  # type: ignore
```
dannyhajj, line 35: It would be preferable to attempt to import `falcon.app_helpers` first, then import `falcon.api_helpers` if the first fails.
Otherwise, the current code will keep complaining with deprecation warnings that api_helpers has been renamed.

sentry_sdk/integrations/falcon.py
bartolootrit: Falcon 3 and Falcon 2 `Request` classes have `_media` attribute and `media` property. How would this `hasattr` check look like?

antonpirker:
Hey @bartolootrit !
Thanks for your continued work. 
If you need help with running the tests locally, than I am here to help

sentry_sdk/integrations/falcon.py
antonpirker: I guess this is then fine as it is. It is not the most elegant code, but it should work.

antonpirker:
Hey @bartolootrit !
Everything looks great now. 
What about the comment from @dannyhajj about the order of imports above? Is this something we need to address, or can we merge this as is?

PR #1746, opened 2022-11-17
Expose proxy_headers as top level config and use in ProxyManager

### LINKED ISSUE #676: Support for Proxy Authentication
Hi there! Thanks for all you do with Sentry -- it adds incredible value to our team.

I have a question about support for HTTP proxy authentication in `sentry_sdk`.

**TL;DR:** Is proxy authentication supported? If not, can it be? Without support for proxy auth, we can't use Sentry from a significant portion of our services.

**More details:**

I see that we can pass a `https_proxy` kwarg to `sentry_sdk.init()`, such as:

```
sentry_sdk.init("https://...@foo.ingest.sentry.io/...", https_proxy="http://user:pass@proxy")
```

However, it looks like that proxy string eventually gets passed to `urllib3.ProxyManager()` here: https://github.com/getsentry/sentry-python/blob/0da369f839ee2c383659c91ea8858abcac04b869/sentry_sdk/transport.py#L293

Unfortunately, `urllib3.ProxyManager` does not appear to respect credentials passed in this string. The [documentation](https://urllib3.readthedocs.io/en/latest/reference/#urllib3.poolmanager.ProxyManager) isn't super clear on this, but it does allude to it in the description of the `proxy_headers` kwarg: "Could be used for proxy authentication." We verified with packet captures that `urllib3.ProxyManager` is not adding a `Proxy-Authorization` HTTP header to requests when the proxy string includes credentials.

Thanks so much for your consideration and help with this!

sl0thentr0py:
fixes #676

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -81,6 +81,7 @@ class ClientConstructor(object):
         auto_session_tracking=True,  # type: bool
         send_client_reports=True,  # type: bool
         _experiments={},  # type: Experiments  # noqa: B006
+        proxy_headers=None,  # type: Optional[Dict[str, str]]
     ):
         # type: (...) -> None
         pass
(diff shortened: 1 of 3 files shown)


antonpirker (approved):
lgtm

PR #1597, opened 2022-09-03
Feat: add Starlite integration

Goldziher:
This PR is a WIP. I'm adding it as a draft for two reasons:

1. to allow for comments.
2. to allow maintainers of Sentry to merge it into a local branch and take over the work.

It addresses #1549

PR #1748, opened 2022-11-18
Feat: add Starlite integration

### LINKED ISSUE #1549: [Feat] Starlite Integration
### Problem Statement

[Starlite](https://github.com/starlite-api/starlite) us an ASGI framework based partially on Starlette. The current Starlette integration is not compatible with it because Starlite has its own routing and exception handling solutions in place. So the feature request here is to create an integration for Starlite.

### Solution Brainstorm

Starlite has both exception handlers and life-cycle hooks (see [the docs](https://starlite-api.github.io/starlite/usage/0-the-starlite-app/0-the-starlite-app/) for more details) and these can be used to create a sentry integration. Furthermore, if need be we can add more lifecycle hooks to Starlite as required.

gazorby:
Supplant #1597 and resolves #1549.


Goldziher:
I guess this needs to be updated @gazorby

sentry_sdk/integrations/starlite.py
```diff
@@ -0,0 +1,261 @@
...
+            Middleware,
+            Receive,
+            Scope,
+            Send,
+            WebSocketReceiveMessage,
+        )
+
+        from sentry_sdk._types import Event
+except ImportError:
+    raise DidNotEnable("Starlette is not installed")
```
Goldziher, line 35: should be Starlite

```suggestion
    raise DidNotEnable("Starlite is not installed")
```

setup.py
```diff
@@ -61,6 +61,7 @@ def get_file_text(file_name):
         "chalice": ["chalice>=1.16.0"],
         "httpx": ["httpx>=0.16.0"],
         "starlette": ["starlette>=0.19.1"],
+        "starlite": ["starlite>=1.39.0"],
```
Goldziher, line 64: can be updated to a later version I think

Goldziher (approved):

setup.py
gazorby: latest?

antonpirker:
Hey @Goldziher and @gazorby ! I ran the test suite and the test for Python 3.7 and 3.8 are failing. Could you have a look: https://github.com/getsentry/sentry-python/actions/runs/3866938112/jobs/6611898016

gazorby:
Should be fixed now @antonpirker !

sentry_sdk/integrations/starlite.py
```diff
@@ -0,0 +1,263 @@
...
+
+    user_info = retrieve_user_from_scope(scope)
+    if user_info and isinstance(user_info, dict):
+        with hub.configure_scope() as sentry_scope:
+            sentry_scope.set_user(user_info)
+
+    event, hint = event_from_exception(
+        exc,
+        client_options=hub.client.options if hub.client else None,
+        mechanism={"type": StarliteIntegration.identifier, "handled": True},
```
antonpirker, line 261: `handled` should be set to `False` here, because the exception was not handled by the users code and thus was captured by Sentry.

sentry_sdk/integrations/starlite.py
```diff
@@ -0,0 +1,263 @@
...
+
+    return None
+
+
+def exception_handler(exc: Exception, scope: "Scope", _: "State"):
+    hub = Hub.current
+    if hub.get_integration(StarliteIntegration) is None:
+        return
+
+    user_info = retrieve_user_from_scope(scope)
```
antonpirker, line 253: I would prefer to check for `_should_send_default_pii()` here instead of checking for it in `retrieve_user_from_scope` because it is easier to read.

antonpirker: (But this is kind of nitpicking...)

antonpirker (changes requested):
First round of review.

sentry_sdk/integrations/starlite.py
```diff
@@ -0,0 +1,264 @@
...
+                route_handler = scope.get("route_handler")
+
+                request_info = event.get("request", {})
+                request_info["content_length"] = len(scope.get("_body", b""))
+                if _should_send_default_pii():
+                    request_info["cookies"] = extracted_request_data["cookies"]
+                if request_data is not None:
+                    request_info["data"] = request_data
+
+                tx_name = route_handler.name or transaction_from_function(
```
gazorby, line 207: I think that's because `route_handler.name` is `None` here

gazorby, line 207: It's because `route_handler.fn` is a `starlite.Ref` in which the actual route handler fn is wrapped, which may itself be wrapped in a `functools.partial`. Will fix that asap!

setup.py
Goldziher: 1.48.1

sentry_sdk/integrations/starlite.py
antonpirker: We have  a (not really nice an clean) function that tries to get a name from something that could be a function or a partial, maybe this can inspire a solution: https://github.com/getsentry/sentry-python/blob/b300b10df5aff2f4822b4ba8a75e62ee5f8798fb/sentry_sdk/integrations/django/signals_handlers.py#L30

antonpirker:
I started the tests again, and will review again tomorrow! Thanks for the quick fixes!

sentry_sdk/integrations/starlite.py
antonpirker: Now we have nice transaction names!

antonpirker:
Ok, we have now nice transaction names!

Screenshot 2023-01-11 at 09 58 37

antonpirker:
All the tests are also green now! Nice work!

The things missing now @gazorby is in my two comments in the review. (One is just nitpicking, so feel free to ignore)

One thing I also noticed is that there is only `endpoint` in `transaction_style`. In other integrations we have also `url`, but most of the time `endpoint` is the default, so I guess this is fine. If a lot of users demand url style transaction names in the future, we can always add it.

tests/utils/test_transaction.py
```diff
@@ -18,3 +38,26 @@ def test_transaction_from_function():
...
+    assert (
+        x(my_partial) == "partial(<function tests.utils.test_transaction.my_partial>)"
+    )
+    assert (
+        x(my_partial_lambda)
+        == "partial(<function tests.utils.test_transaction.<lambda>>)"
+    )
+
+
+@pytest.mark.skipif(sys.version_info < (3, 4))
```
antonpirker, line 51: You have to give a `reason` here. See for example https://github.com/getsentry/sentry-python/blob/master/tests/integrations/django/test_basic.py#L804

antonpirker (changes requested):

antonpirker:
One last thing @gazorby . See above!

tests/utils/test_transaction.py
gazorby: Yup, got it when testing locally, didn't push the fix yet

gazorby:
@antonpirker As I refactored `utils.transaction_from_function` to support partials, `django.signal_handlers._get_receiver_name` 
(and maybe other stuff as well) may benefit from it, should we deduplicate here or do this in another PR?

antonpirker:
please in another PR.

antonpirker:
I think we are good to go! Amazing work @gazorby, thanks a lot!

We will try to ship a new version of the SDK this week.

If you send me your shipping address to anton (dot) pirker {at} sentry .dot. io I can send you a little bit of swag as a small "thank you".

gazorby:
Thanks @antonpirker!

antonpirker:
Starlite support has been released: 
https://github.com/getsentry/sentry-python/blob/master/CHANGELOG.md

antonpirker:
No problem! Do not forget to send me your shipping address and t-shirt size to anton.pirker@sentry.io

PR #1751, opened 2022-11-22
Move set_transaction_name out of event processor in fastapi/starlette

sl0thentr0py:
Doing this in the event processor breaks the transaction name contract if the user sets the transaction name manually on scope.


antonpirker (approved):
Tested it with my FastAPI demo projects. Looks good!

PR #1756, opened 2022-11-25
fix(profiling): Resolve inherited method class names

Zylphrex:
Methods may be inherited from a parent class. If multiple classes inherit from the same class and uses the inherited method, we'd want it to report the parent class's name instead of the individual child classes since they'd have the same filename and lineno of the parent class and not the children.

### DIFF
--- a/sentry_sdk/profiler.py
+++ b/sentry_sdk/profiler.py
@@ -211,7 +211,9 @@ def get_frame_name(frame):
             and f_code.co_varnames[0] == "self"
             and "self" in frame.f_locals
         ):
-            return "{}.{}".format(frame.f_locals["self"].__class__.__name__, name)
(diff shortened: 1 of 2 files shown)


sl0thentr0py (approved):

sl0thentr0py:
just a footnote: 
I'm wondering if we're missing some standardization in extracting meaningful names from frames, all of these seems like extra logic that should probably live somewhere more general upstream?

Zylphrex:
> just a footnote: I'm wondering if we're missing some standardization in extracting meaningful names from frames, all of this seems like extra logic that should probably live somewhere more general upstream?

Yeah, this could potentially be something we port over to errors as well since names like `__init__` isn't particularly meaningful without the class name. This was a major pain point when looking at profiles but if this can enhance the experience for errors, I think it makes sense for this logic to be more general.

PR #1757, opened 2022-11-25
feat(profiling): Add profiling instrumentation for ASGI apps

Zylphrex:
This makes the minimal changes to allow profiling ASGI apps. To allow profiling all transactions, there will be more work needed as there is ongoing work to decouple profiles and transactions, that will be pushed back until we have a clearer idea what that'll look like before making major changes.

### DIFF
--- a/sentry_sdk/integrations/asgi.py
+++ b/sentry_sdk/integrations/asgi.py
@@ -14,6 +14,7 @@ from sentry_sdk.consts import OP
 from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk.integrations._wsgi_common import _filter_headers
 from sentry_sdk.integrations.modules import _get_installed_modules
+from sentry_sdk.profiler import start_profiling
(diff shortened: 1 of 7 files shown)


sentry_sdk/integrations/asgi.py
```diff
@@ -175,7 +176,7 @@ async def _run_app(self, scope, callback):
 
                     with hub.start_transaction(
                         transaction, custom_sampling_context={"asgi_scope": scope}
-                    ):
+                    ), start_profiling(transaction, hub):
```
Zylphrex, line 179: I'm not too familiar with how ASGI works but I tried to run a Django app in ASGI using gunicorn + uvicorn and I noticed that this happens on the main thread and the request is handled on another thread. Ideally, we can detect the thread handling the request so we can show it in the UI by default. Right now, it's just showing the main thread which isn't very helpful.

Zylphrex:
gonna close this and target each framework independently as there's some monkey patching involved to get each one working

PR #1764, opened 2022-11-29
feat(profiling): Introduce active thread id on scope

Zylphrex:
Upto this point, simply taking the current thread when the transaction/profile was started was good enough. When using ASGI apps with non async handlers, the request is received on the main thread. This is also where the transaction or profile was started. However, the request is handled on another thread using a thread pool. To support this use case, we want to be able to set the active thread id on the scope where we can read it when we need it to allow the active thread id to be set elsewhere.

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -429,7 +429,7 @@ class _Client(object):
 
             if is_transaction:
                 if profile is not None:
-                    envelope.add_profile(profile.to_json(event_opt, self.options))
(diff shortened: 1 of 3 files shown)


Zylphrex:
@sl0thentr0py do you know why the `CI / Lint Sources` job is failing? It looks unrelated to my changes.

sl0thentr0py:
@Zylphrex this was the problem, pushed fix
https://stackoverflow.com/a/58121288

sl0thentr0py (approved):

PR #1766, opened 2022-11-30
Add instrumenter config to switch between Otel and Sentry instrumentation.

antonpirker:


sentry_sdk/consts.py
```diff
@@ -57,6 +87,7 @@ def __init__(
         server_name=None,  # type: Optional[str]
         shutdown_timeout=2,  # type: float
         integrations=[],  # type: Sequence[Integration]  # noqa: B006
+        instrumenter=INSTRUMENTER.SENTRY,  # type: Optional[str]
```
sl0thentr0py, line 90: new arguments should always go at the end

sentry_sdk/api.py
```diff
@@ -210,5 +211,5 @@ def start_transaction(
     transaction=None,  # type: Optional[Transaction]
     **kwargs  # type: Any
 ):
-    # type: (...) -> Transaction
+    # type: (...) -> Union[Transaction, NoOpSpan]
```
sl0thentr0py, line 214: does this count as a breaking change then?
In ruby this could always return nil so it wasn't a problem.

sentry_sdk/tracing.py
```diff
@@ -213,6 +217,15 @@ def start_child(self, **kwargs):
         trace id, sampling decision, transaction pointer, and span recorder are
         inherited from the current span/transaction.
         """
+        hub = self.hub or sentry_sdk.Hub.current
+        client = hub.client
+
+        instrumenter = kwargs.get("instrumenter", INSTRUMENTER.SENTRY)
+        configuration_instrumenter = client and client.options["instrumenter"]
+
+        if instrumenter != configuration_instrumenter:
```
sl0thentr0py, line 226: why is this check duplicated here? imo doing it in hub.py is sufficient?

sentry_sdk/hub.py
```diff
@@ -464,6 +465,12 @@ def start_span(
         for every incoming HTTP request. Use `start_transaction` to start a new
         transaction when one is not already in progress.
         """
+        instrumenter = kwargs.get("instrumenter", INSTRUMENTER.SENTRY)
```
sl0thentr0py, line 468: I would prefer an explicit arg `instrumenter=INSTRUMENTER.SENTRY` if you remove the redundant check in tracing.py, then this doesn't need to passed down to the constructor.

sl0thentr0py (changes requested):

sentry_sdk/api.py
antonpirker: I think it is not a breaking change. A NoOpSpan has the same interface as a normal Span, but just does nothing. But I think I have to think more about the implications.

sentry_sdk/hub.py
antonpirker: Main reason for having it in `kwargs` was that I did not want to change the function definition to not break user code eventually. But after thinking a bit I guess when setting a default in the function parameter there is little chance to break user code

sentry_sdk/api.py
sl0thentr0py: ok yea I think it should be fine, are there tests for NoOpSpan later in some other PR?

sl0thentr0py (approved):
if tests are in another PR, it's fine, otherwise please add some tests for checking that the no-ops work correctly

sentry_sdk/api.py
antonpirker: Not yet, but I will add some.

sentry_sdk/tracing.py
antonpirker: As discussed I will leave this check here, to disable all manual instrumentation when the otel instrumenter is selected.

PR #1767, opened 2022-11-30
API Changes for Otel integration

antonpirker:
Allowing:
* set contexts in spans
* start/end times for spans 
* extract sentry trace information


antonpirker:
Closing this PR in favor of atomic PRs per api change.

PR #1769, opened 2022-11-30
Add API to set arbitrary context in Transaction.

antonpirker:


sl0thentr0py (changes requested):
this should be on `Transaction`, not `Span`

antonpirker:
Oh, did not know that, but makes sense. Changed it.

sl0thentr0py (approved):

PR #1770, opened 2022-11-30
Add API to set custom Span timestamps

antonpirker:
* When creating a Span/Transaction a `start_timestamp` can be given. 
* When a Span/Transaction is finished a `end_timestamp` can be given.

Both default to "now" in UTC.

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -473,7 +473,7 @@ class Span(object):
         # type: () -> bool
         return self.status == "ok"
 
-    def finish(self, hub=None, **kwargs):
+    def finish(self, hub=None, end_timestamp=None):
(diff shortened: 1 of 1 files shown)


sentry_sdk/tracing.py
```diff
@@ -472,8 +473,8 @@ def is_success(self):
         # type: () -> bool
         return self.status == "ok"
 
-    def finish(self, hub=None):
-        # type: (Optional[sentry_sdk.Hub]) -> Optional[str]
+    def finish(self, hub=None, **kwargs):
```
sl0thentr0py, line 476: just make it explicit `end_timestamp=None`, I don't like the opaqueness of kwargs and it's only one argument here anyway

antonpirker: done.

sl0thentr0py (approved):

PR #1771, opened 2022-11-30
Add function to extract sentry trace info from traceparent string.

antonpirker:
This is basically the opposite of `to_traceparent()`.


antonpirker:
Oh, good to know! Will close and use the existing thing!

PR #1779, opened 2022-12-09
feat(profiling): Enable profiling for all transactions

Zylphrex:
Up to now, we've only been profiling WSGI transactions. This change will enable profiling for all transactions.

#1778 is required for the active thread id to be set correctly for ASGI transactions


Zylphrex:
Actually going to close this PR for now. There are some optimizations I want to get in first and that'll conflict with the changes here quite a bit.

PR #1780, opened 2022-12-09
fix(ci): Github action checks failing 

Zylphrex:
The checks are failing for 2 reasons

1. Github actions dropped python3.6 support on the latest hosted runners.
   https://github.com/actions/setup-python/issues/544#issuecomment-1332535877
2. New release of tox was validation the python version in the environment name
   and the trailing framework version being used in the environment name was
   being treated as a python version and validated causing an issue.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,97 +9,97 @@ envlist =
     py{2.7,3.4,3.5,3.6,3.7,3.8,3.9,3.10}
 
     # === Integrations ===
-    # General format is {pythonversion}-{integrationname}-{frameworkversion}
+    # General format is {pythonversion}-{integrationname}-v{frameworkversion}
     # 1 blank line between different integrations
     # Each framework version should only be mentioned once. I.e:
-    #   {py3.7,py3.10}-django-{3.2}
-    #   {py3.10}-django-{4.0}
+    #   {py3.7,py3.10}-django-v{3.2}
+    #   {py3.10}-django-v{4.0}
     # instead of:
-    #   {py3.7}-django-{3.2}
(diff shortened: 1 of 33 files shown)


tox.ini
```diff
@@ -359,19 +359,19 @@ basepython =
...
 
     ; https://github.com/more-itertools/more-itertools/issues/578
-    py3.5-flask-{0.11,0.12}: pip install more-itertools<8.11.0
+    py3.5-flask-v{0.11,0.12}: pip install more-itertools<8.11.0
 
     ; use old pytest for old Python versions:
     {py2.7,py3.4,py3.5}: pip install pytest-forked==1.1.3
 
-    py.test --durations=5 {env:TESTPATH} {posargs}
+    python -m pytest --durations=5 {env:TESTPATH} {posargs}
```
Zylphrex, line 377: running `py.text` as an executable was having an issue on tox4 where it fails to import tests

tox.ini
```diff
@@ -9,97 +9,97 @@ envlist =
     py{2.7,3.4,3.5,3.6,3.7,3.8,3.9,3.10}
 
     # === Integrations ===
-    # General format is {pythonversion}-{integrationname}-{frameworkversion}
+    # General format is {pythonversion}-{integrationname}-v{frameworkversion}
```
Zylphrex, line 12: If we suffix the environment name with just the framework version, tox4 will try to validate the python version from it and treats the standalone version number as a python version. By adding the `v`, it prevents tox from doing this.

antonpirker (approved):

PR #1791, opened 2022-12-19
ref(profiling): Remove sample buffer from profiler

Zylphrex:
The sample buffer kept 30s of samples around in memory. This introduces a noticeable memory overhead on systems with less memory available. This change removes the buffer and directly writes to the profile itself where the sample is processed on the fly instead of at the end.


sentry_sdk/profiler.py
```diff
@@ -480,121 +528,78 @@ def make_sampler(self):
...
+
+            # This is the number of profiles we want to pop off.
+            # It's possible another thread adds a new profile to
+            # the list and we spend longer than we want inside
+            # the loop below.
+            #
+            # Also make sure to set this value before extracting
+            # frames so we do not write to any new profiles that
+            # were started after this point.
+            new_profiles = len(self.new_profiles)
```
sl0thentr0py, line 554: mostly just trying to understand, why does this solve the locking/concurrent access issue?
Is there a name for this pattern where we use separate references for reading/writing instead of using a lock?

sl0thentr0py (approved):
lgtm, asked a question for educational purposes

sentry_sdk/profiler.py
Zylphrex: I'm not sure if there's a name for this pattern but the idea is as follows.

The core of the problem is that we have multiple threads trying to append a value to a list, and a single thread that reads that list and may pop values from the list.

The naive implementation here would be something like

```python
def start_profiling(self, profile): 
  "Happens in one of the handler threads"
  self.active_profiles.add(profile)

def stop_profiling(self, profile):
  "Happens in one of the handler threads"
  self.active_profiles.remove(profile)

def _sample_stack():
  "Happens on the profiler thread"
  sample = <take a sample>
  for profile in self.active_profiles:
    profile.write(sample)
```

However, because the modification happens in the handler thread and the iteration happens on the profiler thread, we can potentially run into a issue where while iterating the active profiles to write the samples, one of the handler threads stops profiling resulting in a `RuntimeError: Set changed size during iteration`.

So to get around this issue, the idea is as follows.
1. We remove the iterator on the shared data structure.
2. As a consequence of 1, we now have to buffer the new profiles in a separate data structure because we still need a way of iterating over the active profiles without it's size changing during iteration.

To accomplish this,
1. New profiles are now in a `deque` which allows us to push profiles in order and pop a fixed number of profiles from without the usage of an iterator, this avoids the `RuntimeError: Set changed size during iteration`.
2. Active profiles set are now updated only by the profiler thread, so we can control when the set size changes as we can be certain no other threads will change its size.

sl0thentr0py: oh gotcha, we're solving this by implicit ownership by threads on the two data structures, nice.
Thanks for the detailed explanation!

PR #1794, opened 2022-12-20
feat(django): update django app to be compatible for Django 4.x

DilLip-Chowdary-Codes:
- [The django.db.backends.postgresql_psycopg2 module is removed in Django3.0](https://docs.djangoproject.com/en/4.1/releases/3.0/#features-removed-in-3-0)
- [The get_response argument for django.utils.deprecation.MiddlewareMixin.__init__() is required and doesn’t accept None](https://docs.djangoproject.com/en/4.1/releases/4.0/#features-removed-in-4-0)


antonpirker:
As the only change left has a alias since Django 1.9 (and we officially support only Django 1.11+) this should be good to merge!

antonpirker (approved):

PR #1809, opened 2022-12-27
Attempt to use MonkeyType on sentry-python

vbro:
Benefit seems minimal since this codebase already has lots of type hints


sl0thentr0py:
hey @vbro thanks but we have to support python2.7 so we're still using type hints in comments.
That's a much bigger refactor and we'll need to do that first before we can do fancy new typing stuff.
Closing, sorry!

PR #1810, opened 2022-12-27
feat(integrations): Add method and path pattern transaction style for the Django integration

rynmlng:
The method and path pattern for transaction styles includes the HTTP method as well as request path. This new transaction style for Django allows events to be grouped by this segmentation.


rynmlng:
@antonpirker how can I help move this forward?

sentry_sdk/integrations/django/__init__.py
```diff
@@ -79,7 +79,7 @@ def is_authenticated(request_user):
         return request_user.is_authenticated
 
 
-TRANSACTION_STYLE_VALUES = ("function_name", "url")
+TRANSACTION_STYLE_VALUES = ("function_name", "method_and_path_pattern", "url")
```
antonpirker, line 82: Please call it "url_with_method" so we see it is related to "url".

antonpirker (changes requested):

antonpirker:
Hey @rynmlng thanks for the contribution! 

Sorry for the delay, but we have kind of a lot on our plates right now.

antonpirker:
Hey @rynmlng !
I had a look at your contribution, and I am afraid I can not merge this. This adds to much overhead in maintaining and also maintaining the documentation.

But there is a solution for your problem, you can add the method in a `before_send` hook. Something like this:

```python
import sentry_sdk

def add_method_to_transaction_name(event, hint):
    # modify event here
    event["transaction"] = f"{event["request"]['method']} {event["transaction"]}"
    return event

sentry_sdk.init(
    dsn="...",
    before_send=add_method_to_transaction_name,
)
```

antonpirker:
The solution above is for error events. (And there is a small bug `"request"` in the f-string should be `'request'`.

antonpirker:
If you want to change the transaction name for all kind of events (errors and transactions (the stuff ending up in the "performance" main menu item on sentry.io)) you can write an `event_processor`.

This is a little bit more complicated, because you first have to fetch the HTTP method in a Django middleware because the `event_processor` has no access to the Django request:

1: Middleware to save the http method:

```python
# middleware.py
import sentry_sdk

def sentry_middleware(get_response):
    def middleware(request):
            # Saving HTTP method into Sentry scope
            scope.set_extra("http_method", request.method)
        
        response = get_response(request)
        return response

    return middleware
```

2. Add the event processor:
```python
sentry_sdk.init(
    dsn="...",
)

def make_nice_transaction_name(event, hint):
    # Get HTTP method from Sentry scope
    event["transaction"] = f"{event['extra']['http_method']} {event['transaction']}"
    return event

    scope.add_event_processor(make_nice_transaction_name)
```

antonpirker:
Hope this helps!

PR #1250, opened 2021-11-15
record span and breadcrumb when Django opens db connection

chdsbd:
Here's a screenshot of a trace where Django opens a new connection

Screen Shot 2021-11-14 at 9 04 39 PM

fixes #1249

PR #1818, opened 2023-01-02
add span for response rendering

chdsbd:
Response rendering can be a source of slow performance in Django, especially with Django rest framework and serializers. This change adds a span to track this.

Screenshot 2023-01-02 at 4 30 46 PM

related: #1250

### DIFF
--- a/sentry_sdk/integrations/django/views.py
+++ b/sentry_sdk/integrations/django/views.py
@@ -32,7 +32,9 @@ def patch_views():
     def sentry_patched_render(self):
         # type: (SimpleTemplateResponse) -> Any
         hub = Hub.current
-        with hub.start_span(op=OP.VIEW_RESPONSE_RENDER, description="serialize response"):
+        with hub.start_span(
+            op=OP.VIEW_RESPONSE_RENDER, description="serialize response"
+        ):
             return old_render(self)
 
     @_functools.wraps(old_make_view_atomic)
--- a/sentry_sdk/integrations/django/views.py
+++ b/sentry_sdk/integrations/django/views.py
@@ -27,13 +27,13 @@ def patch_views():
     from sentry_sdk.integrations.django import DjangoIntegration
 
     old_make_view_atomic = BaseHandler.make_view_atomic
-    original_render = SimpleTemplateResponse.render
+    old_render = SimpleTemplateResponse.render
 
-    def wrapped_render(self):
+    def sentry_patched_render(self):
         # type: (SimpleTemplateResponse) -> Any
         hub = Hub.current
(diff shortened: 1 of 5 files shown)


sentry_sdk/integrations/django/views.py
```diff
@@ -23,9 +23,17 @@ def patch_views():
     # type: () -> None
 
     from django.core.handlers.base import BaseHandler
+    from django.template.response import SimpleTemplateResponse
     from sentry_sdk.integrations.django import DjangoIntegration
 
     old_make_view_atomic = BaseHandler.make_view_atomic
+    original_render = SimpleTemplateResponse.render
```
sl0thentr0py, line 30:
```suggestion
    old_render = SimpleTemplateResponse.render
```

sentry_sdk/integrations/django/views.py
```diff
@@ -23,9 +23,17 @@ def patch_views():
...
+
+    def wrapped_render(self):
```
sl0thentr0py, line 32:
```suggestion
    def sentry_patched_render(self):
```

sentry_sdk/integrations/django/views.py
```diff
@@ -23,9 +23,17 @@ def patch_views():
...
+        # type: (SimpleTemplateResponse) -> Any
+        hub = Hub.current
+        with hub.start_span(op="django.response", description="render"):
```
sl0thentr0py, line 35: we're trying to be more consistent in our [Span Ops naming](https://develop.sentry.dev/sdk/performance/span-operations/), so can you make the op `view.response` or `view.response.render` and add it as a constant [here](https://github.com/getsentry/sentry-python/blob/e2e0de10a0614bb8fb8768757849dce584f381cf/sentry_sdk/consts.py#L73)?

sl0thentr0py: also for the description, is there any more specific information we could add instead of just `render`?

sl0thentr0py (changes requested):
thx for the feature @chdsbd ! 
just some nitpicking for naming consistency with the existing code/conventions

sentry_sdk/integrations/django/views.py
chdsbd: I'm not sure on the description. Basically Django serializes the Python objects to HTML, JSON, or plain text

chdsbd: I've updated the PR, ready for another review when you have a moment @sl0thentr0py

sl0thentr0py (approved):
ty @chdsbd !

PR #1822, opened 2023-01-05
feat(profiling): Better gevent support

Zylphrex:
We're missing frames from gevent threads. Using `gevent.threadpool.ThreadPool`
seems to fix that. The monkey patching gevent does is causing the sampler thread
to run in a greenlet on the same thread as the all other greenlets. So when it
is taking a sample, the sampler is current greenlet thus no useful stacks can be
seen.


Zylphrex:
One potential TODO for gevent would be to look into separating the samples by greenlet instead of thread id. Though this has the caveat that since all the greenlets are in the same thread, we're only able to capture a sample from the current greenlet each time. So if that is split across multiple samples, the final flamechart would appear as if we sample at a much lower frequency.

sl0thentr0py (approved):

PR #1824, opened 2023-01-05
feat(profiling): Enable profiling for ASGI frameworks

Zylphrex:
This enables profiling for ASGI frameworks. When running in ASGI sync views, the
transaction gets started in the main thread then the request is dispatched to a
handler thread. We want to set the handler thread as the active thread id to
ensure that profiles will show it on first render.


sentry_sdk/profiler.py
```diff
@@ -305,13 +304,22 @@ def __init__(
...
 
     def __enter__(self):
         # type: () -> None
+        hub = self.hub or sentry_sdk.Hub.current
+
+        _, scope = hub._stack[-1]
+        old_profile = scope.profile
+        scope.profile = self
+
+        self._context_manager_state = (hub, scope, old_profile)
```
sl0thentr0py, line 321: maybe cleaner to do `configure_scope/scope.profile = profile` in the `start_profiling` ctx manager below? Then you don't need this extra state.

sentry_sdk/integrations/fastapi.py
```diff
@@ -62,6 +65,22 @@ def patch_get_request_handler():
 
     def _sentry_get_request_handler(*args, **kwargs):
         # type: (*Any, **Any) -> Any
+        dependant = kwargs.get("dependant")
```
sl0thentr0py, line 68: @antonpirker can you pls review the fastapi part?

sentry_sdk/integrations/quart.py
```diff
@@ -71,18 +77,62 @@ def setup_once():
         got_request_exception.connect(_capture_exception)
         got_websocket_exception.connect(_capture_exception)
 
-        old_app = Quart.__call__
+        patch_asgi_app()
+        patch_scaffold_route()
+
+
+def patch_asgi_app():
```
sl0thentr0py, line 84: can you split quart out into a separate PR? We can get the quart maintainer to take a look as well then so he's aware of the changes.

sentry_sdk/profiler.py
Zylphrex: That context manager is going to be replaced in an upcoming change to enable profiling for all transactions since I need to hook into the transaction which does not allow me to use the context manager.

sentry_sdk/integrations/quart.py
Zylphrex: https://github.com/getsentry/sentry-python/pull/1830

sentry_sdk/integrations/fastapi.py
```diff
@@ -62,6 +65,22 @@ def patch_get_request_handler():
...
+        if dependant and not asyncio.iscoroutinefunction(dependant.call):
+            old_call = dependant.call
```
antonpirker, line 70: It seems that `dependant.call` could be `None`: https://github.com/tiangolo/fastapi/blob/master/fastapi/dependencies/models.py#L27

Do we have to check for that here to prevent calling None as a function dow at line 80?

sentry_sdk/integrations/fastapi.py
```diff
@@ -62,6 +65,22 @@ def patch_get_request_handler():
...
+
+            def _sentry_call(*args, **kwargs):
+                # type: (*Any, **Any) -> Any
+                hub = Hub.current
+                with hub.configure_scope() as sentry_scope:
+                    if sentry_scope.profile is not None:
+                        sentry_scope.profile.active_thread_id = (
+                            threading.current_thread().ident
```
antonpirker, line 78: Because `current_thread()` can be a dummy object (see https://docs.python.org/3/library/threading.html#threading.current_thread) maybe it is saver to use ` threading.get_ident()`?
(But if the dummy object also has a `.ident` my comment can be ignored)

Zylphrex: `_DummyThread` inherits from `Thread` ([source](https://github.com/python/cpython/blob/main/Lib/threading.py#L1437)) and so both should have the `ident` [property](https://docs.python.org/3/library/threading.html#threading.Thread.ident).

sentry_sdk/integrations/fastapi.py
Zylphrex: Good catch! Updated!

sl0thentr0py (approved):

PR #1828, opened 2023-01-09
docs: update README

lebeier:
This is to help those who are having issue with the following error 

```
Sentry is attempting to send 1 pending error messages
```

After looking through the source code, I realized that one should set the `ca_certs` options for self-signed ca certificate.
And the above option was never documented nor tested in the source code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ pip install --upgrade sentry-sdk
 import sentry_sdk
 
 sentry_sdk.init(
-    "https://12927b5f211046b575ee51fd8b1ac34f@o1.ingest.sentry.io/1",
+    dsn="https://12927b5f211046b575ee51fd8b1ac34f@o1.ingest.sentry.io/1",
 
     # Set traces_sample_rate to 1.0 to capture 100%
     # of transactions for performance monitoring.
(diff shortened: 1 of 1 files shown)


sl0thentr0py:
hi @lebeier, thanks for the suggestion!
We have a separate documentation platform where we add these details.
The `ca_certs` option is only shortly documented at https://docs.sentry.io/platforms/python/configuration/options/#ca-certs
If you wish, you could add some clarification text [here](https://github.com/getsentry/sentry-docs/blob/001bede80372e323449931fa5eb89ba82cd7aa98/src/platforms/common/configuration/options.mdx#L352-L356) or alternatively a more complete exampling in [the Troubleshooting section](https://docs.sentry.io/platforms/python/troubleshooting/) [here](https://github.com/getsentry/sentry-docs/blob/master/src/platforms/python/common/troubleshooting.mdx).

lebeier:
> hi @lebeier, thanks for the suggestion! We have a separate documentation platform where we add these details. The `ca_certs` option is only shortly documented at https://docs.sentry.io/platforms/python/configuration/options/#ca-certs If you wish, you could add some clarification text [here](https://github.com/getsentry/sentry-docs/blob/001bede80372e323449931fa5eb89ba82cd7aa98/src/platforms/common/configuration/options.mdx#L352-L356) or alternatively a more complete exampling in [the Troubleshooting section](https://docs.sentry.io/platforms/python/troubleshooting/) [here](https://github.com/getsentry/sentry-docs/blob/master/src/platforms/python/common/troubleshooting.mdx).

@sl0thentr0py thanks for the advice. In that case for better clarification, Will it be ok if I include the link to that page? I will probably update that Troubleshooting section for those who are having similar issues

sl0thentr0py:
yes, just make a PR in https://github.com/getsentry/sentry-docs and we'll take it from there.

> Will it be ok if I include the link to that page?

not sure which page you are referring to but feel free to cross-link inside the docs, but we won't add any links here in the README in this repo.

PR #1830, opened 2023-01-10
feat(profiling): Set active thread id for quart

Zylphrex:
Following up to #1824 to set the active thread id for quart.

### DIFF
--- a/sentry_sdk/integrations/quart.py
+++ b/sentry_sdk/integrations/quart.py
@@ -1,5 +1,8 @@
 from __future__ import absolute_import
 
+import inspect
+import threading
+
 from sentry_sdk.hub import _should_send_default_pii, Hub
 from sentry_sdk.integrations import DidNotEnable, Integration
 from sentry_sdk.integrations._wsgi_common import _filter_headers
@@ -11,6 +14,7 @@ from sentry_sdk.utils import (
     event_from_exception,
 )
 
+from sentry_sdk._functools import wraps
 from sentry_sdk._types import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -34,6 +38,7 @@ try:
         request,
         websocket,
     )
+    from quart.scaffold import Scaffold  # type: ignore
     from quart.signals import (  # type: ignore
         got_background_exception,
         got_request_exception,
@@ -41,6 +46,7 @@ try:
         request_started,
         websocket_started,
     )
+    from quart.utils import is_coroutine_function  # type: ignore
 except ImportError:
     raise DidNotEnable("Quart is not installed")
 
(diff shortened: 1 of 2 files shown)


sl0thentr0py:
@pgjones we're making some changes to the Quart integration to enable our (beta) profiler on ASGI frameworks, in case you want to take a quick look at the changes!

sentry_sdk/integrations/quart.py
```diff
@@ -71,18 +77,62 @@ def setup_once():
...
+                def _sentry_func(*args, **kwargs):
+                    # type: (*Any, **Any) -> Any
+                    hub = Hub.current
+                    integration = hub.get_integration(QuartIntegration)
+                    if integration is None:
+                        return old_func(*args, **kwargs)
+
+                    with hub.configure_scope() as sentry_scope:
+                        if sentry_scope.profile is not None:
+                            sentry_scope.profile.active_thread_id = (
```
pgjones, line 123: What is the active_thread_id used for? Quart doesn't really use threads, so I'm interested to learn what the need is.

sentry_sdk/integrations/quart.py
```diff
@@ -71,18 +77,62 @@ def setup_once():
...
+        middleware = SentryAsgiMiddleware(lambda *a, **kw: old_app(self, *a, **kw))
+        middleware.__call__ = middleware._run_asgi3
+        return await middleware(scope, receive, send)
+
+    Quart.__call__ = sentry_patched_asgi_app
+
+
+def patch_scaffold_route():
+    # type: () -> None
+    old_route = Scaffold.route
```
pgjones, line 102: `route` isn't the only way to create a handler, there is also `add_url_rule` and the corresponding ones for websockets.

Did you consider using a `before_request` function as a place to run some code for every request?

sentry_sdk/integrations/quart.py
Zylphrex: For context, we're working on adding profiling support for the various ASGI frameworks, and when profiling, we have the concept of an active profile which for the purpose of a web server, it is the thread on which the request is handled.

In this case, we're specifically targeting synchronous request handlers because the way they're handled is by dispatching the handler to another thread. So what we're trying to do here is after it's been dispatched, we want to determine the thread id the handler is running on and update the profile metadata to be aware of that.

sentry_sdk/integrations/quart.py
Zylphrex: Looking at `before_request`, I'm not sure it'll serve the use case we want here because it we specifically want to read the thread id of the thread running the request handler and from what I can see, `before_request` is called from the main thread and potentially dispatched to another thread independently from the request handler.

pgjones:
I understand. Had you considered patching the [`app.ensure_async` method](https://github.com/pallets/quart/blob/main/src/quart/app.py#L1206) instead? This would give the correct thread ID in the context for other functions bar route handlers.

Zylphrex:
> I understand. Had you considered patching the [`app.ensure_async` method](https://github.com/pallets/quart/blob/main/src/quart/app.py#L1206) instead? This would give the correct thread ID in the context for other functions bar route handlers.

I did consider that at one point but I noticed that `ensure_async` is called by more than just the route handlers. And if I understood it correctly, that means after setting the thread ID for the handler, other methods can happen and change the thread ID. Is this correct? If so, I don't think patching `ensure_async` is the right thing to do here.

pgjones:
Yes, I think that is correct. However, the other methods could be a before/after function which I think would also ideally be recorded in the transaction? (As the total time to respond will include these). If not I think what is in this PR does make the most sense.

Zylphrex:
> Yes, I think that is correct. However, the other methods could be a before/after function which I think would also ideally be recorded in the transaction? (As the total time to respond will include these). If not I think what is in this PR does make the most sense.

So even if the other calls to `ensure_async` are also captured in the same transaction, I think the justification here is that these other calls are not the "core" work done for that transaction. So if one of these calls happens after the request handler, we'd end up showing a different thread as the active thread for the profile which can be misleading to the end user. So while this approach will only correctly show the right thread for some requests, I think it's better the result is predictable rather than relying on how other code calls `ensure_async` to not change the active thread id.

sl0thentr0py (approved):

sl0thentr0py:
sry this flew under the radar

PR #1831, opened 2023-01-11
feat(profiling): Use co_qualname in python 3.11

Zylphrex:
The `get_frame_name` implementation works well for <3.11 but 3.11 introduced a `co_qualname` that works like our implementation of `get_frame_name` and handles some cases better.

### DIFF
--- a/sentry_sdk/profiler.py
+++ b/sentry_sdk/profiler.py
@@ -24,7 +24,7 @@ from collections import deque
 from contextlib import contextmanager
 
 import sentry_sdk
-from sentry_sdk._compat import PY33
+from sentry_sdk._compat import PY33, PY311
 from sentry_sdk._types import MYPY
 from sentry_sdk.utils import (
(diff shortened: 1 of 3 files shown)


sentry_sdk/profiler.py
```diff
@@ -241,55 +241,60 @@ def extract_frame(frame, cwd):
...
-
-    # nothing we can do if it is a staticmethod (decorated with @staticmethod)
-
-    # we've done all we can, time to give up and return what we have
-    return name
+if PY311:
+
+    def get_frame_name(frame):
+        # type: (FrameType) -> str
+        return frame.f_code.co_qualname  # type: ignore
```
Zylphrex, line 248: https://docs.python.org/3.11/library/inspect.html?highlight=co_qualname was introduced in Python 3.11.

sl0thentr0py (approved):

PR #1833, opened 2023-01-12
fix(serializer): Add support for `bytearray` and `memoryview` types

Tarty:
Both `bytearray` and `memoryview` built-in types weren't explicitly mentioned in the serializer logic, and as they are subtyping Sequence, this led their instances to be enumerated upon, and to be output as a list of bytes, byte per byte.

In the case of `memoryview`, this could also lead to a segmentation fault if the memory referenced was already freed and unavailable to the process by then.

By explicitly adding them as seralizable types, bytearray will be decoded as a string just like bytes, and memoryview will use its __repr__ method instead.

Close GH-1829

-----

I noticed Python 2.7 was still supported, so I also attempted to add the fix for this version.

However `bytearray` and `memoryview` were introduced by Python 2.7, so this PR introduces code that is not supported by Python 2.6 (which reached its end of life in 2013).

I don't know if this is a concern or not?

### DIFF
--- a/sentry_sdk/_compat.py
+++ b/sentry_sdk/_compat.py
@@ -26,6 +26,7 @@ if PY2:
     number_types = (int, long, float)  # noqa
     int_types = (int, long)  # noqa
     iteritems = lambda x: x.iteritems()  # noqa: B301
+    binary_sequence_types = (bytearray, memoryview)
 
     def implements_str(cls):
         # type: (T) -> T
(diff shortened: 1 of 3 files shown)


sl0thentr0py (approved):
ty @Tarty 

> However bytearray and memoryview were introduced by Python 2.7, so this PR introduces code that is not supported by Python 2.6 (which reached its end of life in 2013).

Our support matrix is [here](https://github.com/getsentry/sentry-python/blob/dd8bfe37d2ab369eaa481a93484d4140fd964842/tox.ini#L8-L9) and we only need 2.7, so all good.

PR #1840, opened 2023-01-16
Add `before_send_transaction`

### LINKED ISSUE #1738: Add beforeSendTransaction
## Problem Statement
Similar to `beforeSend`, we will add `beforeSendTransaction` to allow people to strip Pii or drop transactions before they are sent to Sentry.

## Solution Brainstorm
See https://github.com/getsentry/sentry-javascript/pull/6121

antonpirker:
Callback function that is called for all transactions (in opposite to `before_send` which is only called for issue events)

Fixes #1738

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,3 +7,5 @@ coverage:
       python:
         target: 90%
 comment: false
+ignore:
+  - "tests"
--- a/codecov.yml
+++ b/codecov.yml
@@ -9,3 +9,4 @@ coverage:
 comment: false
 ignore:
   - "tests"
+  - "sentry_sdk/_types.py"
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -257,7 +257,7 @@ class _Client(object):
                 logger.info("before send transaction dropped event (%s)", event)
                 if self.transport:
                     self.transport.record_lost_event(
-                        "before_send_transaction", data_category="error"
+                        "before_send", data_category="transaction"
                     )
             event = new_event  # type: ignore
 
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -248,6 +248,19 @@ class _Client(object):
                     )
             event = new_event  # type: ignore
 
+        before_send_transaction = self.options["before_send_transaction"]
+        if before_send_transaction is not None and event.get("type") == "transaction":
+            new_event = None
+            with capture_internal_exceptions():
+                new_event = before_send_transaction(event, hint or {})
+            if new_event is None:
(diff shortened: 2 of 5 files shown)


antonpirker:
Docs: https://github.com/getsentry/sentry-docs/pull/6089

tests/test_basics.py
```diff
@@ -91,7 +91,65 @@ def test_event_id(sentry_init, capture_events):
...
+    assert "before_send_called" not in event["extra"]
+
+
+def test_option_before_send_transaction(sentry_init, capture_events, monkeypatch):
+    def before_send_transaction(event, hint):
+        assert event["type"] == "transaction"
+        event["extra"] = {"before_send_transaction_called": True}
+        return event
+
+    sentry_init(
```
sl0thentr0py, line 128: calling init multiple times in a test is weird, can you use `pytest.mark.parametrize` for the different combinations

sentry_sdk/client.py
```diff
@@ -248,6 +248,19 @@ def _prepare_event(
...
+        before_send_transaction = self.options["before_send_transaction"]
+        if before_send_transaction is not None and event.get("type") == "transaction":
+            new_event = None
+            with capture_internal_exceptions():
+                new_event = before_send_transaction(event, hint or {})
+            if new_event is None:
+                logger.info("before send transaction dropped event (%s)", event)
+                if self.transport:
+                    self.transport.record_lost_event(
+                        "before_send_transaction", data_category="error"
```
sl0thentr0py, line 260:
```suggestion
                        "before_send", data_category="transaction"
```

data_category is self evident
for reason, see [develop docs](https://develop.sentry.dev/sdk/client-reports/#envelope-item-payload) that say that the valid reasons list is in [snuba](https://github.com/getsentry/snuba/blob/4e7cfdddcf7b93eacb762bc74ca2461cec9464e5/snuba/datasets/outcomes_processor.py#L24-L34).

sl0thentr0py (changes requested):

tests/test_basics.py
antonpirker: Removed the second init as a whole, because it is just testing normal sdk behaviour that is tested somewhere else.

codecov.yml
```diff
@@ -7,3 +7,5 @@ coverage:
       python:
         target: 90%
 comment: false
+ignore:
+  - "tests"
```
antonpirker, line 11: I sneaked "Codecov, please ignore the `tests/` directory in here, because I am too lazy to create a new PR to do this two lines.

sl0thentr0py (approved):

PR #1841, opened 2023-01-16
Fix middleware being patched multiple times when using FastAPI

JohnnyDeuss:
When the `StarletteIntegration` is used in a FastAPI app, as instructed in [the docs](https://docs.sentry.io/platforms/python/guides/fastapi/), the `AuthenticationMiddleware` and `ExceptionMiddleware` get patched every time `add_middleware` or `add_exception_handler` is called, since these methods call `build_middleware_stack`, where patched `Middleware` instances are repatched without checking if they were already patched.

If these functions are called enough times, Python registers this as infinite recursion (as we've discovered).

The changes look broad, but amount to the addition of two `if not_yet_patched` checks and indentation.

### DIFF
--- a/sentry_sdk/integrations/starlette.py
+++ b/sentry_sdk/integrations/starlette.py
@@ -167,44 +167,47 @@ def patch_exception_middleware(middleware_class):
     """
     old_middleware_init = middleware_class.__init__
 
-    def _sentry_middleware_init(self, *args, **kwargs):
-        # type: (Any, Any, Any) -> None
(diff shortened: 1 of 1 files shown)


antonpirker (approved):

antonpirker:
Hey @JohnnyDeuss 
Can you fix the "Lint Sources" issues shown above?
You probably just need to reformat the code locally by calling `black tests sentry_sdk` in the root directory.

PR #1866, opened 2023-01-27
feat(crons): Add CronMonitorIntegration

evanpurkhiser:
This reads the SENTRY_MONITOR_ID from the environment and puts it into
the monitor context.

This is in support of https://github.com/getsentry/sentry/issues/43647


evanpurkhiser:
Sounds like we're actually going to be using Relay to extract the trace baggage which will include the monitor ID. Relay can then stick the monitor ID into context of the ingested event.

PR #1870, opened 2023-01-27
tests(gevent): Add workflow to test gevent

Zylphrex:
Run all tests also with `gevent`.


.github/workflows/test-common.yml
```diff
@@ -30,24 +30,7 @@ jobs:
...
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 5
-        # Maps tcp port 5432 on service container to the host
-        ports:
-          - 5432:5432
-    env:
-      SENTRY_PYTHON_TEST_POSTGRES_USER: postgres
-      SENTRY_PYTHON_TEST_POSTGRES_PASSWORD: sentry
-      SENTRY_PYTHON_TEST_POSTGRES_NAME: ci_test
```
Zylphrex, line 50: Nothing in the common tests depends on postgres, removing it speeds things up by ~25s job

.github/workflows/test-common.yml
```diff
@@ -65,7 +48,19 @@ jobs:
...
+    name: All common tests passed or skipped
+    needs: test
+    # Always run this, even if a dependent job failed
+    if: always()
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check for failures
+        if: contains(needs.test.result, 'failure')
+        run: |
+          echo "One of the dependent jobs have failed. You may need to re-run it." && exit 1
```
Zylphrex, line 66: I don't have permission to do this but if we could change the required check to this step like for all the integration tests, that'd be great.

sl0thentr0py:
@antonpirker this is testing so it's yours

.github/workflows/test-common.yml
```diff
@@ -65,7 +48,19 @@ jobs:
...
           coverage erase
 
-          ./scripts/runtox.sh "py${{ matrix.python-version }}$" --cov=tests --cov=sentry_sdk --cov-report= --cov-branch --ignore=tests/integrations
+          ./scripts/runtox.sh "py${{ matrix.python-version }}-${{ matrix.variant }}$" --cov=tests --cov=sentry_sdk --cov-report= --cov-branch --ignore=tests/integrations
           coverage combine .coverage*
           coverage xml -i
           codecov --file coverage.xml
+
+  check_required_tests:
+    name: All common tests passed or skipped
```
antonpirker, line 57: I think we do not need the `variant` here and the `check_required_tests` step.

If we just keep your changes in `tox.ini` and then you run locally `python scripts/split-tox-gh-actions/split-tox-gh-actions.py ` then a new github workflow file `test-gevent.yml` will be created that will run all tests in the specified python version and with the specified `gevent` requirement.

antonpirker (changes requested):

.github/workflows/test-common.yml
antonpirker: For testing purposes, I have reverted your naming change of the common tests. I hope you do not mind if I am committing to your branch.

.github/workflows/test-common.yml
antonpirker: You are right. But you can set the path where to find the tests here: https://github.com/getsentry/sentry-python/blob/master/tox.ini#L377-L379

So I took the freedom to make some changes to your PR to run all the tests with gevent like the other integration tests.

.github/workflows/test-common.yml
antonpirker: Yes! Thanks!

.github/workflows/test-common.yml
antonpirker: resolve because it is outdated now.

.github/workflows/test-common.yml
antonpirker: resolve, because it is outdated now.

antonpirker:
Now all the test run successfully with gevent.
But they fail because the coverage can not be collected.
See: https://stackoverflow.com/questions/22890779/determining-test-coverage-with-py-test-gevent

Can you try to fix this @Zylphrex ?

Zylphrex:
> Now all the test run successfully with gevent. But they fail because the coverage can not be collected. See: https://stackoverflow.com/questions/22890779/determining-test-coverage-with-py-test-gevent
> 
> Can you try to fix this @Zylphrex ?

Thanks, I'll get this fixed 👍

Zylphrex:
@antonpirker the coverage stuff is fixed now, turns out you can't inline comments with the environment variables. It breaks the `posargs` and it doesn't get passed on to pytest

antonpirker (approved):

PR #1873, opened 2023-01-29
Do not overwrite default for username with email address in Flask-Login integration

homeworkprod:
This line seems like a copy/paste error, introduced in 41120009fa7d6cb88d9219cb20874c9dd705639d.

The email address is already set a few lines above for "email".

Unless I'm missing some very clever fallback magic, considering the `try/catch` around this, I think this is a mistake and the email address should not be used as username.


sl0thentr0py (approved):
indeed, thx @homeworkprod !

PR #1876, opened 2023-01-30
feat(pii): Sanitize URLs in Span description and breadcrumbs

### LINKED ISSUE #1742: Sanitize URLs in Span description and breadcrumbs
When we do HTTP requests to third party services we create a breadcrumb with the URL and also create a span that has the URL as a description (and also breadcrumbs including this URL are created.)

We created [RFC-0038](https://github.com/getsentry/rfcs/blob/main/text/0038-scrubbing-sensitive-data.md) to decide how to improve the current situation and creating a spec here: https://github.com/getsentry/develop/pull/773 

Make sure that all integrations that record outgoing or incoming HTTP request structure the data like described in the spec linked above.

antonpirker:
Fixes #1742


sentry_sdk/utils.py
```diff
@@ -1126,6 +1144,74 @@ def from_base64(base64_string):
...
+
+def sanitize_url(url):
+    # type: (str) -> str
+    """
+    Removes all query parameter values and username:password from a given URL.
+    """
+    parsed_url = urlsplit(url)
+    query_params = parse_qs(parsed_url.query, keep_blank_values=True)
+
+    from sentry_sdk.consts import SENSITIVE_DATA_SUBSTITUTE
```
sl0thentr0py, line 1158: why is the import inside?

sentry_sdk/integrations/stdlib.py
```diff
@@ -79,12 +84,18 @@ def putrequest(self, method, url, *args, **kwargs):
...
         span = hub.start_span(
-            op=OP.HTTP_CLIENT, description="%s %s" % (method, real_url)
+            op=OP.HTTP_CLIENT,
+            description="%s %s" % (method, parsed_url.url),
         )
 
         span.set_data("method", method)
-        span.set_data("url", real_url)
+        span.set_data("url", parsed_url.url)
+        span.set_data("http.query", parsed_url.query)
```
sl0thentr0py, line 97: I still think we should send `http.query` as a map/dict and not filter all the values out. 
Relay will currently take care of sensitive stuff in `span.data` if it finds something and later my scrubber can also be used to filter certain keys out.

cc: @cleptric

sl0thentr0py (changes requested):

sentry_sdk/integrations/stdlib.py
sl0thentr0py: fine with merging this for now I guess, don't wanna block or whatever
still think it's annoying to lose all query values as `FILTERED`

cleptric: I remember discussing that with @jjbayer and we agreed to send it as a string.
I opted to not scrub the data in `http.query` on the client, see https://github.com/getsentry/sentry-php/pull/1453.

antonpirker: For now it is a string and filtered by default. I guess when we implement data scrubbers, we can change that.

sentry_sdk/utils.py
antonpirker: Because of a circular import problem: https://github.com/getsentry/sentry-python/actions/runs/4183447955/jobs/7247820509

This is currently happening when we create our API docs, so in the future we have to decide what to do with API docs, because right now they are kind of abandoned.

sl0thentr0py: i think you can move it to the bottom of the file then

sentry_sdk/integrations/stdlib.py
cleptric: @antonpirker Just to make sure, even values in `data.http_query` are filtered?

sl0thentr0py: but like can we be consistent? I'm not sure what you guys agreed on but if Michi didn't do it and you do it here, it's already inconsistent.

cleptric: My understanding was that the whole idea was to move this data into a new property so server-side scrubbing can be applied more easily. If we also filter it by default in the SDKs, this feels like a regression to me.

antonpirker: @cleptric Yes, everything in query is filtered by default. If you turn send_default_pii on, it will stay in there.

antonpirker: But you are right @sl0thentr0py 
Just checked with the docs: https://develop.sentry.dev/sdk/data-handling/#structuring-data

It does not specify if all the values should be removed by default.
So the consistent way would be to ALWAYS remove `username:password` in the URL, but keep all the query parameters as is? (no matter what `send_default_pii` is set to, right?

sl0thentr0py: yes i think that's fine. We can move query params to span.data separately later.

sl0thentr0py: ty!

cleptric: We should move the query params to span data right now, this was again part of our initial idea. And @antonpirker is correct, the span descriptions should not contain any authority, query, or fragments anymore.
We can iterate on client-side filtering for the span data as a follow-up, happy to make changes to docs and PHP SDKs.

sl0thentr0py: ok then please do what Michi says :)

sentry_sdk/utils.py
antonpirker: lets try

sentry_sdk/integrations/stdlib.py
antonpirker: done

sentry_sdk/utils.py
antonpirker: Moved the variable to `utils.py` now because it is mainly used there to fix the circular import.

sl0thentr0py (approved):

PR #1880, opened 2023-01-31
feat(profiling): Report resident set size on each sample

Zylphrex:
This adds the RSS values at each sample to be visualized along side the flamechart to help debug memory leaks.


Zylphrex:
let's close this for now, will be revisiting it at some point though

PR #1882, opened 2023-02-01
feat(cloud): Adding Cloud Resource Context

### LINKED ISSUE #1820: Add cloud resource context
### Problem Statement

When a user gets an error or transacation (performance event) in Sentry they may not know the specific of where that particular service is hosted. Where possible it would be great to know some contextual information about where that service is hosted to speed up problem solving and eventually deploying a fix

example data
```
cloud:
  provider: aws|gcp|azure
  region: ...
  [account|project]_id: ...
  service: ec2, cloud-functions, ...
  resource: <ARN>
```

### Solution Brainstorm

Add this data using the [Contexts Interface](https://develop.sentry.dev/sdk/event-payloads/contexts/) of the event payload

- [x] Define a standard set of contextual data: 
  - example above
- [ ] determine how to fetch that data from top hosting platforms
  - [x] AWS
  - [x] GCP
  - [ ] Azure
  - [ ] Vercel
  - [ ] Netlify
  - [ ] others?
- [ ] Python SDK should check for this context and populate it in the event when possible
 - [ ] relay changes needed prior to

Future considerations:
- contexts should be searchable in product, changes will be needed in sentry backend so this data can be used to filter issues

antonpirker:
Fixes #1820 

This is the cloud context definition in relay: https://github.com/getsentry/relay/pull/1854


sentry_sdk/integrations/cloudresourcecontext.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def __init__(self, cloud_provider=""):
+        # type: (str) -> None
+        CloudResourceContextIntegration.cloud_provider = cloud_provider
+
+    @classmethod
+    def _is_aws(cls):
+        # type: () -> bool
+        try:
+            r = cls.http.request(
```
sl0thentr0py, line 74: these requests will happen synchronously on sentry.init, is this the behaviour we want?

sl0thentr0py (changes requested):
a fundamental question first, review will come after we've discussed that

sentry_sdk/integrations/cloudresourcecontext.py
```diff
@@ -0,0 +1,256 @@
+import json
```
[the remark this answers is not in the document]
antonpirker: renamed.

sentry_sdk/integrations/cloudresourcecontext.py
antonpirker: Good question. 

I thought a bit about it and my conclusion was that this is a longer running server process and there is only done one HTTP request, so I guess it should be OK. 

But I am not 100% happy with it. The other solutions would be to make it async or in a thread or something. What do you think?

If we add this to Lambda functions and the like than this will need to be changed.

antonpirker: As discussed in a call the other day, we will keep this as is for now.

antonpirker:
This is the cloud context definition in relay: https://github.com/getsentry/relay/pull/1854

HazAT (approved):

PR #1894, opened 2023-02-10
Better setting of in-app in stack frames

### LINKED ISSUE #320: Set in_app in integrations
In many integrations we currently don't set in_app, but we should because it affects trimming. See #316

### LINKED ISSUE #1754: Do not mark every frame as in-app
### Problem Statement

Most errors on Sentry with stacktraces have every frame marked as in_app even though many of the files are part of the 3rd party libraries. This causes issues for the commit context and issue details page.

### Solution Brainstorm

This can be avoided if we look at the files that belong to the virtualenv. If a file is under `os.environ.get("VIRTUAL_ENV")` then it should be marked as `in_app: False`. Or even check that the abs path includes `site-packages` (e.g. `/some/path/repo/.venv/lib/python3.10/site-packages/django/views/decorators/csrf.py`).

This would be a huge win for suspect commits (since we always look for the first in-app frame) and issue details page (since we load the first in-app frame).

CC @NisanthanNanthakumar @scttcper @asottile-sentry

antonpirker:
How the `in_app` flag is set in stack trace frames (in `set_in_app_in_frames()`):
- If there is already `in_app` set, it is left untouched.
- If there is a `module` in the frame and it is in the `in_app_includes` -> `in_app=True`
- If there is a `module` in the frame and it is in the `in_app_excludes` -> `in_app=False`
- If there is an `abs_path` in the frame and the path is in `/side-packages/` or `/dist-packages/` -> `in_app=False`
- If there is an `abs_path` in the frame and it starts with the current working directory of the process -> `in_app=True`
- If nothing of the above is true, there will be no `in_app` set.

Fixes #1754 
Fixes #320

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -103,6 +103,15 @@ class _Client(object):
     def __init__(self, *args, **kwargs):
         # type: (*Any, **Any) -> None
         self.options = get_options(*args, **kwargs)  # type: Dict[str, Any]
+
+        try:
+            project_root = os.getcwd()
+        except Exception:
+            project_root = None
+
+        self.configuration = {
+            "project_root": project_root,
+        }
         self._init_impl()
 
     def __getstate__(self):
@@ -222,7 +231,10 @@ class _Client(object):
             event["platform"] = "python"
 
         event = handle_in_app(
-            event, self.options["in_app_exclude"], self.options["in_app_include"]
+            event,
+            self.options["in_app_exclude"],
+            self.options["in_app_include"],
+            self.configuration["project_root"],
         )
 
         # Postprocess the event here so that annotated types do
@@ -446,7 +458,9 @@ class _Client(object):
 
             if is_transaction:
                 if profile is not None:
-                    envelope.add_profile(profile.to_json(event_opt, self.options))
+                    envelope.add_profile(
+                        profile.to_json(event_opt, self.options, self.configuration)
+                    )
                 envelope.add_transaction(event_opt)
             else:
                 envelope.add_event(event_opt)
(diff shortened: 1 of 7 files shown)


sentry_sdk/utils.py
```diff
@@ -773,32 +773,34 @@ def handle_in_app(event, in_app_exclude=None, in_app_include=None):
...
+        else:
+            abs_path = frame.get("abs_path")
+            if abs_path is None:
+                continue
+
+            external_source = (
+                re.search(r"[\\/](?:dist|site)-packages[\\/]", frame["abs_path"])
+                is not None
+            )
+            frame["in_app"] = not external_source
```
Zylphrex, line 803: This sets anything that's not in site packages or dist packages to false. Including frames coming from the standard library which is unexpected at least from the profiling side.

antonpirker: Valid point. Fixed.

Zylphrex (approved):

sl0thentr0py (changes requested):

sentry_sdk/client.py
```diff
@@ -103,6 +103,15 @@ class _Client(object):
     def __init__(self, *args, **kwargs):
         # type: (*Any, **Any) -> None
         self.options = get_options(*args, **kwargs)  # type: Dict[str, Any]
+
+        try:
+            project_root = os.getcwd()
+        except Exception:
+            project_root = None
+
+        self.configuration = {
```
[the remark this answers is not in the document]
antonpirker: Probably makes sense. I put it in configuration because it can not be set but it is just read from os.getcwd(). But it is cleaner in options and also being able to set it would be cool. Will change.

sentry_sdk/client.py
```diff
@@ -90,6 +90,13 @@ def _get_options(*args, **kwargs):
     if rv["instrumenter"] is None:
         rv["instrumenter"] = INSTRUMENTER.SENTRY
 
+    try:
```
sl0thentr0py, line 93: you need this only if
```python
   if rv["project_root"] is None:
```

sl0thentr0py (approved):

PR #1901, opened 2023-02-14
fix(profiling): Delay start of profiler thread

Zylphrex:
When using uWSGI with the master enabled, the profiler thread is started in the master process. However, when uWSGI forks the worker processes, the profiler thread will not be present in the worker process. This means that we never take any samples. To address this, we delay the start of the profiler thread until the first profile to guarantee it is started within the worker threads.


Zylphrex:
closing in favour of #1903

PR #1903, opened 2023-02-15
fix(profiling): Start profiler thread lazily

Zylphrex:
When running with uWSGI, it preforks the process so the profiler thread is started on the master process but doesn't run on the worker process. This means that no samples are ever taken. This change delays the start of the profiler thread to the first profile that is started.


sentry_sdk/profiler.py
```diff
@@ -888,22 +917,49 @@ def __init__(self, frequency):
...
+            # another thread may have tried to acquire the lock
+            # at the same time so it may start another thread
+            # make sure to check again before proceeding
+            if self.running and self.pid == pid:
+                return
+
+            self.pid = pid
+            self.running = True
+
+            self.thread = self.make_thread()
```
antonpirker, line 961: Why can ThreadPool(1) not be called here directly?

antonpirker (approved):

PR #1905, opened 2023-02-16
Fixed checks for structured http data

antonpirker:


cleptric (approved):
May we add a test where the URL contains some query params as well?

PR #1916, opened 2023-02-21
Adds `trace_propagation_targets` option

### LINKED ISSUE #1785: Add `trace_propagation_targets` option.
### Problem Statement

Add an option `trace_propagation_targets` that defines to what targets the trace headers (`sentry-trace` and `baggage`) are added in outgoing HTTP requests. 

Similar to the same feature in JS: https://docs.sentry.io/platforms/node/guides/serverless-cloud/configuration/options/#trace-propagation-targets

### Solution Brainstorm

Implement like it is implemented in Javascript.

antonpirker:
Add an option `trace_propagation_targets` that defines to what targets the trace headers (sentry-trace and baggage) are added in outgoing HTTP requests.

Spec is here: https://docs.sentry.io/platforms/node/guides/serverless-cloud/configuration/options/#trace-propagation-targets

Fixes #1785

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -67,6 +67,8 @@ class OP:
     MIDDLEWARE_STARLITE = "middleware.starlite"
     MIDDLEWARE_STARLITE_RECEIVE = "middleware.starlite.receive"
     MIDDLEWARE_STARLITE_SEND = "middleware.starlite.send"
+    QUEUE_SUBMIT_ARQ = "queue.submit.arq"
+    QUEUE_TASK_ARQ = "queue.task.arq"
     QUEUE_SUBMIT_CELERY = "queue.submit.celery"
     QUEUE_TASK_CELERY = "queue.task.celery"
     QUEUE_TASK_RQ = "queue.task.rq"
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -121,6 +121,7 @@ class ClientConstructor(object):
         before_send_transaction=None,  # type: Optional[TransactionProcessor]
         project_root=None,  # type: Optional[str]
         enable_tracing=None,  # type: Optional[bool]
+        trace_propagation_targets=[],  # type: Sequence[str]
     ):
         # type: (...) -> None
         pass
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -33,7 +33,6 @@ if MYPY:
             "max_spans": Optional[int],
             "record_sql_params": Optional[bool],
             "smart_transaction_trimming": Optional[bool],
-            "custom_measurements": Optional[bool],
             "profiles_sample_rate": Optional[float],
             "profiler_mode": Optional[str],
         },
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -125,7 +125,9 @@ class ClientConstructor(object):
         before_send_transaction=None,  # type: Optional[TransactionProcessor]
         project_root=None,  # type: Optional[str]
         enable_tracing=None,  # type: Optional[bool]
-        trace_propagation_targets=[MATCH_ALL],  # type: Sequence[str]  # noqa: B006
+        trace_propagation_targets=[
+            MATCH_ALL
+        ],  # type: Optional[Sequence[str]]  # noqa: B006
     ):
         # type: (...) -> None
         pass
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -121,7 +121,7 @@ class ClientConstructor(object):
         before_send_transaction=None,  # type: Optional[TransactionProcessor]
         project_root=None,  # type: Optional[str]
         enable_tracing=None,  # type: Optional[bool]
-        trace_propagation_targets=[],  # type: Sequence[str]
+        trace_propagation_targets=[],  # type: Sequence[str]  # noqa: B006
     ):
         # type: (...) -> None
         pass
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -33,7 +33,6 @@ if MYPY:
             "max_spans": Optional[int],
             "record_sql_params": Optional[bool],
             "smart_transaction_trimming": Optional[bool],
-            "propagate_tracestate": Optional[bool],
             "custom_measurements": Optional[bool],
             "profiles_sample_rate": Optional[float],
             "profiler_mode": Optional[str],
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -42,6 +42,8 @@ if MYPY:
 DEFAULT_QUEUE_SIZE = 100
 DEFAULT_MAX_BREADCRUMBS = 100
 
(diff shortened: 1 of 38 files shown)


sentry_sdk/integrations/httpx.py
```diff
@@ -52,13 +53,18 @@ def send(self, request, **kwargs):
             span.set_data("http.query", parsed_url.query)
             span.set_data("http.fragment", parsed_url.fragment)
 
-            for key, value in hub.iter_trace_propagation_headers():
-                logger.debug(
-                    "[Tracing] Adding `{key}` header {value} to outgoing request to {url}.".format(
-                        key=key, value=value, url=request.url
+            if should_propagate_trace(
+                str(request.url),
```
untitaker, line 57: What is the difference between `parsed_url.url` and this?

sentry_sdk/integrations/httpx.py
```diff
@@ -91,13 +97,17 @@ async def send(self, request, **kwargs):
...
+                hub.client.options["trace_propagation_targets"] if hub.client else [],
+            ):
+                for key, value in hub.iter_trace_propagation_headers():
```
sl0thentr0py, line 104: `iter_trace_propagation_headers` is a hub method and already has some early return checks, so I would just add this logic there, and then we don't have to replicate it in every integration.

sl0thentr0py (changes requested):

sentry_sdk/integrations/httpx.py
sl0thentr0py: sry ignore, i'm stupid

sentry_sdk/tracing_utils.py
```diff
@@ -376,6 +377,24 @@ def serialize(self, include_third_party=False):
         return ",".join(items)
 
 
+def should_propagate_trace(url, trace_propagation_targets):
+    # type: (str, List[str]) -> bool
+    """
+    Returns True if trace_propagation_targets are not set, or if the url matches any of the set targets. Otherwise, returns False.
+    """
+    if trace_propagation_targets is None or trace_propagation_targets == []:
```
sl0thentr0py, line 385: can we define semantics more precisely, if `None` is the default that's fine and you can treat it as `allow all` but empty array should be empty array. Having both `None` and `[]` take the same semantics is not clean here.

sl0thentr0py (changes requested):

sentry_sdk/tracing_utils.py
antonpirker: Empty array is the default and means: send to everywhere.
The check for None is just a bit paranoid so in case that someone sets `sentry_sdk.init(trace_propagation_targets=None)` nothing breaks.

sentry_sdk/integrations/httpx.py
antonpirker: `parsed_url` is a named tuple `(url, query_string, fragment)`
And the the `parsed_url.url` is just the base url without query_string and fragment

sentry_sdk/tracing_utils.py
antonpirker: I have now made the code better readable.

cleptric:
Semantics for this option are defined here https://docs.sentry.io/platforms/node/guides/serverless-cloud/configuration/options/#trace-propagation-targets

antonpirker:
Now I changed the behavior to have a `[MATCH_ALL]` default that matches everything so traces are propagated everywhere. If it is set to `[]` or `None` then traces will not be propagated anywhere.

Matching is also done by substring like described in the linked develop docs.

sl0thentr0py (changes requested):

sentry_sdk/integrations/httpx.py
```diff
@@ -52,13 +53,20 @@ def send(self, request, **kwargs):
...
             span.set_data("http.fragment", parsed_url.fragment)
 
-            for key, value in hub.iter_trace_propagation_headers():
-                logger.debug(
-                    "[Tracing] Adding `{key}` header {value} to outgoing request to {url}.".format(
-                        key=key, value=value, url=request.url
+            if should_propagate_trace(
+                str(request.url),
+                hub.client.options["trace_propagation_targets"]
+                if hub.client
```
untitaker, line 59: For what it's worth: Line 42 already "proves" there is a client. It is just a typing problem that you have to match on client here.

I was thinking that at some point, `hub.get_integration()` would be replaced with a method that returns both the integration and the current client so this check can be done all at once without typing issues

antonpirker: Good Point. Have now changed the code to cast the hub.client. Now mypy is happy and the code is easier.

antonpirker: And also good idea about having those checks in one place! 👍

sentry_sdk/integrations/httpx.py
```diff
@@ -52,13 +53,20 @@ def send(self, request, **kwargs):
             span.set_data("http.query", parsed_url.query)
             span.set_data("http.fragment", parsed_url.fragment)
 
-            for key, value in hub.iter_trace_propagation_headers():
-                logger.debug(
-                    "[Tracing] Adding `{key}` header {value} to outgoing request to {url}.".format(
-                        key=key, value=value, url=request.url
+            if should_propagate_trace(
+                str(request.url),
+                hub.client.options["trace_propagation_targets"]
```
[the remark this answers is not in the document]
antonpirker: I have now implemented a suggestion from markus below, which makes the code way easier.
I rather not do the "fetching the config" in function, so the function does just does one thing, check if the trace should be for the given url and trace_propagation_targets.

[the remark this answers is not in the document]
antonpirker: I was thinking a bit. How about we do a `hub.should_propagate_trace(url)` method? Is the hub the right "thing" to ask if the the trace to a given url should be propagated?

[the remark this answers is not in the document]
antonpirker: Maybe you @untitaker also have an opinion about my idea above?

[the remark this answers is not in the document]
untitaker: I don't have a strong opinion on this at all, I think there's no strong pattern to follow here. We do however have a few freestanding functions lying around the entire codebase that just take a hub as their first argument. That might be a bit better because then you can at least define `should_propagate_trace` in some helper module rather than the ever-growing `hub.py`

antonpirker:
I have now implemented @sl0thentr0py suggestion to fetch the options in the function. Now calling the function looks nicer.

sentry_sdk/consts.py
```diff
@@ -123,6 +125,7 @@ def __init__(
         before_send_transaction=None,  # type: Optional[TransactionProcessor]
         project_root=None,  # type: Optional[str]
         enable_tracing=None,  # type: Optional[bool]
+        trace_propagation_targets=[MATCH_ALL],  # type: Sequence[str]  # noqa: B006
```
sl0thentr0py, line 128: since we're checking for None below, the type should be optional[sequence]

sl0thentr0py (approved):
one typing comment

PR #1917, opened 2023-02-21
Updated outdated HTTPX test matrix

antonpirker:


tox.ini
```diff
@@ -79,10 +79,10 @@ envlist =
     {py3.7}-gcp
 
     # HTTPX
-    {py3.6,py3.7,py3.8,py3.9,py3.10,py3.11}-httpx-v{0.16,0.17}
```
sl0thentr0py, line 82: why can't we keep testing the older versions?

antonpirker: no reason. but you are right, I added them again. All the versions 0.16+

sl0thentr0py (approved):

PR #1918, opened 2023-02-21
Updated werkzeug used in tests

antonpirker:
Because it has some vulnerabilities.


cc-jj:
Werkzeug 2.1.0 has a breaking change causing a sentry internal error, see here:
https://github.com/getsentry/sentry-python/issues/1936

antonpirker:
Wont do this now. It is only used in our docs and we have more important things to do right now.

PR #1924, opened 2023-02-27
Rename 'with_locals'  to 'include_local_variables'

antonpirker:
Created an alias 'include_local_variables' for the 'with_locals' options.
Updated tests to make sure everything still works as expected.

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -87,7 +87,7 @@ class ClientConstructor(object):
     def __init__(
         self,
         dsn=None,  # type: Optional[str]
-        with_locals=True,  # type: bool
+        with_locals=True,  # type: bool   # DEPRECATED in favor of `include_local_variables`
         max_breadcrumbs=DEFAULT_MAX_BREADCRUMBS,  # type: int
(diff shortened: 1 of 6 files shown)


sentry_sdk/consts.py
```diff
@@ -89,7 +89,6 @@ class ClientConstructor(object):
     def __init__(
         self,
         dsn=None,  # type: Optional[str]
-        with_locals=True,  # type: bool
```
sl0thentr0py, line 92: removing something from here is a breaking change. This we have to do in a major.

I would keep this here and if present, set the other one in client.py

sl0thentr0py (changes requested):

sentry_sdk/consts.py
antonpirker: The users can still use `with_locals` in `sentry_sdk.init()` it will just be then copied to the new var name here: https://github.com/getsentry/sentry-python/blob/antonpirker/1854-include-local-variables/sentry_sdk/client.py#L74-L82

sl0thentr0py (approved):

PR #1934, opened 2023-03-01
fix: Rename MYPY to TYPE_CHECKING

untitaker:
we have a lot of conditionals in our codebase that are supposed to
separate the code that mypy is supposed to see from the code that we
actually want to execute.

In the specific case of sentry_sdk.configure_scope, this means that
pyright does not handle with the overloads correctly because it only
recognizes TYPE_CHECKING as a special variable name, not MYPY.

Rename MYPY to TYPE_CHECKING so pyright typechecks configure_scope
correctly.

### DIFF
--- a/scripts/init_serverless_sdk.py
+++ b/scripts/init_serverless_sdk.py
@@ -10,11 +10,11 @@ import sys
 import re
 
 import sentry_sdk
-from sentry_sdk._types import MYPY
+from sentry_sdk._types import TYPE_CHECKING
 from sentry_sdk.utils import Dsn
 from sentry_sdk.integrations.aws_lambda import AwsLambdaIntegration
 
-if MYPY:
+if TYPE_CHECKING:
     from typing import Any
 
 
--- a/sentry_sdk/_types.py
+++ b/sentry_sdk/_types.py
@@ -4,6 +4,10 @@ except ImportError:
     TYPE_CHECKING = False
 
 
+# Re-exported for compat, since code out there in the wild might use this variable.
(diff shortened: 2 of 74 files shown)


scripts/init_serverless_sdk.py
```diff
@@ -10,11 +10,11 @@
 import re
 
 import sentry_sdk
-from sentry_sdk._types import MYPY
+from sentry_sdk._types import TYPE_CHECKING
 from sentry_sdk.utils import Dsn
 from sentry_sdk.integrations.aws_lambda import AwsLambdaIntegration
 
-if MYPY:
+if TYPE_CHECKING:
```
sl0thentr0py, line 17: should we still expose `MYPY` as an alias just in case someone's importing it currently?

untitaker: oh.... yeah. probably. they shouldn't do that, right?

sl0thentr0py: shouldn't.. but I would remove it in a major just to be safe

sl0thentr0py (approved):

sentry_sdk/_types.py
```diff
@@ -1,10 +1,14 @@
 try:
-    from typing import TYPE_CHECKING as MYPY
+    from typing import TYPE_CHECKING as TYPE_CHECKING
 except ImportError:
-    MYPY = False
+    TYPE_CHECKING = False
```
aqeelat, line 4: two questions: 
1. why the try-catch?
2. why `from x import y as y`?

Issue #1936, opened 2023-03-02
FlaskIntegration JSON BadRequest with Werkzeug >= 2.1.0


cc-jj:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.16.0

### Steps to Reproduce

1. Create anFlask application. Werkzeug must be >= 2.1.0.

Werkzeug 2.1.0 changed the default behavior of `Request.get_json()`. See [here](https://werkzeug.palletsprojects.com/en/2.2.x/changes/#version-2-1-0):

```
Request.get_json() will raise a 400 BadRequest error if the Content-Type header is not application/json. This makes a very common source of confusion more visible. [#2339](https://github.com/pallets/werkzeug/issues/2339)
```

```
> python3 -m pip install --upgrade 'sentry-sdk[flask]'
> python3 -m pip freeze
blinker==1.5
certifi==2022.12.7
click==8.1.3
Flask==2.2.3
itsdangerous==2.1.2
Jinja2==3.1.2
MarkupSafe==2.1.2
sentry-sdk==1.16.0
urllib3==1.26.14
Werkzeug==2.2.3
```
Application code:
```python
import sentry_sdk
from flask import Flask
from sentry_sdk.integrations.flask import FlaskIntegration


sentry_sdk.init(
    integrations=[FlaskIntegration()],
    traces_sample_rate=1.0,
    debug=True,
)

app = Flask("app")


@app.route("/", methods=["GET"])
def index():
    return "ok", 200


if __name__ == "__main__":
    app.run("localhost", port=5000)
```

2. Run the flask server
```
> python3 app.py
```
3. Send a request
```
> curl http://localhost:5000/
```

### Expected Result

There should not be any internal server error in sentry_sdk.

### Actual Result

There was an internal server error in sentry_sdk. From the application logs:
```
[sentry] DEBUG: [Tracing] Starting <http.server> transaction <generic WSGI request>
 [sentry] DEBUG: [Profiling] Discarding profile because profiler was not started.
 [sentry] ERROR: Internal error in sentry_sdk
Traceback (most recent call last):
  File "/Users/cjgasper/third-party/sentry-bug/env/lib/python3.11/site-packages/flask/wrappers.py", line 128, in on_json_loading_failed
    return super().on_json_loading_failed(e)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1939, opened 2023-03-02
fix(integration): fix FlaskIntegration for Werkzeug > 2.1.0

michielderoos:
As explained in [this bug report](https://github.com/getsentry/sentry-python/issues/1936), recent versions of werkzeug [introduced](https://github.com/pallets/werkzeug/pull/2355) the behaviour to `get_json` where it will throw an error when content_type isn't set.

### DIFF
--- a/sentry_sdk/integrations/flask.py
+++ b/sentry_sdk/integrations/flask.py
@@ -173,7 +173,7 @@ class FlaskRequestExtractor(RequestExtractor):
 
     def json(self):
         # type: () -> Any
-        return self.request.get_json()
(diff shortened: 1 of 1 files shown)


cc-jj:
Fixes: https://github.com/getsentry/sentry-python/issues/1936

antonpirker (commented):
So in old Werkzeug doing `.get_json()` never resulted in an error and now an error will be thrown unless you add `.get_json(silent=True)` which then makes get_json behave like in the old Werkzeug? Is this correct?

antonpirker (approved):

PR #1943, opened 2023-03-06
Implement EventScrubber

### LINKED ISSUE #1897: Implement `config.event_scrubber`
### Problem Statement

Add an event scrubber as per https://github.com/getsentry/rfcs/pull/62

### Solution Brainstorm

^

sl0thentr0py:
Add a new `EventScrubber` class that scrubs certain potentially sensitive interfaces with a `DEFAULT_DENYLIST`. The default scrubber is automatically run if `send_default_pii = False`.
 
### Usage
```python
from sentry_sdk.scrubber import EventScrubber, DEFAULT_DENYLIST

# custom denylist
denylist = DEFAULT_DENYLIST + ["my_sensitive_var"]

sentry_sdk.init(
    # ...
    send_default_pii=False,
    event_scrubber=EventScrubber(denylist=denylist),
)
```

As outlined in https://github.com/getsentry/rfcs/blob/main/text/0062-controlling-pii-and-credentials-in-sd-ks.md
closes #1897 
docs: https://github.com/getsentry/sentry-docs/pull/6489

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -29,6 +29,7 @@ from sentry_sdk.utils import ContextVar
 from sentry_sdk.sessions import SessionFlusher
 from sentry_sdk.envelope import Envelope
 from sentry_sdk.profiler import has_profiling_enabled, setup_profiler
+from sentry_sdk.scrubber import EventScrubber
 
 from sentry_sdk._types import TYPE_CHECKING
 
@@ -110,6 +111,9 @@ def _get_options(*args, **kwargs):
 
     if rv["enable_tracing"] is True and rv["traces_sample_rate"] is None:
         rv["traces_sample_rate"] = 1.0
+
+    if rv["event_scrubber"] is None:
+        rv["event_scrubber"] = EventScrubber()
 
     return rv
 
@@ -248,6 +252,11 @@ class _Client(object):
             self.options["in_app_include"],
             self.options["project_root"],
         )
+
+        if event is not None:
+            event_scrubber = self.options["event_scrubber"]
+            if event_scrubber and not self.options["send_default_pii"]:
+                event_scrubber.scrub_event(event)
 
         # Postprocess the event here so that annotated types do
         # generally not surface in before_send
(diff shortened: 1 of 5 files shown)


sentry_sdk/serializer.py
```diff
@@ -254,6 +254,8 @@ def _serialize_node_impl(
         obj, is_databag, should_repr_strings, remaining_depth, remaining_breadth
     ):
         # type: (Any, Optional[bool], Optional[bool], Optional[int], Optional[int]) -> Any
+        if isinstance(obj, AnnotatedValue):
```
antonpirker, line 257: Could this check be in `_should_repr_strings()`?

antonpirker (commented):
Why not just walk the whole event structure recursively and apply `scrub_dict()`?

As the implementation is now, when we (or somenone else) adds a top level key "super_secret_stuff" it will not be touched.

antonpirker:
I get that having those seperate scrubbing functions makes it convenient for a user to derive a custom scrubber that works like the original one, but does not scrub the breadcrumbs.

antonpirker:
But maybe add a `scrub_others` or `scrub_rest`

That does something like:
```
for key in event.keys():
    if key not in ["request", "extra", "user", ...]:
        self.scrub_dict(event[key])
```

sl0thentr0py:
> Why not just walk the whole event structure recursively and apply scrub_dict()?

Only python has the `Event` as an opaque dict/object without interfaces, no other SDK does this, so I just went with a structure that would work for all SDKs. The recursive walking can only work in python sadly.

I think this is fine for a first version and we can see how and if users use this at all and wait for feedback.
I just wanted to have something like this as a first-class abstraction and this is a good first step.

antonpirker:
Ok. 
How can we ensure, that if we add some top level key to the event in the future, that we do not forget to update the default EventScrubber? Because we (or whoever is than working on this) will forget about this.

sentry_sdk/client.py
```diff
@@ -249,6 +253,11 @@ def _prepare_event(
             self.options["project_root"],
         )
 
+        if event is not None:
```
cleptric, line 256: Shall we run this after the `before_send` callbacks?

sl0thentr0py: we always said `before_send` sees what actually gets sent so let's not change that semantic

antonpirker (approved):

PR #1954, opened 2023-03-14
Added top level api to get current span and transaction

antonpirker:
Idea originated in this comment of a code review: https://github.com/getsentry/sentry-python/pull/1089#discussion_r1133985202

### DIFF
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -54,8 +54,6 @@ __all__ = [
     "set_level",
     "set_measurement",
     "get_current_span",
-    "get_current_transaction",
-    "get_current_span_or_transaction",
 ]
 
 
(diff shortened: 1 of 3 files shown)


sl0thentr0py (approved):

sentry_sdk/api.py
```diff
@@ -53,6 +53,9 @@ def overload(x):
     "set_user",
     "set_level",
     "set_measurement",
+    "get_current_span",
+    "get_current_transaction",
+    "get_current_span_or_transaction",
```
sl0thentr0py, line 58: you also have to add these to `__init__.py`

sl0thentr0py (changes requested):

sentry_sdk/api.py
antonpirker: done.

sl0thentr0py (approved):

sl0thentr0py (changes requested):
as discussed in slack, we just need `get_current_span`, everything else about the transaction is internal details

sl0thentr0py (approved):

PR #1967, opened 2023-03-21
Celery Beat auto monitoring

antonpirker:
Automatically monitor Celery Beat tasks with Sentry Crons.

How we do this:
- we dynamically create a function that listens to the `celery_beat_init` hook. In the hook we do two things:
- 1.) patch existing scheduled tasks (in `sender.scheduler.schedule`):
    - Each scheduled task is patched to contain information about the Sentry monitor (the monitor slug and config (timezone, schedule, ...) in its headers.
    - We then stop Celery Beat and replace the scheduled tasks with the new patched scheduled tasks  
    - We restart Celery Beat to enable our patched tasks
- 2.) Connect each task to the following hooks to send information about the task to sentry: `task_prerun`, `task_success`, `task_failure`, `task_retry`. (config is sent by the tasks in its headers we set up in 1))

Docs: https://github.com/getsentry/sentry-docs/pull/6526


sentry_sdk/integrations/celery.py
```diff
@@ -46,10 +64,13 @@
 class CeleryIntegration(Integration):
     identifier = "celery"
 
-    def __init__(self, propagate_traces=True):
-        # type: (bool) -> None
+    def __init__(self, propagate_traces=True, celery_app=None):
```
sl0thentr0py, line 67: can't we auto discover the app?

sl0thentr0py (dismissed):
can you please summarize the important moving parts in the PR description? at least a high level overview of what we're patching etc

sentry_sdk/integrations/celery.py
antonpirker: Hm, maybe I can get the app from the sender in the hook. I will have a look.

antonpirker: Auto discovering works and look also better :-)

sentry_sdk/integrations/celery.py
```diff
@@ -46,10 +64,13 @@
 class CeleryIntegration(Integration):
     identifier = "celery"
 
-    def __init__(self, propagate_traces=True):
-        # type: (bool) -> None
+    def __init__(self, propagate_traces=True, monitor_beat_tasks=False):
```
antonpirker, line 67: Now we need a flag for enabling monitoring of our Beat schedules, because we auto discover the celery app. fyi @sl0thentr0py

HazAT (approved):
Rubberstamp ✅

Issue #1974, opened 2023-03-24
Do not truncate SQL queries in Span description


antonpirker:
### Problem Statement

Python SDK truncates SQL queries and appends an ellipsis at the end. Depending on the length of the parameter list, the query might be truncated in different places, which affects issue grouping in performance issues.

### Solution Brainstorm

Why not just always put the whole (parameterized) SQL query in there?

antonpirker:
Related #1953

antonpirker:
Done

PR #1983, opened 2023-03-31
Do not trim span descriptions.

antonpirker:
- Made sure that span descriptions are never trimmed. (for all `op` values, not just `db` spans.)
- Removed the experimental `smart_transaction_trimming` option
- Also removed some dead code that was never executed because the experimental option defaults to `False`.

Refs #1974

### DIFF
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -8,13 +8,9 @@ from sentry_sdk.utils import (
     capture_internal_exception,
     disable_capture_event,
     format_timestamp,
-    json_dumps,
     safe_repr,
     strip_string,
 )
-
-import sentry_sdk.utils
-
 from sentry_sdk._compat import (
     text_type,
     PY2,
@@ -23,12 +19,9 @@ from sentry_sdk._compat import (
     iteritems,
     binary_sequence_types,
 )
-
 from sentry_sdk._types import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from datetime import timedelta
-
     from types import TracebackType
 
     from typing import Any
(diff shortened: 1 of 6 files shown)


sentry_sdk/serializer.py
```diff
@@ -365,113 +356,23 @@ def _serialize_node_impl(
...
-        # potentially long SQL queries that are most useful when not truncated.
-        # Because arbitrarily large events may be discarded by the server as a
-        # protection mechanism, we dynamically limit the description length
-        # later in _truncate_span_descriptions.
-        if (
-            smart_transaction_trimming
-            and len(path) == 3
-            and path[0] == "spans"
-            and path[-1] == "description"
-        ):
```
antonpirker, line 380: The real change is changing this if. Because `smart_transaction_trimming` evaluated always to "False" the span description was always truncated with `strip_string`. 

By removing the check for `smart_transaction_trimming` the span description will never be truncated

sentry_sdk/serializer.py
```diff
@@ -365,113 +356,23 @@ def _serialize_node_impl(
...
+    # Start of serialize() function
+    #
     disable_capture_event.set(True)
     try:
-        rv = _serialize_node(event, **kwargs)
-        if meta_stack and isinstance(rv, dict):
-            rv["_meta"] = meta_stack[0]
-
-        sum_span_description_bytes = sum(span_description_bytes)
-        if smart_transaction_trimming and sum_span_description_bytes > 0:
```
antonpirker, line 446: Everything in this IF probably never run.

sl0thentr0py (approved):
lgtm, nice work!
let's just hope no body complains about the experimental option being removed, but it's fine

PR #1994, opened 2023-04-08
Add `loguru` integration

### LINKED ISSUE #653: Add support of loguru logger.
Please add support for [loguru](https://github.com/Delgan/loguru). I can make pull request but I don't know how to correctly integrate structure loggers.

PerchunPak:
Actually, this is the solution in comments under #653 adapted to codebase and tested as well.
https://github.com/getsentry/sentry-python/issues/653#issuecomment-788854865

I also changed `logging` integration to use methods instead of functions in handlers, as in that way we can easily overwrite parts that are different in `loguru` integration. It shouldn't be a problem, as those methods are private and used only in that file.

Please let me know if I missed something, or you have more ideas how to test this.

Fixes #653.

### DIFF
--- a/.github/workflows/test-integration-loguru.yml
+++ b/.github/workflows/test-integration-loguru.yml
@@ -1,0 +1,73 @@
+name: Test loguru
+
+on:
+  push:
+    branches:
+      - master
+      - release/**
+
+  pull_request:
+
+# Cancel in progress workflows on pull_requests.
+# https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+env:
+  BUILD_CACHE_KEY: ${{ github.sha }}
+  CACHED_BUILD_PATHS: |
+    ${{ github.workspace }}/dist-serverless
+
+jobs:
+  test:
+    name: loguru, python ${{ matrix.python-version }}, ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 45
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.5","3.6","3.7","3.8","3.9","3.10","3.11"]
+        # python3.6 reached EOL and is no longer being supported on
+        # new versions of hosted runners on Github Actions
+        # ubuntu-20.04 is the last version that supported python3.6
+        # see https://github.com/actions/setup-python/issues/544#issuecomment-1332535877
+        os: [ubuntu-20.04]
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Setup Test Env
+        run: |
+          pip install codecov "tox>=3,<4"
+
+      - name: Test loguru
+        timeout-minutes: 45
+        shell: bash
+        run: |
+          set -x # print commands that are executed
+          coverage erase
+
+          ./scripts/runtox.sh "py${{ matrix.python-version }}-loguru" --cov=tests --cov=sentry_sdk --cov-report= --cov-branch
+          coverage combine .coverage*
+          coverage xml -i
+          codecov --file coverage.xml
+
+  check_required_tests:
+    name: All loguru tests passed or skipped
+    needs: test
+    # Always run this, even if a dependent job failed
+    if: always()
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check for failures
+        if: contains(needs.test.result, 'failure')
+        run: |
+          echo "One of the dependent jobs have failed. You may need to re-run it." && exit 1
--- a/.github/workflows/test-integration-loguru.yml
+++ b/.github/workflows/test-integration-loguru.yml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Setup Test Env
         run: |
-          pip install codecov "tox>=3,<4"
+          pip install coverage "tox>=3,<4"
 
       - name: Test loguru
         timeout-minutes: 45
@@ -55,10 +55,15 @@ jobs:
           set -x # print commands that are executed
           coverage erase
 
+          # Run tests
           ./scripts/runtox.sh "py${{ matrix.python-version }}-loguru" --cov=tests --cov=sentry_sdk --cov-report= --cov-branch
           coverage combine .coverage*
           coverage xml -i
-          codecov --file coverage.xml
+
+      - uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage.xml
 
   check_required_tests:
     name: All loguru tests passed or skipped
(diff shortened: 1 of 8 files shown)


shanamatthews:
@antonpirker - want to take a look at this?

antonpirker:
I am a bit swamped right now, so this could take some time.

PerchunPak:
I do not really understand the error. This doesn't look like my problem.

antonpirker:
Sometimes our tests are a bit flaky. I have updated your branch with the current master, so the tests will run again.

antonpirker:
Hey @PerchunPak 

I finally had some time to play with this integration. Nice work!

I create a small test project for testing:
```python
from loguru import logger as loguru_logger

import sentry_sdk
import sentry_sdk.integrations.loguru
sentry_sdk.init(
    dsn="...",
    integrations=[
        sentry_sdk.integrations.loguru.LoguruIntegration(),
    ],
    traces_sample_rate=1.0, 
    debug=True,
)

def main():
    loguru_logger.debug("Guru: I am ignored")
    loguru_logger.info("Guru: I am a breadcrumb")
    loguru_logger.error("Guru: I am an event", extra=dict(bar=43))
    loguru_logger.exception("Guru: An exception happened")    


if __name__ == "__main__":
    main()
```

So the `error` log entry creates an "information like" event in Sentry and the `exception` log entry creates an exception in Sentry. Very nice!

Screenshot 2023-04-26 at 12 03 54

The error message contains all the formatted Loguru message with all the information. This is not the best developer experience. Could you please change this, to only contain the message from the call to "loguru_logger.exception()"?

One other thing I noticed is, that `loguru_logger.exception()` does not attach an exception object, so this code is never reached: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/logging.py#L209-L213

Is this just how Loguru works, or am I doing something wrong?

PerchunPak:
> The error message contains all the formatted Loguru message with all the information. This is not the best developer experience. Could you please change this, to only contain the message from the call to "loguru_logger.exception()"?

Unfortunately, looks like no. Loguru parses the message before it is passed to our handler, so we get only a formatted message. We can somehow cut unnecessary parts (e.g. regex or build the unnecessary parts of message by ourselves to cut it) or we can change the format of messages to contain only plain messages (`loguru.add(..., format="{message}")`). The second way will make it impossible to include loguru's data into messages.

image

> Is this just how Loguru works, or am I doing something wrong?

Both, you don't catch any exception, so loguru doesn't provide it. This code works, and also doesn't send formatted messages by loguru in the report name.

```py
try:
    1 / 0
except ZeroDivisionError:
    loguru_logger.exception("Guru: An exception happened")
```

If there is no exception, loguru prints `None` to the console (is it a bug?)

```py
>>> from loguru import logger
>>> logger.exception("Guru: An exception happened") 
2023-04-26 21:29:02.780 | ERROR    | __main__:<module>:1 - Guru: An exception happened
NoneType: None
```

---

I also do not understand what I missed in `tox.ini` so loguru [is not installed](https://github.com/getsentry/sentry-python/actions/runs/4805762527/jobs/8552508507?pr=1994#step:5:269) in CI. Could you look at it please?

antonpirker:
> or we can change the format of messages to contain only plain messages (loguru.add(..., format="{message}")).

We should not do this. We should never change the behavior of the users code, just add functionality on top of it.
As the users can set whatever formatting that she likes the regex approach seems not possible. 

Can we "build the unnecessary parts of message" with the formatting the users has set so it always works, no matter what the configuration of the user looks like?

antonpirker:
Or one other option: can we hook into something that is run before our handler, where we have access to the raw message?

PerchunPak:
`loguru` works different from `logging` module. In `logging`, you can set the format of message globally, while in `loguru` - every handler can have a customized message (handler added with `logger.add`). We can't get the user's message, because the user can configure multiple loggers, that will output to multiple terminals/files. E.g. the user wants to have logs in files too, so they will just do

```python
loguru.add("logs.txt", format="some custom format")  # and maybe some rotation, doesn't matter
```

So if we will monkeypatch `loguru`'s global handler - we will get duplicates because

```python
logger.info("hi!")
# calls all handlers:
# 0 - default (stderr logger)
# 1 - what we added upper (log into file)
# 2 - sentry's breadcrumb handler   # only if INFO >= logging_level > ERROR
# 3 - sentry's event handler        # only if ERROR >= logging_level
```

All of them can have custom format, so we can't catch what the user actually sees in the terminal. BUT we can provide configuration options that we will send into `loguru`.

---

Also, as I understand, all info that provides `loguru` in the console, Sentry provides too in the UI.

```
>>> import sentry_sdk
... import sentry_sdk.integrations.loguru
... sentry_sdk.init(
...     dsn="...",
...     integrations=[
...         sentry_sdk.integrations.loguru.LoguruIntegration(),
...     ],
...     traces_sample_rate=1.0,
...     debug=True,
... )
 [sentry] DEBUG: Setting up integrations (with default = True)
 ...
>>> from loguru import logger
>>> import sys
>>> logger.add(sys.stdout, format="{time} {level} {message}")
>>> logger.info("hi")
2023-04-28T09:02:25.169723+0200 INFO hi
2023-04-28 09:02:25.169 | INFO     | __main__:<module>:1 - hi
>>> logger.error("test")
2023-04-28 09:02:36.507 | ERROR    | __main__:<module>:1 - test
 [sentry] DEBUG: Sending envelope [envelope with 1 items (error)] project:... host:....ingest.sentry.io
2023-04-28T09:02:36.507423+0200 ERROR test
 [sentry] DEBUG: [Tracing] Adding `sentry-trace` header ...-...- to outgoing request to https://....ingest.sentry.io/api/.../envelope/.
```

image

---

> can we hook into something that is run before our handler, where we have access to the raw message?

I did this via appending this code to the end of our loguru integration module

```python
import loguru._simple_sinks

class PatchedStandardSink(loguru._simple_sinks.StandardSink):
    def write(self, message):
        parsed_record = None
        actual_message = message.record["message"]

        def handler_hande(record):
            nonlocal parsed_record
            parsed_record = record

        original_handler_handle = self._handler.handle
        self._handler.handle = handler_hande
        original_write(self, message)

        parsed_record.message = actual_message
        parsed_record.msg = actual_message

        original_handler_handle(parsed_record)


original_write = loguru._simple_sinks.StandardSink.write
loguru._simple_sinks.StandardSink.write = PatchedStandardSink.write
```

But this gives absolutely the same result as just setting `format="{message}"`.

image

antonpirker (approved):
Looks good. It does what it should and the mentioned problems we can not fix because of the nature of loguru, so this is good to be merged.

antonpirker:
Hey @PerchunPak !

Again: Great work, thanks again for this contribution!

We now need also some documentation for this. Could you please create a PR in our docs repo. You can just copy the Redis docs and change it to fit Loguru: https://github.com/getsentry/sentry-docs/blob/master/src/platforms/python/common/configuration/integrations/redis.mdx#L19

Thanks!

And if you want to have e small thank you gift please send me your shipping address and t-shirt size to anton.pirker@sentry.io

PerchunPak:
Thank you for the merge!

> We now need also some documentation for this. Could you please create a PR in our docs repo.

It is already done [here](https://github.com/getsentry/sentry-docs/pull/6617).

antonpirker:
Hey @PerchunPak !

Your contribution has been released: https://github.com/getsentry/sentry-python/releases/tag/1.23.0
Congrats!

If you want a little thank you for your contribution please send me your shipping address (and tshirt size) to anton.pirker@sentry.io!

Issue #1990, opened 2023-04-06
Sentry crashes on every endpoint request


GENL:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

quart==0.18.3
sentry-sdk[quart]==1.19.1

### Steps to Reproduce

We have a Sentry Quart integration that worked correctly until today. Since this morning Sentry fails on every API request.
This is the implementation:
```python
from quart import Quart
from sentry_sdk import init as init_sentry
from sentry_sdk.integrations.quart import QuartIntegration

app = Quart(__name__)

init_sentry(
    before_send=filter_sentry_event,
    integrations=[QuartIntegration()],
    traces_sample_rate=0.001,
)

@app.route('/health', methods=['GET'])
def health():
    return {'status': 'pass'}

```

We have an API with a very simple health check endpoint that returns a success message. Now, when we call that endpoint or any other of our endpoints, Sentry crashes with the stack trace below.
From my testing, `before_send` never has the chance to be executed. So Sentry fails before `before_send` gets called.

Note: There was no change to our Sentry's implementation; we haven't updated either the version or the code. It just stopped working.

Thank you.

### Expected Result

What is expected is sentry not crashing on API calls.

### Actual Result

Sentry crashes with the following stack trace:

```
2023-04-06 12:23:02,961 [ERROR] hypercorn.error: Error in ASGI Framework
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/hypercorn/asyncio/task_group.py", line 23, in _handle
    await app(scope, receive, send, sync_spawn, call_soon)
  File "/usr/local/lib/python3.11/site-packages/hypercorn/app_wrappers.py", line 33, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.11/site-packages/sentry_sdk/integrations/quart.py", line 83, in sentry_patched_asgi_app
    return await middleware(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sentry_sdk/integrations/asgi.py", line 139, in _run_asgi3
    return await self._run_app(scope, lambda: self.app(scope, receive, send))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1996, opened 2023-04-09
Change the Quart signal receivers to coroutine functions

pgjones:
This will ensure that the integration works best with Quart 0.19 (not yet released) and pre 0.19 versions. As the latter would have implicitly converted the signal receivers to coroutine functions anyway. There is sadly no way to support Quart <= 0.18 and Blinker >= 1.6.

@antonpirker something to consider.

### DIFF
--- a/sentry_sdk/integrations/quart.py
+++ b/sentry_sdk/integrations/quart.py
@@ -151,7 +151,7 @@ def _set_transaction_name_and_source(scope, transaction_style, request):
         pass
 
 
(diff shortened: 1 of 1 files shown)


antonpirker:
Hey @pgjones !

Thanks for the fix! 
I have created a new PR including your fix and also updating how we test our Quart integration here:
https://github.com/getsentry/sentry-python/issues/1990

I will close this PR in favor for the other one.

PR #2001, opened 2023-04-11
Celery Beat monitoring without restarting the Beat process

antonpirker:


sentry_sdk/integrations/celery.py
```diff
@@ -387,123 +392,47 @@ def _get_monitor_config(celery_schedule, app):
...
-    app.Beat(schedule=cloned_schedule.name).run()
+        check_in_id = capture_checkin(
+            monitor_slug=monitor_name,
+            monitor_config=monitor_config,
+            status=MonitorStatus.IN_PROGRESS,
+        )
+        headers.update({"sentry-monitor-check-in-id": check_in_id})
 
+        schedule_entry.options.update(headers)
+        original_apply_entry(*args, **kwargs)
```
sl0thentr0py, line 424: do you have to return the value to the original function?

sl0thentr0py (approved):

sentry_sdk/integrations/celery.py
antonpirker: good catch. The original one does not have an return value, but it can not hurt. Makes it forward compatible

PR #2003, opened 2023-04-11
Fixed support for Quart

antonpirker:
- Changed Quart signal receivers to async functions
- Fixed test setup for Quart

Fixes https://github.com/getsentry/sentry-python/issues/1990


sl0thentr0py (approved):
please describe the actual change in the PR title/description, like

> Change quart signal receivers to async

instead of the current generic title.

PR #2009, opened 2023-04-12
Send all events to /envelope endpoint when tracing is enabled

### LINKED ISSUE #2008: Use `/envelope` endpoint for all events when performance is enabled
(do not use `/store` for those events)

Also make sure the dynamic sampling context is also sent with all event going to `/envelope`

antonpirker:
When tracing is enabled we can safely send all events to `/envelope`. The dynamic sampling context (DSC) is sent with all envelopes.

The legacy `/store` endpoint will be removed in the future.

Fixes #2008

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -18,7 +18,7 @@ from sentry_sdk.utils import (
     logger,
 )
 from sentry_sdk.serializer import serialize
-from sentry_sdk.tracing import trace
+from sentry_sdk.tracing import trace, has_tracing_enabled
 from sentry_sdk.transport import make_transport
 from sentry_sdk.consts import (
     DEFAULT_OPTIONS,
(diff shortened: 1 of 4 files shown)


cleptric (commented):
We considered this change only to apply if "propagation" is enabled.
As we didn't define what this will be, I'm not in favour of merging this just yet.
Especially since this will be a hidden change in behaviour and not a new opt-in mechanism.

sentry_sdk/client.py
```diff
@@ -495,25 +495,27 @@ def capture_event(
         if not is_transaction and not self._should_sample_error(event):
             return None
 
+        tracing_enabled = has_tracing_enabled(self.options)
```
cleptric, line 498: Unrelated to the general discussion, `has_tracing_enabled` relies on `enable_tracing` being set, hence we do not solve the immediate need of sending all events via `/envelope` when performance is enabled.

sl0thentr0py (approved):

HazAT (approved):

PR #2013, opened 2023-04-14
chore(profiling): Remove profile context from sdk

Zylphrex:
The profile context can be populated by the relay automatically. No need to do this in the SDK. This also means that if the profile has to be dropped by relay due to rate limits or any reason, we won't render a bad link on the transaction to a non-existent profile.


antonpirker (approved):
lgtm

PR #2025, opened 2023-04-21
Add support for ExceptionGroups

### LINKED ISSUE #1788: Add support for ExceptionGroup
### Problem Statement

ExceptionGroup was introduced in Python 3.11:
https://peps.python.org/pep-0664/#features-for-3-11
https://peps.python.org/pep-0654/

Our current SDK v1.12.0 does not support capturing all exceptions separately with its own stack traces so they can be shown in Sentry UI.

This issue was originally raised here on a general manner to expand the feature to other SDKs:  https://github.com/getsentry/sentry/issues/37716

### Solution Brainstorm

.

antonpirker:
With Python 3.11 ExceptionGroups was introduced. This adds support for catching them and displaying them in a meaningful way.

See also the related RFC: https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md

Fixes #1788


sl0thentr0py:
@antonpirker pls rebase/merge master

sentry_sdk/utils.py
```diff
@@ -658,19 +664,56 @@ def single_exception_from_error_tuple(
...
+
+    if is_root_exception and "type" not in exception_value["mechanism"]:
+        exception_value["mechanism"]["type"] = "generic"
+
+    if exc_type == ExceptionGroup:
+        exception_value["mechanism"]["is_exception_group"] = True
+
+    exception_value["module"] = get_type_module(exc_type)
+    exception_value["type"] = get_type_name(exc_type)
+    exception_value["value"] = (
```
sl0thentr0py, line 712: before this was `safe_str(exc_value)` without any conditions, this doesn't seem to be the same logic as before.
Please explain what you're doing here.

sentry_sdk/utils.py
```diff
@@ -658,19 +664,56 @@ def single_exception_from_error_tuple(
...
+            "errno", {}
+        ).setdefault("number", errno)
+
+    if source is not None:
+        exception_value["mechanism"]["source"] = source
+
+    is_root_exception = exception_id == 0
+    if not is_root_exception:
+        exception_value["mechanism"]["parent_id"] = parent_id
+        del exception_value["mechanism"]["handled"]
```
sl0thentr0py, line 700: why the `del`?

sl0thentr0py (changes requested):
couple of questions to start, I need to read the PEP to review this properly, will get back to it.

sentry_sdk/utils.py
antonpirker: The RFC states, that the `handled` should always be taken from the root exception and I thought I make it extra explicit. 

See https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md#issue-titles

sl0thentr0py: I would just leave this for the server to process, the SDK can just send those over and have the handled in each separate exception of the group, it's extra metadata that doesn't hurt to have in case we change the product in the future. cc @mattjohnsonpint what do you think?

sl0thentr0py: I would just leave this for the server to process, the SDK can just send those over and have the handled in each separate exception of the group, it's extra metadata that doesn't hurt to have in case we change the product in the future. cc @mattjohnsonpint what do you think?

sentry_sdk/utils.py
sl0thentr0py, line 1: since `walk_exception_chain` is no longer being used, this part of the logic is not being handled at all as before?
https://github.com/getsentry/sentry-python/blob/9272f9f12f2df791ebda14999ce08d9f27790b2b/sentry_sdk/utils.py#L738

can you maybe write this in a way that all the older code branches are preserved and the new exception group logic is applied separately so the changes are clear?

This way you're changing old code in a core part of the product which is always risky imo.

sl0thentr0py (changes requested):

sentry_sdk/utils.py
```diff
@@ -658,19 +664,56 @@ def single_exception_from_error_tuple(
...
+
+        exception_value["mechanism"]["type"] = "chained"
+
+    if is_root_exception and "type" not in exception_value["mechanism"]:
+        exception_value["mechanism"]["type"] = "generic"
+
+    if exc_type == ExceptionGroup:
```
sl0thentr0py, line 707: should this be `BaseExceptionGroup`?

sentry_sdk/utils.py
mattjohnsonpint: The overall "handled" of the issue should come from the root exception, but it's ok if each individual exception has its own handled flag.  Indeed, if an exception is caught and a new exception is thrown, then the first exception (the child) was handled - even if the parent exception wasn't.

antonpirker: Ok. Will set handled everywhere.

sentry_sdk/utils.py
antonpirker: In case of an ExceptionGroup they have a message and it will use this message instead of the value because it is nicer to read. See also https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md#exception-value

sentry_sdk/utils.py
antonpirker: I have now implemented a check for `__suppress_context__` and also the use of `__suppress_context__` like it was in the original implementation.

sentry_sdk/utils.py
antonpirker: yes

sentry_sdk/utils.py
```diff
@@ -740,24 +774,133 @@ def walk_exception_chain(exc_info):
...
 
-    rv.reverse()
+    else:
+        exceptions = []
+        for exc_type, exc_value, tb in walk_exception_chain(exc_info):
+            exceptions.append(
+                single_exception_from_error_tuple(
+                    exc_type, exc_value, tb, client_options, mechanism
+                )
+            )
```
antonpirker, line 900: Here now preserving the old behavior for non group exceptions.

antonpirker: discussed in a meeting

mattjohnsonpint:
This can be shipped as soon as https://github.com/getsentry/sentry/pull/48653 is completed and released to production.  

Also, for the best experience with exception groups, self-hosted Sentry users should take the corresponding update when it is available.

sentry_sdk/utils.py
sl0thentr0py: I still think this is a confusing way of writing this condition. 
If the requirement is to take the `exc_value.message` if it exists, I'd write it the other way as below so that it's clear when the older behavior is preserved. The way you wrote it, if `exc_value` is false-y it will go to `exc_value.message` and throw an error.

```python
    exception_value["value"] = (
         exc_value.message
         if exc_value and hasattr(exc_value, "message")
         else safe_str(exc_value)
     )
```

You can even simplify to 
```python
getattr(exc_value, "message", safe_str(exc_value))
```
which does the same as above.

sentry_sdk/utils.py
```diff
@@ -740,24 +778,137 @@ def walk_exception_chain(exc_info):
...
+    # type: (...) -> Tuple[int, List[Dict[str, Any]]]
+    """
+    Creates the list of exceptions.
+    This can include chained exceptions and exceptions from an ExceptionGroup.
+
+    See the Exception Interface documentation for more details:
+    https://develop.sentry.dev/sdk/event-payloads/exception/
+    """
+
+    # TODO: implement also the seen_exception_ids from walk_exception_chain above.
```
sl0thentr0py, line 800: stale TODO?

sentry_sdk/utils.py
```diff
@@ -740,24 +778,137 @@ def walk_exception_chain(exc_info):
...
+                exception_id=exception_id,
+                source="__context__",
+            )
+            exceptions.extend(child_exceptions)
+
+    # Add exceptions from an ExceptionGroup.
+    is_exception_group = exc_value and hasattr(exc_value, "exceptions")
+    if is_exception_group:
+        for idx, e in enumerate(exc_value.exceptions):  # type: ignore
+            (exception_id, child_exceptions) = exceptions_from_error(
```
sl0thentr0py, line 863: I think you have to passthrough `client_options` to each of these recursive `exceptions_from_error` calls?

sentry_sdk/utils.py
```diff
@@ -740,24 +778,137 @@ def walk_exception_chain(exc_info):
...
+                exc_type=type(e),
+                exc_value=e,
+                tb=getattr(e, "__traceback__", None),
+                mechanism=mechanism,
+                exception_id=exception_id,
+                parent_id=parent_id,
+                source="exceptions[%s]" % idx,
```
sl0thentr0py, line 870: shouldn't the source id be `exception_id`? Each `enumerate` will start over from 0 so if you have multiple nestings, the source won't be correct this way I think.

tests/test_exceptiongroup.py
```diff
@@ -0,0 +1,241 @@
...
+try:
+    # Python 3.11
+    from builtins import ExceptionGroup  # type: ignore
+except ImportError:
+    # Python 3.10 and below
+    ExceptionGroup = None
+
+
+def test_exceptiongroup():
+    exception_groups_available = ExceptionGroup is not None
```
sl0thentr0py, line 13: please use `pytest.mark.skipif` like this to skip instead of return in tests

https://github.com/getsentry/sentry-python/blob/9457282e02e2fec6d3d658f4e19c45a336e9e70b/tests/integrations/asyncio/test_asyncio_py3.py#L11-L13

tests/test_exceptiongroup.py
```diff
@@ -0,0 +1,241 @@
...
+            "include_local_variables": True,
+        },
+        mechanism={"type": "test_suite", "handled": False},
+    )
+
+    values = event["exception"]["values"]
+
+    # For this test the stacktrace and the module is not important
+    for x in values:
+        if "stacktrace" in x:
```
sl0thentr0py, line 52: this is fine to test just the exception group nesting structure, but I would add another test with a simpler exception group to make sure the stacktrace/traceback etc is also correct

sl0thentr0py (changes requested):

sentry_sdk/utils.py
antonpirker: yes. will remove

sentry_sdk/utils.py
antonpirker: Good catch! (or in Austrian: "Pfoa, der Hund is gut!")

sentry_sdk/utils.py
antonpirker: Nice. way cleaner!

sentry_sdk/utils.py
antonpirker: No. the source is where the exception is from. And it is from the exceptions[] array of the source groups. (source is not an id, but __cause__, __context__ or exceptions[i]). So idx is correct here..

tests/test_exceptiongroup.py
antonpirker: ok (there was some reason why I did it like that, but I cant remember. Lets try the decorator and see if the reason pops up again or not.)

tests/test_exceptiongroup.py
antonpirker: done.

sentry_sdk/utils.py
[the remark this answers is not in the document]
antonpirker: Why is that? To what? I think I am missing something...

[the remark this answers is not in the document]
sl0thentr0py: nvm they all derive from `BaseException`, ignore
```
>>> builtins.ExceptionGroup.__bases__
(<class 'BaseExceptionGroup'>, <class 'Exception'>)
>>> builtins.BaseExceptionGroup.__bases__
(<class 'BaseException'>,)
>>> builtins.Exception.__bases__
(<class 'BaseException'>,)
```

andyljones:
Hooray! I'm excited to start using this. I updated to 1.24 and tried it out, and I'm not seeing the nested error in the UI yet; it bottoms out at the nursery still. Is there anything I need to do to switch this on?

```python
import trio


async def f():
    raise ValueError("echo")


async def run():
    async with trio.open_nursery() as nursery:
        nursery.start_soon(f)


trio.run(run)
```

image

image

mattjohnsonpint:
@andyljones - That's because you're only raising one error, and the default behavior for trio is to only raise an exception group if there's more than one error in the group.

To raise an exception group from Trio, you can do either of the following:

- Do at least two things that fail:

  ```python
  async def run():
      async with trio.open_nursery() as nursery:
          nursery.start_soon(f)
          nursery.start_soon(f)
  ```

- Tell Trio that you want to use "strict" mode to always get an exception group regardless of number of errors.

  ```python
  async def run():
      async with trio.open_nursery(strict_exception_groups=True) as nursery:
          nursery.start_soon(f)
  ```

See ["Strict" versus "loose" ExceptionGroup semantics](https://trio.readthedocs.io/en/stable/reference-core.html#strict-versus-loose-exceptiongroup-semantics) in the Trio docs.

The [section right before that](https://trio.readthedocs.io/en/stable/reference-core.html#errors-in-multiple-child-tasks) is also a great introduction to exception groups in Trio.

mattjohnsonpint:
There's nothing else needed in Sentry, other than updating to 1.24.0 or newer.  An event with two `ValueErrors` raised by Trio looks like this in Sentry:

image

Note that `NonBaseMultiError` is an exception defined by Trio that inherits from `ExceptionGroup`.

mattjohnsonpint:
Also, just to point out some of the extra work that went into this on the Sentry side, note that an error such as the one shown above will be titled and grouped by the first inner exception:

image

That's because all exceptions within the group are of the same type and have similar values.   Without this, the error would be titled by the root exception, "NonBaseMultiError: multiple tasks failed" - which could make it difficult to distinguish between issues on the issues overview.  Also different quantities of inner exceptions of the same type and similar values will be grouped together, rather than separately. (i.e. having `ValueError:echo` two, three, or four times in the group shouldn't make them different issues.)

Conversely, when the inner exceptions differ significantly - _then_ we'll use the exception group as the title, and group issues based on the combination of the types and values of the inner exceptions.  For example:

image

Will get titled as:

image

If another of the same exception comes through but the `KeyError` and `IndexError` exceptions are swapped, it will still get grouped together as another occurrence of the same issue.

PR #2035, opened 2023-04-25
feat: Add `db.system` indicators to redis, sqlalchemy, and django

AbhiPrasad:
ref https://github.com/getsentry/team-webplatform-meta/issues/60

dependent on https://github.com/getsentry/develop/pull/917

Add this information to span data field as db.system, matching [OpenTelemetry's well known conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#notes-and-well-known-identifiers-for-dbsystem).

For example, db.system of postgresql would indicate that it's a postgres database.

### DIFF
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -11,7 +11,7 @@ from sentry_sdk.hub import Hub, _should_send_default_pii
 from sentry_sdk.scope import add_global_event_processor
 from sentry_sdk.serializer import add_global_repr_processor
 from sentry_sdk.tracing import SOURCE_FOR_STYLE, TRANSACTION_SOURCE_URL
-from sentry_sdk.tracing_utils import record_sql_queries, get_db_system
+from sentry_sdk.tracing_utils import record_sql_queries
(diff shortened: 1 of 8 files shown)


sentry_sdk/integrations/django/__init__.py
```diff
@@ -578,7 +580,10 @@ def execute(self, sql, params=None):
 
         with record_sql_queries(
             hub, self.cursor, sql, params, paramstyle="format", executemany=False
-        ):
+        ) as span:
+            vendor = connection.vendor
```
AbhiPrasad, line 584: I realized this strategy doesn't work if you have multiple connections open, but not sure how common that is.

antonpirker: In Django it is common to have one connection to a read only replica of the database and one you use to change the data, so probably on bigger installations quite common.

AbhiPrasad:
Splitting this up in favour of 4 PRs:

- Adding `db.system` constant: https://github.com/getsentry/sentry-python/pull/2037
- Add redis
- Add sqlalchemy
- Add django

PR #2049, opened 2023-04-27
fix(profiling): Do not keep reference to frame

### LINKED ISSUE #1980: Abnormal memory consumption in celery integration
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.16.0-1.18.0

### Steps to Reproduce

So I have created celery application with sentry integration and worker that allocates large chunks of memory.
I observe that when I use `sentry-sdk==1.15.0` my worker allocates memory ones and stays roughly at the same level for each consecutive task, whereas with version  `>1.15.0` (`1.16.0`, `1.17.0`, `1.18.0`) memory consumption increases with each task. 

### Expected Result

I expect that with newer version of `sentry-sdk` will behave like `1.15.0` in my case.

### Actual Result

I've prepared repository with a code that could help to reproduce a bug.
https://github.com/oleks-popovych/sentry-leak

Zylphrex:
The profiler can capture frames from it's own thread. When it does so, it holds on to a reference to the frame in the previous sample. One of the frames it holds on it is a frame from the profiler itself, which prevents the references to other frames to other frames from being freed. A consequence of this is that the local variables of those frames are not able to be freed either. This change ensures we do not keep a reference to the profiler around in order to prevent this issue.

Closes #1980

### DIFF
--- a/sentry_sdk/_lru_cache.py
+++ b/sentry_sdk/_lru_cache.py
@@ -1,0 +1,156 @@
+"""
+A fork of Python 3.6's stdlib lru_cache (found in Python's 'cpython/Lib/functools.py')
+adapted into a data structure for single threaded uses.
+
+https://github.com/python/cpython/blob/v3.6.12/Lib/functools.py
+
+
+Copyright (c) 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010,
+2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020 Python Software Foundation;
+
+All Rights Reserved
+
+
+PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2
+--------------------------------------------
+
+1. This LICENSE AGREEMENT is between the Python Software Foundation
+("PSF"), and the Individual or Organization ("Licensee") accessing and
+otherwise using this software ("Python") in source or binary form and
+its associated documentation.
+
+2. Subject to the terms and conditions of this License Agreement, PSF hereby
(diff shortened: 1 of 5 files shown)


sentry_sdk/profiler.py
```diff
@@ -791,12 +795,7 @@ def make_sampler(self):
         # type: () -> Callable[..., None]
         cwd = os.getcwd()
 
-        # In Python3+, we can use the `nonlocal` keyword to rebind the value,
-        # but this is not possible in Python2. To get around this, we wrap
-        # the value in a list to allow updating this value each sample.
-        last_sample = [
-            {}
-        ]  # type: List[Dict[int, Tuple[RawStackId, RawStack, Deque[FrameType]]]]
+        cache = LRUCache(max_size=256)
```
Zylphrex, line 798: Need to assess what the right size for this cache is.

sentry_sdk/_lru_cache.py
```diff
@@ -0,0 +1,156 @@
+"""
```
Zylphrex, line 1: Not sure if this will suffice but it's copied from https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/_queue.py

sl0thentr0py (approved):

sentry_sdk/profiler.py
Zylphrex: 256 should be a sufficiently large cache, we want a cache larger than the stack limit of 128 so if 2 unique stacks are captured, it doesn't completely invalidate the cache. At 256, it should have enough room for the common frames to be kept in the cache.

PR #2053, opened 2023-04-28
ref: Use constants for http.query and http.fragment

AbhiPrasad:
#skip-changelog

while working toward https://github.com/getsentry/team-webplatform-meta/issues/60, this is a small refactor to clean up our http related span data (documented in https://develop.sentry.dev/sdk/performance/span-data-conventions/) by making sure they all use the same central constant (and adding docstrings as appropriate). 

In a follow up PR, I'm going to update the spans to start setting `http.method` instead of `method`, because this is the schema we want to use moving forward.


sentrivana (approved):
LGTM 👍

PR #2054, opened 2023-04-28
feat: Use `http.method` instead of `method`

AbhiPrasad:
~Blocked on https://github.com/getsentry/sentry-python/pull/2053 merging, so draft for now.~

As per https://develop.sentry.dev/sdk/performance/span-data-conventions/, we want to rename `method` being set on spans to `http.method`. This better aligns with otel, and also removes ambiguity when operating in Relay.


AbhiPrasad:
@antonpirker this is now ready to review!

antonpirker (approved):

PR #2057, opened 2023-04-28
Add `cache.hit` and `cache.item_size` to Django

### LINKED ISSUE #2043: Add `cache.hit` data to Django spans
### Problem Statement

We want to add information on if a configured cache was hit or missed in Django. 

See also https://github.com/getsentry/team-webplatform-meta/issues/60

### Solution Brainstorm

There are multiple ways to cache data in Django, we probably have to monkey patch Django in various places:
- per-site cache (middlewares) https://docs.djangoproject.com/en/4.2/topics/cache/#the-per-site-cache
- per-view cache (decorator) https://docs.djangoproject.com/en/4.2/topics/cache/#the-per-view-cache
- Template fragment caching (template tag) https://docs.djangoproject.com/en/4.2/topics/cache/#template-fragment-caching
- low level cache api https://docs.djangoproject.com/en/4.2/topics/cache/#the-low-level-cache-api
- async flavor of low level cache api https://docs.djangoproject.com/en/4.2/topics/cache/#asynchronous-support

Maybe just patching the low level cache api will handle all the cases.

antonpirker:
In Django we want to add information to spans if a configured cache was hit or missed and if hit what the item_size of the object in the cache was. 

Fixes #2043


sentrivana (approved):
lgtm!

sentry_sdk/integrations/django/caching.py
```diff
@@ -0,0 +1,101 @@
...
+
+    setattr(cache, method_name, sentry_method)
+
+
+def _patch_cache(cache):
+    # type: (CacheHandler) -> None
+    from sentry_sdk.integrations.django import DjangoIntegration
+
+    integration = Hub.current.get_integration(DjangoIntegration)
+    if integration and integration.cache_spans:
```
sl0thentr0py, line 66: can you move this to `patch_caching` and early return there?

sentry_sdk/integrations/django/caching.py
```diff
@@ -0,0 +1,101 @@
...
+        if not hasattr(cache, "_sentry_patched"):
```
sl0thentr0py, line 67: is this and the other `_sentry_patched` there for a reason or just defensive?

sentry_sdk/integrations/django/caching.py
```diff
@@ -0,0 +1,101 @@
...
+    "get",
+    "get_many",
+]
+
+
+def _instrument_call(cache, method_name, original_method, args, kwargs):
+    # type: (CacheHandler, str, Callable[..., Any], Any, Any) -> Any
+    description = "{} {}".format(method_name, " ".join(args))
+
+    with Hub.current.start_span(op=OP.CACHE, description=description) as span:
```
sl0thentr0py, line 27: you need a hub/integration check before here too because this is during execution of each function and not just during `setup_once` in the beginning.

```python
        hub = Hub.current
        integration = hub.get_integration(DjangoIntegration)
        if integration is None or not integration.cache_spans:
            return None
```

see template example
https://github.com/getsentry/sentry-python/blob/6c7625c1ae7cae1a81761f7576a59f00556d15c2/sentry_sdk/integrations/django/templates.py#L87-L96

sentry_sdk/integrations/django/caching.py
```diff
@@ -0,0 +1,101 @@
...
+from sentry_sdk.consts import OP, SPANDATA
+from sentry_sdk._compat import text_type
+
+
+if TYPE_CHECKING:
+    from typing import Any
+    from typing import Callable
+
+
+METHODS_TO_INSTRUMENT = [
```
sl0thentr0py, line 17: do we want to patch any others here? no setters right?

sentry_sdk/integrations/django/caching.py
```diff
@@ -0,0 +1,101 @@
...
+    "get_many",
+]
+
+
+def _instrument_call(cache, method_name, original_method, args, kwargs):
+    # type: (CacheHandler, str, Callable[..., Any], Any, Any) -> Any
+    description = "{} {}".format(method_name, " ".join(args))
+
+    with Hub.current.start_span(op=OP.CACHE, description=description) as span:
+        cache._sentry_recording = True
```
sl0thentr0py, line 28: this is stored on the `cache` object so if multiple threads are doing cache stuff, I don't think this works?

 also why is this check necessary? is it because `get_many` calls `get` for some types of caches?

sentry_sdk/integrations/django/caching.py
```diff
@@ -0,0 +1,101 @@
...
+
+METHODS_TO_INSTRUMENT = [
+    "get",
+    "get_many",
+]
+
+
+def _instrument_call(cache, method_name, original_method, args, kwargs):
+    # type: (CacheHandler, str, Callable[..., Any], Any, Any) -> Any
+    description = "{} {}".format(method_name, " ".join(args))
```
sl0thentr0py, line 25: what does `args` look like here? Can you post a sample screenshot maybe?

sl0thentr0py (changes requested):

sentry_sdk/integrations/django/caching.py
antonpirker: got it

sentry_sdk/integrations/django/caching.py
antonpirker: currently it is just the getters, because we want to record cache hit or miss.

sentry_sdk/integrations/django/caching.py
antonpirker: Yes, this is done because there is a `get_or_set` function that first calls `get` and then `set` and I only want to record the cache hit/miss of functions the user actually called and that where not called by other functions.

antonpirker: About the thread safety. You are probably right. Did not think about threading. I probably can add a `with treading.lock()` or something to make sure it is thread save. (Will first fix all the other stuff than think about threading)

sentry_sdk/integrations/django/caching.py
antonpirker: to prevent patching things multiple time. making it idempotent

sentry_sdk/integrations/django/caching.py
antonpirker: In `patch_caching` the integration is always None (I guess because this is run in `setup_once()`. But I have now moved the check right before the calls to `_patch_cache`. So it is now not patching if not needed.

sentry_sdk/integrations/django/caching.py
antonpirker: It is (mostly, I guess) just the cache key: https://sentry-sdks.sentry.io/performance/django-otel:14130998f01b457799bf724fbbcc9a74/?project=6539771&query=http.method%3AGET&referrer=performance-transaction-summary&showTransactions=recent&statsPeriod=14d&transaction=%2F&unselectedSeries=p100%28%29

antonpirker: if we later instrument more than just the cache getter it could also be the value that needs to be cached, then we need to check for PII stuff.

sentry_sdk/integrations/django/caching.py
antonpirker: Ok, I will just remove the `_sentry_recording` flag. Because the question we want to have answer is "is my caching setup working great?". And for this it is also important that we record hit/misses no matter if run by the user or by a command the user was calling.

sentry_sdk/integrations/django/caching.py
```diff
@@ -0,0 +1,105 @@
...
+def _patch_cache_method(cache, method_name):
+    # type: (CacheHandler, str) -> None
+    from sentry_sdk.integrations.django import DjangoIntegration
+
+    def _instrument_call(cache, method_name, original_method, args, kwargs):
+        # type: (CacheHandler, str, Callable[..., Any], Any, Any) -> Any
+        hub = Hub.current
+        integration = hub.get_integration(DjangoIntegration)
+        if integration is None or not integration.cache_spans:
+            return None
```
sl0thentr0py, line 32:
```suggestion
            return original_method(*args, **kwargs)
```

sl0thentr0py (changes requested):
wrong return value

sl0thentr0py (approved):

sentry_sdk/integrations/django/caching.py
```diff
@@ -0,0 +1,105 @@
...
+        integration = hub.get_integration(DjangoIntegration)
+        if integration is None or not integration.cache_spans:
+            return original_method(*args, **kwargs)
+
+        description = "{} {}".format(method_name, " ".join(args))
+
+        with hub.start_span(op=OP.CACHE, description=description) as span:
+            value = original_method(*args, **kwargs)
+
+            if value:
```
nicootto, line 39: @antonpirker you can't call `__bool__` of the returned value, it might not be implemented. You should check if a default value was returned, in which case it is a cache miss, or if there is not default value configure you should check `value is not None`.  

I recently reported the same issue on `ddtrace`, you can check of examples on how to reproduce there https://github.com/DataDog/dd-trace-py/issues/5460 and here is how they handle it https://github.com/DataDog/dd-trace-py/pull/5477

Issue #2052, opened 2023-04-27
Cron monitor check-in does not show up in UI when using sentry_sdk.monitor


benjaminrussell-americor:

antonpirker:
Moved to the SDK repo, because it has to be fixed in the SDK.

antonpirker:
According to internal crons people this is due to not sending `monitor_config` or sending a empty one and then relay discards the check in. 

Fix: always send a `monitor_config`

evanpurkhiser:
Thanks for opening this @benjaminrussell-americor -- Fix here https://github.com/getsentry/sentry-python/pull/2058

PR #2058, opened 2023-04-28
fix(crons): Do not send monitor_config when unset

evanpurkhiser:
This fixes a regression introduced in 1.18 which caused monitor
check-ins without a `monitor_config` to be rejected by relay.

Fixes https://github.com/getsentry/sentry-python/issues/2052

### DIFF
--- a/tests/test_crons.py
+++ b/tests/test_crons.py
@@ -90,3 +90,46 @@ def test_capture_checkin_new_id(sentry_init):
         )
 
         assert check_in_id == "a8098c1af86e11dabd1a00112444be1e"
+
+
+def test_end_to_end(sentry_init, capture_envelopes):
+    sentry_init()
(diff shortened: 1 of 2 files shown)


tests/test_crons.py
```diff
@@ -90,3 +90,46 @@ def test_capture_checkin_new_id(sentry_init):
...
+    }
+
+    capture_checkin(monitor_slug="abc123", monitor_config=monitor_config)
+    check_in = envelopes[0].items[0].payload.json
+
+    # Check for final checkin
+    assert check_in["monitor_slug"] == "abc123"
+    assert check_in["monitor_config"] == monitor_config
+
+    # Without passing a monitor_config the field is not in the checkin
```
evanpurkhiser, line 130: This is the case that catches the regression

HazAT (approved):

PR #2079, opened 2023-05-05
Add additional details section in GitHub bug template

EpicWink:


sl0thentr0py:
uhh @EpicWink thanks for this, but our templates are consistent across SDK repos and you can always just add in more info to the issue after creating it and editing. Sorry, closing!

PR #2088, opened 2023-05-08
Add support for cron jobs in ARQ integration

lewazo:
This adds support for logging exceptions that occurs in ARQ cron jobs to Sentry.

Previously, only the tasks defined in the functions argument of the ARQ Worker Settings class were supported. Cron jobs were being ignored.

Since Cron Jobs run automatically by the `worker` object, we need to patch the `worker` instead of the `func` function that was previously patched.


iam-abbas (approved):

tests/integrations/arq/test_arq.py
```diff
@@ -105,36 +127,69 @@ async def division(_, a, b=0):
...
     job = await pool.enqueue_job("division", 1, b=int(not job_fails))
     await worker.run_job(job.job_id, timestamp_ms())
 
-    if job_fails:
-        error_event = events.pop(0)
-        assert error_event["exception"]["values"][0]["type"] == "ZeroDivisionError"
-        assert error_event["exception"]["values"][0]["mechanism"]["type"] == "arq"
+    loop = asyncio.get_event_loop()
+    task = loop.create_task(worker.async_run())
+    await asyncio.sleep(1)
```
lewazo, line 144: This is needed because the cron job automatically execute after the `worker` starts.

There's no way to actually wait for a cron job to finish other than adding some delay before asserting results.

antonpirker:
Hey @lewazo !
Thanks for the PR! 
It will take some time until we have the capacity to review this, please be patient.

antonpirker:
ping

antonpirker (approved):
Nice work!
Looks good to me, and does what it should!

antonpirker:
Btw @lewazo  and @iam-abbas Sentry has now a "Crons" feature, where cron jobs can submit their config (when it should run) on creation of a cron job and then can have check ins on each run (and on each success and failure) and then one has a nice UI in Sentry to see if the crons are running on time and if there are errors. Here is part of the implementation in Celery: https://d655584d05f14c58b86e9034aab6817f@o447951.ingest.sentry.io/5461230

Maybe you want to add this too for ARQ?

lewazo:
Thanks @antonpirker that looks like a nice feature indeed and sounds perfect for this integration.

I'll reserve some time to take a look at implementing it for ARQ

PR #2098, opened 2023-05-10
fix django caching error for values without truthy __bool__

oizik:
Regression in: efa55d

### DIFF
--- a/sentry_sdk/integrations/django/caching.py
+++ b/sentry_sdk/integrations/django/caching.py
@@ -32,6 +32,17 @@ def _get_span_description(method_name, args, kwargs):
     return description
 
 
+def _check_truthy_value(value):
+    # type: (Any) -> bool
+    if value is not None:
+        try:
+            return bool(value)
+        except BaseException:
+            return True
+
+    return False
+
+
 def _patch_cache_method(cache, method_name):
     # type: (CacheHandler, str) -> None
     from sentry_sdk.integrations.django import DjangoIntegration
@@ -48,7 +59,7 @@ def _patch_cache_method(cache, method_name):
         with hub.start_span(op=OP.CACHE_GET_ITEM, description=description) as span:
             value = original_method(*args, **kwargs)
 
-            if value:
+            if _check_truthy_value(value):
                 span.set_data(SPANDATA.CACHE_HIT, True)
 
                 size = len(text_type(value).encode("utf-8"))
--- a/tests/integrations/django/myapp/urls.py
+++ b/tests/integrations/django/myapp/urls.py
@@ -36,6 +36,11 @@ urlpatterns = [
         name="view_with_cached_template_fragment",
     ),
     path(
+        "view-with-cache-of-non-bool-object",
+        views.view_with_cache_of_non_bool_object,
+        name="view_with_cache_of_non_bool_object",
+    ),
+    path(
         "read-body-and-view-exc",
         views.read_body_and_view_exc,
         name="read_body_and_view_exc",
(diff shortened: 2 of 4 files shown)


sentry_sdk/integrations/django/caching.py
```diff
@@ -32,6 +32,17 @@ def _get_span_description(method_name, args, kwargs):
     return description
 
 
+def _check_truthy_value(value):
+    # type: (Any) -> bool
+    if value is not None:
+        try:
+            return bool(value)
+        except BaseException:
+            return True
```
antonpirker, line 41: Why is the value "True" when there is an exception?

oizik: When the `value` is not `None` (which is checked on line 37) and an exception is raised in `bool(value)` then this exception originated from an overriden `__bool__`  method of the `value` object. 

A common case is if `value` is a `DataFrame` which will raise an exception i.e.

```
import pandas as pd
df = pd.DataFrame()
bool(df)
```

will result in `ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().` . Hence it can not be use in an `if` statement:

```
if value:
   ...
```

I added a test case specifically for this in `tests/integrations/django/myapp/views.py`: https://github.com/getsentry/sentry-python/blob/d3f3157321cd0f2ce3ee3e790c1d144cf6910086/tests/integrations/django/myapp/views.py#L77 and https://github.com/getsentry/sentry-python/blob/d3f3157321cd0f2ce3ee3e790c1d144cf6910086/tests/integrations/django/myapp/views.py#LL87C3-L87C3

Happy to change if you suggest a better solution?

antonpirker: I am not suggesting a better solution, the solution looks fine. I just do not get why a `bool(something)` that raises and exception is thruthy, wy not falsy?

oizik: My reasoning was that it should be `True` so that the if statement executes in the following code:

```python
with hub.start_span(op=OP.CACHE_GET_ITEM, description=description) as span:
  value = original_method(*args, **kwargs)
  
  if _check_truthy_value(value):
      span.set_data(SPANDATA.CACHE_HIT, True)
  
      size = len(text_type(value).encode("utf-8"))
      span.set_data(SPANDATA.CACHE_ITEM_SIZE, size)
else:
      span.set_data(SPANDATA.CACHE_HIT, False)
```

Line [62 here](https://github.com/getsentry/sentry-python/blob/d3f3157321cd0f2ce3ee3e790c1d144cf6910086/sentry_sdk/integrations/django/caching.py#LL59C9-L66C62)

The previous code 

```python
with hub.start_span(op=OP.CACHE_GET_ITEM, description=description) as span:
    value = original_method(*args, **kwargs)

    if value:
        span.set_data(SPANDATA.CACHE_HIT, True)

        size = len(text_type(value).encode("utf-8"))
        span.set_data(SPANDATA.CACHE_ITEM_SIZE, size)

    else:
        span.set_data(SPANDATA.CACHE_HIT, False)
```

Line [48 here](https://github.com/getsentry/sentry-python/blob/565567f9c440d9f9e0e20396a91571961a4337c5/sentry_sdk/integrations/django/caching.py#L48-L58) raises an exception, when for example the type of `value` is a `DataFrame` (or anything in general that overrides the `__bool__` method and raises an exception. 

Again using `DataFrame` as an example, 

```python
try:
    return bool(value)
except BaseException:
    return True
```

Would correctly return `True` because even the truth value of a `DataFrame` is ambiguous it still exists and is not `None` and hence should be recorded in the span.

PR #2120, opened 2023-05-17
This disables Django Cache spans by default.

antonpirker:
Users were telling us that the cache spans are adding to much overhead onto their servers, so we are disabling cache spans by default and make them opt-in.

### DIFF
--- a/tests/integrations/django/test_basic.py
+++ b/tests/integrations/django/test_basic.py
@@ -1017,8 +1017,6 @@ def test_cache_spans_middleware(
     use_django_caching_with_middlewares,
     settings,
 ):
-    client.application.load_middleware()
-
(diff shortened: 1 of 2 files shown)


tests/integrations/django/test_basic.py
```diff
@@ -1029,6 +1027,8 @@ def test_cache_spans_middleware(
         ],
         traces_sample_rate=1.0,
     )
+
+    client.application.load_middleware()
```
antonpirker, line 1031: By changing the default it was uncovered, that this test only worked by coincidence. Had to load the middleware after sdk init, to have the options set in DjangoIntegration.

sl0thentr0py (approved):

PR #2130, opened 2023-05-22
Celery beat exclude option

### LINKED ISSUE #2119: Exclude Celery Beat tasks from auto-instrumenting
### Problem Statement

Currently if Celery Beat auto-instrumentation is enabled (with `monitor_beat_tasks=True` option) ALL tasks in the beat schedule are instrumented.  Have the possibility to exclude some tasks from being instrumented.

### Solution Brainstorm

Create a new option `exclude_beat_tasks=[]` that is a list and defaults to an empty list. 
It contains strings of Celery beat task names, the same names that are used in the beat schedule.
If an task is in that list, it will not be autoinstrumented.

antonpirker:
Adds a `exclude_beat_tasks` option to `CeleryIntegration` where users can exclude Celery Beat tasks from being auto-instrumented.

Fixes #2119

### DIFF
--- a/sentry_sdk/crons/api.py
+++ b/sentry_sdk/crons/api.py
@@ -32,6 +32,7 @@ def _create_check_in_event(
     if monitor_config:
         check_in["monitor_config"] = monitor_config
 
+    print(f"check_in: {check_in}")
     return check_in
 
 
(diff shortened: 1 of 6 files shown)


antonpirker:
Changed to always match the full task names.

sl0thentr0py (approved):

PR #2131, opened 2023-05-23
Swap UUID generation to be lazy

bayangan1991:
Running py-spy while the sentry sdk in a django project shows that there is a lot of time spent generating un-used uuid.

This PR simply takes the same UUID format and puts it's behind some custom getter and setter code.

Just in a fairly large django project start I saw number of UUID's used go from several hundred thousand to hundreds.


sl0thentr0py:
hey @bayangan1991 thanks for the PR, that's a valuable observation. 

I'm actually working on some more general 'make the SDK as low overhead as possible', so I'll take this into account for that. I'll keep this PR open for now and come back to you in some time on if we take this in or I incorporate avoiding UUID / span creation in some other way.

sentry_sdk/tracing.py
```diff
@@ -162,6 +162,32 @@ def init_span_recorder(self, maxlen):
...
+    @property
+    def span_id(self):
+        # type: () -> str
+        if not self._span_id:
+            self._span_id = uuid.uuid4().hex[16:]
+
+        return self._span_id
+
+    @span_id.setter
+    def span_id(self, value: str) -> None:
```
sl0thentr0py, line 187:
```suggestion
    def span_id(self, value) -> None:
```

our tests need to pass on 2.7 as well

sl0thentr0py (changes requested):
hey @bayangan1991 I think we'll take this in!
can you fix the one type signature for 2.7?

sentry_sdk/tracing.py
bayangan1991: Ah yeah ok. Force of habit. I dropped the extra `-> None` as well.

sl0thentr0py:
ty! I'll merge this on monday after some quick sanity testing on django

sl0thentr0py:
im gonna reopen this separately since we dont have rights for the fork

PR #2153, opened 2023-06-06
Load AWS Lambda secrets in Github CI

antonpirker:
Make sure our AWS Lambda test setup is correct and the tests work as expected and also in a timely manner.

We run our test in AWS Lambda and then parse the log output to see what events/envelopes where sent. Because Lambda truncates this log output to 4kb, I had to change the tests to make the events/envelopes smaller in size to get the whole event/envelop in the log output.

When the AWS env vars where not set, the tests where skipped but it looked like they where successful. I made them now fail loudly in that case, so we see if they do not run.

Also made the code easier to comprehend. 

Somehow related docs PR: https://github.com/getsentry/sentry-docs/pull/8246 

If this has been merged AND released, this PR in `sentry` can be merged: https://github.com/getsentry/sentry/pull/58321


tests/integrations/aws_lambda/client.py
```diff
@@ -68,110 +172,128 @@ def run_lambda_function(
...
 
-        assert 200 <= response["StatusCode"] < 300, response
-        return response
+# This is for inspecting new Python runtime environments in AWS Lambda
+# If you need to debug a new runtime, use this REPL to run arbitrary Python or bash commands
+# in that runtime in a Lambda function:
+#
+#    pip3 install click
+#    python3 tests/integrations/aws_lambda/client.py --runtime=python4.0
+#
```
sentrivana, line 296: Can't comment on the line directly but on 322 we say "name of the runtime to use, eg python3.8" -- that should probably also be updated to 3.9 if 3.8 is not supported anymore

sentrivana (approved):
Looks good to me!
Awesome that you've fixed this. 🎖️ 
Left a few small comments.

aws-lambda-layer-requirements.txt
```diff
@@ -0,0 +1,7 @@
+certifi
+
+# In Lambda functions botocore is used, and botocore is not
+# yet supporting urrlib3 1.27.0 never mind 2+.
```
sentrivana, line 4:
```suggestion
# yet supporting urllib3 1.27.0 never mind 2+.
```

scripts/aws-deploy-local-layer.sh
```diff
@@ -22,7 +22,7 @@ aws lambda publish-layer-version \
     --region "eu-central-1" \
     --zip-file "fileb://dist/$ZIP" \
     --description "Local test build of SentryPythonServerlessSDK (can be deleted)" \
-    --compatible-runtimes python3.6 python3.7 python3.8 python3.9
+    --compatible-runtimes python3.8 python3.9 python3.10 python3.11 \
```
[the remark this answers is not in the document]
antonpirker: i made a change and now we can support all the python version usesd in Lambda!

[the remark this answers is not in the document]
sentrivana: Awesome!

tests/integrations/aws_lambda/client.py
antonpirker: fixed.

PR #2167, opened 2023-06-12
 feat(integrations): Add integration for clickhouse-driver

mimre25:
Adds an integration that automatically facilitates tracing/recording of all queries, their parameters, data, and results.

Closes https://github.com/getsentry/sentry-python/issues/2154
Companion PR for docs: https://github.com/getsentry/sentry-docs/pull/7129


shanamatthews:
@antonpirker - I was going to ping you to remind you about the docs PR for this, but it looks like you weren't even tagged in this PR yet 😆

antonpirker:
Hey @mimre25 !

I finally had time to look at your PR. Really great work! I added some code for adding more db connection data to the span (we need this to some magic stuff on the Sentry backend) and also made sure that the `db.params` are only sent to Sentry when `send_default_pii` parameter of the `sentry_sdk.init()` is set to `True`. (so by default we will not leak personal data.

Also added the tests to our test mastrix. There is still some work to do, but I will have a look at this tomorrow.

Do you know if there is a way to mock clickhouse for the tests? (I am not sure it is a good idea to spin up clickhouse in CI...)

mimre25:
Unfortunately, I don't know anything to mock clickhouse for tests. 
At work we even spin up a whole cluster (dockerized) for testing (unit + integration), because we haven't found anything when we started to adopt clickhouse

If you find something, though, please let me know, I would be very interested in that as well!

antonpirker:
Hey @mimre25 !
I think we are there! 
(I just made sure all the tests are run and work in CI, plus added some more tests)

When all the tests are green this will be merged and released this week!
Thanks again for the great work. Really amazing!

antonpirker (approved):
🚢

PR #2172, opened 2023-06-15
Do not support sub-minute cron intervals

### LINKED ISSUE #2168: Sub minute intervals for Celery Cron integrations
We do NOT support sub minute intervals in Sentry Crons. There when where would be "interval" type schedule that is sub minute, we should print a warning and return an empty monitor config:

See code here:
[[_get_monitor_confi](https://github.com/getsentry/sentry-python/blob/a1c27600625d20803fc58a008172dcc11d053a01/sentry_sdk/integrations/celery.py#L396-L407)](https://github.com/getsentry/sentry-python/blob/a1c27600625d20803fc58a008172dcc11d053a01/sentry_sdk/integrations/celery.py#L396-L407)

In the `elif` there should be warning and a return {} like in the `else` below in case the interval is under a minute.

antonpirker:
In Celery you can run task in intervals shorter than one minute. Sentry crons does not support this, so in a case where a sub-minute interval is specified, just print a warning and do not set a schedule configuration.

Fixes #2168

### DIFF
--- a/sentry_sdk/integrations/celery.py
+++ b/sentry_sdk/integrations/celery.py
@@ -374,7 +374,7 @@ def _get_humanized_interval(seconds):
             interval = int(seconds / divider)
             return (interval, unit)
 
-    return (1, "minute")
+    return (int(seconds), "second")
 
 
 def _get_monitor_config(celery_schedule, app):
@@ -398,6 +398,12 @@ def _get_monitor_config(celery_schedule, app):
         (schedule_value, schedule_unit) = _get_humanized_interval(
             celery_schedule.seconds
         )
+
+        if schedule_unit == "second":
+            logger.warning(
+                "Intervals shorter than one minute are not supported by Sentry Crons."
+            )
+            return {}
 
     else:
         logger.warning(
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/celery.py
```diff
@@ -399,6 +399,12 @@ def _get_monitor_config(celery_schedule, app):
             celery_schedule.seconds
         )
 
+        if schedule_unit == "second":
+            logger.warning(
+                "Intervals shorter than one minute are not supported by Sentry Crons."
+            )
+            return {}
```
antonpirker, line 406: And when a second interval is returned, we print a warning and do not return a monitor config (this is the behavior the backend guys requested, because then they ignore the checkins)

sentry_sdk/integrations/celery.py
```diff
@@ -440,24 +446,27 @@ def sentry_apply_entry(*args, **kwargs):
...
-        )
-
-        check_in_id = capture_checkin(
-            monitor_slug=monitor_name,
-            monitor_config=monitor_config,
-            status=MonitorStatus.IN_PROGRESS,
-        )
-        headers.update({"sentry-monitor-check-in-id": check_in_id})
+        is_supported_schedule = bool(monitor_config)
+        if is_supported_schedule:
```
antonpirker, line 450: We now only create the "in_progress" checkin for supported schedule types (so if the schedule is sub-minute interval, the check in will not be sent)

Also because the headers are not set in case it is an unsupported type, the ok/failure checkins will also not be sent.

sentrivana (approved):
lgtm

Issue #2186, opened 2023-06-22
Inject tracing HTML `meta` tags in rendered HTML


antonpirker:
### Problem Statement

We have a function `trace_propagation_meta()` (https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/hub.py#L723) but it is never used anywhere.

### Solution Brainstorm

- [X] Refactor `trace_propagation_meta()` to use the new `get_traceparent()` and `get_baggage()` apis (https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/api.py#L252): https://github.com/getsentry/sentry-python/pull/2202
- [x] Flask: Update template helper: https://github.com/getsentry/sentry-python/pull/2203
- [x] Django: Add template helper: https://github.com/getsentry/sentry-python/pull/2222
- [x] Starlette/FastAPI (Jinja) Add template helper: https://github.com/getsentry/sentry-python/pull/2225
- [x] Update docs to include template helper for Flask/Django/Starlette/FastAPI: https://github.com/getsentry/sentry-docs/pull/7325

This will make it WAAY easier for users to setup distributed tracing for use cases when the server is the head of trace and renders HTML and the JS in the frontend then picks up the trace.

sl0thentr0py:
duplicate
https://github.com/getsentry/sentry-python/issues/1341

antonpirker:
Thanks for linking the other issue @sl0thentr0py 
So we have this in Flask already. Nice. 
I will probably remove `trace_propagation_meta()` with something new that uses the new top level functions and then make flask, fastaapi and django use this new function. Cool with you to remove `trace_propagation_meta`?

antonpirker:
Was just thinking this, keeping it and just change the internals.

antonpirker:
This is done, closing the issue

PR #2202, opened 2023-06-27
Use new top level api in `trace_propagation_meta`

antonpirker:
Use new top level api in trace_propagation_meta and also move the functions into the Hub, so they can be used in the Hub. (following the pattern of other top level API)

Refs #2186


sentrivana (approved):
Looks very nice.

PR #2209, opened 2023-06-28
Take trace id always from propagation context

antonpirker:
Make sure that the trace information is always taken from propagation context. (was not the case if you create a span without a transaction, which is happening if you have a vanilla Python without any Integrations and you make an outgoing HTTP request)

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -130,14 +130,16 @@ class Span(object):
         start_timestamp=None,  # type: Optional[datetime]
     ):
         # type: (...) -> None
+        import ipdb
+
+        ipdb.set_trace()
         if trace_id:
             self.trace_id = trace_id
         elif hub:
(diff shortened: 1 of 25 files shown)


sentry_sdk/tracing.py
```diff
@@ -130,7 +130,16 @@ def __init__(
         start_timestamp=None,  # type: Optional[datetime]
     ):
         # type: (...) -> None
-        self.trace_id = trace_id or uuid.uuid4().hex
+        if trace_id:
+            self.trace_id = trace_id
+        elif hub:
+            traceparent = hub.get_traceparent()
```
sl0thentr0py, line 136: the constructor already takes an explicit `trace_id`, so the hub based logic should be in `hub.start_span` and not here. Please leave this constructor unchanged.

sl0thentr0py (changes requested):

sl0thentr0py (approved):

PR #2227, opened 2023-07-04
Support newest starlette versions

antonpirker:
Update our test matrix to support current Starlette versions. (only testing every other version, because otherwise it would be too much versions to test)


sentrivana (approved):
Nice!

PR #2246, opened 2023-07-12
ref(crons): Add information to short-interval cron error message

lobsterkatie:
This adds both the name of the problematic monitor and a suggestion of how to fix the problem to the error message about crons with < 1 min intervals.


lobsterkatie:
@antonpirker - Any theories about why just the one `gevent` test suite is failing? I took a look and it doesn't _seem_ to be at all related to this change, though maybe there's a connection I'm missing. Would you mind taking a look, please?

antonpirker (approved):
Thanks for this @lobsterkatie !

lgtm

PR #2314, opened 2023-08-19
feat(integrations): Add integration for asyncpg

### LINKED ISSUE #1278: asynpg support?
Would it be possible to add support for [asyncpg](https://github.com/MagicStack/asyncpg) in the performance monitoring?
I notice that SQLAlchemy for example is automatically set up, so I was wondering if this support could be extended to convert asyncpg.
Hopefully this is the right place to ask this, thanks

mimre25:
Closes #1278 
I think this is close to being ready (see https://github.com/getsentry/sentry-python/pull/2314#issuecomment-1703792668) for an update.

So far this records every statement that is directly issued, as well as the SQL statements that are used for cursors and prepared statements.

I have a few open questions:

1. I noticed `asyncpg` uses the postgres binary protocol to communicate to postgres. I haven't had any encounter with that yet, but it's making it more challenging to record the creation of cursors, transactions, and prepared statements properly. Do you have a suggestion what to do about this? Keep in mind that all of this is done in Cython, ie we would need to tap into that lower level for recording the exact statements and reverse parsing the binary protocol.
2. For cursors and prepared statements, what is the desired outcome? Should the creation be recorded, and every subsequent use, or just the creation, or just the usages?
3. I've noticed that `tracing_utils:record_sql_queries` checks the `hub.client.options["_experiments"]["record_sql_params"]` and strips out the parameters if it's `True`. However, I did not want to rely on it to strip out parameters, as there is a `TODO` right with it, that hints at a future removal of that check. Therefore I strip them out per default, but provide a config option to the integration for recording the parameters. 
4. I've noticed that the `test-requirements.txt` doesn't contain `pytest-asyncio`, yet the `pytest.ini` configures the `asyncio_mode` to `strict`. Am I missing something here, or is this on purpose?

@antonpirker please let me know what you think :)


antonpirker:
Hey @mimre25 !

Finally I have some time to review. Looks very clean at the first glance! Great job!
Will test it with a local demo project to see what the spans look like.

One thing I noticed, maybe we need to set more data on the span like this: https://github.com/getsentry/sentry-python/blob/antonpirker/2262-traces-sampler-url-instead-of-generic-asgi-request/sentry_sdk/integrations/sqlalchemy.py#L57-L71

antonpirker:
Screenshot 2023-08-30 at 14 45 55

I get `db` spans from your integration! Yay! \o/

antonpirker:
I also added `asyncpg` to the test matrix so the tests run in CI.

antonpirker:
> 1. I noticed asyncpg uses the postgres binary protocol to communicate to postgres. I haven't had any encounter with that yet, but it's making it more challenging to record the creation of cursors, transactions, and prepared statements properly. Do you have a suggestion what to do about this? Keep in mind that all of this is done in Cython, ie we would need to tap into that lower level for recording the exact statements and reverse parsing the binary protocol.

For the first iteration of the asyncpg support it is enough if we record spans for execute, executemany (and maybe db connect if it is easy) The rest we can ignore for now. This will already give a lot of value to users.

> 2. For cursors and prepared statements, what is the desired outcome? Should the creation be recorded, and every subsequent use, or just the creation, or just the usages?

I think we can omit those for now. See above.

antonpirker:
All in all really great work @mimre25 !

If you have addressed my comments I will do another round of test/review!

happyleavesaoc:
I tried this out too, and it worked great. I did notice that the span `duration` is effectively zero in my test (and also the screenshot above).

mimre25:
Alright, the new commits do the following:
- [x] Reading test connection params from env
- [x] Fix the incorrect span durations (see below)
- [x] Record additional information (executemany, cursor, etc)

The span durations were incorrectly recorded because I returned the coroutines instead of awaiting them and returning them.
<details>

<summary>Screenshot of span duration of old implementation:</summary>

image

</details>

<details>

<summary>Screenshot of span duration of new implementation:</summary>

image

</details>


There are a few things that I still plan to do, but don't have the time right now:

- [ ] Recording `connect` statements
- [ ] Record `execute` queries without parameters (I noticed this doesn't happen at the moment)

sentry_sdk/integrations/asyncpg.py
```diff
@@ -60,3 +67,64 @@ async def _inner(*args: P.args, **kwargs: P.kwargs) -> T:
...
+            self._query,
+            params_list,
+            param_style,
+            executemany=False,
+            record_cursor_repr=True,
+        ) as span:
+            try:
+                res = await f(self)
+            except StopAsyncIteration:
+                span.set_data("db.cursor.exhausted", True)
```
mimre25, line 125: I'm really confused why this is not recorded. 

The code is reached, but it doesn't matter what data I'm setting on this span, it won't be recorded.
What am I missing here?

sentry_sdk/tracing_utils.py
```diff
@@ -132,6 +134,8 @@ def record_sql_queries(
         data["db.paramstyle"] = paramstyle
     if executemany:
         data["db.executemany"] = True
+    if record_cursor_repr and cursor is not None:
+        data["db.cursor"] = cursor
```
mimre25, line 138: I decided to add this to the tracing utils, to allow recording the cursor instance. 

I can see this as being useful if someone does subsequent fetches from a cursor.

sentry_sdk/tracing_utils.py
```diff
@@ -107,6 +107,8 @@ def record_sql_queries(
     params_list,  # type:  Any
     paramstyle,  # type: Optional[str]
     executemany,  # type: bool
+    *,
+    record_cursor_repr: bool = False
```
mimre25, line 111: Not sure how to go about adding this to the interface of `record_sql_queries`.

sentry_sdk/integrations/asyncpg.py
mimre25: Ok, I figured out, that we don't check for the data of the span in the test but for the data of the breadcrumb.

Is there a way to test for the data for a span?

sentry_sdk/integrations/asyncpg.py
```diff
@@ -0,0 +1,236 @@
...
+            res = await f(self, n, timeout)
+        return res
+
+    return _exec
+
+
+def _wrap_cursoriterator_anext(
+    f: Callable[..., Awaitable[T]]
+) -> Callable[..., Awaitable[T]]:
+    async def __await__(self: CursorIterator) -> T:  # noqa: N807
```
antonpirker, line 159: We will not instrument this. Because if someone does a query returning one million rows, then one million spans will be created. That is just too much. (And in the UI it looks like the same query was run one million times. Which then would trigger Sentry's n+1 query detector, what we also do not want.)

Could you please remove the instrumentation of `__await__` of the CursorIterators anext?

antonpirker (changes requested):
Hey! Did a second round of review. Looks very good already!

I updated the data we store on a span, to make it include all the information Sentry needs for doing some magic stuff with database spans. 

There is one instrumentation of the `__await__` that we can not release (see comment in the review)

And could you please fix the type problems in the tests that break the tests? Thanks!

sentry_sdk/integrations/asyncpg.py
```diff
@@ -0,0 +1,62 @@
+from typing import ParamSpec, TypeVar, Callable
```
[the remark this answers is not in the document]
antonpirker: I removed now `ParamSpec` and replaced it with `Any`. I know it is not optimal, but we need to support older Python versions that do not support `ParamSpec`

antonpirker:
I also added a PR to the docs to add the new integration: https://github.com/getsentry/sentry-docs/pull/7756

mimre25:
I've removed the cursor instrumentation, and now only record the creation of a cursor instead:
image

image


All typing and lint fixes are done, and tox passes locally for `py3.{7-11}-asyncpg` & `linters`

sentry_sdk/integrations/asyncpg.py
mimre25: Removed it :slightly_smiling_face: 

I was thinking of maybe leaving it in and cutting it after the 10th iteration, but since the `CursorIterator` uses slots, it would be too much of a mess to fiddle the counting in.

I could, however, see this as a future improvement.

sentry_sdk/integrations/asyncpg.py
mimre25: This is removed now, since we don't instrument every iteration of a cursor anymore.

antonpirker:
Oh, CI tests can not connect to postgres. I will fix that.

Issue #2291, opened 2023-08-04
Add .viscose to .gitignore


szokeasaurusrex:
### Problem Statement

The `.vscode` directory, which is used to store local VSCode configuration options, is being tracked by Git. Since my local settings differ from what is stored in the repo, I always need to make sure not to stage my `.vscode` directory when commuting changes in Git. Since different developers may have different VSCode configurations, I don’t see any reason why we should be tracking this directory in our repo. 

### Solution Brainstorm

Add the `.vscode` directory to `.gitignore`. If we implement this change, we should probably also delete the `.vscode` directory from the GitHub repo.

PR #2317, opened 2023-08-23
Enhancement/add .vscode to .gitignore

shoaib-mohd:
Add .vscode to .gitignore #2291

As [szokeasaurusrex](https://github.com/szokeasaurusrex) Said in the problem statement mentioned below

Problem Statement
The .vscode directory, which is used to store local VSCode configuration options, is being tracked by Git. Since my local settings differ from what is stored in the repo, I always need to make sure not to stage my .vscode directory when commuting changes in Git. Since different developers may have different VSCode configurations, I don’t see any reason why we should be tracking this directory in our repo.

Solution Brainstorm
Add the .vscode directory to .gitignore. If we implement this change, we should probably also delete the .vscode directory from the GitHub repo.

I have successfully deleted the .vscode directory from the project and also added .vscode directory in .gitignore file.


szokeasaurusrex (changes requested):
It looks like your PR includes several changes unrelated to adding .vscode to .gitignore. Please remove these changes from your PR so we can merge it.

shoaib-mohd:
Hello [szokeasaurusrex](https://github.com/szokeasaurusrex)
I am trying to revert the unwanted changes but these changes are just the next line at the end of every file.
These changes are done by running the "pre-commit run --all-files" command.

PR #2341, opened 2023-09-01
Fix transaction name in Starlette and FastAPI

### LINKED ISSUE #2262: `traces_sampler` receives always "generic ASGI request" as transaction name.
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.11.1

### Steps to Reproduce

I have manually instrumented traces_sampler counted on transaction name in sampling context. At some point with sentry-sdk version upgrade, it stopped work - there is always "generic ASGI request" in transaction name. But at the performance page, transaction has a proper name.

How we initiate sentry FastAPI integration:
```
sentry_sdk.init(
    dsn=SENTRY_DSN,
    environment="dev",
    traces_sampler=sentry_traces_sampler,
    integrations=[
        starlette.StarletteIntegration(transaction_style="endpoint"),
        fastapi.FastApiIntegration(transaction_style="endpoint"),`
    ]
)
```

The traces_sampler function:
```
def sentry_traces_sampler(sampling_context):
    transaction_sampling_mapper = {
        "app.api.api_v1.routers.users.some_name": 0.1,
        "app.api.api_v1.routers.messages.some_name_2": 0.5,
    }
    return transaction_sampling_mapper.get(
        sampling_context["transaction_context"]["name"], 1.0
    )
```

### Expected Result

-

### Actual Result

Possible solution: Give the plain URL of the request instead of "generic ASGI request" as the transaction name to `traces_sampler`.

### LINKED ISSUE #2263: FastAPI requests handled by middleware are always "generic ASGI request"
I have a problem, that requests in FastAPI that get handled by a middleware, such as 404s or OPTIONS for CORS, are reported as "generic ASGI request" for transaction name. It seems like there should be an fix - all of these requests have enough information to name them at least a bit more specific, such as by the name of the middleware, the URI + method, or similar.

antonpirker:
Set the url as a transaction name instead of 'generic ASGI request' in the beginning, so traces_sampler has something to work with that is more meaningful than 'generic ASGI request'.

Closes #2262 
Closes #2263 

### New Behaviour:

Note: transaction names can be two styles, "url" or "endpoint". (set by the `transaction_style` parameter of the Integrations)

Note 2: See also `@pytest.mark.parametrize` decorator in the new tests as reference. 

- vanilla ASGI: set URL instead of always "generic ASGI request"
- Starlette:
  1. normal request: transaction name is function name or route (depending on `transaction_style` setting)
  2. `traces_sampler`: *always* receives the raw URL as the transaction name (no matter the `transaction_style` setting. because we do not know more at the time the traces_sampler is called.)
  3. requests that end in a middleware (like 404, CORS): the functions name or the raw URL (depending on `transaction_style` setting)
- FastAPI
  1. normal request: transaction name is function name or route (depending on `transaction_style` setting)
  2. `traces_sampler`: *always* receives the raw URL as the transaction name (no matter the `transaction_style` setting. because we do not know more at the time the traces_sampler is called.)
  3. requests that end in a middleware (like 404, CORS): the functions name or the raw URL (depending on `transaction_style` setting)
- There used to be "generic ASGI request" transactions being created at the server startup (when a "lifespan" ASGI message was received.) Those transactions are not created anymore. (we can think of creating propper "Server was started/stopped" transactions in the future)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ venv
 relay
 pip-wheel-metadata
 .mypy_cache
+.vscode/
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,0 @@
-{
-    "python.pythonPath": ".venv/bin/python",
-    "python.formatting.provider": "black",
-    "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
-}
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -258,14 +258,14 @@ class _Client(object):
             SDK_INFO["name"] = sdk_name
             logger.debug("Setting SDK name to '%s'", sdk_name)
 
+            if has_profiling_enabled(self.options):
+                try:
+                    setup_profiler(self.options)
+                except Exception as e:
+                    logger.debug("Can not set up profiler. (%s)", e)
+
         finally:
             _client_init_debug.set(old_debug)
-
-        if has_profiling_enabled(self.options):
-            try:
-                setup_profiler(self.options)
-            except ValueError as e:
-                logger.debug(str(e))
 
         self._setup_instrumentation(self.options.get("functions_to_trace", []))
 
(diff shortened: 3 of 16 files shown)


sentrivana (approved):
LGTM! 🎖️

PR #2346, opened 2023-09-04
Fix Falcon Integration

### LINKED ISSUE #1946: sentry-sdk stopped working with falcon 3+ asgi app
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.16.0

### Steps to Reproduce

I created a test repo where the bug can be reproduced -> https://github.com/detoyz/sentry_falcon_asgi_bug . There are also details on how to setup project & reproduce the error.

However a small summary here:
1. Install sentry-sdk 1.16.0
2. Install falcon
3. Init `falcon.asgi.App()` and try to use it via [falcon-inspect-app](https://falcon.readthedocs.io/en/stable/api/inspect.html)
4. **See error from sentry's falcon patching middleware**
5. Downgrade sentry sdk to 1.15.0
6. Your `falcon.asgi.App()` and `falcon-inspect-app` are working as expected

### Expected Result

By using sentry-sdk anywhere in the code it should not break during runtime because of any `falcon.asgi.App()` isntances

### Actual Result

Runtime errors while using `falcon.asgi.App()` isntances with sentry-sdk client
```
TypeError: _patch_prepare_middleware.<locals>.sentry_patched_prepare_middleware() takes from 0 to 2 positional arguments but 3 were given
```

szokeasaurusrex:
Fixes #1946

### DIFF
--- a/sentry_sdk/integrations/_falcon_async_py3.py
+++ b/sentry_sdk/integrations/_falcon_async_py3.py
@@ -1,0 +1,23 @@
+"""
+This file was created so we could have a process_request_async method on our
+SentryFalconMiddleware class, so we could properly support ASGI with Falcon.
+However, since process_request_async needs to be an async function, we cannot
+define it in Python 2.7, as this produces a syntax error.
+
+So, we have split the process_request_async implementation into this separate
+file, which can only be loaded in Python 3 and above.
+"""
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class SentryFalconMiddlewareBase(ABC):
+    @abstractmethod
+    def process_request(self, req: Any, resp: Any, *args: Any, **kwargs: Any) -> None:
+        pass
+
+    async def process_request_async(
+        self, req: Any, resp: Any, *args: Any, **kwargs: Any
+    ) -> None:
+        self.process_request(req, resp, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


szokeasaurusrex:
TODO: Add unit test, potentially

sentry_sdk/integrations/_falcon_async_py3.py
```diff
@@ -0,0 +1,23 @@
...
+file, which can only be loaded in Python 3 and above.
+"""
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class SentryFalconMiddlewareBase(ABC):
+    @abstractmethod
+    def process_request(self, req: Any, resp: Any, *args: Any, **kwargs: Any) -> None:
```
szokeasaurusrex, line 17: Is it okay that I used Python 3-style type hints here? This file should never get loaded in Python 2

szokeasaurusrex:
Should we be adding unit tests for this change? Or is it simple enough for unit tests to be unnecessary?

sentrivana (approved):
Great job figuring out how to deal with the py2/3 compat issue! 🎖️ 

Approving as the fix itself looks good, but please don't merge yet -- I think we do need testcases.

Even though the change looks small we're actually adding a major feature, which is support for ASGI Falcon, so I think that warrants some testcases. We could create a dummy ASGI Falcon app and basically check that the tests that work for WSGI also work for the async app. I see that there's a `falcon.testing` module -- maybe there's a test asgi app that we could use?

szokeasaurusrex:
Converting back to a draft until test cases are completed

szokeasaurusrex:
This PR has been superseded by #2359

PR #2348, opened 2023-09-04
Add new performance API

### LINKED ISSUE #2345: New Performance API
Considering work done on JS and the connected RFC. We would like to implement a new performance API which should simplify the process for manual instrumentation, and remove the need for understanding underlying SDK concepts like hubs and scopes.

https://github.com/getsentry/sentry-javascript/issues/8724

antonpirker:
Fixes #2345


antonpirker:
It was decided that we put this on hold for now.

PR #2359, opened 2023-09-08
Prevent Falcon integration from breaking ASGI apps

### LINKED ISSUE #1946: sentry-sdk stopped working with falcon 3+ asgi app

szokeasaurusrex:
Fixes #1946 and supersedes #2346

### DIFF
--- a/sentry_sdk/integrations/falcon.py
+++ b/sentry_sdk/integrations/falcon.py
@@ -206,18 +206,12 @@ def _patch_prepare_middleware():
     original_prepare_middleware = falcon_helpers.prepare_middleware
 
     def sentry_patched_prepare_middleware(
-        middleware=None,
-        independent_middleware=False,
-        asgi=False,
-        *args,  # args and kwargs for future compatibility
-        **kwargs
+        middleware=None, independent_middleware=False, asgi=False
     ):
-        # type: (Any, Any, bool, *Any, **Any) -> Any
+        # type: (Any, Any, bool) -> Any
         if asgi:
             # We don't support ASGI Falcon apps, so we don't patch anything here
-            return original_prepare_middleware(
-                middleware, independent_middleware, asgi, *args, **kwargs
-            )
+            return original_prepare_middleware(middleware, independent_middleware, asgi)
 
         hub = Hub.current
         integration = hub.get_integration(FalconIntegration)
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/falcon.py
```diff
@@ -206,13 +206,20 @@ def _patch_prepare_middleware():
...
 
     def sentry_patched_prepare_middleware(
-        middleware=None, independent_middleware=False
+        middleware=None, independent_middleware=False, asgi=False
     ):
-        # type: (Any, Any) -> Any
+        # type: (Any, Any, bool) -> Any
+        if asgi:
+            # We don't support ASGI Falcon apps, so we don't patch anything here
+            return original_prepare_middleware(middleware, independent_middleware, asgi)
```
antonpirker, line 214: Please always add the Sentry middleware to the call to `original_prepare_middleware`, so our middleware runs, no matter if asgi or not. (because it sets the transaction name and this probalby works in wsgi and asgi.)

antonpirker (changes requested):

sentry_sdk/integrations/falcon.py
szokeasaurusrex: We should omit the `SentryFalconMiddleware` here, since by adding it, we would cause an error since its `process_request` method is a synchronous function. We would need to add an async version of `process_request`, which is what I was having difficulty getting to work in the previous PR (#2346)

antonpirker (approved):
lgtm

PR #2394, opened 2023-09-27
feat(integrations): add support for cluster clients from redis sdk

md384:
This change adds support for cluster clients from the redis sdk (as opposed to the rediscluster library).

This has also been tested in my own app which uses clusters (but not asyncio clusters).

Fixes GH-2523


md384:
I have updated this PR with the db data on the span changes. For redis clusters, I have used the default_node but it might be better to not provide that data at all.

From redis docs - https://redis.readthedocs.io/en/stable/clustering.html#creating-clusters

> RedisCluster instance can be directly used to execute Redis commands. When a command is being executed through the cluster instance, the target node(s) will be internally determined. When using a key-based command, the target node will be the node that holds the key’s slot. Cluster management commands and other commands that are not key-based have a parameter called ‘target_nodes’ where you can specify which nodes to execute the command on. In the absence of target_nodes, the command will be executed on the default cluster node. As part of cluster instance initialization, the cluster’s default node is randomly selected from the cluster’s primaries, and will be updated upon reinitialization. Using r.get_default_node(), you can get the cluster’s default node, or you can change it using the ‘set_default_node’ method.

This likely means that it is difficult to get the actual node information and the default node will not always be the one used.

wgiddens:
fixes #2523

szokeasaurusrex (commented):

I took an initial pass at your PR, and I added a few comments and a question. Please address these, and once I hear back, I will review again

szokeasaurusrex:
@md384, to get your PR merged as quickly as possible, we would appreciate if you could [enable edits from repository maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) if you are okay with doing so. This way, we can help with any minor changes that need to be made before merging your PR (such as resolving merge conflicts). Thanks!

tests/integrations/redis/asyncio/test_redis_asyncio.py
```diff
@@ -9,7 +9,11 @@
 
 @pytest.mark.asyncio
 async def test_async_basic(sentry_init, capture_events):
-    sentry_init(integrations=[RedisIntegration()])
+    sentry_init(
+        integrations=[RedisIntegration()],
+        traces_sample_rate=1.0,
+        send_default_pii=True,
```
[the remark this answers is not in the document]
md384: My mistake, this change is not needed. I will revert.

md384:
> @md384, to get your PR merged as quickly as possible, we would appreciate if you could [enable edits from repository maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) if you are okay with doing so. This way, we can help with any minor changes that need to be made before merging your PR (such as resolving merge conflicts). Thanks!

@szokeasaurusrex I have made your suggested changes and enabled access for maintainers. Thanks!

szokeasaurusrex:
Thank you! I will take a look at your PR again when I have time

szokeasaurusrex (commented):
Hey, thank you for addressing my previous suggestions. I have added a few more comments now; once you address them, I think this PR will be ready to merge!

sentry_sdk/integrations/redis/__init__.py
```diff
@@ -135,8 +140,34 @@ def _set_db_data(span, connection_params):
...
+    # type: (Span, Redis) -> None
+    _set_db_data_on_span(span, redis_instance.connection_pool.connection_kwargs)
+
+
+def _set_cluster_db_data(span, redis_cluster_instance):
+    # type: (Span, RedisCluster) -> None
+    default_node = redis_cluster_instance.get_default_node()
+    if default_node:
+        _set_db_data_on_span(
+            span, {"host": default_node.host, "port": default_node.port}
```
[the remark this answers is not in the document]
md384: It does not, there is not much consistency across the redis modules

szokeasaurusrex (commented):
Thank you for addressing the previous comments. I just tried to resolve the merge conflict between your branch and master, so your PR would be ready to merge, but I was unsure how to resolve it (please see my inline comment).

sentry_sdk/integrations/redis/__init__.py
```diff
@@ -189,7 +220,7 @@ def sentry_patched_execute_command(self, name, *args, **kwargs):
             description = description[: integration.max_data_size - len("...")] + "..."
 
         with hub.start_span(op=OP.DB_REDIS, description=description) as span:
-            _set_db_data(span, self.connection_pool.connection_kwargs)
+            set_db_data_fn(span, self)
```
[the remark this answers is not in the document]
md384: I think we will need it for the `_set_db_data` but the other `set_db_data_fn`s will not - they check existence or use attributes that are required to have values. I will  change`_set_db_data`.

szokeasaurusrex:
Hi @md384, now that I have resolved the merge conflict between your branch and master, I was able to run our CI checks on your code. It looks like your code is failing our linter ("CI / Lint Sources") and our Redis tests. The aws_lambda test is also failing for now, but this is expected, since I need to allow it to run separately; once you have fixed all the other tests, I will run the aws_lambda test.

Please fix the "CI / Lint Sources" and Redis tests, and let me know if you need help with anything!

tests/integrations/redis/cluster/test_redis_cluster.py
```diff
@@ -0,0 +1,143 @@
...
+    capture_message("hi")
+
+    (event,) = events
+    crumbs = event["breadcrumbs"]["values"]
+
+    # on initializing a RedisCluster, a COMMAND call is made - this is not important for the test
+    # but must be accounted for
+    assert len(crumbs) <= 2
+    if len(crumbs) == 2:
+        assert event["breadcrumbs"]["values"][0]["message"] == "COMMAND"
```
md384, line 38: As I mention in the comment, initializing the RedisCluster executes a COMMAND call to the redis server to get available commands. This only happens on the first test run of a parameterized run (more of a problem for test_rediscluster_basic), perhaps there is some caching mechanism somewhere

sentry_sdk/integrations/redis/__init__.py
```diff
@@ -155,15 +158,20 @@ def _set_cluster_db_data(span, redis_cluster_instance):
...
 
 def _set_async_cluster_pipeline_db_data(span, async_redis_cluster_pipeline_instance):
-    # type: (Span, AsyncClusterPipeline) -> None
-    _set_async_cluster_db_data(span, async_redis_cluster_pipeline_instance._client)
+    # type: (Span, AsyncClusterPipeline[Any]) -> None
+    _set_async_cluster_db_data(
+        span,
+        # the AsyncClusterPipeline has always had a `_client` attr but it is private so potentially problematic and mypy
+        # does not recognize it - see https://github.com/redis/redis-py/blame/v5.0.0/redis/asyncio/cluster.py#L1386
+        async_redis_cluster_pipeline_instance._client,  # type: ignore[attr-defined]
```
szokeasaurusrex, line 173: Is there a way we could try to avoid using `_client` here, and instead access what we need through a public API? I would strongly prefer to avoid using an external library's private API, since the private API could change at any time and cause problems in the future.

tests/integrations/redis/cluster/test_redis_cluster.py
szokeasaurusrex: I would rewrite the assertions as follows:
```suggestion
    # on initializing a RedisCluster, a COMMAND call is made - this is not important for the test
    # but must be accounted for
    assert len(crumbs) in (1, 2)
    assert len(crumbs) == 1 || crumbs[0]["message"] == "COMMAND"
```

The first assertion should be changed because the test should explicitly verify that at least one breadcrumb was created; the previous assertion would have succeeded with no breadcrumbs and the test would only fail later with an IndexError.

Regarding the second assertion, I think my suggested revision conveys more clearly the intent to verify that either there is only one breadcrumb, or if there are two, then the first breadcrumb should be a command.

If you prefer the structure with the `if` statement for the second assertion, it is okay with me to keep that structure, but in either case, please replace `event["breadcrumbs"]["values"][0]["message"]` with `crumbs[0]["message"]`

tests/integrations/redis/cluster/test_redis_cluster.py
```diff
@@ -29,7 +29,15 @@ def test_rediscluster_breadcrumb(sentry_init, capture_events):
...
-    (crumb,) = event["breadcrumbs"]["values"]
+    crumbs = event["breadcrumbs"]["values"]
+
+    # on initializing a RedisCluster, a COMMAND call is made - this is not important for the test
+    # but must be accounted for
+    assert len(crumbs) <= 2
+    if len(crumbs) == 2:
+        assert event["breadcrumbs"]["values"][0]["message"] == "COMMAND"
+
+    crumb = event["breadcrumbs"]["values"][-1]
```
szokeasaurusrex, line 40:
```suggestion
    crumb = crumbs[-1]
```

tests/integrations/redis/cluster/test_redis_cluster.py
```diff
@@ -65,7 +73,15 @@ def test_rediscluster_basic(sentry_init, capture_events, send_default_pii, descr
...
 
     (event,) = events
-    (span,) = event["spans"]
+    spans = event["spans"]
+
+    # on initializing a RedisCluster, a COMMAND call is made - this is not important for the test
+    # but must be accounted for
+    assert len(spans) <= 2
+    if len(spans) == 2:
+        assert event["spans"][0]["description"] == "COMMAND"
```
szokeasaurusrex, line 82: My previous comment applies here, as well

szokeasaurusrex (commented):
I have a few comments on your most recent changes, but I think this is getting really close

sentry_sdk/integrations/redis/__init__.py
szokeasaurusrex: Okay, in that case, we can use `capture_internal_exceptions` here. I also checked that one of your unit tests is verifying this function, so we will get a test failure if `_client` ever gets removed

szokeasaurusrex (approved):
Looks good! Thanks for the contribution!

Issue #1064, opened 2021-03-24
Add util to send user feedback


ThiefMaster:
Applications which don't already use the client-side sentry SDK may want to display their own UI to send user feedback to associate with an error that was logged to sentry.

It would be nice if the [user feedback api](https://docs.sentry.io/api/projects/submit-user-feedback/) was exposed in a clean way. For example, it should automatically infer the correct host/path from the DSN.

Not sure if this doesn't actually require changes on Sentry itself... right now I don't see how one would get the data needed for `organization_slug` and `project_slug` using just the DSN...

vagi8:
Hey I am planning to take it up. I have taken a look at the Dsn class in utils but I was unable to find a method for feedback api url. Nor I could find any organization / project slug values in the Dsn nor in Auth.  What am I missing ?

antonpirker:
Hey @vagi8 

You are right. Right now the SDK only knows the DSN but not the `organization_slug` and `project_slug`.  I looked at how the browser JS SDK does it [1] and it seems it creates an envelope.

So I guess this is not a minor change as we thought. We will need to discuss this internally if we want to have this feature in the Python SDK.

1: https://github.com/getsentry/sentry-javascript/blob/f63036ba6d896f624328c857d2170f8a6843f5e1/packages/browser/src/client.ts#L97-L109

antonpirker:
We discussed this internally and because this is not only about giving the user the Feddback api URL, but implementing a feature that puts user feedback in an envelope and send it to Sentry, we will not implement this right now or in the short to mid term future. We have more pressing things on our agenda. So I will close this issue.

theCapypara:
@antonpirker Would this still be OK to contribute as a PR, or is this a feature you currently don't want implemented at all for some reason?

I'm not sure what you mean by "not only about giving the user the Feddback api URL, but implementing a feature that puts user feedback in an envelope and send it to Sentry". Isn't this also what this issue was originally about? Exposing the user feedback API via the SDK so developers can show their own forms to the users and then the SDK will send it to Sentry via the `user-feedback` api?

I would be interested in contributing such a feature, but I'm not sure what you mean about the envelope, the only thing I can find is the official API documentation (https://docs.sentry.io/api/projects/submit-user-feedback/) which I thought would be the one to use?

theCapypara:
Nevermind, found info on the topic: https://develop.sentry.dev/sdk/envelopes/

[...]

PR #2442, opened 2023-10-13
feat(api): Support capturing user feedback

theCapypara:
This adds an API to the Sentry Python SDK that captures user feedback via envelope.

This is implemented very similiarly to how it is done for the JavaScript SDK, see https://github.com/getsentry/sentry-javascript/pull/7729.

Fixes GH-1064

Also see my comments at https://github.com/getsentry/sentry-python/issues/1064#issuecomment-1758485928

---

The usage is pretty much like it's [explained](https://docs.sentry.io/platforms/javascript/enriching-events/user-feedback/#user-feedback-api) for the JavaScript SDK.

Example usage:
```py
import sentry_sdk

test_event = sentry_sdk.capture_message("test_message")
assert test_event is not None

sentry_sdk.capture_user_feedback(
    {
        "event_id": test_event,  # or sentry_sdk.last_event_id()
        "name": "Tester",
        "email": "test@example.com",
        "comments": "Test comments",
    }
)

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -12,6 +12,7 @@ Capturing Data
 .. autofunction:: sentry_sdk.api.capture_event
 .. autofunction:: sentry_sdk.api.capture_exception
 .. autofunction:: sentry_sdk.api.capture_message
+.. autofunction:: sentry_sdk.api.capture_user_feedback
 
 
 Enriching Events
--- a/sentry_sdk/__init__.py
+++ b/sentry_sdk/__init__.py
@@ -22,6 +22,7 @@ __all__ = [  # noqa
     "capture_event",
     "capture_message",
     "capture_exception",
+    "capture_user_feedback",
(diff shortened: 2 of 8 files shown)


cleptric:
We are currently working on an entirely new User Feedback product, which is tied together with Session Replays. The feature is currently in a closed Alpha. There is a bit more info about it in this discussion https://github.com/getsentry/sentry/discussions/54956.

I would hold off on supporting the old version, as it will be obsolete soon.

theCapypara:
The new user feedback only supports browser based apps and the "old" way is still recommended for non-browser based apps:

https://docs.sentry.io/product/user-feedback/setup/#supported-sdks-for-crash-report-api

Does this mean this could be merged?

theCapypara:
I withdraw this contribution, since there doesn't seem to be any interest.

PR #2444, opened 2023-10-14
feat: Add ray integration support (#2400)

### LINKED ISSUE #2400: Please add Ray framework integration for sentry/apm.
### Problem Statement

https://www.ray.io/

I want to annotate that kind of code with spans:

```import logging
import ray
import os
import time
import sentry_sdk
import random
import asyncio

logging.basicConfig(level=logging.DEBUG)

def step1():
    with sentry_sdk.start_span(description="step1"):
        time.sleep(random.randint(200,500)/1000.)
def step2():
    with sentry_sdk.start_span(description="step2"):
        time.sleep(random.randint(200,500)/1000.)
def step3():
    with sentry_sdk.start_span(description="step3"):
        time.sleep(random.randint(200,500)/1000.)

@ray.remote
def retrieve_task(item):
    step1()
    step2()
    step3()
    return 1

async def main():
    sentry_sdk.init(
        dsn=os.getenv('SENTRY_DSN'),
        traces_sample_rate=1.0,
        integrations=[
            AsyncioIntegration(),
        ],
    )
    #---
    ray.init()
    with sentry_sdk.start_transaction(op="task", name="example ray task transaction"):
        object_references = [
            retrieve_task.remote(item) for item in range(8)
        ]
        data = ray.get(object_references)
        print(data)

if __name__ ==  '__main__':
    asyncio.run(main())
```
    
But I only see one big transaction trace without those 3 spans.

### Solution Brainstorm

I think new integration needs to be done.

glowskir:
Adds a basic instrumentation for the Ray framework (https://www.ray.io/)

Closes #2400 

Documentation PR: https://github.com/getsentry/sentry-docs/pull/11029


leokster:
Is it planned to merge this PR?

szokeasaurusrex:
@leokster Taking a look now – likely this PR will require some changes before we can merge

glowskir:
@szokeasaurusrex you can do the review, I will adjust accordingly

smeubank:
Might be interesting to review the PR for Langchain https://github.com/getsentry/sentry-python/pull/2911

I am not familiar with Ray, but if it is a framework similar to Langchain, then following a similar model should be ideal

cc @colin-sentry if you have any thoughts on this integration

szokeasaurusrex:
Hey @glowskir, thanks for reaching out again and for contributing this PR.

We are currently working on an initiative to [support AI tools](https://github.com/getsentry/sentry-python/discussions/3007) in the Sentry SDK, and we would be interested in learning how you are using Ray and what you hope to be able to achieve by integrating Ray with Sentry.

If you are interested, we can set up a customer call with you and anyone else interested in seeing a Ray integration in Sentry, so that we can learn how you are using Ray and how a Ray integration in the SDK could help you. You can reach me at daniel.szoke@sentry.io – please let me know when you would have time.

glowskir:
@szokeasaurusrex This PR introduces RayIntegration to enable out of the box distributed tracing support. Ray has high level task interface and we want to see chain of tasks as single transaction in performance tab.  At the moment of creating this PR there was no Opentelemetry support, now in newer versions of ray one can use Opentelemetry facitlities provided by Ray. I don't know how Sentry integrations relate to Opentelemetry hooks tbh, whether you are planning to adapt them or ignore them but this is a thing to consider.

szokeasaurusrex:
@glowskir, understood. In that case, we can help to get this PR merged. 

Before I do an in depth review, I wanted to mention a few things that might be helpful here. First of all, since it looks like Ray is mostly used with AI, it might make sense to design this integration so it is compatible with our [LLM Monitoring](https://docs.sentry.io/product/llm-monitoring/) product. If you need some inspiration for how to do this, you can look at our existing Langchain integration, which works with the LLM Monitoring feature.

Additionally, I noticed while looking at your PR that you are using `sentry_sdk.Hub` in several places. `sentry_sdk.Hub` is deprecated as of 2.0.0, and although some of our existing code uses it, we are going to migrate all of these uses away from the `Hub` API and would like to avoid introducing any new `Hub` usages. Please see [here](https://docs.sentry.io/platforms/python/migration/1.x-to-2.x#custom-instrumentation) for a guide of how to migrate `Hub` calls to the new `Scope` API.

Please let me know whether you have any questions! I am happy to help

antonpirker:
Hi @glowskir , i tried to use your integration with a sample ray app (I have never used ray before)

Can you tell me how I can initialize sentry in the ray worker that is started when I call `ray.init()`? Is there an event/signal I can listen to?

antonpirker:
Ok, got it to work. Did some more cleanup and the transactions/spans look nice:

Screenshot 2024-08-02 at 08 19 48

antonpirker:
The integration when one defines a task as a function. If a task is defined as a class (called an Actor) then the integrations breaks the ray application. 

I think we can not release it in this state. 

@glowskir do you have time to look into why Actors do not work with the integration? 
(When doing an `counter = Counter.remote()` from the example in the docs then `counter` is of type `ray._raylet.ObjectRef` instead of the correct `ray._raylet.ActorHandle`)

antonpirker:
This is the `ray` app I used for testing: https://github.com/antonpirker/testing-sentry/tree/main/test-ray

antonpirker:
I fixed the crash when using Actors. Performance data is still not captured when using Actors but at least we are not crashing users code anymore. I also mention this in the docs, so everyone knows that Actors are not yet supported.

sentrivana (approved):
See one comment, else LGTM

ZacharyHampton:
Thanks for the work on this ya'll, excited to use this when merged.

PR #2453, opened 2023-10-18
Update compatible runtimes for AWS Lambda layer

antonpirker:
Set the compatible runtimes in the Sentry AWS Lambda Layer to also include Python 3.10 and 3.11.

Related PRs in `sentry`:
* https://github.com/getsentry/sentry/pull/58321
* https://github.com/getsentry/sentry/pull/58320

Related docs update:
* https://github.com/getsentry/sentry-docs/pull/8246

Merge this, before merging the PRs in `sentry`.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -22,6 +22,8 @@ targets:
           - python3.7
           - python3.8
           - python3.9
+          - python3.10
+          - python3.11
     license: MIT
   - name: sentry-pypi
     internalPypiRepo: getsentry/pypi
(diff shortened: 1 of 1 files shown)


sentrivana (approved):

antonpirker:
Had to update the platforms again, because it turns out, that we do not support below 3.9 with the new urllib version we have in our layers. (I will document this so we know exactly what lambda layer versions for which python versions)

PR #2454, opened 2023-10-18
ref: Patched functions decorator for integrations

szokeasaurusrex:
This PR introduces two new decorators in `sentry_sdk.utils` that we can use in our integrations to automate the checks for whether the integration is still enabled. Since these decorators use the new scopes API, adopting these decorators may simplify the change to the new Scopes API.

For example, we could replace the below sample, which uses the old Hub API:
```python
async def _sentry_app(*args, **kwargs):
     # type: (*Any, **Any) -> Any
     hub = Hub.current
     integration = hub.get_integration(FastApiIntegration)
     if integration is None:
         return await old_app(*args, **kwargs)

    ... # rest of function body
```

with the code below, using these decorators:
```python
@ensure_integration_enabled_async(FastApiIntegration, old_app)
async def _sentry_app(*args, **kwargs):
    # type: (*Any, **Any) -> Any
    ... # rest of function body
```

The above example is taken from https://github.com/getsentry/sentry-python/pull/2836; the full code is available at that PR.


antonpirker:
I really like the idea, this can make our integration's code way cleaner and easier to read!

sentry_sdk/utils.py
```diff
@@ -1563,6 +1566,26 @@ def parse_version(version):
     return release_tuple
 
 
+def integration_patched(original_function, integration=None):
```
antonpirker, line 1569: I would called it `ensure_integration_enabled(integration=None, original_function=None)` to make it easier to understand what is happening. (And if `original_function` is None, just fail with an error)

sentry_sdk/utils_py3.py
```diff
@@ -0,0 +1,31 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+    from typing import Optional
+    from sentry_sdk.integrations import Integration
+
+    from sentry_sdk._types import GenericAsyncCallable
+
+
+def integration_patched_async(original_function, integration=None):
```
antonpirker, line 14: I would called it `ensure_integration_enabled(integration=None, original_function=None)` to make it easier to understand what is happening. (And if `original_function` is None, just fail with an error)

sentry_sdk/utils.py
```diff
@@ -1563,6 +1566,26 @@ def parse_version(version):
...
+    # type: (GenericCallable, Optional[type[Integration]]) -> Callable[[GenericCallable], GenericCallable]
+    def patcher(sentry_patched_function):
+        # type: (GenericCallable) -> GenericCallable
+        @wraps(original_function)
+        def runner(*args, **kwargs):
+            # type: (*object, **object) -> object
+            if (
+                integration is not None
+                and sentry_sdk.hub.Hub.current.get_integration(integration) is None
```
antonpirker, line 1578: There should be the same condition in the ifs of both versions of the decorator

antonpirker (commented):
Looks already very good. Left some comments.

antonpirker:
Some tests are still failing. I guess the GQL you can fix easily. For the Starlette tests we need to check if the behavior changed now because of the `wraps` (I guess) if so, we would need to wait for a major release

szokeasaurusrex:
> Some tests are still failing. I guess the GQL you can fix easily. For the Starlette tests we need to check if the behavior changed now because of the `wraps` (I guess) if so, we would need to wait for a major release

If adding `wraps` is a breaking change, we could also remove `wraps` from this PR, and add it back in when we have a new major.

szokeasaurusrex:
@antonpirker The GQL tests that are failing are themselves appear to be broken; I don't think that the fact that they are failing indicates that this is a breaking change.

The reason the tests fail is because they are doing some strange monkey patching to verify that the method we monkeypatch is still being called by the integration. I don't think we do tests like this anywhere else; looks like I added them six months ago when I was less familiar with how to best test integrations.

I think we should just delete the failing tests since they provide little value, but clearly are failing even when they should not. I will open a new PR to make this change.

sentry_sdk/utils.py
```diff
@@ -1622,6 +1631,74 @@ def reraise(tp, value, tb=None):
...
+    def patch_my_function():
+        with sentry_sdk.start_transaction(...):
+            return my_function()
+    ```
+    """
+
+    def patcher(sentry_patched_function):
+        # type: (Callable[P, R]) -> Callable[P, R]
+        @wraps(original_function)
+        def runner(*args: "P.args", **kwargs: "P.kwargs"):
```
szokeasaurusrex, line 1664: `P.args` and `P.kwargs` can only be used as Python 3-style type hints; they are apparently incompatible with the type comment syntax we usually use.

sentry_sdk/utils.py
```diff
@@ -1622,6 +1631,74 @@ def reraise(tp, value, tb=None):
...
+):
+    # type: (...) -> Callable[[Callable[P, R]], Callable[P, R]]
+    """
+    Ensures a given integration is enabled prior to calling a Sentry-patched function.
+
+    The function takes as its parameters the integration that must be enabled and the original
+    function that the SDK is patching. The function returns a function that takes the
+    decorated (Sentry-patched) function as its parameter, and returns a function that, when
+    called, checks whether the given integration is enabled. If the integration is enabled, the
+    funciton calls the decorated, Sentry-patched funciton. If the integration is not enabled,
```
antonpirker, line 1646:
```suggestion
    function calls the decorated, Sentry-patched function. If the integration is not enabled,
```

Issue #1036, opened 2021-03-01
Sentry with eventlet


Randomneo:
### Environment

self-hosted sentry 9.0.0

python 3.9/3.7
sentry_sdk version: 0.20.3
eventlet version: 0.30.1
greenlet version: 1.0.0

### Steps to Reproduce

    import eventlet
    import sentry_sdk
    
    eventlet.monkey_patch()
    
    sentry_sdk.init('<sentry_sdk_dsn>')
    sentry_sdk.capture_message('aoe')

### Expected Result

getting message in sentry

### Actual Result

Nothing happens. No errors thrown. No clear documentation for that problem. Here [issue](https://github.com/eventlet/eventlet/issues/689) for eventlet.

If sentry_sdk can't be used with eventlet and there no possible way to combine them it should be in documentation.
Please correct me if i'm missing something.

untitaker:
Can you add `sentry_sdk.flush()` as last line?

Randomneo:
Same. Nothing happens. Without flush I got.

    Sentry is attempting to send 0 pending error messages
    Waiting up to 2 seconds
    Press Ctrl-C to quit

Program ends by itself instantly.

With flush. No output and no message in sentry

Randomneo:
Problem definitely in eventlet.monkey_patch() if that line is commented everything works fine.

temoto:
Just for record, copying workaround from our tracker.

Calling monkey_patch before even importing sentry_sdk works.

```
import eventlet
eventlet.monkey_patch()

import sentry_sdk
...
```

I understand it's well below acceptable and yet don't have a better solution right now.

Successful fix would start with analyzing what side effects are produced by `import sentry_sdk`

Randomneo:
@temoto Thanks alot. I had no idea that import sentry_sdk was the problem :)

untitaker:
only work on import is this: https://github.com/getsentry/sentry-python/blob/37105d981fb116c60df2ea3d1e58a87b9c65fc21/sentry_sdk/utils.py#L840

However, flush etc should not really break in either case. So I think it's more related to threading patches that would affect `sentry_sdk.worker`.

untitaker:
@Randomneo can you check `Hub.current.client` after init? Maybe locals *are* broken in a really bad way.

Randomneo:
@untitaker I not quite get how I supposed to check it. But this code outputs:

    import eventlet
    import sentry_sdk
    from sentry_sdk import Hub
    
[...]

PR #2464, opened 2023-10-25
Patch eventlet under Sentry SDK

szokeasaurusrex:
Pull request containing the same changes as #1258, created since #1258 does not allow edits from maintainers and is currently out of date with master, and therefore cannot be merged even though the PR is approved.

From @guiscaranse (original author on #1258):

> This patches the eventlet library to allow it to run without issues under the Sentry SDK when using greenlet>=0.5. Eventlet doesn't pin the greenlet dependency and https://github.com/eventlet/eventlet/issues/663 (that's why the sdk runs nicely under Gevent for example). So, the eventlet library itself is not a good marker to whatever the contextvars are broken or not.
>
> This should fix https://github.com/getsentry/sentry-python/issues/1036


sentry_sdk/utils.py
```diff
@@ -1159,9 +1159,13 @@ def _is_contextvars_broken():
         pass
 
     try:
+        import greenlet  # type: ignore
         from eventlet.patcher import is_monkey_patched  # type: ignore
 
-        if is_monkey_patched("thread"):
+        version_tuple = tuple(
+            [int(part) for part in greenlet.__version__.split(".")[:2]]
+        )
```
antonpirker, line 1167: We should use `parse_version` for parsing the version. This function: https://github.com/getsentry/sentry-python/blob/antonpirker/fix-aws-secrects/sentry_sdk/utils.py#L1514

antonpirker (approved):

PR #2474, opened 2023-10-31
Make parse_version work in utils.py itself.

antonpirker:


szokeasaurusrex (approved):
Looks good. Might be worth adding a test to ensure the same issue doesn't happen again, but we can merge without a test and add later if you think it's needed

PR #2497, opened 2023-11-09
Read timezone for Crons monitors from `celery_schedule` if existing

### LINKED ISSUE #2472: Crons django-celery-beat timezone glitch
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.29.2

### Steps to Reproduce

1. Setup django-celery-beat
2. Create a Periodic task via the admin and set a wacked timezone for it
3. setup Sentry with CeleryIntegration(monitor_beat_tasks=True),
4. See the resulting cron monitor in sentry

The current implementation of the `_get_monitor_config` helper does not consider for the posibility that the schedule might be coming from django_celery_beat (this is understandable since this is all new).
Here is the implementation for it
https://django-celery-beat.readthedocs.io/en/latest/_modules/django_celery_beat/tzcrontab.html#TzAwareCrontab
which has a timezone property. The resulting schedule could have a different timezone from app.conf.timezone and could be per-schedule.

We traced it to here https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/celery.py#L447 and think that considering that schedule could be an instance of TzAwareCrontab would be the way to go.

Not sure how you would like to code the possible import failure of django_celery_beat and so on, so stopped there.

### Expected Result

Expectation is that the cron monitor in sentry has the Americas/Chicago timezone we have setup in django admin for our task

### Actual Result

The task cron monitor in sentry has Americas/Los_Angeles as timezone as that is the app.conf.timezone

antonpirker:
Fixes #2472

### DIFF
--- a/sentry_sdk/integrations/celery.py
+++ b/sentry_sdk/integrations/celery.py
@@ -445,9 +445,13 @@ def _get_monitor_config(celery_schedule, app, monitor_name):
         monitor_config["schedule"]["unit"] = schedule_unit
 
     monitor_config["timezone"] = (
-        str(celery_schedule.tz)
-        if hasattr(celery_schedule, "tz")
-        else app.conf.timezone or "UTC"
+        (
+            hasattr(celery_schedule, "tz")
+            and celery_schedule.tz is not None
+            and str(celery_schedule.tz)
+        )
+        or app.conf.timezone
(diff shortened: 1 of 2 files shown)


sentrivana (approved):
🚀

PR #2510, opened 2023-11-15
Do not create a span when task is triggered by Celery Beat

### LINKED ISSUE #1096: parent_span_id being attached to root celerybeat transactions
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
sentry-python==1.0.0

### Steps to Reproduce

https://gist.github.com/wmak/fffab40bdbe5329f2ef8bd0a4ff44659
1. Repro project attached, a very basic flask app with an endpoint that starts a celery task + a celerybeat that starts the same task
2. Setup the repro project
3. Run redis on 6379
4. Start the celery worker `celery -A main.celery worker`
5. Start celery beat `celery -A main.celery beat --loglevel=INFO --pidfile=''`
6. Wait for the task to be scheduled

### Expected Result

I don't expect for celery tasks started by celerybeat to have the parent_span_id set to a span_id, especially when there isn't a corresponding transaction with that span to make the association, the repro attached here is flask + celerybeat. But also happens with Django + celerybeat

### Actual Result

The parent_span_id was set in the trace context, so the transaction appears to be orphaned, when it in fact should be a root event instead
image

antonpirker:
We create a span for submitting a Celery task for execution (when `apply_async()` is called). In cases where web frameworks are calling `apply_async()` this is fine, because the web framework created a transaction where the span is attached. 

When Celery Beat wakes up and is calling `apply_async()` this is not good, because there is no transaction and then the span ID of the newly created span will be given to the task as `parent_span_id` leading to orphaned transactions.

So in case `apply_async()` is called by Celery Beat, we do not create a span for submitting the task for execution.

Fixes #1096

### DIFF
--- a/scripts/aws-cleanup.sh
+++ b/scripts/aws-cleanup.sh
@@ -3,7 +3,7 @@
 # Helper script to clean up AWS Lambda functions created
 # by the test suite (tests/integrations/aws_lambda/test_aws.py).
 #
-# This will delete all Lambda functions named `test_*`.
+# This will delete all Lambda functions named `test_function_*`.
 #
 
 export AWS_DEFAULT_REGION="us-east-1"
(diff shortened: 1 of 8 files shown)


sentrivana (approved):
lgtm

PR #2511, opened 2023-11-16
Fixed local var not present when error in users error_sampler function

antonpirker:
When a user has an error in her `error_sampler` function (like having the wrong number of function parameters specified) the SDK raised an error in line 468 because the `sample_rate` var was never assigned something.


sentry_sdk/client.py
```diff
@@ -459,11 +459,11 @@ def _should_sample_error(
         # type: (...) -> bool
         sampler = self.options.get("error_sampler", None)
 
+        sample_rate = self.options["sample_rate"]
```
szokeasaurusrex, line 462: I would prefer to find a slightly different way to solve this bug, since with this change, we will be handling the case where the `error_sampler` returns an invalid value differently from the case where it raises an error, even though it would make the most sense to handle both cases identically.

Specifically, with this change, returning an invalid value from the sampler triggers the `except TypeError` block on line 470, where we log a warning message informing the user of the invalid return value, and then we default to always sampling the event. However, if the sampler function raises an error, we fall back to the `sample_rate` option since the variable assignment from line 462 is never overwritten. If the user has set a custom `sample_rate`, the event might not end up being sampled.

I think we should either (1) default to always sampling when the `error_sampler` has an invalid value (i.e., raises an error, returns invalid value, or is not callable) or (2) fall back to the `sample_rate` and then default to always sampling only if the `sample_rate` is also invalid.

I would slightly prefer option (1), since it also matches the [documented behavior](https://docs.sentry.io/platforms/python/configuration/options/#error-sampler) that the `sample_rate` is ignored when an `error_sampler` is provided. If we instead pick option (2), we would also need to tweak the docs to state that we ignore the `sample_rate` when a _valid_ `error_sampler` is provided.

To implement option (1), I think we could first undo this change completely, and then add an `except NameError` block to handle the case where the `sample_rate` does not get set because the sampler raised an error. Lastly, to ensure a consistent handling when the user has set the `error_sampler` to something that is not a function, we should change the `if callable(sampler)` line to `if sampler is not None`; then, we will end up handling the case of of the user supplying a non-callable `error_sampler` in the same `except NameError` block that also handles an `error_sampler` that raises an error.

szokeasaurusrex (changes requested):
Please see inline comment

sentry_sdk/client.py
antonpirker: makes sense. will change this to option (1). Great review @szokeasaurusrex

antonpirker: @szokeasaurusrex updated my code. have you had something like this is mind? Please re-review.

sentry_sdk/client.py
```diff
@@ -466,20 +466,28 @@ def _should_sample_error(
         hint,  # type: Hint
     ):
         # type: (...) -> bool
-        sampler = self.options.get("error_sampler", None)
+        error_sampler = self.options.get("error_sampler", None)
 
-        if callable(sampler):
+        if callable(error_sampler):
```
szokeasaurusrex, line 471:
```suggestion
        if error_sampler is not None:
```
I think we should check that the `error_sampler` is not None here rather than check whether it is callable. This way, if the user supplies an `error_sampler` that cannot be called, we handle it in the same way as if the `error_sampler` raises an error when called

szokeasaurusrex (commented):
Looks about right, just one more suggestion

sentry_sdk/client.py
antonpirker: This would change existing behavior. Up to now if someone submitted a string as `error_sampler` it took the `sample_rate` from the options. With this change we would not take the `sample_rate` from options but default to always sampling the event. We should not change existing behavior.

szokeasaurusrex: Good point -- this is probably an edge case anyways, so I am fine with this change. Will approve the PR

szokeasaurusrex (approved):

PR #2524, opened 2023-11-22
feat: Send to Spotlight sidecar

HazAT:
Add Spotlight option to SDK:

```python
sentry_sdk.init(
        spotlight=True,
    )
    
# OR

sentry_sdk.init(
        spotlight='https://mycustom.ur',
    )
```

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -587,14 +587,17 @@ class _Client(object):
         tracing_enabled = has_tracing_enabled(self.options)
         attachments = hint.get("attachments")
 
-
         trace_context = event_opt.get("contexts", {}).get("trace") or {}
         dynamic_sampling_context = trace_context.pop("dynamic_sampling_context", {})
 
         # If tracing is enabled all events should go to /envelope endpoint.
         # If no tracing is enabled only transactions, events with attachments, and checkins should go to the /envelope endpoint.
         should_use_envelope_endpoint = (
-            tracing_enabled or is_transaction or is_checkin or bool(attachments) or bool(self.spotlight)
+            tracing_enabled
+            or is_transaction
+            or is_checkin
+            or bool(attachments)
+            or bool(self.spotlight)
         )
         if should_use_envelope_endpoint:
             headers = {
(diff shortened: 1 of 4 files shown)


sentry_sdk/spotlight.py
```diff
@@ -0,0 +1,51 @@
...
+            )
+            req.close()
+        except Exception as e:
+            logger.exception(str(e))
+
+
+def setup_spotlight(options):
+    # type: (Dict[str, Any]) -> Optional[SpotlightSidecar]
+
+    url = options["spotlight"]
```
antonpirker, line 42:
```suggestion
    url = options.get("spotlight")
```

sentry_sdk/client.py
```diff
@@ -268,6 +269,10 @@ def _capture_envelope(envelope):
                 ],
             )
 
+            self.spotlight = None
+            if self.options["spotlight"]:
```
antonpirker, line 273:
```suggestion
            if self.options.get("spotlight"):
```

sentry_sdk/spotlight.py
```diff
@@ -0,0 +1,51 @@
...
+if TYPE_CHECKING:
+    from typing import Any
+    from typing import Dict
+    from typing import Optional
+
+from sentry_sdk.utils import logger
+from sentry_sdk.envelope import Envelope
+
+
+class SpotlightSidecar(object):
```
antonpirker, line 15: That's more the `SpotlightClient` rather then the `SpotlightSidecar` right?

antonpirker (approved):
Looks good. Just some minor nitpicks

sentry_sdk/client.py
```diff
@@ -616,9 +623,18 @@ def capture_event(
             for attachment in attachments or ():
                 envelope.add_item(attachment.to_envelope_item())
 
+            if self.spotlight:
+                self.spotlight.capture_envelope(envelope)
+
+            if self.transport is None:
+                return None
```
HazAT, line 630: This moved here otherwise we would've returned too early and the envelope wouldn't even be assembled. Yes it's less efficient but anyone using the SDK shouldn't notice anything.

PR #2529, opened 2023-11-23
Add source to custom 404 transactions in django

### LINKED ISSUE #2520: Django custom 404 transaction do not have a transaction source
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.35.0

### Steps to Reproduce

If you look at the code here we are setting the transaction, but not the transaction source: https://github.com/getsentry/sentry-python/blob/63e2f774500c08033b0991f8c913609608849823/sentry_sdk/integrations/django/__init__.py#L396-L401

Add the transaction with something like this: 
https://github.com/getsentry/sentry-python/pull/2519/files

antonpirker:
Fixes #2520


sentrivana (approved):
LGTM!

What I don't understand though is how this was not caught by tests. We have a [test case for custom 404 handlers](https://github.com/getsentry/sentry-python/blob/ddf37a335c16e0b8e07c5904cc49011aea7264dd/tests/integrations/django/test_basic.py#L703), but it doesn't seem to be testing the right thing?

PR #2535, opened 2023-11-27
fix(metrics): Normalize dashes to underscores in metrics

jan-auer:
Dashes are not supported in metric names on Sentry DDM. This updates
validation and replaces dashes with underscores. Tag keys in sentry
support dashes, too.


jan-auer:
Marking this as draft. We are going to normalize this in Relay, so we can continue sending this too and later on allow dashes even.

jan-auer:
Closing in favor of full normalization in https://github.com/getsentry/relay/pull/2769

PR #2539, opened 2023-11-28
feat(metrics): Add source context to code locations

jan-auer:
The product is going to show source context with code locations on metrics. This
PR adds source context to the code location submission payload in the same
format that stack frames use.

Relay supports this with https://github.com/getsentry/relay/pull/2781.

### DIFF
--- a/sentry_sdk/metrics.py
+++ b/sentry_sdk/metrics.py
@@ -76,7 +76,7 @@ def get_code_location(stacklevel):
         return None
 
     return serialize_frame(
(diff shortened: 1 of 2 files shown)


antonpirker (approved):
Looks ok. 

Just one thing: this will add considerable overhead because it loads the source.
So, your call!

jan-auer:
Thanks, great callout. In that case, we'll collect one more utilization metric first and then merge this.

PR #2557, opened 2023-12-01
Hash AWS Lambda test functions based on current revision

sentrivana:


HansAarneLiblik:
@antonpirker I'm guessing regarding this, we now get these errors in Sentry UI, if we're deploying our staging code for testing / demo purposes (At least - Have not tested with real production code with a bit different conf). These deploy's send errors to sentry, but do not have 

image

sentrivana:
Hey @HansAarneLiblik, will fix that.

sentrivana:
@HansAarneLiblik SDK version 1.39.1 that should fix this is now available -- thanks for reporting!

PR #2564, opened 2023-12-04
(4) Move transaction/span related functions from Hub to Scope

antonpirker:

- moved `start_transaction` from Hub to Scope
- moved `start_span` from Hub to Scope
- moved `continue_trace` from Hub to Scope

Depends on: https://github.com/getsentry/sentry-python/pull/2558 (please review the linked one first)

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -6,7 +6,7 @@ import uuid
 
 from sentry_sdk.attachments import Attachment
 from sentry_sdk._compat import datetime_utcnow
-from sentry_sdk.consts import FALSE_VALUES
+from sentry_sdk.consts import FALSE_VALUES, INSTRUMENTER
 from sentry_sdk._functools import wraps
 from sentry_sdk.session import Session
 from sentry_sdk.tracing_utils import (
@@ -18,6 +18,7 @@ from sentry_sdk.tracing_utils import (
 from sentry_sdk.tracing import (
     BAGGAGE_HEADER_NAME,
     SENTRY_TRACE_HEADER_NAME,
+    NoOpSpan,
     Transaction,
 )
 from sentry_sdk._types import TYPE_CHECKING
@@ -34,6 +35,7 @@ if TYPE_CHECKING:
     from typing import Optional
     from typing import Tuple
     from typing import TypeVar
+    from typing import Union
 
     from sentry_sdk._types import (
         Breadcrumb,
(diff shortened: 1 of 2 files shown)


sentrivana (approved):

sentry_sdk/scope.py
```diff
@@ -636,6 +637,155 @@ def add_breadcrumb(self, crumb=None, hint=None, **kwargs):
...
+        configuration_instrumenter = client and client.options["instrumenter"]
+
+        if instrumenter != configuration_instrumenter:
+            return NoOpSpan()
+
+        custom_sampling_context = kwargs.pop("custom_sampling_context", {})
+
+        # if we haven't been given a transaction, make one
+        if transaction is None:
+            kwargs.setdefault("hub", hub)
```
sl0thentr0py, line 680: so this shows that these Transaction functions have a dependency on hub. We will need to think about how to get rid of that.

The only non-trivial usage is in the context manager 
https://github.com/getsentry/sentry-python/blob/ddf37a335c16e0b8e07c5904cc49011aea7264dd/sentry_sdk/tracing.py#L197-L216

sentry_sdk/scope.py
```diff
@@ -636,6 +637,155 @@ def add_breadcrumb(self, crumb=None, hint=None, **kwargs):
...
+
+        if instrumenter != configuration_instrumenter:
+            return NoOpSpan()
+
+        # THIS BLOCK IS DEPRECATED
+        # TODO: consider removing this in a future release.
+        # This is for backwards compatibility with releases before
+        # start_transaction existed, to allow for a smoother transition.
+        if isinstance(span, Transaction) or "transaction" in kwargs:
+            deprecation_msg = (
```
sl0thentr0py, line 734: let us make sure to remove these in the major. In general, we should massively simplify this span behaviour - we don't need to have 5 different ways of using this function. One well documented way is sufficient.

sl0thentr0py (approved):
lgtm, just some observations

sentry_sdk/scope.py
antonpirker: True. Will change that in a separate pr.

sentry_sdk/scope.py
antonpirker: Yes, it is part of this PR: https://github.com/getsentry/sentry-python/pull/2581

PR #2570, opened 2023-12-07
(5) Add `reset()` to thread local ContextVar and no-op `copy_context()`

antonpirker:
Improves the capabilities of our threadlocal context variables, we use when no "real" context variables are available (for example in old Python versions)

- Adds a no-op [copy_context](https://docs.python.org/3/library/contextvars.html#contextvars.copy_context) function to use in environments without context vars
- Adds [reset](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar.reset) functionality to our threadlocal based context vars.


sentry_sdk/utils.py
```diff
@@ -1275,28 +1298,28 @@ def _get_contextvars():
...
+                return True, ContextVar, copy_context
             except ImportError:
                 pass
 
     # Fall back to basic thread-local usage.
 
     from threading import local
 
-    return False, _make_threadlocal_contextvars(local)
+    return False, _make_threadlocal_contextvars(local), _make_noop_copy_context()
```
50-Course, line 1319: Looks good to me.

50-Course:
Following this MR 👍🏼 looking forward to the updates

sentry_sdk/utils.py
```diff
@@ -1248,24 +1249,49 @@ def _make_threadlocal_contextvars(local):
...
+            token = str(random.getrandbits(64))
+            original_value = self.get()
+            setattr(self._original_local, token, original_value)
             self._local.value = value
+            return token
+
+        def reset(self, token):
+            # type: (Any) -> None
+            self._local.value = getattr(self._original_local, token)
+            setattr(self._original_local, token, None)
```
sl0thentr0py, line 1274: isn't it better to delete here so that it gets GCd?

sl0thentr0py (approved):
I don't yet know how you will use `copy_context` but this change is fine self-contained

sentry_sdk/utils.py
antonpirker: @50-Course If it is not garbage collected it stays in memory. With a lot sets and resets the memory usage could grow without memory being freed.

50-Course: Oh okay got it!

PR #2600, opened 2023-12-14
Arq integration ctx

ivanovart:
Arq is broken if the integration is missing because the context is not being passed, here is my tiny fix and a test

---

## General Notes


sentrivana (approved):
Thank you @ivanovart!

PR #2606, opened 2023-12-18
[Experiment] New scope step3

antonpirker:
Third step is switching out the existing functionaly with the new Scope handling.


antonpirker:
Closing in favor of 
https://github.com/getsentry/sentry-python/pull/2609
https://github.com/getsentry/sentry-python/pull/2610

Issue #2576, opened 2023-12-10
Django integration does not respect SCRIPT_NAME


ethrgeist:
### How do you use Sentry?

Self-hosted/on-premise

### Version

self-hosted-23.11.2

### Steps to Reproduce

1. Django setup with `SCRIPT_NAME=/foo/bar` under domain `example.net`
2. Full url `https://example.net/foo/bar/`
3. Triggern an error in a view `https://example.net/foo/bar/my-view/`
4.  URL in Sentry UI is just `https://example.net/my-view/`, missing the `SCRIPT_NAME`

### Expected Result

URL should respect SCRIPT_NAME and the link in the UI should be `https://example.net/foo/bar/my-view/`

### Actual Result

URL in Sentry UI is just `https://example.net/my-view/`

antonpirker:
Hey @ethrgeist ! Thanks for reporting this. Makes total sense. I put it on our list, but also marked it as good first issue and help wanted, so maybe someone wants to jump on this.

Nandika-A:
Hello! Can I work on this issue?

sarvaSanjay:
Apologies for the multiple PRs, got a bit confused but it seems like we need to add the `Trigger: tests using secrets` label to the PR in order for checks to pass. Could anyone help me out with that?

szokeasaurusrex:
> Apologies for the multiple PRs, got a bit confused but it seems like we need to add the `Trigger: tests using secrets` label to the PR in order for checks to pass. Could anyone help me out with that?

Yes @sarvaSanjay – only maintainers can add this label. We will add the label once all other tests are passing and the change has been approved. 

I took a look at your PR and added some comments. Once you address these comments and get the tests to pass, we can hopefully get this change merged.

PR #2622, opened 2024-01-07
fix(wsgi): WSGI integrations respect SCRIPT_NAME env variable

sarvaSanjay:
URLs generated using Sentry's WSGI Middleware should include SCRIPT_NAME in the event's url

Fixes https://github.com/getsentry/sentry-python/issues/2576

Noticed that the problem described in the above issue wasn't specific to Django per se but general to all WSGI applications which is why I have made modifications to the SentryWsgiMiddleware class that should resolve this issue for Django and other integrations as well. As far as I can tell this problem should not arise for ASGI applications as the root_path is accounted for.

## General Notes

#### For maintainers

### DIFF
--- a/sentry_sdk/integrations/wsgi.py
+++ b/sentry_sdk/integrations/wsgi.py
@@ -58,7 +58,14 @@ def get_request_url(environ, use_x_forwarded_for=False):
     return "%s://%s/%s" % (
         environ.get("wsgi.url_scheme"),
         get_host(environ, use_x_forwarded_for),
-        wsgi_decoding_dance(environ.get("PATH_INFO") or "").lstrip("/"),
+        wsgi_decoding_dance(
+            "/".join(
+                [
+                    environ.get("SCRIPT_NAME", "").rstrip("/"),
+                    environ.get("PATH_INFO", "").lstrip("/"),
+                ]
+            )
+        ).lstrip("/"),
     )
 
 
--- a/sentry_sdk/integrations/wsgi.py
+++ b/sentry_sdk/integrations/wsgi.py
@@ -55,17 +55,14 @@ def get_request_url(environ, use_x_forwarded_for=False):
     # type: (Dict[str, str], bool) -> str
     """Return the absolute URL without query string for the given WSGI
     environment."""
+    script_name = environ.get("SCRIPT_NAME", "").rstrip("/")
+    path_info = environ.get("PATH_INFO", "").lstrip("/")
+    path = f"{script_name}/{path_info}"
+
     return "%s://%s/%s" % (
         environ.get("wsgi.url_scheme"),
         get_host(environ, use_x_forwarded_for),
-        wsgi_decoding_dance(
-            "/".join(
-                [
-                    environ.get("SCRIPT_NAME", "").rstrip("/"),
-                    environ.get("PATH_INFO", "").lstrip("/"),
-                ]
-            )
-        ).lstrip("/"),
+        wsgi_decoding_dance(path).lstrip("/"),
     )
 
 
--- a/tests/integrations/wsgi/test_wsgi.py
+++ b/tests/integrations/wsgi/test_wsgi.py
@@ -61,6 +61,29 @@ def test_basic(sentry_init, crashing_app, capture_events):
     }
 
 
+def test_basic_script_name_is_respected(sentry_init, crashing_app, capture_events):
+    sentry_init(send_default_pii=True)
+    app = SentryWsgiMiddleware(crashing_app)
+    client = Client(app)
+    events = capture_events()
+
+    with pytest.raises(ZeroDivisionError):
+        # setting url with PATH_INFO: bark/, HTTP_HOST: dogs.are.great and SCRIPT_NAME: woof/woof/
+        client.get("bark/", "https://dogs.are.great/woof/woof/")
+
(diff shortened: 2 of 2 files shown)


sarvaSanjay:
Hey, its my first contribution. It seems as if we need to add a label to this PR for any of the checks to execute, could any of the collaborators add the same?

szokeasaurusrex:
> Hey, its my first contribution. It seems as if we need to add a label to this PR for any of the checks to execute, could any of the collaborators add the same?

I have approved most of the tests to run. For now, the AWS permissions check will still fail – we will add the necessary label once we have reviewed and approved the change

sentry_sdk/integrations/wsgi.py
```diff
@@ -62,7 +62,14 @@ def get_request_url(environ, use_x_forwarded_for=False):
...
         get_host(environ, use_x_forwarded_for),
-        wsgi_decoding_dance(environ.get("PATH_INFO") or "").lstrip("/"),
+        wsgi_decoding_dance(
+            "/".join(
+                [
+                    environ.get("SCRIPT_NAME", "").rstrip("/"),
+                    environ.get("PATH_INFO", "").lstrip("/"),
+                ]
+            )
+        ).lstrip("/"),
```
szokeasaurusrex, line 72: The following might be more readable. For readability, I would also suggest computing the path before the return statement and storing it in a variable (named `path` or something similar); then, the call could be replaced with `wsgi_decoding_dance(path).lstrip("/")`.
```suggestion
        wsgi_decoding_dance(
            "%s/%s" % (
                environ.get("SCRIPT_NAME", "").rstrip("/"),
                environ.get("PATH_INFO", "").lstrip("/"),
            )
        ).lstrip("/"),
```

tests/integrations/wsgi/test_wsgi.py
```diff
@@ -67,6 +67,29 @@ def test_basic(sentry_init, crashing_app, capture_events):
...
+    events = capture_events()
+
+    with pytest.raises(ZeroDivisionError):
+        # setting url with PATH_INFO: bark/, HTTP_HOST: dogs.are.great and SCRIPT_NAME: woof/woof/
+        client.get("bark/", "https://dogs.are.great/woof/woof/")
+
+    (event,) = events
+
+    assert event["transaction"] == "generic WSGI request"
+    print(event["request"])
```
szokeasaurusrex, line 83:
```suggestion
```
This `print` statement should probably be removed

tests/integrations/wsgi/test_wsgi.py
```diff
@@ -67,6 +67,29 @@ def test_basic(sentry_init, crashing_app, capture_events):
...
+    sentry_init(send_default_pii=True)
+    app = SentryWsgiMiddleware(crashing_app)
+    client = Client(app)
+    events = capture_events()
+
+    with pytest.raises(ZeroDivisionError):
+        # setting url with PATH_INFO: bark/, HTTP_HOST: dogs.are.great and SCRIPT_NAME: woof/woof/
+        client.get("bark/", "https://dogs.are.great/woof/woof/")
+
+    (event,) = events
```
szokeasaurusrex, line 80:
```suggestion
    assert len(events) == 1
    (event,) = events
```
Assuming we are expecting to have exactly one event, we should explicitly assert this expectation to document this expectation in case the test fails. Otherwise, if there isn't exactly one event, the `(event,) = event` line will error, and it might be unclear to whoever is debugging the error whether there is an error in the code or whether the test is implemented incorrectly.

tests/integrations/wsgi/test_wsgi.py
```diff
@@ -67,6 +67,29 @@ def test_basic(sentry_init, crashing_app, capture_events):
...
+
+    assert event["transaction"] == "generic WSGI request"
+    print(event["request"])
+    assert event["request"] == {
+        "env": {"SERVER_NAME": "dogs.are.great", "SERVER_PORT": "443"},
+        "headers": {"Host": "dogs.are.great"},
+        "method": "GET",
+        "query_string": "",
+        "url": "https://dogs.are.great/woof/woof/bark/",
+    }
```
szokeasaurusrex, line 90: Since the purpose of this test is to check that we set the URL correctly, I would suggest only checking that the URL is set correctly, like so:
```suggestion
    assert event["request"]["url"] == "https://dogs.are.great/woof/woof/bark/"
```

This way, we don't have to update the test in case we were to add/change the other fields on the event request dictionary.

tests/integrations/wsgi/test_wsgi.py
```diff
@@ -67,6 +67,29 @@ def test_basic(sentry_init, crashing_app, capture_events):
     }
 
 
+def test_basic_script_name_is_respected(sentry_init, crashing_app, capture_events):
```
szokeasaurusrex, line 70: Maybe we could add additional test cases for edge cases

szokeasaurusrex (commented):
Thank you for the PR! It looks good, but I think we should make a couple of improvements before we merge. Please see my inline comments

szokeasaurusrex:
Also, please address the linter and test failures (except the AWS Lambda tests – you can ignore these for now); we can only merge the changes once all these tests pass

szokeasaurusrex (approved):

PR #2702, opened 2024-02-02
Guard against sentry initialization mid sqlalchemy cursor

apmorton:
In an application that is executing sqlalchemy cursors on a thread concurrently with another thread calling `sentry_sdk.init` it is possible for `_after_cursor_execute` to be called on a cursor that never had `_before_cursor_execute` called.

```
  File "site-packages/sentry_sdk/integrations/sqlalchemy.py", line 98, in _after_cursor_execute
    span = context._sentry_sql_span
           ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'PGExecutionContext_psycopg2' object has no attribute '_sentry_sql_span'
```


sentrivana (approved):
LGTM, thank you @apmorton!

PR #2704, opened 2024-02-05
GrapheneIntegration: Track operation name for tracing

keithhackbarth:
It is helpful to have the operation's name if trying to track the individual performance of GraphQL transactions using Sentry. This PR implements the approach suggested in [this article](https://jerrynsh.com/how-to-monitor-python-graphql-api-with-sentry/), as part of the GrapheneIntegration()

Note: I looked and wasn't initially sure how to write a good test for this and also wasn't sure if this PR is even a good idea in the first place. So leaving a light version here for feedback.

### DIFF
--- a/.github/workflows/test-integrations-ai.yml
+++ b/.github/workflows/test-integrations-ai.yml
@@ -45,31 +45,39 @@ jobs:
       - name: Test anthropic latest
         run: |
           set -x # print commands that are executed
-          ./scripts/runtox.sh "py${{ matrix.python-version }}-anthropic-latest" --cov=tests --cov=sentry_sdk --cov-report= --cov-branch
+          ./scripts/runtox.sh "py${{ matrix.python-version }}-anthropic-latest"
       - name: Test cohere latest
         run: |
           set -x # print commands that are executed
-          ./scripts/runtox.sh "py${{ matrix.python-version }}-cohere-latest" --cov=tests --cov=sentry_sdk --cov-report= --cov-branch
+          ./scripts/runtox.sh "py${{ matrix.python-version }}-cohere-latest"
       - name: Test langchain latest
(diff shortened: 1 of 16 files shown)


czyber:
Hi @keithhackbarth, i would love to see this change! 
Mind if I look into writing tests for it on the weekend?

czyber:
Hey @keithhackbarth, i added a test and made a PR on your fork 😄

sentry_sdk/integrations/graphene.py
```diff
@@ -53,7 +53,8 @@ def _sentry_patched_graphql_sync(schema, source, *args, **kwargs):
         with hub.configure_scope() as scope:
             scope.add_event_processor(_event_processor)
 
-        result = old_graphql_sync(schema, source, *args, **kwargs)
+        with hub.start_transaction(op="graphql", name=kwargs.get("operation_name")):
+            result = old_graphql_sync(schema, source, *args, **kwargs)
```
szokeasaurusrex, line 57: We should be starting a span here rather than a transaction, since Graphene is likely to be called by a web server's response handler, which would already have a transaction automatically started for it (we integrate with the popular web frameworks, like Flask, Django, etc). Creating a transaction within a transaction leads to undefined behavior; they cannot be nested, so we need a span, not a transaction here.

Since spans do not have a `name`, the operation name should be stored in the span's `description` instead.

Lastly, the `op` should be one of the values listed under the GraphQL category in the "Usage" column of [this table](https://develop.sentry.dev/sdk/performance/span-operations/#web-server).

szokeasaurusrex (changes requested):
Thank you for the contribution! This PR will require some changes before we can merge it; I have described these in an inline comment.

PR #2753, opened 2024-02-21
ref: Event Type

szokeasaurusrex:
Implements type hinting for `Event` via a TypedDict. This PR mainly adjusts type hints; however, there are also some minor code changes to make the code type-safe following the new changes.

Some items in the `Event` could have their types expanded by being defined as TypedDicts themselves. These items have been indicated with `TODO` comments. 

Fixes GH-2357


szokeasaurusrex:
Will need to either:
  - Rebase on 2.0 branch
  - Move to functional TypedDict syntax

sentry_sdk/client.py
```diff
@@ -667,7 +675,7 @@ def capture_event(
             envelope = Envelope(headers=headers)
 
             if is_transaction:
-                if profile is not None:
+                if isinstance(profile, Profile):
```
szokeasaurusrex, line 678: `profile` should only ever be getting set to a `Profile`; however, we can only type the profile as `object` because we otherwise have a circular import when building the Sphinx docs.

We check `isinstance` explicitly now to ensure type safety – behavior should however remain the same since we currently only set `profile` to either a `Profile` or `None`.

sentry_sdk/integrations/tornado.py
```diff
@@ -164,7 +164,7 @@ def tornado_processor(event, hint):
 
         with capture_internal_exceptions():
             method = getattr(handler, handler.request.method.lower())
-            event["transaction"] = transaction_from_function(method)
+            event["transaction"] = transaction_from_function(method) or ""
```
antonpirker, line 167: I checked quickly in we never check for `None` anywhere so I guess this is OK.

sentry_sdk/profiler.py
```diff
@@ -683,6 +684,10 @@ def to_json(self, event_opt, options):
             options["project_root"],
         )
 
+        trace_id = None
+        with capture_internal_exceptions():
+            trace_id = event_opt["contexts"]["trace"]["trace_id"]  # type: ignore
```
antonpirker, line 689: strictly not necessary I guess, but as it prevents bugs from happening OK :-)

antonpirker (commented):
Great work @szokeasaurusrex looks great!
I have a few questions.

sentry_sdk/integrations/tornado.py
szokeasaurusrex: Yeah, `event["transaction"]` has a type of `str`, so if we set it, we need to set it to a string. `transaction_from_function` can return `None` (although I think it is rare), hence why I added the `or ""`.

sentry_sdk/profiler.py
szokeasaurusrex: I undid this change, again with the updated `contexts` type, this is now type safe. Of course, we cannot guarantee whether these keys are present in the dictionary, but if this has not caused problems until now, I see no reason to change it.

szokeasaurusrex (commented):
Thank you for the feedback! I replied to your comments and made some changes

sentry_sdk/integrations/logging.py
```diff
@@ -225,7 +226,9 @@ def _emit(self, record):
 
         hint["log_record"] = record
 
-        event["level"] = self._logging_to_event_level(record)
+        level = self._logging_to_event_level(record)
+        if level in {"debug", "info", "warning", "error", "critical", "fatal"}:
```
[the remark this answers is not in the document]
antonpirker: Ok, then I think this is ok.

sentry_sdk/_types.py
```diff
@@ -19,13 +23,69 @@
...
+            "sdk": Mapping[str, object],
+            "server_name": str,
+            "spans": list[dict[str, object]],
+            "stacktrace": dict[
+                str, object
+            ],  # We access this key in the code, but I am unsure whether we ever set it
+            "start_timestamp": Union[datetime, int],
+            "status": Optional[str],
+            "tags": MutableMapping[
+                str, str
```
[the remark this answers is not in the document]
antonpirker: I quickly checked in relay, and I am not sure this is the right place, but I guess they are strings: https://github.com/getsentry/relay/blob/master/relay-event-schema/src/protocol/tags.rs#L16-L17

antonpirker (approved):
Typing rocks!

ods:
Looks like a breaking change. 

Is there a way to use functions with `Event` type arguments without `type: ignore` or erasing type with `Any`? And example from our flow that worked before this:
```python
def before_send(event: dict[str, Any], _: dict[str, Any]) -> dict[str, Any]:
    for frame in iter_event_frames(event):
        do_smth(frame)
    return event
```
Now checking with mypy gives an errors:
```
error: Argument 1 to "iter_event_frames" has incompatible type "dict[str, Any]"; expected "Event"  [arg-type]
```
I'd be happy to switch to using `Event` typed dict, but 1) it's not publicly exposed, 2) it's defined behind `if TYPE_CHECKING` which is not always acceptable.

szokeasaurusrex:
Hi @ods, thank you for reaching out!

> I'd be happy to switch to using Event typed dict, but 1) it's not publicly exposed, 2) it's defined behind if TYPE_CHECKING which is not always acceptable.

We released `1.43.0` earlier today, which re-exports `Event` as a public type in `sentry_sdk.types` (see #2829).

However, because we need to maintain compatibility with old Python versions, we need to define the `Event` type behind an `if TYPE_CHECKING` block. You should be able, however, to use the type hint in your code by importing `Event` from within an `if TYPE_CHECKING` block, and then enclosing the type in quotes, like so:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from sentry_sdk.typing import Event

def before_send(event: "Event", _: dict[str, Any]) -> "Event":
    ...
```

> Looks like a breaking change.

I understand your frustration, and while I certainly believe we made a mistake by using a private type on publicly exposed functions, our reasoning with this change is that it is not breaking because it only affects type hints, which are not evaluated at runtime and not enforced by the Python language. The code is fully backwards-compatible at runtime, and the runtime public API has remained the same as previously – any code written before this change will continue to work the same now after the change.

szokeasaurusrex:
> Strictly speaking it's not the only option. The type can be defined unconditionally, but imported inside `if TYPE_CHECKING` block. This way it won't break in old Python versions, because module is not imported.

This is a good idea! I can certainly look into implementing a conditional import that ensures Event is always defined.

fraser-langton:
Hi @szokeasaurusrex was there a reason this type isn't exposed in the api or just wasn't considered?

szokeasaurusrex:
Hi @fraser-langton, it is exposed in the public API via a [re-export in `sentry_sdk.types`](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/types.py).

PR #2772, opened 2024-02-27
More compatibility tests

antonpirker:
Have some tests that check full events/transactions between SDK 1.x and 2.x


sentrivana (approved):
Nice, lgtm!

PR #2773, opened 2024-02-27
fix(metrics): Replace invalid tag values with an empty string instead of _

markushi:
[According to our spec](https://develop.sentry.dev/sdk/metrics/#normalization) invalid tag values should be replaced with no string.

> Tag Values: regex [^\w\d_:/@\.{}\[\]$-]+ with no replacement character.

I cross checked with [PHP](https://github.com/getsentry/sentry-php/blob/6f8cce5441da92829e84c3c2549625860f11787f/src/Serializer/EnvelopItems/MetricsItem.php#L59) and [JS](https://github.com/getsentry/sentry-javascript/blob/2a8ef4094ff283a207d3522db3e4c3357d73bb30/packages/core/src/metrics/utils.ts#L65): both of them also use no replacement character.


sentrivana (approved):
👍🏻 Thanks for the fix!

PR #2787, opened 2024-03-05
ref(api): Type hinting for start_transaction kwargs

szokeasaurusrex:
This PR adds be type hints for the `**kwargs` that can be passed to `sentry_sdk.start_transaction`, thereby clearly documenting the parameters that can be passed directly in the code.

Ref https://github.com/getsentry/sentry-docs/issues/5082
  - We intend to add to the docs page at least the most useful arguments defined in the `TransactionKwargs` type that this PR introduces.

### DIFF
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -25,7 +25,8 @@ if TYPE_CHECKING:
         ExcInfo,
         MeasurementUnit,
     )
-    from sentry_sdk.tracing import Span, TransactionKwargs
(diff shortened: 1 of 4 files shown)


szokeasaurusrex:
Need to fix syntax errors. Class TypedDict syntax is not supported in old Pythons!

szokeasaurusrex:
Closed in favor of #2796, which applies these changes on the 2.0 branch (getting TypedDict to work in the way I would like it to here would be extremely difficult below Python 3.6)

PR #2788, opened 2024-03-05
feat(graphene): Add span for grapqhl operation

### LINKED ISSUE #2765: GrapheneIntegration: Enhance Transaction Naming for Errors in Graphene GraphQL Operations
### Problem Statement

Currently when an exception or error occurs during the processing of a GraphQL operation by Graphene, the transaction name assigned to the reported issue is often not as descriptive as it could be. This lack of specificity in the transaction naming makes it challenging to quickly identify and address the source of the problem, and there are already existing workarounds such as described in [this article](https://jerrynsh.com/how-to-monitor-python-graphql-api-with-sentry/).

image

### Solution Brainstorm

Setting the transaction name of the current scope to the operation type and the operation name, would add more meaning to the issues, and one could see which operation was called at first glance.

image

This should be just a small change - i'm happy to do it 😄

czyber:
This commit adds a span for a graphql operation to the graphene integration

Fixes #2765

---

## General Notes

#### For maintainers


szokeasaurusrex (dismissed):
Thank you for the contribution, and thanks for your patience as we work through our backlog of community PRs!

I noticed that your PR is still using our old, deprecated `Hub` APIs. We are going to need to replace these with the new corresponding `Scope` APIs before we can merge this PR. More information on how to do this is available in our [Migration Guide](https://docs.sentry.io/platforms/python/migration/1.x-to-2.x#custom-instrumentation). 

Is this a change you are comfortable with making yourself, or do you need us to take over the PR?

antonpirker:
I take the freedom to remove the hubs from the pr

antonpirker (approved):
I think is is looking good now!

sentry_sdk/integrations/graphene.py
```diff
@@ -17,7 +20,7 @@
 
 
 if TYPE_CHECKING:
-    from typing import Any, Dict, Union
+    from typing import Any, Dict, Generator, Union
```
szokeasaurusrex, line 23:
```suggestion
    from collections.abc import Generator
    from typing import Any, Dict, Union
```

`typing.Generator` is a [deprecated alias](https://docs.python.org/3/library/typing.html#typing.Generator) to `collections.abc.Generator`

sentry_sdk/integrations/graphene.py
```diff
@@ -106,3 +114,42 @@ def _event_processor(event, hint):
...
+    else:
+        _graphql_span = sentry_sdk.start_span(op=op, description=operation_name)
+
+    _graphql_span.set_data("graphql.document", source)
+    _graphql_span.set_data("graphql.operation.name", operation_name)
+    _graphql_span.set_data("graphql.operation.type", operation_type)
+
+    yield
+
+    _graphql_span.finish()
```
szokeasaurusrex, line 155: We probably need to make sure we finish the span even in the case that we have an error in the `yield`
```suggestion
    try:
        yield
    finally:
         _graphql_span.finish()
```

szokeasaurusrex:
Mostly looks good, see above

antonpirker:
Committed your suggestions! Thanks a lot for the quick review!

PR #2790, opened 2024-03-06
docs(scope): Correct `use_scope` comment

szokeasaurusrex:
The example in the `use_scope` comment appears to have been copied directly from the `new_scope` example verbatim.

This change fixes the example, so the example demonstrates how to call `use_scope` rather than how to call `new_scope`.

@antonpirker, please make sure that I got the usage in the example correct; I think this is how `use_scope` is used but I'm less than 100% sure.


antonpirker (approved):
lgtm

PR #2791, opened 2024-03-06
OpenAI integration

colin-sentry:
OpenAI integration for performance monitoring.

This commit instruments two OpenAI functions (for now) with monkeypatches, and creates span objects automatically.

---

## General Notes

#### For maintainers


sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,230 @@
...
+    # type: (Callable[..., Any]) -> Callable[..., Any]
+    @wraps(f)
+    def new_chat_completion(*args, **kwargs):
+        # type: (*Any, **Any) -> Any
+        hub = Hub.current
+        integration = hub.get_integration(OpenAIIntegration)
+        if integration is None:
+            return f(*args, **kwargs)
+
+        if "messages" not in kwargs:
```
Zylphrex, line 119: Can't they pass `messages` as a positional argument instead?

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,230 @@
...
+            return f(*args, **kwargs)
+
+        kwargs["messages"] = list(kwargs["messages"])
+        messages = kwargs["messages"]
+        model = kwargs.get("model")
+        streaming = kwargs.get("stream")  # TODO handle streaming
+
+        span = hub.start_span(op="openai", description="Chat Completion")
+        span.__enter__()
+        res = f(*args, **kwargs)
```
Zylphrex, line 136: In the event that `f` throws, I think you'll end up with a span that never finishes.

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,230 @@
...
+
+    if total_tokens == 0:
+        total_tokens = prompt_tokens + completion_tokens
+
+    if completion_tokens != 0:
+        span.set_data(COMPLETION_TOKENS, completion_tokens)
+    if prompt_tokens != 0:
+        span.set_data(PROMPT_TOKENS, prompt_tokens)
+    if total_tokens != 0:
+        span.set_data(TOTAL_TOKENS, total_tokens)
```
Zylphrex, line 106: Not sure how to do it exactly but I believe we now allow measurements on spans. Storing these numeric values as a measurement will allow us to do more with them server side.

sentry_sdk/integrations/openai.py
colin-sentry: No, OpenAI blocks that in the definition
```
 def create(
        self,
        *,
        messages: Iterable[ChatCompletionMessageParam],
```
(the star in line 3 means no positional arguments)

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
...
+                res._iterator = new_iterator()
+            else:
+                span.set_tag("unknown_response", True)
+                span.__exit__(None, None, None)
+            return res
+
+    return new_chat_completion
+
+
+def _wrap_enbeddings_create(f):
```
indragiek, line 185: nit: typo in `enbeddings`

indragiek (commented):
looks great! will defer to @Zylphrex and others for final review

tests/integrations/openai/test_openai.py
```diff
@@ -0,0 +1,157 @@
...
+from sentry_sdk.integrations.openai import OpenAIIntegration
+
+try:
+    from unittest import mock  # python 3.3 and above
+except ImportError:
+    import mock  # python < 3.3
+
+COMPLETION_TOKENS = "completion_tоkens"
+PROMPT_TOKENS = "prompt_tоkens"
+TOTAL_TOKENS = "total_tоkens"
```
antonpirker, line 18: Most of our span data fields are prefixed with something for easier identification. We should do this here to. 
Existing span data fields are documented here: https://develop.sentry.dev/sdk/performance/span-data-conventions/

(we try to align as closely to OpenTelemetry's semantic conventions for attributes (link in the docs page I linked) as possible, so maybe there is already something in there. so please check those)

tests/integrations/openai/test_openai.py
```diff
@@ -0,0 +1,157 @@
...
+from openai.types.chat.chat_completion_chunk import ChoiceDelta, Choice as DeltaChoice
+from openai.types.create_embedding_response import Usage as EmbeddingTokenUsage
+
+from sentry_sdk import start_transaction
+from sentry_sdk.integrations.openai import OpenAIIntegration
+
+try:
+    from unittest import mock  # python 3.3 and above
+except ImportError:
+    import mock  # python < 3.3
```
antonpirker, line 14: We can just use `from unittest import mock` because it will never run in old pythons (and we drop support for older pythons in general with the upcoming SDK 2.0)

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+
+    def count_tokens(s):
+        # type: (str) -> int
+        return 0
+
+
+COMPLETION_TOKENS = "completion_tоkens"
+PROMPT_TOKENS = "prompt_tоkens"
+TOTAL_TOKENS = "total_tоkens"
```
antonpirker, line 47: Most of our span data fields are prefixed with something for easier identification. We should do this here to. 
Existing span data fields are documented here: https://develop.sentry.dev/sdk/performance/span-data-conventions/

(we try to align as closely to OpenTelemetry's semantic conventions for attributes (link in the docs page I linked) as possible, so maybe there is already something in there. so please check those)

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
+from __future__ import absolute_import
```
antonpirker, line 1: can be removed because it will not run in that old python versions.

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
...
+    def count_tokens(s):
+        # type: (str) -> int
+        return len(enc.encode_ordinary(s))
+
+    logger.debug("[OpenAI] using tiktoken to count tokens")
+except ImportError:
+    logger.info(
+        "The Sentry Python SDK requires 'tiktoken' in order to measure token usage from some OpenAI APIs"
+        "Please install 'tiktoken' if you aren't receiving token usage in Sentry."
+        "See https://docs.sentry.io/platforms/python/guides/openai/ for more information."
```
antonpirker, line 37: Docs page not existing yet, but I guess you will add it soon.

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
...
+            # invalid call (in all versions), messages must be iterable
+            return f(*args, **kwargs)
+
+        kwargs["messages"] = list(kwargs["messages"])
+        messages = kwargs["messages"]
+        model = kwargs.get("model")
+        streaming = kwargs.get("stream")
+
+        span = hub.start_span(op="openai", description="Chat Completion")
+        span.__enter__()
```
antonpirker, line 134: It would be cleaner if you do something like this (and remove your `span.__exit__()` calls)
```python 
import sentry_sdk  # do this at top of file

with sentry_sdk.start_span(op="openai", description="Chat Completion")
    # your whole "with capture_internal_exceptions()" block
```

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
...
+        res = f(*args, **kwargs)
+        with capture_internal_exceptions():
+            span.set_data("messages", messages)
+            span.set_tag("model", model)
+            span.set_tag("streaming", streaming)
```
antonpirker, line 139: SDKs should not set tags anymore. (because they are expensive to index). Please use `set_data` instead. 
If it is necessary for the product to have tags for this, we have to add something to `relay` to create the tags from the data. 

Please also check here if your data could match existing span data fields: https://develop.sentry.dev/sdk/performance/span-data-conventions/

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
...
+
+    @wraps(f)
+    def new_embeddings_create(*args, **kwargs):
+        # type: (*Any, **Any) -> Any
+        hub = Hub.current
+        integration = hub.get_integration(OpenAIIntegration)
+        if integration is None:
+            return f(*args, **kwargs)
+
+        with hub.start_span(op="openai", description="Embeddings Creation") as span:
```
antonpirker, line 196: The `op` should also be something like `openai.embeddings.create` and be documented here: https://develop.sentry.dev/sdk/performance/span-operations/
And also added here: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/consts.py#L195

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
...
+        except TypeError:
+            # invalid call (in all versions), messages must be iterable
+            return f(*args, **kwargs)
+
+        kwargs["messages"] = list(kwargs["messages"])
+        messages = kwargs["messages"]
+        model = kwargs.get("model")
+        streaming = kwargs.get("stream")
+
+        span = hub.start_span(op="openai", description="Chat Completion")
```
antonpirker, line 133: The `op` should also be something like `openai.completion.create` and be documented here: https://develop.sentry.dev/sdk/performance/span-operations/
And also added here: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/consts.py#L195

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
...
+        hub = Hub.current
+        integration = hub.get_integration(OpenAIIntegration)
+        if integration is None:
+            return f(*args, **kwargs)
+
+        with hub.start_span(op="openai", description="Embeddings Creation") as span:
+            if "input" in kwargs and isinstance(kwargs["input"], str):
+                span.set_data("input", kwargs["input"])
+            if "model" in kwargs:
+                span.set_tag("model", kwargs["model"])
```
antonpirker, line 200: Same as above. Please not tags.

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
...
+            if "dimensions" in kwargs:
+                span.set_tag("dimensions", kwargs["dimensions"])
```
antonpirker, line 202: Same as above. Please not tags.

sentry_sdk/integrations/openai.py
antonpirker, line 196: Also same as mentioned above. Please use `sentry_sdk.start_span(...)` instead of calling it from the Hub. We are in the process of removing the hub and this way it will be forward compatible.

sentry_sdk/integrations/openai.py
```diff
@@ -0,0 +1,228 @@
...
+                            )
+                            span.set_data("responses", all_responses)
+                            _calculate_chat_completion_usage(
+                                messages, res, span, all_responses
+                            )
+                    span.__exit__(None, None, None)
+
+                res._iterator = new_iterator()
+            else:
+                span.set_tag("unknown_response", True)
```
antonpirker, line 178: Same as above. Please not tags.

antonpirker (changes requested):
Yay! An OpenAI integration! Great work! Requested some changes to have the data in all the right places. 

One important general question:

Is it possible that this will send personal data of any kind?
If yes, please guard it like we do here: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/flask.py#L186

sentry_sdk/integrations/openai.py
colin-sentry: The span exits asynchronously sometimes (after some streaming operation has been done) - it can't use a context manager I think.

sentry_sdk/integrations/openai.py
colin-sentry: PR open now.

sentry_sdk/integrations/openai.py
antonpirker: Ok. Thanks for explaining!

tests/integrations/openai/test_openai.py
```diff
@@ -0,0 +1,187 @@
...
+from sentry_sdk import start_transaction
+from sentry_sdk.integrations.openai import (
+    OpenAIIntegration,
+    COMPLETION_TOKENS_USED,
+    PROMPT_TOKENS_USED,
+    TOTAL_TOKENS_USED,
+)
+
+from unittest import mock
+
```
antonpirker, line 18: Please add some tests that check for PII. Something like:
* init() with OpenAIIntegration() -> no prompts
* init(send_default_pii=True) with OpenAIIntegration() -> prompts
* init(send_default_pii=True) with OpenAIIntegration(include_prompts=False) -> no prompts
* init() or init(send_default_pii=False) with OpenAIIntegration(include_prompts=True) -> no prompts

(we will add a note in the integrations docs to make sure users get this)

antonpirker (changes requested):
Wow, that was a fast update! 

I took the liberty do add a small fix to make the test pass. Please pull your `openai` branch.

Some little changes we will still have to make, but then we are good to go!

antonpirker (approved):
Yes, looks great now! Thanks for all the hard work!
You can merge this now!

PR #2792, opened 2024-03-07
feat(integrations): Propagate sentry-trace and baggage to huey tasks

cnschn:
This PR enables passing `sentry-trace` and `baggage` headers to background tasks using the Huey task queue.

This allows easily correlating what happens inside a background task with whatever transaction (e.g. a user request in a Django application) queued the task in the first place.

Periodic tasks do not get these headers, because otherwise each execution of the periodic task would be tied to the same parent trace (the long-running worker process).

---

## General Notes

#### For maintainers


jedie (approved):

antonpirker (approved):
Looks good! Thanks for the work!

PR #2796, opened 2024-03-07
ref(api): Type hinting for start_transaction kwargs

szokeasaurusrex:
This PR adds be type hints for the **kwargs that can be passed to sentry_sdk.start_transaction, thereby clearly documenting the parameters that can be passed directly in the code.

Ref https://github.com/getsentry/sentry-docs/issues/5082

  - We intend to add to the docs page at least the most useful arguments defined in the TransactionKwargs type that this PR introduces.

This PR replaces #2787

### DIFF
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -17,6 +17,8 @@ if TYPE_CHECKING:
     from typing import ContextManager
     from typing import Union
 
+    from typing_extensions import Unpack
+
     from sentry_sdk.client import BaseClient
     from sentry_sdk._types import (
         Event,
@@ -26,7 +28,7 @@ if TYPE_CHECKING:
         ExcInfo,
         MeasurementUnit,
     )
-    from sentry_sdk.tracing import Span
+    from sentry_sdk.tracing import Span, TransactionKwargs
 
     T = TypeVar("T")
     F = TypeVar("F", bound=Callable[..., Any])
@@ -278,7 +280,7 @@ def start_span(
 @scopemethod
 def start_transaction(
     transaction=None,  # type: Optional[Transaction]
-    **kwargs,  # type: Any
+    **kwargs,  # type: Unpack[TransactionKwargs]
 ):
     # type: (...) -> Union[Transaction, NoOpSpan]
     return Scope.get_current_scope().start_transaction(transaction, **kwargs)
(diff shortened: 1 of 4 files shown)


antonpirker (commented):
Hm. Does not seem to work for me in vscode. (in both api and hub)

Screenshot 2024-03-08 at 08 01 51

szokeasaurusrex:
> Hm. Does not seem to work for me in vscode. (in both api and hub)

@antonpirker Can you try adding this line to the file where the type hinting is broken?

```python
reveal_type(start_transaction)
```

Then, run `mypy filename` and see whether mypy identifies the type correctly

antonpirker:
It seams mypy understands the type:
```
tests/test_api.py:46: note: Revealed type is "def (transaction: Union[sentry_sdk.tracing.Transaction, None] =, **kwargs: Unpack[TypedDict('sentry_sdk.scope.StartTransactionKwargs', {'trace_id'?: builtins.str, 'span_id'?: builtins.str, 'parent_span_id'?: builtins.str, 'same_process_as_parent'?: builtins.bool, 'sampled'?: builtins.bool, 'op'?: builtins.str, 'description'?: builtins.str, 'status'?: builtins.str, 'containing_transaction'?: Union[sentry_sdk.tracing.Transaction, None], 'start_timestamp'?: Union[datetime.datetime, builtins.float, None], 'scope'?: sentry_sdk.scope.Scope, 'name'?: builtins.str, 'source'?: builtins.str, 'parent_sampled'?: builtins.bool, 'baggage'?: sentry_sdk.tracing_utils.Baggage, 'client'?: Union[sentry_sdk.client.Client, None], 'custom_sampling_context'?: builtins.dict[builtins.str, Any]})]) -> Union[sentry_sdk.tracing.Transaction, sentry_sdk.tracing.NoOpSpan]"
```

but vscode does not pick it up. Do I need to do anything to make it work in vscode? (setting an venv or selecting the right interpretor or something like this?)

szokeasaurusrex:
@antonpirker For me it worked out of the box, although sometimes the correct type did not get picked up immediately. Perhaps you could try quitting and reopening VSCode, maybe it is some sort of caching issue. 

If that does not work, make sure you have the venv activated, and make sure you are on a new version of Python (3.11 and 3.12 worked for me).

antonpirker (approved):
looks good. If it works for you out of the box, my vscode setup is probably a problem.

PR #2800, opened 2024-03-11
Add a method for normalizing data passed to set_data

colin-sentry:
Some data (like newer OpenAI versions' outputs) comes in some proprietary format (like pydantic) - this adds a way to recursively convert messy input data into clean json-formattable data.


antonpirker (approved):
Looks good!

I moved the _normalize_data function to the openAI integration. It could be that some users pass a `data` object with a `model_dump` method that does something different and we should never break users code. This way it is saver.

PR #2821, opened 2024-03-14
Update sanic integration

hamedsh:
in Sanic, for some exceptions, quiet variables are set to True (https://github.com/hamedsh/sanic/blob/b8ec9ed3e6f63f4c61fd45d3e09cfc9457a53b82/sanic/exceptions.py#L9) these kinds of exceptions, do not exist in the stderr, but exist in the sentry panel. 

by this modification, these kinds of exceptions will not sent to the sentry hub.


antonpirker (approved):
Looks good!

PR #2826, opened 2024-03-15
Make UUID generation lazy

sl0thentr0py:
same as https://github.com/getsentry/sentry-python/pull/2131

for #2827

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -92,8 +92,8 @@ class Span(object):
     Spans can have multiple child spans thus forming a span tree."""
 
     __slots__ = (
-        "trace_id",
-        "span_id",
+        "_trace_id",
+        "_span_id",
         "parent_span_id",
         "same_process_as_parent",
         "sampled",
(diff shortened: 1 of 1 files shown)


sl0thentr0py:
note that we also have uuid generation on `scope.propagation_context` now and to make that lazy we first need to encapsulate `propagation_context` in a class and not just a raw dict as currently implemented.

https://github.com/getsentry/sentry-python/blob/64e0ec53cbc2e4f3b57a5b45562583ce1130b049/sentry_sdk/scope.py#L225-L232

Without also making those lazy, this will not really have a performance impact but we can merge it anyway.

anonrig:
Is there a reason to not land this pull request right now?

sl0thentr0py:
we can merge but without the prop context changes it won't have full impact, and that might conflict with the v2 refactor, I will let @sentrivana and @antonpirker decide

antonpirker:
I will refactor the propagation context first and then merge this change

antonpirker:
This has found its way into the codebase with the refactoring of the propagation context: https://github.com/getsentry/sentry-python/pull/2970

filips123:
UUIDs in [`tracing.py`](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/tracing.py#L283-L284) are still always generated, as #2970 only made the generation lazy for propagation contexts. Is this intentional?

PR #2828, opened 2024-03-15
Small APIdocs improvement

antonpirker:
This PR makes sure all apidocs are recreated always (by deleting an eventually existing `docs/_build` folder) and also adds some minor changes to `set_level` and `set_tag` to make the types of parameters clear. 

This is related to this Docs PR: https://github.com/getsentry/sentry-docs/issues/5432

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,7 @@ lint: .venv
 apidocs: .venv
 	@$(VENV_PATH)/bin/pip install --editable .
 	@$(VENV_PATH)/bin/pip install -U -r ./docs-requirements.txt
+	rm -rf docs/_build
 	@$(VENV_PATH)/bin/sphinx-build -vv -W -b html docs/ docs/_build
 .PHONY: apidocs
 
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -443,10 +443,17 @@ class Scope(object):
     def level(self, value):
         # type: (LogLevelStr) -> None
         """
-        When set this overrides the level. Deprecated in favor of set_level.
+        When set this overrides the level.
+
+        .. deprecated:: 1.0.0
+            Use :func:`set_level` instead.
 
         :param value: The level to set.
         """
+        logger.warning(
+            "Deprecated: use .set_level() instead. This will be removed in the future."
+        )
+
         self._level = value
 
     def set_level(self, value):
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -51,6 +51,7 @@ lint: .venv
 apidocs: .venv
 	@$(VENV_PATH)/bin/pip install --editable .
 	@$(VENV_PATH)/bin/pip install -U -r ./docs-requirements.txt
+	rm -rf docs/_build
```
antonpirker, line 54: Making sure all the docs are rebuilt every time. (For CI this is not important because it is run in a new runner anyways, more for local apidocs generation)

sentry_sdk/scope.py
```diff
@@ -441,13 +441,21 @@ def clear(self):
...
     def level(self, value):
-        # type: (Optional[LogLevelStr]) -> None
-        """When set this overrides the level. Deprecated in favor of set_level."""
+        # type: (LogLevelStr) -> None
+        """
+        When set this overrides the level. Deprecated in favor of set_level.
+
+        :param value: The level to set.
+        """
         self._level = value
```
szokeasaurusrex, line 450: Should we have a deprecation warning?

szokeasaurusrex (approved):
Looks good, see comment

sentry_sdk/scope.py
antonpirker: Good idea. Added it. (turns out it has been deprecated since 1.0 :-) )

szokeasaurusrex: Great! Have we deleted it in 2.0? If not, we could

sentry_sdk/scope.py
```diff
@@ -441,13 +441,28 @@ def clear(self):
...
+        # type: (LogLevelStr) -> None
+        """
+        When set this overrides the level.
+
+        .. deprecated:: 1.0.0
+            Use :func:`set_level` instead.
+
+        :param value: The level to set.
+        """
+        logger.warning(
```
szokeasaurusrex, line 453: @antonpirker Why are you emitting the warning with `logger`? When I have had to emit deprecation warnings, I usually do so as follows:
```python
warnings.warn("blah blah blah", DeprecationWarning, stacklevel=2)
```

Is there any significant difference with using `logger.warning` vs `warnings.warn`? Do we have a preference in the SDKs?

antonpirker: Just used it because all the other deprecation warnings in that file do it the same way.

PR #2829, opened 2024-03-15
Re-export `Event` in `types.py`

szokeasaurusrex:
End-users may need to use the `Event` type for their type hinting to work following the `Event` type changes. However, we define `Event` in a private module `sentry_sdk._types`, which provides no stability guarantees.

Therefore, this PR creates a new public module `sentry_sdk.types`, where we re-export the `Event` type, and explicitly make it available as public API via `sentry_sdk.types.Event`. The new `sentry_sdk.types` module includes a docstring to inform users that we reserve the right to modify types in minor releases, since we consider types to be a form of documentation (they are not enforced by the Python language), but that we guarantee that we will only remove type definitions in a major release.

### DIFF
--- a/sentry_sdk/types.py
+++ b/sentry_sdk/types.py
@@ -1,0 +1,14 @@
+"""
+This module contains type definitions for the Sentry SDK's public API.
+The types are re-exported from the internal module `sentry_sdk._types`.
+
+Disclaimer: Since types are a form of documentation, type definitions
+may change in minor releases. Removing a type would be considered a
+breaking change, and so we will only remove type definitions in major
+releases.
+"""
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from sentry_sdk._types import Event  # noqa: F401
--- a/sentry_sdk/types.py
+++ b/sentry_sdk/types.py
@@ -11,4 +11,4 @@ releases.
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from sentry_sdk._types import Event  # noqa: F401
+    from sentry_sdk._types import Event, Hint  # noqa: F401


sentry_sdk/types.py
```diff
@@ -0,0 +1,14 @@
...
+Disclaimer: Since types are a form of documentation, type definitions
+may change in minor releases. Removing a type would be considered a
+breaking change, and so we will only remove type definitions in major
+releases.
+"""
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from sentry_sdk._types import Event  # noqa: F401
```
antonpirker, line 14: Lets also re-export the `Hint` type, because those are the two used by before_send/before_send_transaction. 
(I am aware the Hint ist just a dict right now, but we can improve on this later)

antonpirker (approved):
Looks good!

szokeasaurusrex:
Fixed your comment, @antonpirker.

Also, I know we had discussed maybe changing our internal imports of `Event` to import from `types` instead of `_types`; however, it looks like we would need to update many files manually to make this change. 

Perhaps, it would make sense just to change the import in `api`, since that file is likely the one that external users would interact with the most anyway?

szokeasaurusrex:
@antonpirker, yes of course, I was planning to make any import changes within a separate PR :)

mdczaplicki:
What version is it available in?

szokeasaurusrex:
@mdczaplicki This PR is not released yet. It should be included in our next release!

mdczaplicki:
I see. When do you plan to release?

szokeasaurusrex:
@mdczaplicki, there is no date set yet for the next release. However, we will probably release a new version including these changes sometime this week.

mdczaplicki:
Would you say that 1.42 should be yanked? If we use `dict[str, Any]` for type hints then `mypy` is screaming that types are not matching. If we however use `Event` or `Hint` from `sentry_sdk._types` then test suite screams that those cannot be imported.

szokeasaurusrex:
@mdczaplicki Apologies for the inconvenience. We release a new version including these changes later today.

However, we will not yank 1.42 because the code itself remains backwards compatible. If you are encountering problems with the new type hinting, we would recommend updating your `mypy` configuration (e.g. you could import `Event` from `sentry_sdk._types` with a `# type: ignore` comment), or ideally updating to this new version once it is released later today.

HansAarneLiblik:
Shouldn't you add the following to the file also? Otherwise non-typechecking environment can't do the import

```py
else:
  Event = dict[str, Any]
  Hint = ...
```

PR #2830, opened 2024-03-15
feat(profiling): Introduce continuous profiling mode

Zylphrex:
This is a new profiling mode that is mutually exclusive from the existing profiling modes. In the current profiling modes, a profile is always directly attached to a transaction. This new mode will continuously emit chunks of profiling data that will be connected to the span data.


tests/test_continuous_profiler.py
```diff
@@ -0,0 +1,237 @@
+import threading
```
sentrivana, line 1: If you decide to go with the new `profiler` folder in the sdk that I suggested above, it'd make sense to do the same with `test_continuous_profiler.py` and `test_profiler.py` in the tests.

sentrivana (commented):
Left a few comments, but in general looks great to me

sentry_sdk/client.py
```diff
@@ -378,6 +379,14 @@ def _capture_envelope(envelope):
...
                 except Exception as e:
                     logger.debug("Can not set up profiler. (%s)", e)
+            else:
+                try:
+                    setup_continuous_profiler(
+                        self.options,
+                        capture_func=_capture_envelope,
+                    )
+                except Exception as e:
+                    logger.debug("Can not set up continuous profiler. (%s)", e)
```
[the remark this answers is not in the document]
sentrivana: Ah ok, the user has to explicitly provide `_experiments["continuous_profiling_auto_start"]` for the profiler to kick off right? Then disregard.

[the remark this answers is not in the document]
Zylphrex: yes, we just need to bootstrap some stuff initially so when the profiler is started, everything is ready to go

sentry_sdk/continuous_profiler.py
```diff
@@ -0,0 +1,543 @@
...
+    else:
+        default_profiler_mode = ThreadContinuousScheduler.mode
+
+    experiments = options.get("_experiments", {})
+
+    profiler_mode = (
+        experiments.get("continuous_profiling_mode") or default_profiler_mode
+    )
+
+    frequency = DEFAULT_SAMPLING_FREQUENCY
```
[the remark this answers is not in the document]
Zylphrex: Not today, because the intention is to bill by profiler seconds and increasing the frequency will increase costs.

sentry_sdk/profiler/transaction_profiler.py
```diff
@@ -60,15 +60,20 @@
     from typing import List
     from typing import Optional
     from typing import Set
-    from typing import Sequence
-    from typing import Tuple
     from typing_extensions import TypedDict
 
-    import sentry_sdk.tracing
+    from sentry_sdk.profiler.utils import (
```
Zylphrex, line 65: @sentrivana I moved this file into its own folder as suggested. One thing that I am wondering here is if this is considered a breaking change since users may be importing from it. This entire file was originally meant to be purely internal, and users should not be using it directly so I assume it's fine but just wanted to double check because I ran into this previously when I moved an utility some time ago.

sentrivana: Yeah that's a good point. We're not explicitly marking this as internal, but at the same time, the public profiler API is basically what's exposed via `sentry_sdk.init` options. @antonpirker what do you think? (TL;DR: We're moving the contents of `profiler.py` to `profiler/transaction_profiler.py`.)

I'm thinking if it makes sense to expose more stuff from this file in `profiler/__init__.py` so that at least if someone is importing from `sentry_sdk.profiler` for whatever reason, the interface is more or less the same.

antonpirker: I did a quick [github code search](https://github.com/search?type=code&auto_enroll=true&q=%22from+sentry_sdk.profiler%22) and only a hand full of people seem to import it (others are just forks of the SDK)

But to be on the safe side, we should do what @sentrivana said and expose the old API in `profiler/__init__.py`.

sentrivana (approved):

PR #2834, opened 2024-03-18
Fixed OpenAI tests

antonpirker:
This will prevent the streaming reponse OpenAI tests to fail.


szokeasaurusrex (approved):
looks reasonable

PR #2838, opened 2024-03-18
ref(gql): Use new scopes API in GQL Integration

szokeasaurusrex:


sentry_sdk/integrations/gql.py
```diff
@@ -85,25 +90,22 @@ def _patch_execute():
...
-            scope.add_event_processor(_make_gql_event_processor(self, document))
+        scope = Scope.get_isolation_scope()
+        scope.add_event_processor(_make_gql_event_processor(self, document))
 
         try:
             return real_execute(self, document, *args, **kwargs)
         except TransportQueryError as e:
             event, hint = event_from_exception(
                 e,
-                client_options=hub.client.options if hub.client is not None else None,
```
szokeasaurusrex, line 102: It looks to me like this `None` check should be unnecessary with the new API – is this correct?

sentry_sdk/integrations/gql.py
```diff
@@ -85,25 +90,22 @@ def _patch_execute():
...
+                client_options=scope.client.options,
                 mechanism={"type": "gql", "handled": False},
             )
 
-            hub.capture_event(event, hint)
+            scope.capture_event(event, hint)
```
antonpirker, line 108: please use `sentry_sdk.capture_event` here, because it will take the correct scope (the current scope for capture_event)

sentry_sdk/integrations/gql.py
antonpirker: Yes, the client has now always options. But please use `sentry_sdk.get_client()` to get the client, because it could be that this scope has no client.

antonpirker (changes requested):
When possible we should always use `sentry_sdk.*` to call the top level apis, because then you always get the right scope or client.

antonpirker (approved):
looks good!

PR #2839, opened 2024-03-18
test: Fork `test_cloud_resource_context` test

szokeasaurusrex:
This test is flaky. Seems to be because of a concurrency issue because it fails nondeterministically with different call counts on reruns against the same code (e.g. [here](https://github.com/getsentry/sentry-python/actions/runs/8327184318/job/22784714862?pr=2454)). 

Let's add `pytest.mark.forked` to see if it fixes the flakiness!


antonpirker:
Fixed in here: https://github.com/getsentry/sentry-python/pull/2454

PR #2840, opened 2024-03-18
Moved should_send_default_pii into client

### LINKED ISSUE #2819: [SDK 2.0] Fix Client `should_send_default_pii` implementation
`should_send_default_pii` is defined on the `BaseClient` with a simple implementation that always returns `False`. This implementation should be overridden in the `_Client` class.

antonpirker:
Moved functionality from `_should_send_default_pii()` in Hub into `should_send_default_pii` on the Client.

Fixes #2819

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -403,6 +403,15 @@ class _Client(BaseClient):
         """
         return True
 
+    def should_send_default_pii(self):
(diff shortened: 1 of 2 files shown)


szokeasaurusrex (approved):
I added a comment to the old `_should_send_default_pii` so we remember not to use it going forward

PR #2843, opened 2024-03-18
feat(profiling): Add thread data to spans

Zylphrex:
As per getsentry/rfcs#75 getsentry/rfc#75, this adds the thread data to the spans. This will be needed for the continuous profiling mode in #2830.

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -193,13 +193,13 @@ class SPANDATA:
 
     THREAD_ID = "thread.id"
     """
-    The thread id within which the span was started. This should be a string.
+    Identifier of a thread from where the span originated. This should be a string.
     Example: "7972576320"
     """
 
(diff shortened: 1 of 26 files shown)


antonpirker (approved):
Looks ok.

This somewhat weakens our tests, but after talking with the other Python people and also confirmed with Ingest that if there is an onknown value in span data it is just stored anyhow. I think this is fine.

Zylphrex:
> Looks ok.
> 
> This somewhat weakens our tests, but after talking with the other Python people and also confirmed with Ingest that if there is an onknown value in span data it is just stored anyhow. I think this is fine.

👍 I'm not sure there's a stronger test we can add here as the thread ids are unstable. The only alternative I can think of is to mock it in each place which seems worse because now every test is concerned about mocking the thread id data appropriately to pass. Which is why I opted for each test to just assert that the relevant properties are added to the span data.

PR #2844, opened 2024-03-19
ref: `should_send_default_pii` shortcut

szokeasaurusrex:
Currently, with the new scope API, calls to `hub._should_send_default_pii` need to be replaced with calls to `scope.Scope.get_client().should_send_default_pii`.

This PR introduces `scope.should_send_default_pii` as a drop-in replacement for `hub._should_send_default_pii`, so we don't need to type out `Scope.get_client().should_send_default_pii` everywhere we need to check `should_send_default_pii`.


sentrivana (approved):
Nice!

antonpirker (approved):

PR #2905, opened 2024-03-26
ref: Correct `api.start_transaction` method signature

szokeasaurusrex:
Adds the `instrumenter` argument to `sentry_sdk.start_transaction`. Should unblock #2865.


sentrivana (approved):
🙏🏻

PR #2914, opened 2024-03-27
ref: Define types at runtime

szokeasaurusrex:
Set types in `sentry_sdk.types` to `None` at runtime. This allows the types to be imported from outside `if TYPE_CHECKING` guards.

Fixes GH-2909

### DIFF
--- a/sentry_sdk/types.py
+++ b/sentry_sdk/types.py
@@ -11,7 +11,7 @@ releases.
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from sentry_sdk._types import Event, Hint  # noqa: F401
+    from sentry_sdk._types import Event, Hint
 else:
     # The lines below allow the types to be imported from outside `if TYPE_CHECKING`
     # guards. The types in this module are only intended to be used for type hints.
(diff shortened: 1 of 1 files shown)


sentry_sdk/types.py
```diff
@@ -12,5 +12,10 @@
 
 if TYPE_CHECKING:
     from sentry_sdk._types import Event, Hint  # noqa: F401
```
asottile-sentry, line 14: you can remove the noqa comment here now

sentry_sdk/types.py
```diff
@@ -12,5 +12,10 @@
...
+else:
+    # The lines below allow the types to be imported from outside `if TYPE_CHECKING`
+    # guards. The types in this module are only intended to be used for type hints.
+    Event = None
+    Hint = None
 
-    __all__ = ["Event", "Hint"]
+__all__ = ["Event", "Hint"]
```
asottile-sentry, line 21:
```suggestion
__all__ = ("Event", "Hint")
```

it doesn't really make much of a perf difference here but it's a good habit anyway

asottile-sentry (approved):

PR #2917, opened 2024-03-28
ref: Remove deprecated `Transaction` creation method

szokeasaurusrex:
This has been deprecated for 4 years, so I suppose we can remove it now in `2.0`.

Removing the `__new__` method also fixes our API docs for the `Span` and `Transaction` constructors. The `__new__` method causes the API docs to show both classes as only taking opaque `**kwargs`, like so:

image

Following this change, the arguments are displayed correctly in the API docs:

image

Partially addresses https://github.com/getsentry/sentry-docs/issues/5082


sentrivana (approved):
👍🏻

antonpirker (approved):

PR #2920, opened 2024-03-28
ref: Improve `start_transaction` docs

szokeasaurusrex:
This PR adds documentation for all `start_transaction` parameters. It also proposes removing the `StartTransactionKwargs` typeddict, instead placing the `custom_sampling_context` parameter directly on the method and deleting the `client` parameter, which appears not to be used.

ref https://github.com/getsentry/sentry-docs/issues/5082


antonpirker (approved):
Looks fine.

PR #2923, opened 2024-03-28
docs: Document Transaction and Span kwargs typed dicts

szokeasaurusrex:
Repeating the doc comments also on the kwargs typed dicts enables better hinting in VSCode. See example below.

Before:
Screenshot 2024-03-28 at 15 16 08

After:
Screenshot 2024-03-28 at 15 14 44

I wish there was a way to do this without simply repeating the documentation both on the classes and in the Kwargs typeddicts, but this is the best way I found.

ref: https://github.com/getsentry/sentry-docs/issues/5082

### DIFF
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -29,9 +29,9 @@ if TYPE_CHECKING:
         ExcInfo,
         MeasurementUnit,
         LogLevelStr,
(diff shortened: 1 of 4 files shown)


antonpirker (approved):
Yay!

Issue #2926, opened 2024-04-02
Set `Event` to something other than `None` at runtime


szokeasaurusrex:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.44.0

### Steps to Reproduce

See https://github.com/getsentry/sentry-python/issues/2909#issuecomment-2028659630. Simply running the code (i.e. with `python` not `pytest`) also causes the problem.

### Expected Result

No error

### Actual Result

```
TypeError: unsupported operand type(s) for |: 'NoneType' and 'NoneType'
```

PR #2928, opened 2024-04-02
fix(types): Fixed `Event | None` runtime `TypeError`

szokeasaurusrex:
Change `Event`'s runtime value to `typing.Any`, since the previous value of None caused the expression `Event | None` to result in a `TypeError` at runtime, even when the `Event | None` expression was used as a type hint. Also, add a test to make sure we don't reintroduce this bug.

Fixes GH-2926

### DIFF
--- a/sentry_sdk/types.py
+++ b/sentry_sdk/types.py
@@ -13,9 +13,11 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from sentry_sdk._types import Event, Hint
 else:
+    from typing import Any
+
     # The lines below allow the types to be imported from outside `if TYPE_CHECKING`
     # guards. The types in this module are only intended to be used for type hints.
-    Event = None
-    Hint = None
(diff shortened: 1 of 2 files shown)


antonpirker (approved):

antonpirker:
One question: Why are we not importing the types from `_types` no matter what `TYPE_CHECKING` is?

szokeasaurusrex:
> One question: Why are we not importing the types from `_types` no matter what `TYPE_CHECKING` is?

All the types are defined within an `if TYPE_CHECKING` block. So, the types only exist in `TYPE_CHECKING` environments. Attempting the import at runtime results in an `ImportError`:

```
>>> from sentry_sdk._types import Event
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'Event' from 'sentry_sdk._types' (/Users/dszoke/Development/SDK/sentry-python/sentry_sdk/_types.py)
```

PR #2935, opened 2024-04-04
Speed up grpc test suite. Use `grpcio` instead of grpcio-tools in tests 

antonpirker:
Installing just `grpcio` instead of `grpcio-tools` which is not needed by tests because no protobufs are compiled during running tests.


antonpirker:
not needed. see sdk 2.0 branch...

PR #2954, opened 2024-04-09
fix(metrics): Change `data_category` from `statsd` to `metric_bucket`

cleptric:
The event category for emitted metrics is `metric_bucket` and not `statsd`.


antonpirker:
@cleptric can you add the "why" to the PR description, so we have a nice commit message, thank you!

antonpirker (approved):

PR #2962, opened 2024-04-10
Add comments for args and kwargs

szokeasaurusrex:
PR with some comments on where we could add the data needed for the enhancements of the queues module described in https://github.com/getsentry/sentry-python/issues/2951. This PR is based on Sentry SDK 2.0.

**Do not merge!**


szokeasaurusrex:
No longer needed, as #2951 is completed

Issue #2827, opened 2024-03-15
Make UUID generation Lazy / Create PropagationContext class


antonpirker:
We want to lazily create UUIDs to decrease the number of CPU cycles we use. 

There are two important places where UUIDs are created:

1. Span/Transaction creation: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/tracing.py#L145-L146
2. When a propagation context is created: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/scope.py#L225-L232
3. When creating a Profile: https://github.com/getsentry/sentry-python/blob/sentry-sdk-2.0/sentry_sdk/profiler.py#L455
4. When creating a Session: https://github.com/getsentry/sentry-python/blob/sentry-sdk-2.0/sentry_sdk/session.py#L49

## Solution brainstorm

1. There is a PR that already started the work: https://github.com/getsentry/sentry-python/pull/2826
This already fixes point 1. from above. 
2. For point 2. we need to create a class `PropagationContext` and create the uuid4 only when the `trace_id` or the `span_id` is read.
3. Implement similar to 1.
4. Implement similar to 1.

sl0thentr0py:
PR https://github.com/getsentry/sentry-python/pull/2826

PR #2970, opened 2024-04-11
Refactoring propagation context

antonpirker:
Create a class for the `PropagationContext`. Make the class generate the UUIDs lazily.

Fixes https://github.com/getsentry/sentry-python/issues/2827

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -1,6 +1,5 @@
 import os
 import sys
-import uuid
 from copy import copy
 from collections import deque
 from contextlib import contextmanager
@@ -18,6 +17,7 @@ from sentry_sdk.tracing_utils import (
     extract_sentrytrace_data,
     has_tracing_enabled,
     normalize_incoming_data,
+    PropagationContext,
 )
 from sentry_sdk.tracing import (
     BAGGAGE_HEADER_NAME,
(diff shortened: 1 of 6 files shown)


sentry_sdk/scope.py
```diff
@@ -431,77 +430,28 @@ def _load_trace_data_from_env(self):
...
-                baggage_header
-            ).dynamic_sampling_context()
-
-        sentry_trace_header = normalized_data.get(SENTRY_TRACE_HEADER_NAME)
-        if sentry_trace_header:
-            sentrytrace_data = extract_sentrytrace_data(sentry_trace_header)
-            if sentrytrace_data is not None:
-                context.update(sentrytrace_data)
-
-        only_baggage_no_sentry_trace = (
```
sl0thentr0py, line 451: this case is no longer there in the new logic, is it not needed?

sl0thentr0py (approved):
one comment, ty for refactoring!

sentry_sdk/tracing_utils.py
```diff
@@ -318,6 +319,109 @@ def _format_sql(cursor, sql):
...
+        if baggage_header:
+            propagation_context = PropagationContext()
+            propagation_context.dynamic_sampling_context = Baggage.from_incoming_header(
+                baggage_header
+            ).dynamic_sampling_context()
+
+        sentry_trace_header = normalized_data.get(SENTRY_TRACE_HEADER_NAME)
+        if sentry_trace_header:
+            sentrytrace_data = extract_sentrytrace_data(sentry_trace_header)
+            if sentrytrace_data is not None:
```
antonpirker, line 378: It is here @sl0thentr0py

sentry_sdk/scope.py
antonpirker: see other comment below. it is there.

Issue #2961, opened 2024-04-10
Add queue or topic to Celery transactions/spans


szokeasaurusrex:
The queue should be sent with the producer span and the consumer transaction according to the [Otel Specs](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#destinations). We first need to figure out how to obtain the queue.

```[tasklist]
### Tasks
- [ ] #2984
- [x] ~`messaging.destination_publish.name`: Queue task is published to originally (medium priority)~ not required per https://develop-git-queue-module.sentry.dev/sdk/performance/modules/queue/
```

szokeasaurusrex:
Logic for setting data needed to route to correct queue in the producer side. From here, it seems that the producer only sometimes will know the queue; in other cases, the broker determines the queue.

https://github.com/celery/celery/blob/481cf8e19ca2b5d34eb74a6954974e9dc92ebda3/celery/app/task.py#L622-L645

The question, therefore, is whether we can determine the queue on the consumer side.

PR #2984, opened 2024-04-16
feat(celery): Send queue name to Sentry

szokeasaurusrex:
Send the queue name to Sentry for Celery tasks using the default exchange. The queue name is sent as span data with the key `messaging.destination.name`.

Closes GH-2961

### DIFF
--- a/sentry_sdk/integrations/celery/__init__.py
+++ b/sentry_sdk/integrations/celery/__init__.py
@@ -4,7 +4,7 @@ from functools import wraps
 import sentry_sdk
 from sentry_sdk import isolation_scope
 from sentry_sdk.api import continue_trace
-from sentry_sdk.consts import OP
+from sentry_sdk.consts import OP, SPANDATA
 from sentry_sdk.integrations import Integration, DidNotEnable
 from sentry_sdk.integrations.celery.beat import (
     _patch_beat_apply_entry,
@@ -325,6 +325,18 @@ def _wrap_tracer(task, f):
     return _inner  # type: ignore
 
 
+def _set_messaging_destination_name(task, span):
+    # type: (Any, Span) -> None
+    """Set "messaging.destination.name" tag for span"""
+    with capture_internal_exceptions():
+        delivery_info = task.request.delivery_info
+        routing_key = delivery_info.get("routing_key")
+        if delivery_info.get("exchange") == "" and routing_key is not None:
+            # Empty exchange indicates the default exchange, meaning the tasks
+            # are sent to the queue with the same name as the routing key.
+            span.set_data(SPANDATA.MESSAGING_DESTINATION_NAME, routing_key)
+
+
 def _wrap_task_call(task, f):
     # type: (Any, F) -> F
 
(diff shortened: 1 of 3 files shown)


szokeasaurusrex:
Due to the current way we test Celery, it appears there is no straightforward way to test this change in our test suite. The problem is that our tests run with Celery's eager mode, which means the tasks are run synchronously in the same process, without any queueing. Since there is no queue, we never set `messaging.destination.name`.

We should improve our Celery tests, and add a test for this change, in a separate PR.

sentry_sdk/integrations/celery/__init__.py
```diff
@@ -311,20 +311,38 @@ def _inner(*args, **kwargs):
...
+    # to add @functools.wraps(f) here.
     # https://github.com/getsentry/sentry-python/issues/421
-    @wraps(f)
+    @ensure_integration_enabled(CeleryIntegration, f)
     def _inner(*args, **kwargs):
         # type: (*Any, **Any) -> Any
         try:
-            return f(*args, **kwargs)
+            with sentry_sdk.start_span(
+                op=OP.QUEUE_TASK_CELERY, description=task.name
```
szokeasaurusrex, line 342: Is there a better `op` we could use here? `queue.task.celery` seemed like the most fitting `op`; however, since the parent transaction also has `op` set to `queue.task.celery`, this leads to the span having the same `op` as its parent, which could be somewhat confusing

antonpirker (changes requested):
The one assert in the tests I do not get. 

And because of the "always-eager" there is no way we can tests for `messaging.destination.name` right? Do you have any other PR plan on how we could run tests with a queue?

tests/integrations/celery/test_celery.py
```diff
@@ -209,7 +209,17 @@ def dummy_task(x, y):
     else:
         assert execution_event["contexts"]["trace"]["status"] == "ok"
 
-    assert execution_event["spans"] == []
+    assert len(execution_event["spans"]) == 1
+    assert (
```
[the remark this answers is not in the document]
szokeasaurusrex: This assert uses a `>=` on the dictionary items, which asserts that the `execution_event["spans"][0].items()` is a superset of the items of the dictionary provided in the test. So, this assert checks that all of the key-value pairs I listed (`trace_id`, `same_process_as_parent`, `op`, `description`, and `data`) are present in the `execution_event["spans"][0]` dictionary, and that each of them have the values I specified. However, the `execution_event["spans"][0]` can contain any additional key-value pairs.

The reason I wrote the assert this way was to be explicit about checking only what is relevant to this test case. The assert that was already here (on line 223) solves this problem by looking up the items we don't care about (e.g. `parent_span_id`) in the dictionary itself, which I find confusing. My way should also be more maintainable, since no changes to the test are required if we ever add additional keys to the span dictionary.

szokeasaurusrex:
> And because of the "always-eager" there is no way we can tests for `messaging.destination.name` right? Do you have any other PR plan on how we could run tests with a queue?

I could not find an obvious way to test `messaging.destination.name`, since with the "always-eager" mode, the queueing information does not get set properly. We would need to find how to have a real queue in the tests. Or, perhaps, we could work around this with some mocking logic

tests/integrations/celery/test_celery.py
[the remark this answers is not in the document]
antonpirker: Now I got it! Thanks for explaining.

sentry_sdk/integrations/celery/__init__.py
cleptric: Will be `queue.process` if I get the context here right.
We'll put `celery` on `messaging.system`.

szokeasaurusrex: So this would be a new span `op` @cleptric? I do not see `queue.process` listed anywhere on our [list of span `op`s](https://develop.sentry.dev/sdk/performance/span-operations/).

szokeasaurusrex:
@antonpirker Turns out that we can test for `messaging.destination.name` by mocking out the task request. I just added some commits to test the changes.

antonpirker (approved):

antonpirker:
Failing tests should be fixed by this pr: https://github.com/getsentry/sentry-python/pull/3030

edwardgou-sentry (approved):
worked for me in testing

PR #2999, opened 2024-04-22
Make it work with old and new newrelic versions

antonpirker:
This fixes our tests that rely on the `newrelic` package. In their version `v9.9.0` they removed a function that we called so we had to update our tests.


sentrivana (approved):
Thanks for fixing!

Issue #2990, opened 2024-04-18
Transactions started with `continue_trace()` do not inherit sampling


jwhitaker-gridcog:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.45.0

### Steps to Reproduce

Run the following python script, setting SENTRY_DSN accordingly.

Alternatively, clone from https://github.com/jwhitaker-gridcog/sentry-bug .

<details><summary>Python script</summary>

```py
import subprocess
import sentry_sdk
import multiprocessing
import argparse
import sys
import os

# set SENTRY_DSN in env


# worker: this worker is used in two demos: first by multiprocessing, second by subprocess.
def worker(baggage_and_tp):
    sentry_sdk.init(traces_sample_rate=1.0, environment="jarrad-local", debug=True)
    baggage, traceparent = baggage_and_tp
    try:
        with sentry_sdk.continue_trace(
            {"baggage": baggage, "sentry-trace": traceparent}, op="task", name="child"
        ) as worker_tx:
            # # workaround:
            # if worker_tx.parent_sampled:
            #     worker_tx.sampled = True
            #     worker_tx.init_span_recorder(1000)
            sentry_sdk.capture_message("hi")
    finally:
        sentry_sdk.Hub.current.flush()


# demo 1: worker is started with multiprocessing.Process(). baggage is passed with multiprocessing magic.
def main_multiprocessing():
    sentry_sdk.init(traces_sample_rate=1.0, environment="jarrad-local", debug=True)

    with sentry_sdk.start_transaction(op="task", name="parent") as tx:
        baggage, traceparent = sentry_sdk.get_baggage(), sentry_sdk.get_traceparent()

        proc = multiprocessing.Process(
            target=worker, name="worker", args=[(baggage, traceparent)]
        )
        proc.start()
        proc.join()


# helper for demo 2: pull baggage from env and pass to worker.
def worker_subprocess():
    baggage, traceparent = (
        os.environ["SENTRY_BAGGAGE"],
        os.environ["SENTRY_TRACEPARENT"],
    )
    worker((baggage, traceparent))


# demo 2: worker is started with subprocess. baggage is passed with env vars.
def main_subprocess():
    sentry_sdk.init(traces_sample_rate=1.0, environment="jarrad-local", debug=True)

    with sentry_sdk.start_transaction(op="task", name="parent") as tx:
        baggage, traceparent = sentry_sdk.get_baggage(), sentry_sdk.get_traceparent()

[...]

PR #3003, opened 2024-04-22
Warn when transaction entered without calling `start_transaction`

szokeasaurusrex:
This change should prevent confusion like what led to #2990 being opened.

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -714,8 +714,27 @@ class Transaction(Span):
             )
         )
 
+    def _possibly_started(self):
+        # type: () -> bool
+        """Returns whether the transaction might have been started.
+
+        If this returns False, we know that the transaction was not started
+        with sentry_sdk.start_transaction, and therefore the transaction will
+        be discarded.
+        """
+
+        # We must explicitly check self.sampled is False since self.sampled can be None
+        return self._span_recorder is not None or self.sampled is False
+
     def __enter__(self):
         # type: () -> Transaction
+        if not self._possibly_started():
+            logger.warning(
+                "Transaction was entered without being started with sentry_sdk.start_transaction."
+                "The transaction will not be sent to Sentry. To fix, start the transaction by"
+                "passing it to sentry_sdk.start_transaction."
+            )
+
         super().__enter__()
 
         if self._profile is not None:
--- a/tests/tracing/test_misc.py
+++ b/tests/tracing/test_misc.py
@@ -401,3 +401,19 @@ def test_transaction_dropeed_sampled_false(sentry_init):
         mock_logger.debug.assert_any_call(
             "Discarding transaction because it was not started with sentry_sdk.start_transaction"
         )
+
+
+def test_transaction_not_started_warning(sentry_init):
+    sentry_init(enable_tracing=True)
+
+    tx = Transaction()
+
+    with mock.patch("sentry_sdk.tracing.logger") as mock_logger:
+        with tx:
+            pass
+
+    mock_logger.warning.assert_any_call(
+        "Transaction was entered without being started with sentry_sdk.start_transaction."
+        "The transaction will not be sent to Sentry. To fix, start the transaction by"
+        "passing it to sentry_sdk.start_transaction."
+    )


sentry_sdk/tracing.py
```diff
@@ -695,8 +695,31 @@ def __repr__(self):
             )
         )
 
+    def _possibly_stated(self):
```
antonpirker, line 698:
```suggestion
    def _possibly_started(self):
```

sentry_sdk/tracing.py
```diff
@@ -695,8 +695,31 @@ def __repr__(self):
...
+        with sentry_sdk.start_transaction, and therefore the transaction will
+        be discarded.
+        """
+
+        # We must explicitly check self.sampled is False since self.sampled can be None
+        return self._span_recorder is not None or self.sampled is False
+
     def __enter__(self):
         # type: () -> Transaction
+        if not self._possibly_stated():
```
antonpirker, line 712:
```suggestion
        if not self._possibly_started():
```

sentry_sdk/tracing.py
```diff
@@ -695,8 +695,31 @@ def __repr__(self):
...
+        # type: () -> bool
+        """Returns whether the transaction might have been started.
+
+        If this returns False, we know that the transaction was not started
+        with sentry_sdk.start_transaction, and therefore the transaction will
+        be discarded.
+        """
+
+        # We must explicitly check self.sampled is False since self.sampled can be None
+        return self._span_recorder is not None or self.sampled is False
```
antonpirker, line 708: `sampled=False` is a sign that the transaction has been started because it is set in `_set_inital_sampling_decision` that is only called in start_transaction, right?

tests/tracing/test_misc.py
```diff
@@ -401,3 +401,23 @@ def test_transaction_dropeed_sampled_false(sentry_init):
...
+            pass
+
+    mock_logger.warning.assert_any_call(
+        """Transaction was entered without being started with sentry_sdk.start_transaction.
+                   The transaction will not be sent to Sentry. To fix, start the transaction by
+                   passing it to sentry_sdk.start_transaction, like so:
+
+                   with sentry_sdk.start_transaction(transaction):
+                       # code here
+                """
```
antonpirker, line 422:
```suggestion
        The transaction will not be sent to Sentry. To fix, start the transaction by
        passing it to sentry_sdk.start_transaction, like so:

        with sentry_sdk.start_transaction(transaction):
        # code here
        """
```

antonpirker (approved):
Looks good, just one question.

sentry_sdk/tracing.py
szokeasaurusrex: No, `sampled is False` tells us nothing about whether the transaction has been started – `self.sampled` could also be set in the `__init__` (i.e. if the user passes `sampled=False` to the constructor). However, if the user calls `start_transaction` on a transaction that has `sampled=False`, `self._span_recorder` will be `None`, even though the transaction was started with `start_transaction`.

As I attempted to explain in the docstring, this function's purpose is to tell us whether the transaction _might_ have been started with `start_transaction`. In other words:
  - If the function returns `True`, we _do not know_ whether the transaction was started with `start_transaction`, but it might have been.
      - Specifically, if `self.sampled is False`, we have no way to tell whether the user started the transaction with `start_transaction`, so we return `True`.
  - If the function returns `False`, we _know_ that the transaction _was not started_ with `start_transaction`.

tests/tracing/test_misc.py
szokeasaurusrex: This spacing is intentional; it makes the message appear nicely in the logs, like so:

```
 [sentry] WARNING: Transaction was entered without being started with sentry_sdk.start_transaction.
                   The transaction will not be sent to Sentry. To fix, start the transaction by
                   passing it to sentry_sdk.start_transaction, like so:

                   with sentry_sdk.start_transaction(transaction):
                       # code here            
```

sentry_sdk/tracing.py
antonpirker: Got it, thanks for the explanation!

sentry_sdk/tracing.py
```diff
@@ -695,8 +695,31 @@ def __repr__(self):
...
+        if not self._possibly_started():
+            logger.warning(
+                """Transaction was entered without being started with sentry_sdk.start_transaction.
+                   The transaction will not be sent to Sentry. To fix, start the transaction by
+                   passing it to sentry_sdk.start_transaction, like so:
+
+                   with sentry_sdk.start_transaction(transaction):
+                       # code here
+                """
+            )
```
antonpirker, line 721: Log messages should not be multi-line because it makes it hard or impossible to grep through the logs (or do some other line based processing)

Please change the log message to be one line.

antonpirker (changes requested):

szokeasaurusrex:
Made the requested change @antonpirker

antonpirker (approved):
Great, thanks for the change!

PR #3005, opened 2024-04-23
Restore original behavior by always creating a span

antonpirker:
In the original implementation of celery trace propagation we had code to only create a span for the task if it was NOT started by Celery Beat (because there is no transaction created in the beat process, so also not span should be created).

See this code: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/celery.py#L187-L200

Turns out this has never worked and `task_started_from_beat` has always been `False` meaning a span was ALWAYS created.

(This did never break anything or caused any troube. When looking into a transaction less future this is also absolutely fine.)

So this PR restores now the original behavior by always creating a span.

### DIFF
--- a/sentry_sdk/integrations/celery/__init__.py
+++ b/sentry_sdk/integrations/celery/__init__.py
@@ -243,15 +243,7 @@ def _wrap_apply_async(f):
 
         task = args[0]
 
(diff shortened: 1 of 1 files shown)


sentrivana (approved):
LGTM. Why did the check never work in the first place, do we know?

Issue #2964, opened 2024-04-10
Caching module update for DjangoIntegration


antonpirker:
To make spans created by the `DjangoIntegration` (when `cache_spans` is enabled) eligible for the "Cache Module" in Sentry Performance include more data when `cache_spans` is enabled. 

Please follow the specification found here:
https://develop.sentry.dev/sdk/performance/modules/caches/

Part of https://github.com/getsentry/sentry-python/issues/2949

Bonus:
- [ ] https://github.com/getsentry/sentry-python/issues/3058

PR #3009, opened 2024-04-24
Django caching instrumentation update

antonpirker:
This adds more data to the cache spans and makes adding the cache item size optional. 
This implements parts of following spec https://develop.sentry.dev/sdk/performance/modules/cache/

Documentation for manual instrumentation for the Cache module can be found here:
https://github.com/getsentry/sentry-docs/pull/9926

Load testing of the feature was done here: https://github.com/getsentry/sentry-python/issues/3058

Fixes https://github.com/getsentry/sentry-python/issues/2964


sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,138 @@
...
-                span.set_data(SPANDATA.CACHE_HIT, True)
-
-                size = len(str(value))
-                span.set_data(SPANDATA.CACHE_ITEM_SIZE, size)
-
-            else:
-                span.set_data(SPANDATA.CACHE_HIT, False)
+            with capture_internal_exceptions():
+                if address is not None:
+                    # Strip eventually contained username and password in url style location
```
szokeasaurusrex, line 68: Can you please explain what this comment means? I am struggling to understand it

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,138 @@
...
+                            parsed_url.netloc or "",
+                            parsed_url.path or "",
+                        )
+
+                    span.set_data(SPANDATA.NETWORK_PEER_ADDRESS, address)
+
+                if port is not None:
+                    span.set_data(SPANDATA.NETWORK_PEER_PORT, int(port))
+
+                key = _get_key(args, kwargs)
```
szokeasaurusrex, line 82: Can `key` ever be `None` in the args or kwargs? Asking because `str(None) == "None"`, so just wondering if this is a case we need to handle

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,138 @@
...
+                    span.set_data(SPANDATA.CACHE_KEY, key)
+
+                item_size = None
+                if is_get_operation:
+                    if value:
+                        if integration.cache_spans_add_item_size:
+                            item_size = len(str(value))
+                        span.set_data(SPANDATA.CACHE_HIT, True)
+                    else:
+                        span.set_data(SPANDATA.CACHE_HIT, False)
```
szokeasaurusrex, line 93:
```suggestion
                if is_get_operation:
                    if integration.cache_spans_add_item_size:
                        item_size = len(str(value))
                    span.set_data(SPANDATA.CACHE_HIT, bool(value))  # Or maybe `value is not None` would be better
```

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,138 @@
...
+                else:
+                    if integration.cache_spans_add_item_size:
+                        item_size = len(str(args[1]))
```
szokeasaurusrex, line 96:
```suggestion
                elif integration.cache_spans_add_item_size:
                    item_size = len(str(args[1]))
```

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,138 @@
...
 
-def _patch_cache_method(cache, method_name):
-    # type: (CacheHandler, str) -> None
+def _patch_cache_method(cache, method_name, address, port):
+    # type: (CacheHandler, str, Optional[str], Optional[str]) -> None
     from sentry_sdk.integrations.django import DjangoIntegration
 
     original_method = getattr(cache, method_name)
 
     @ensure_integration_enabled(DjangoIntegration, original_method)
```
szokeasaurusrex, line 51:
```suggestion
```
I think it makes more sense to place this on the `sentry_method` function. It is inaccurate to say that this method maps `original_method` because it has a different signature. Although, perhaps it would be more appropriate to make this change in a separate PR.

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,138 @@
...
+                else:
+                    if integration.cache_spans_add_item_size:
+                        item_size = len(str(args[1]))
+
+                if item_size is not None:
+                    span.set_data(SPANDATA.CACHE_ITEM_SIZE, item_size)
 
             return value
 
     @functools.wraps(original_method)
```
szokeasaurusrex, line 103:
```suggestion
    @ensure_integration_enabled(DjangoIntegration, original_method)
```

To be applied alongside my comment on (original version) line 41; also, likely makes sense to do in separate PR.

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,138 @@
...
+            _patch_cache_method(cache, method_name, address, port)
         cache._sentry_patched = True
 
 
+def _get_address_port(settings):
+    # type: (dict[str, Any]) -> Tuple[Optional[str], Optional[str]]
+    address, port = None, None
+    location = settings.get("LOCATION")
+    # TODO: location can also be an array of locations
+    # see: https://docs.djangoproject.com/en/5.0/topics/cache/#redis
```
szokeasaurusrex, line 126: Are you planning to add code to handle this case in this PR?

tests/integrations/django/test_basic.py
```diff
@@ -1158,240 +1126,3 @@ def dummy(a, b):
...
-            "something",
-            {},
-            "get s",
-        ),  # this should never happen, just making sure that we are not raising an exception in that case.
-    ],
-)
-def test_cache_spans_get_span_description(
-    method_name, args, kwargs, expected_description
-):
-    assert _get_span_description(method_name, args, kwargs) == expected_description
```
szokeasaurusrex, line 1397: I am assuming all the deleted tests were moved to `test_cache_module.py`, correct?

tox.ini
```diff
@@ -94,7 +94,7 @@ envlist =
     {py3.6,py3.9}-django-v{2.2}
     # - Django 3.x
     {py3.6,py3.9}-django-v{3.0}
-    {py3.6,py3.11}-django-v{3.2}
+    {py3.6,py3.9,py3.11}-django-v{3.2}
```
szokeasaurusrex, line 97: Why add `3.9` here?

szokeasaurusrex (commented):
Have some questions and suggestions

tests/integrations/django/test_basic.py
antonpirker: yes, exactly.

tox.ini
antonpirker: to improve code coverages coverage reports. we need a python version where tests for Django 3.0 and Django 3.2 are running, else codecov always tells us that one part of a `if django_version < 3.2` in 
`sentry_sdk/integrations/django/caching.py` is not covered by tests.

sentry_sdk/integrations/django/caching.py
antonpirker: No. For V0 of caches module we ignore cache clusters (at least in Django, not sure about direct redis instrumenation)

sentry_sdk/integrations/django/caching.py
antonpirker: I will keep my longer version, because it is easier to read

sentry_sdk/integrations/django/__init__.py
```diff
@@ -119,19 +130,23 @@ def __init__(
         signals_spans=True,
         cache_spans=False,
         signals_denylist=None,
+        cache_spans_add_item_size=False,
```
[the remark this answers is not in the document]
antonpirker: The main idea was to reduce overhead out of the box. I will do some load tests with and without item_sizes, and if there is no or not much difference I will remove the option, so give the best experience for the user out of the box.

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,138 @@
...
 import sentry_sdk
 from sentry_sdk.consts import OP, SPANDATA
-from sentry_sdk.utils import ensure_integration_enabled
+from sentry_sdk.utils import capture_internal_exceptions, ensure_integration_enabled
 
 
 if TYPE_CHECKING:
     from typing import Any
     from typing import Callable
+    from typing import Tuple
```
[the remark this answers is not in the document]
antonpirker: will do!

sentry_sdk/integrations/django/caching.py
antonpirker: done.

sentry_sdk/integrations/django/caching.py
antonpirker: I do not think so. The only way it would be `"None"` would be if kwarg["key"] would return None (in _get_key()). And this code  is there since a very long time and no user has ever reported that they see `"None"` as the key.

sentry_sdk/integrations/django/__init__.py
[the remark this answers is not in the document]
antonpirker: removed the option, because load testing did not reveal any performance impact

sentry_sdk/integrations/django/caching.py
antonpirker: Yes, probably better a new PR. Can you please create a ticket @szokeasaurusrex ?

sentry_sdk/integrations/django/caching.py
szokeasaurusrex: ah okay, did not realize this code was here before

sentry_sdk/integrations/django/caching.py
szokeasaurusrex: Do we have an issue for adding this feature? If yes, it would be nice if we could link the issue in the `TODO` comment. If there is no issue yet, it might be a good idea to create the issue and link it

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,152 @@
...
+        key = args[0]
+        if isinstance(key, dict):
+            # Do not leak sensitive data
+            # `set_many()` has a dict {"key1": "value1", "key2": "value2"} as first argument.
+            # Those values could include sensitive data so we replace them with a placeholder
+            key = {x: SENSITIVE_DATA_SUBSTITUTE for x in key}
+        return str(key)
     elif kwargs is not None and "key" in kwargs:
-        description += str(kwargs["key"])
+        return str(kwargs["key"])
```
szokeasaurusrex, line 42: Why do we do the sensitive data stripping above but not here?

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,152 @@
...
+
+    return ""
+
 
+def _get_span_description(method_name, args, kwargs):
+    # type: (str, Any, Any) -> str
```
szokeasaurusrex, line 48:
```suggestion
    # type: (str, list[Any], dict[str, Any]) -> str
```

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,152 @@
...
 
     original_method = getattr(cache, method_name)
 
     @ensure_integration_enabled(DjangoIntegration, original_method)
-    def _instrument_call(cache, method_name, original_method, args, kwargs):
-        # type: (CacheHandler, str, Callable[..., Any], Any, Any) -> Any
+    def _instrument_call(
+        cache, method_name, original_method, args, kwargs, address, port
+    ):
+        # type: (CacheHandler, str, Callable[..., Any], Any, Any, Optional[str], Optional[int]) -> Any
```
szokeasaurusrex, line 66:
```suggestion
        # type: (CacheHandler, str, Callable[..., Any], list[Any], dict[str, Any], Optional[str], Optional[int]) -> Any
```

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,152 @@
...
+                if port is not None:
+                    span.set_data(SPANDATA.NETWORK_PEER_PORT, port)
+
+                key = _get_key(args, kwargs)
+                if key != "":
+                    span.set_data(SPANDATA.CACHE_KEY, key)
+
+                item_size = None
+                if is_get_operation:
+                    if value:
```
szokeasaurusrex, line 89:
```suggestion
                    if value is not None:
```
Or, is there a reason why `if value` is more appropriate here?

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,152 @@
...
+                        span.set_data(SPANDATA.CACHE_HIT, True)
+                    else:
+                        span.set_data(SPANDATA.CACHE_HIT, False)
+                else:
+                    try:
+                        # 'set' command
+                        item_size = len(str(args[1]))
+                    except IndexError:
+                        # 'set_many' command
+                        item_size = len(str(args[0]))
```
szokeasaurusrex, line 100: Is there a way to check the command directly?

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,152 @@
...
+    # type: (dict[str, Any]) -> tuple[Optional[str], Optional[int]]
+    address, port = None, None
+    location = settings.get("LOCATION")
+    # TODO: location can also be an array of locations
+    # see: https://docs.djangoproject.com/en/5.0/topics/cache/#redis
+    if isinstance(location, str):
+        if ":" in location:
+            address, port = location.rsplit(":", 1)
+        else:
+            address, port = location, None
```
szokeasaurusrex, line 135: This logic is broken. For example, if `location == "https://example.com"`, then `":" in location` is `True`, so we will end up with `address == "https"` and `port == "//example.com"`.

Perhaps, there is a library that can handle the URL parsing for us? Otherwise, we should use a regular expression here, or at least ensure this case is handled.

Please also add a unit test that ensures that this function handles URLs like https://example.com correctly 🙂

sentry_sdk/integrations/django/caching.py
```diff
@@ -1,80 +1,152 @@
...
+    # remove the username and password from URL to not leak sensitive data.
+    if address is not None and "://" in address:
+        parsed_url = urlparse(address)
+        address = "{}://{}{}".format(
+            parsed_url.scheme or "",
+            parsed_url.netloc or "",
+            parsed_url.path or "",
+        )
+
+    return address, int(port) if port is not None else None
```
szokeasaurusrex, line 147: [nitpick] the logic would be clearer to me if we return early with `None, None` when `location` is not a `str`. However, I am also fine with the original code, if you prefer it

```suggestion
def _get_address_port(settings):
    # type: (dict[str, Any]) -> tuple[Optional[str], Optional[int]]
    location = settings.get("LOCATION")
    if not isinstance(location, str):
        return None, None
    
    if ":" in location:  # TODO: Still need to fix this logic
        address, port = location.rsplit(":", 1)
    else:
        address, port = location, None

    # If address is a URL (could also be a string identifier or path to a Unix socket)
    # remove the username and password from URL to not leak sensitive data.
    if "://" in address:
        parsed_url = urlparse(address)
        address = "{}://{}{}".format(
            parsed_url.scheme or "",
            parsed_url.netloc or "",
            parsed_url.path or "",
        )

    return address, int(port) if port is not None else None
```

sentry_sdk/integrations/redis/__init__.py
```diff
@@ -218,6 +218,11 @@ def sentry_patched_execute_command(self, name, *args, **kwargs):
         if data_should_be_truncated:
             description = description[: integration.max_data_size - len("...")] + "..."
 
+        # TODO: Here we could also create the caching spans.
+        # TODO: also need to check if the `name` (if this is the cache key value) matches the prefix we want to configure in __init__ of the integration
+        #       Questions:
+        #       -) We should probablby have the OP.DB_REDIS span and a separate OP.CACHE_GET_ITEM (or set_item) span, right?
+        #       -) We probably need to research what redis commands are used by caching libs.
```
szokeasaurusrex, line 225: Will these be addressed in this PR? If not, I would suggest linking an issue to each – the questions also should likely only go in the issue, not the comment.

sentry_sdk/integrations/redis/__init__.py
```diff
@@ -358,6 +363,7 @@ class RedisIntegration(Integration):
     def __init__(self, max_data_size=_DEFAULT_MAX_DATA_SIZE):
         # type: (int) -> None
         self.max_data_size = max_data_size
+        # TODO: add some prefix that users can set to specify a cache key
```
szokeasaurusrex, line 366: Same comment regarding `TODO`

sentry_sdk/integrations/redis/asyncio.py
```diff
@@ -59,6 +59,10 @@ async def _sentry_execute_command(self, name, *args, **kwargs):
 
         description = _get_span_description(name, *args)
 
+        # TODO: Here we could also create the caching spans.
+        #       Questions:
+        #       -) We should probablby have the OP.DB_REDIS span and a separate OP.CACHE_GET_ITEM (or set_item) span, right?
+        #       -) We probably need to research what redis commands are used by caching libs.
```
szokeasaurusrex, line 65: Same comment regarding `TODO`

tests/integrations/django/test_cache_module.py
```diff
@@ -0,0 +1,562 @@
...
+        "default": {
+            "BACKEND": "django.core.cache.backends.dummy.DummyCache",
+            "LOCATION": [
+                "redis://127.0.0.1:6379",
+                "redis://127.0.0.2:6378",
+                "redis://127.0.0.3:6377",
+            ],
+        }
+    }
+
```
szokeasaurusrex, line 81: Please also add a test with URL lacking a port, e.g. https://example.com, as requested in a previous comment

szokeasaurusrex (changes requested):
I found some logic that appears to be broken – this should be fixed, and we should also add tests for it.

Besides that, I have also added some more minor suggestions and questions.

sentry_sdk/integrations/django/caching.py
antonpirker: depending on the cache backend and if `get` or `get_many` is called it could be also `[]` `{}`, `""`

sentry_sdk/integrations/redis/__init__.py
antonpirker: done

sentry_sdk/integrations/redis/__init__.py
antonpirker: done.

sentry_sdk/integrations/redis/asyncio.py
antonpirker: done.

sentry_sdk/integrations/django/caching.py
antonpirker: yes, linked

sentry_sdk/integrations/django/caching.py
antonpirker: Good catch! ⭐

sentry_sdk/integrations/django/caching.py
antonpirker: probably. but because `set_many` is more likely a rare edge case (because mostly used when using the low level Django cache api by hand) it is good as it is.

sentry_sdk/integrations/django/caching.py
antonpirker: I refactored the function to always strip sensitive data if it is a dict

tests/integrations/django/test_cache_module.py
antonpirker: done

sentry_sdk/integrations/django/caching.py
antonpirker: done

sentrivana (approved):

Issue #3040, opened 2024-05-02
[Quart] sentry_patched_asgi_app() missing 1 required positional argument: 'send'


KingsleyKelly:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

latest

### Steps to Reproduce

Requirements 
```
aiohttp==3.8.1
annotated-types==0.6.0
anyio==4.2.0
blinker==1.7.0
beautifulsoup4==4.10.0
cachetools==5.3.2
certifi==2023.11.17
charset-normalizer==2.1.1
click==8.1.7
distro==1.9.0
quart
google-ads==21.3.0
google-api-core==2.15.0
google-api-python-client==2.15.0
google-auth==2.25.1
google-auth-httplib2==0.1.0
google-auth-oauthlib==0.4.6
googleapis-common-protos==1.61.0
grpcio==1.59.3
grpcio-status==1.59.3
hypercorn
h11==0.14.0
httpcore==1.0.2
httplib2==0.22.0
httpx==0.26.0
idna==3.6
itsdangerous==2.1.2
Jinja2==3.1.2
MarkupSafe==2.1.3
oauthlib==3.2.2
openai==1.6.1
packaging==23.2
proto-plus==1.22.3
protobuf==4.25.1
psycopg2-binary==2.9.9
pyasn1==0.5.1
pyasn1-modules==0.3.0
pydantic==2.5.3
pydantic_core==2.14.6
pyparsing==3.1.1
pytest==6.2.5
PyYAML==6.0.1
requests==2.31.0
requests-oauthlib==1.3.1
rsa==4.9
six==1.16.0
sniffio==1.3.0
tqdm==4.66.1
typing_extensions==4.9.0
uritemplate==3.0.1
urllib3==2.1.0
Werkzeug==3.0.1
protobuf==4.25.1
cachetools==5.3.2
mock==5.1.0
pytest-env==0.6.2
sentry-sdk[quart]
python-dotenv==0.19.1
```

Note that sentry, hypercorn and quart arent locked.

With the following invocation

```
import sentry_sdk
from sentry_sdk.integrations.quart import QuartIntegration
sentry_sdk.init(
    dsn="https://94e1b1ff4eb4da3636a3e8d55c2c0acf@o4506739227426816.ingest.sentry.io/4506739299254272",
    auto_enabling_integrations=False,
    integrations=[QuartIntegration()],
    # Set traces_sample_rate to 1.0 to capture 100%
[...]

PR #3043, opened 2024-05-03
fix(quart): Fix Quart integration

szokeasaurusrex:
The Quart integration was completely broken prior to this commit, as it caused every request to fail with a 500 error. The reason was that we were using the non-async `ensure_integration_enabled` decorator on the async `sentry_patched_asgi_app` function. This commit fixes the issue by removing the use of that decorator, instead replacing it with a manual check for the integration being enabled.

Fixes GH-3040

### DIFF
--- a/sentry_sdk/integrations/quart.py
+++ b/sentry_sdk/integrations/quart.py
@@ -87,9 +87,11 @@ def patch_asgi_app():
     # type: () -> None
     old_app = Quart.__call__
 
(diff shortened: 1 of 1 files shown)


szokeasaurusrex:
It is very surprising that our test suite did not detect that this integration was completely broken. Perhaps, this is a limitation of the Quart `test_client` we use in testing. We should certainly add tests to make sure we do not break in this way again, but that can likely be done later, most important is to quickly fix the broken integration ASAP

Issue #3038, opened 2024-05-02
Misleading set metric value type hint


elramen:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2

### Steps to Reproduce

1. Write sentry_sdk.metrics.set and hover over set to see that the type hint for value is MetricValue. Go to the definition of MetricValue to see its definition: Union[int, float, str]. 

2. Emit a set metric two times using the same metric name but values that differ in the decimals, e.g., sentry_sdk.metrics.set("test", 1.2)
sentry_sdk.metrics.set("test", 1.6)

### Expected Result

Metric shows unique count of 2 in the sentry.io dashboard.

### Actual Result

Metric shows unique count of 1 in the sentry.io dashboard.

PR #3048, opened 2024-05-06
ref(metrics): Improve type hints for set metrics

elramen:
Remove MetricValue from the set metric's value type hint. MetricValue is a union that includes floatwhich is correct based on how other parts of the code uses MetricValue. However, as floats are intentionally floored for set metrics, the current type hint might be confusing for a user as it looks like floats would work as unique values in a set.

Fixes GH-3038


szokeasaurusrex (approved):
Looks good! But, before merging, please rename the commit to `ref(metrics) ...`

PR #3052, opened 2024-05-07
Added option to disable middleware spans in Starlette

antonpirker:
When `middleware_spans` is set to `False`, no spans will be recorded for Starlette middleware. (analogue to how the DjangoIntegration works)

Docs update: https://github.com/getsentry/sentry-docs/pull/10523

---

## General Notes

#### For maintainers


tests/integrations/starlette/test_starlette.py
```diff
@@ -636,20 +636,51 @@ def test_middleware_spans(sentry_init, capture_events):
...
     for span in transaction_event["spans"]:
-        if span["op"] == "middleware.starlette":
-            assert span["description"] == expected[idx]
-            assert span["tags"]["starlette.middleware_name"] == expected[idx]
+        if span["op"].startswith("middleware.starlette"):
+            assert (
+                span["tags"]["starlette.middleware_name"]
+                == expected_middleware_spans[idx]
+            )
             idx += 1
```
szokeasaurusrex, line 658: I guess this `idx` stuff was already here, but maybe we can consider changing this `for` loop to use `zip` or `enumerate` instead, so we can get rid of it?

sentrivana (approved):
LGTM! Daniel's suggestion makes sense to me too but already looks good as is.

Issue #3050, opened 2024-05-07
Add support for `add_note()` / `__notes__`


arjenzorgdoc:
### Problem Statement

Python 3.11 added support for notes:

```python
async def f4():
    try:
        await f3()
    except TypeError as e:
        e.add_note('this failed while trying to blah')
        raise
```

(Source: https://blog.sentry.io/python-3-11-release-top-5-things-to-know/#adding-notes-to-exceptions)

I don't see them in sentry.

### Solution Brainstorm

Check if an exception has a `__notes__` property and add them to the "message" field.

szokeasaurusrex:
Good idea! Seems like something we should look into supporting. We would also be happy to accept PRs for this!

PR #3056, opened 2024-05-08
Add support for add_note()

arjenzorgdoc:
See issue #3050

This adds support for `__notes__` and adds them to the exception value.

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -713,11 +713,15 @@ def get_errno(exc_value):
 
 def get_error_message(exc_value):
     # type: (Optional[BaseException]) -> str
-    return (
+    value = (
         getattr(exc_value, "message", "")
         or getattr(exc_value, "detail", "")
         or safe_str(exc_value)
     )
+    notes = getattr(exc_value, "__notes__", [])
+    if notes:
+        value = "\n".join([value] + [safe_str(note) for note in notes])
+    return value
 
 
 def single_exception_from_error_tuple(
--- a/tests/test_basics.py
+++ b/tests/test_basics.py
@@ -999,3 +999,47 @@ def test_hub_current_deprecation_warning():
 def test_hub_main_deprecation_warnings():
     with pytest.warns(sentry_sdk.hub.SentryHubDeprecationWarning):
         Hub.main
+
+
+@pytest.mark.skipif(sys.version_info < (3, 11), reason="add_note() not supported")
+def test_notes(sentry_init, capture_events):
+    sentry_init()
+    events = capture_events()
+    try:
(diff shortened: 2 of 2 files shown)


sentry_sdk/utils.py
```diff
@@ -655,11 +655,15 @@ def get_errno(exc_value):
...
     # type: (Optional[BaseException]) -> str
-    return (
+    value = (
         getattr(exc_value, "message", "")
         or getattr(exc_value, "detail", "")
         or safe_str(exc_value)
     )
+    notes = getattr(exc_value, "__notes__", [])
+    if notes:
+        value = "\n".join([value] + [safe_str(note) for note in notes])
```
szokeasaurusrex, line 665: I am unsure whether it makes sense to add the note to the exception message; perhaps setting this as [extra data](https://docs.sentry.io/concepts/key-terms/enrich-data/) would be better

arjenzorgdoc:
The use case for the notes is to add text to the exception massage. 

I think `notes_new()` is the new way to do `notes_old()`:

```python
def failure():
    raise ValueError("Nope.")


MESSAGE = """\
Nope. It's wrong!
See https://example.com/why-nope.html for more info
"""


def notes_old():
    try:
        failure()
    except Exception as e:
        raise RuntimeError(f"{e}\n{MESSAGE}\n") from e


def notes_new():
    try:
        failure()
    except Exception as e:
        e.add_note(MESSAGE)
        raise
```

`notes_old()` renders like this:

image

I tried to have `notes_new()` render the same.

BTW python cli renders the notes after the Exception message as well:

```
  File ".../sentry_notes.py", line 7, in failure
    raise ValueError("Nope.")
ValueError: Nope.
Nope. It's wrong!
See https://example.com/why-nope.html for more info
```

sentry_sdk/utils.py
arjenzorgdoc: I responded in the pull request conversation.

I think the notes add context that developers expect to be shown directly with the error message. That is what CPython does in the console.

szokeasaurusrex: I don't think we need this `safe_str` stuff here, notes should anyways just be strings. This is guaranteed when using `add_note`. Although you correctly point out in your test that it is possible to add non-string values to the `__notes__` list manually, this is likely a misuse of `__notes__`, and is not something we need to support; although, we should still not crash if there is an invalid value in the `__notes__`. 

I would suggest we just filter out any non-string values from the `__notes__`.

szokeasaurusrex:
Even though the PR says it can be edited by maintainers, I am getting a 403 when I try to push to this PR branch. So, I am closing this PR in favor of #3620

PR #3057, opened 2024-05-08
Bring back `last_event_id`

### LINKED ISSUE #3049: No more last_event_id
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.1.0

### Steps to Reproduce

The method last_event_id does not exist anymore. I understand that's been deprecated but, how are we supposed to show the event_id to the user now?
from sentry_sdk import last_event_id

### Expected Result

Be able to get the last event id from a non controlled exception and show it to the final user.
In django for example, the recommended way was this:

```
from sentry_sdk import last_event_id

from django.shortcuts import render


def custom_handler500(request):
    context = {
        "request": request,
        "title": "Error 500",
        "last_event_id": last_event_id(),
    }
    template_name = "500.html"
    return render(request, template_name, context, status=500)
```

### Actual Result

`ImportError: cannot import name 'last_event_id' from 'sentry_sdk' (/app/.venv/lib/python3.11/site-packages/sentry_sdk/__init__.py)`

szokeasaurusrex:
`last_event_id` will return the ID of the most recently captured error event

Closes #3049

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -205,6 +205,16 @@ class BaseClient:
         # type: (*Any, **Any) -> None
         return None
 
+    def last_event_id(self):
+        # type: () -> Optional[str]
+        """
+        .. versionadded:: 2.1.2
+
+        Returns the most recently captured error event's ID. If this client has
+        not captured an error event yet, returns `None`.
+        """
+        return None
+
     def __enter__(self):
         # type: () -> BaseClient
         return self
@@ -378,6 +388,8 @@ class _Client(BaseClient):
                     setup_profiler(self.options)
                 except Exception as e:
                     logger.debug("Can not set up profiler. (%s)", e)
+
+            self._last_event_id = None  # type: Optional[str]
 
         finally:
             _client_init_debug.set(old_debug)
@@ -709,6 +721,7 @@ class _Client(BaseClient):
 
         is_transaction = event_opt.get("type") == "transaction"
         is_checkin = event_opt.get("type") == "check_in"
+        is_error = not is_transaction and not is_checkin
 
         if (
             not is_transaction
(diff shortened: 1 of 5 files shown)


sentry_sdk/client.py
```diff
@@ -750,6 +763,9 @@ def capture_event(
         if self.transport is None:
             return None
 
+        if is_error:
+            self._last_event_id = event_id
```
szokeasaurusrex, line 767: Do we want `last_event_id` only to return the ID of the most recent error event? Or, should we expand it to all events? Error event seems like the better option to align with #3049

sentry_sdk/client.py
```diff
@@ -709,6 +721,7 @@ def capture_event(
 
         is_transaction = event_opt.get("type") == "transaction"
         is_checkin = event_opt.get("type") == "check_in"
+        is_error = not is_transaction and not is_checkin
```
antonpirker, line 724: I guess this would also include profiles? Maybe make a check similar to the one for transaction?

sentry_sdk/client.py
antonpirker: Yes, I think only errors is fine. Because the error ID is important for user feedback form.

szokeasaurusrex:
I am going to rewrite this PR, since we should be setting the `last_event_id` on the isolation scope, rather than the client, to align with the old `last_event_id` implementation which stored the last event ID on the hub.

szokeasaurusrex:
Following a discussion with @lforst, @Lms24, and @HazAT, we will reintroduce this functionality to align as closely as possible with how the `last_event_id` worked in 1.x. We will need to make some changes to align the functionality

szokeasaurusrex:
Closing in favor of #3064

PR #3064, opened 2024-05-10
feat(scope): Add last_event_id to Scope

### LINKED ISSUE #3049: No more last_event_id

szokeasaurusrex:
Fixes #3049


sentry_sdk/scope.py
```diff
@@ -1089,7 +1103,12 @@ def capture_event(self, event, hint=None, scope=None, **scope_kwargs):
         """
         scope = self._merge_scopes(scope, scope_kwargs)
 
-        return Scope.get_client().capture_event(event=event, hint=hint, scope=scope)
+        event_id = Scope.get_client().capture_event(event=event, hint=hint, scope=scope)
+
+        if event_id is not None and event.get("type") != "transaction":
+            self.get_isolation_scope()._last_event_id = event_id
```
andreiborza, line 1109: We were talking in the frontend team about when to best place this and decided to do it at the latest possible point in the sdk to avoid having `last_event_id`s set for events that potentially don't exist anymore (e.g. user dropped events).
Is that something the python sdk could align with?

szokeasaurusrex: We are already aligned with this behavior; `capture_event` would return `None` if the event was dropped.

antonpirker (approved):
nice!

PR #3070, opened 2024-05-14
fix(tracing): Only propagate headers from spans within transactions

### LINKED ISSUE #3068: Dynamic sampling context might be missing from Celery Beat transactions
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

Latest 2.x

### Steps to Reproduce

Update: [See here](https://github.com/szokeasaurusrex/issue-reproductions/tree/main/getsentry/sentry-python/3068) for a simple demonstration of the broken behavior.

Under investigation – if the dynamic sampling context is missing from transactions started in Celery Beat, this would explain strange behavior we observe in Sentry where the parent (Celery Beat) transaction is missing while the child transactions are all present

### Expected Result

Celery Beat transactions contain a dynamic sampling context

### Actual Result

In some cases, the dynamic sampling context might be missing

szokeasaurusrex:
This change ensures that we only propagate trace headers from spans that are within a transaction. This fixes a bug where any child transactions of a span created outside a transaction are missing a dynamic sampling context and are part of a trace missing a root transaction (because the root is the span).

Fixes #3068

### DIFF
--- a/tests/integrations/celery/test_update_celery_task_headers.py
+++ b/tests/integrations/celery/test_update_celery_task_headers.py
@@ -77,33 +77,6 @@ def test_span_with_transaction(sentry_init):
             )
 
 
-def test_span_with_no_transaction(sentry_init):
-    sentry_init(enable_tracing=True)
-    headers = {}
-
-    with sentry_sdk.start_span(op="test_span") as span:
-        updated_headers = _update_celery_task_headers(headers, span, False)
-
-        assert updated_headers["sentry-trace"] == span.to_traceparent()
-        assert updated_headers["headers"]["sentry-trace"] == span.to_traceparent()
-        assert "baggage" not in updated_headers.keys()
-        assert "baggage" not in updated_headers["headers"].keys()
-
-
-def test_custom_span(sentry_init):
-    sentry_init(enable_tracing=True)
-    span = sentry_sdk.tracing.Span()
-    headers = {}
-
-    with sentry_sdk.start_transaction(name="test_transaction"):
-        updated_headers = _update_celery_task_headers(headers, span, False)
-
-        assert updated_headers["sentry-trace"] == span.to_traceparent()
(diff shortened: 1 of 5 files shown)


tests/tracing/test_propagation.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_span_in_span_in_transaction(sentry_init):
+    sentry_init(enable_tracing=True)
+
+    with sentry_sdk.start_transaction(op="test"):
+        with sentry_sdk.start_span(op="test2"):
+            with sentry_sdk.start_span(op="test3") as span_inner:
+                # Ensure the headers are there
+                next(span_inner.iter_headers())
```
szokeasaurusrex, line 40: I put these tests in a separate file because I couldn't find any other file where they would fit better. Do you think they could belong better somewhere else?

antonpirker: new file is fine!

antonpirker (approved):

antonpirker:
This breaks now a lot of tests because they assume `sentry-trace` to be there, when there is no transaction it seems...

tests/integrations/celery/test_update_celery_task_headers.py
```diff
@@ -77,33 +77,6 @@ def test_span_with_transaction(sentry_init):
             )
 
 
-def test_span_with_no_transaction(sentry_init):
```
szokeasaurusrex, line 80: I deleted several tests in this file because they no longer seem to make much sense with the new behavior. Do you agree, or should we rewrite these tests somehow? It looks like we are already doing similar tests with the span enclosed in a transaction

antonpirker: As the coverage of the changed code is still 100% I think it is ok.

antonpirker (approved):
lgtm!

PR #3079, opened 2024-05-16
feat(celery): Queues module producer implementation

### LINKED ISSUE #3078: Queues module: Celery producer instrumentation
Message ID and queue name should be provided on the producer side

szokeasaurusrex:
Depends on #3082 

Fixes #3078

### DIFF
--- a/sentry_sdk/integrations/celery/__init__.py
+++ b/sentry_sdk/integrations/celery/__init__.py
@@ -447,7 +447,11 @@ def _patch_producer_publish():
         # type: (Producer, *Any, **Any) -> Any
         kwargs_headers = kwargs.get("headers", {})
         if not isinstance(kwargs_headers, Mapping):
-            # Ensure kwargs_headers is a Mapping, so we can safely call get()
+            # Ensure kwargs_headers is a Mapping, so we can safely call get().
+            # We don't expect this to happen, but it's better to be safe. Even
+            # if it does happen, only our instrumentation breaks. This line
+            # does not overwrite kwargs["headers"], so the original publish
+            # method will still work.
             kwargs_headers = {}
 
         task_name = kwargs_headers.get("task")
--- a/sentry_sdk/integrations/celery/__init__.py
+++ b/sentry_sdk/integrations/celery/__init__.py
@@ -1,4 +1,5 @@
 import sys
+from collections.abc import Mapping
 from functools import wraps
 
 import sentry_sdk
@@ -47,6 +48,7 @@ try:
         Retry,
         SoftTimeLimitExceeded,
     )
+    from kombu import Producer  # type: ignore
 except ImportError:
     raise DidNotEnable("Celery not installed")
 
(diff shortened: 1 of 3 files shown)


tests/integrations/celery/test_celery.py
```diff
@@ -722,3 +735,27 @@ def task(): ...
...
+    with start_transaction():
+        task.apply_async()
+
+    (event,) = events
+    span = next(span for span in event["spans"] if span["op"] == "queue.publish")
+    assert span["data"]["messaging.system"] == system
+
+    assert span["data"]["messaging.destination.name"] == "celery"
+    assert "messaging.message.id" in span["data"]
+    assert span["data"]["messaging.message.retry.count"] == 0
```
szokeasaurusrex, line 761: I put all of these in a single test, rather than splitting into multiple tests like for the consumer, since it appears I need to fork them in order for my `init_celery_producer_only` fixture to work

edwardgou-sentry (approved):
tested locally, from a product perspective I think this looks good. Thanks!

sentry_sdk/integrations/celery/__init__.py
```diff
@@ -433,3 +436,44 @@ def sentry_workloop(*args, **kwargs):
...
+
+def _patch_producer_publish():
+    # type: () -> None
+    original_publish = Producer.publish
+
+    @ensure_integration_enabled(CeleryIntegration, original_publish)
+    def sentry_publish(self, *args, **kwargs):
+        # type: (Producer, *Any, **Any) -> Any
+        kwargs_headers = kwargs.get("headers", {})
+        if not isinstance(kwargs_headers, Mapping):
```
antonpirker, line 449: If `kwargs_headers` is a list `[("key", "value"), ("k", "v")]` then it would be reset to an empty dict?

I am not sure if this list of tuples could happen, but I do not understand this if.

antonpirker (commented):
Just one question, but otherwise looks very good!

sentry_sdk/integrations/celery/__init__.py
szokeasaurusrex: Basically, what I am trying to do here is to make sure that the `get` operations (line 453 to 455) are successful. 

[`Mapping`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) (full name `collections.abc.Mapping`) is the abstract base class that all [mapping types](https://docs.python.org/3/glossary.html#term-mapping) in Python inherit from. `dict` inherits from `Mapping`, so this `isinstance` is true for any dictionary, but it would also be possible to create a custom `Mapping` type, which provides similar functionality to a `dict`, but which does not inherit from `dict`. This `isinstance` also evaluates to `True` for any of these `dict`-like data structures.

Regarding your example, `isinstance([("key", "value"), ("k", "v")], Mapping)` would evaluate to `False` since `list` does not inherit from `Mapping` (it does not have a `get` method, which all `Mapping` types have). So, we would replace it with an empty dictionary to ensure that the `get` calls don't raise an exception.

antonpirker: now I get it!

antonpirker (approved):

PR #3083, opened 2024-05-17
fix: Add conditional check for delivery_info's existence

cmanallen:
From @JoshFerge "[Throws errors] millions of times a minute".  If its null we can't do anything with it so let's check first.


JoshFerge (approved):
🔥

antonpirker (approved):

PR #3102, opened 2024-05-23
feat: Add getter methods for Scope class

iKlotho:
The getter methods would allow users to check if a specific tag/context/extra is already set or not. I have a use case like this and I don't want to access the protected members of a class. This PR adds getter methods for the following variables: `_tags`, `_contexts`, and `_extras`. If adding getter methods for _contexts and _extras doesn't make sense, I can remove them; however, I think they will be useful.

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -849,14 +849,14 @@ class Scope(object):
         self._tags.pop(key, None)
 
     def get_tag(self, key):
-        # type: (str) -> Any
+        # type: (str) -> (Any | None)
         """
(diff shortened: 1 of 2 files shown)


szokeasaurusrex:
@iKlotho I would prefer to find an alternative solution to your problem, rather than add new public APIs to the `Scope` that we would need to maintain.

Can you please explain exactly why you need these getter methods?

szokeasaurusrex:
Closing issue due to lack of response to my [previous question](https://github.com/getsentry/sentry-python/pull/3102#issuecomment-2200617046).

PR #3114, opened 2024-05-30
Fix last_event_id() in nested scopes

### LINKED ISSUE #3113: last_event_id() crash: AttributeError: 'Scope' object has no attribute '_last_event_id'.
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

master

### Steps to Reproduce

Run this file:

```python
import sentry_sdk

sentry_sdk.init()

with sentry_sdk.push_scope() as scope:
    sentry_sdk.last_event_id()
```

### Expected Result

No crash.

### Actual Result

Crash:

```
$ python example.py
  File "/.../example.py", line 6, in <module>
    sentry_sdk.last_event_id()
  File "/.../sentry_sdk/api.py", line 343, in last_event_id
    return Scope.last_event_id()
           ^^^^^^^^^^^^^^^^^^^^^
  File "/.../sentry_sdk/scope.py", line 330, in last_event_id
    return cls.get_isolation_scope()._last_event_id
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Scope' object has no attribute '_last_event_id'. Did you mean: 'last_event_id'?
```

#3064 added `last_event_id()` and `_last_event_id` but I think it missed copying `_last_event_id` in `Scope.__copy__`.

adamchainz:
Fixes #3113.

I added two tests, one that generally checks `Scope.__copy__()` copies everything that it should, and one top-level one that reproduces the bug seen. On `master`, both fail with:

```
AttributeError: 'Scope' object has no attribute '_last_event_id'. Did you mean: 'last_event_id'?
```

@szokeasaurusrex please take a look :)

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -208,9 +208,6 @@ class Scope(object):
         incoming_trace_information = self._load_trace_data_from_env()
         self.generate_propagation_context(incoming_data=incoming_trace_information)
 
-        # self._last_event_id is only applicable to isolation scopes
-        self._last_event_id = None  # type: Optional[str]
-
     def __copy__(self):
         # type: () -> Scope
         """
@@ -243,6 +240,8 @@ class Scope(object):
         rv._attachments = list(self._attachments)
 
         rv._profile = self._profile
+
+        rv._last_event_id = self._last_event_id
 
         return rv
 
@@ -676,6 +675,9 @@ class Scope(object):
         self._profile = None  # type: Optional[Profile]
 
         self._propagation_context = None
+
+        # self._last_event_id is only applicable to isolation scopes
+        self._last_event_id = None  # type: Optional[str]
 
     @_attr_setter
     def level(self, value):
--- a/tests/test_basics.py
+++ b/tests/test_basics.py
@@ -800,3 +800,11 @@ def test_last_event_id_transaction(sentry_init):
         pass
 
     assert last_event_id() is None, "Transaction should not set last_event_id"
+
+
+def test_last_event_id_scope(sentry_init):
+    sentry_init(enable_tracing=True)
+
+    # Should not crash
+    with push_scope() as scope:
+        assert scope.last_event_id() is None
--- a/tests/test_scope.py
+++ b/tests/test_scope.py
@@ -25,6 +25,10 @@ def test_copying():
     s1.set_tag("foo", "bar")
 
     s2 = copy.copy(s1)
+    # Check all attributes are copied
+    for name in set(Scope.__slots__) - {"client"}:
+        getattr(s2, name)
+
     assert "foo" in s2._tags
 
     s1.set_tag("bam", "baz")


sentry_sdk/scope.py
```diff
@@ -208,9 +208,6 @@ def __init__(self, ty=None, client=None):
         incoming_trace_information = self._load_trace_data_from_env()
         self.generate_propagation_context(incoming_data=incoming_trace_information)
 
-        # self._last_event_id is only applicable to isolation scopes
-        self._last_event_id = None  # type: Optional[str]
-
```
szokeasaurusrex, line 213: While I agree that it likely makes sense to move these lines to the `clear` function, I believe this change should be considered in a separate PR, since it does not appear to be needed to fix #3113; the tests you added pass even if I undo this change.

sentry_sdk/scope.py
```diff
@@ -677,6 +676,9 @@ def clear(self):
 
         self._propagation_context = None
 
+        # self._last_event_id is only applicable to isolation scopes
+        self._last_event_id = None  # type: Optional[str]
+
```
szokeasaurusrex, line 681: Likely belongs in separate PR, see previous comment.

tests/test_scope.py
```diff
@@ -25,6 +25,10 @@ def test_copying():
     s1.set_tag("foo", "bar")
 
     s2 = copy.copy(s1)
+    # Check all attributes are copied
+    for name in set(Scope.__slots__) - {"client"}:
+        getattr(s2, name)
+
```
szokeasaurusrex, line 31: I would prefer to add this check to a separate unit test, since it is somewhat unrelated to the tag copying checks that this existing test is checking. The comment should also be updated to something like "Check that all attributes, except `client`, are present on the new scope", as the check only verifies the attributes' presence, and does not check whether the values were actually copied.

sl0thentr0py:
@antonpirker assigning this to you since I'm not sure what the desired semantics here should be (since it only should be on the isolation scope and so on..)

tests/test_basics.py
```diff
@@ -800,3 +800,11 @@ def test_last_event_id_transaction(sentry_init):
         pass
 
     assert last_event_id() is None, "Transaction should not set last_event_id"
+
+
+def test_last_event_id_scope(sentry_init):
+    sentry_init(enable_tracing=True)
+
+    # Should not crash
+    with push_scope() as scope:
```
szokeasaurusrex, line 809:
```suggestion
    with isolation_scope() as scope:
```

`push_scope` is [deprecated](https://github.com/getsentry/sentry-python/blob/master/CHANGELOG.md#deprecated) since version 2.0.0, so let's use the new `isolation_scope` function here, instead. Using `isolation_scope` also fails on master but passes on this branch.

szokeasaurusrex (commented):
Thank you for the contribution! I have requested a few changes, please see my comments.

szokeasaurusrex:
@sl0thentr0py, removing @antonpirker's assignment since I have already reviewed – I wrote the original Scope-based `last_event_id` implementation (#3064), so I am familiar with this code

szokeasaurusrex:
Also, I split your changes to clear `_last_event_id` in the `clear` method into #3124

adamchainz:
Thank you for the fix. I didn’t have to look at this because I wasn’t working on the related contract.

PR #3124, opened 2024-06-03
fix(scope): Clear last_event_id on scope clear

szokeasaurusrex:
This PR includes a change that was split off from #3114.


antonpirker (approved):
nice

Issue #3118, opened 2024-06-02
Fix Migration Guide regarding removal of last_event_id()


lovetox:
### How do you use Sentry?

Self-hosted/on-premise

### Version

2.2.0

### Steps to Reproduce

Migration Guide says, method last_event_id() was removed, but it was reintroduced in 2.2.0

### Expected Result

Correct Migration Guide

### Actual Result

> The API function last_event_id() was removed

szokeasaurusrex:
Good catch, @lovetox – I already opened a PR to fix the migration guide as you have suggested

PR #3126, opened 2024-06-03
docs: Update `last_event_id` in migration guide

szokeasaurusrex:
Update the migration guide to mention that `last_event_id` has been reintroduced.

Fixes GH-3118


sl0thentr0py:
if we brought it back I'd just not mention it anymore to reduce confusion

szokeasaurusrex:
Fair enough @sl0thentr0py – my thought was what if someone tried to upgrade to `2.0.0`, which does not have `last_event_id` – in that case, this mention could be helpful. But, it is probably unlikely that someone tries to do that, and if they do, I suppose they can just read the changelog

sl0thentr0py (approved):

PR #3156, opened 2024-06-11
ref(metrics): Update type hints for tag values

elramen:
Remove `Tuple` and `List` from the `MetricTagValue` type as these are not supported and might confuse the user. See https://github.com/getsentry/relay/issues/3691 for more information.


antonpirker (approved):

szokeasaurusrex (approved):
Looks good! Do we need to update any Python SDK docs in response to this change?

PR #3160, opened 2024-06-12
feat(transport): Use env vars for default CA cert bundle location

DragoonAethis:
Many libraries use the SSL_CERT_FILE environment variable to point at a CA bundle to use for HTTPS certificate verification. This is often used in corporate environments with internal CAs or HTTPS hijacking proxies, where the Sentry server presents a certificate not signed by one of the CAs bundled with Certifi. Additionally, Requests, Python's most popular HTTP client library, uses the REQUESTS_CA_BUNDLE variable instead.

Use the SSL_CERT_FILE or REQUESTS_CA_BUNDLE vars if present to set the default CA bundle.

---

One thing I'm not sure about is how to test this - this would require setting up a custom self-signed CA, generate bundle + self-signed certs for the devserver, run tests with env vars against it...


sl0thentr0py (commented):
you're right that testing this is tricky but we can just merge, it's mostly a harmless change.

One remark:

I assume there's no way to keep the precedence order to have `certifi.where()` first because it returns a string, this would be better to ensure backwards compat on weird systems that might have both.

DragoonAethis:
> I assume there's no way to keep the precedence order to have `certifi.where()` first because it returns a string, this would be better to ensure backwards compat on weird systems that might have both.

`certifi.where()` always returns a valid string because certifi ships a fallback certificate bundle with the package. The alternative is to set `ca_cert_dir` which may be a directory that contains multiple loose CA certs, but then some preprocessing is required (building a dir with certifi + provided CAs in a temporary directory).

sl0thentr0py (approved):
alright let's ship it

DragoonAethis:
Thanks!

PR #3135, opened 2024-06-05
Add `origin` to transactions/spans created by integrations

antonpirker:
Add the span `origin` to all our integrations. 

Refs #2244
Refs https://github.com/getsentry/team-sdks/issues/14

PR #3163, opened 2024-06-12
Add `origin` to transactions/spans created by integrations 2

antonpirker:
Add the span origin to all our integrations.

Same as https://github.com/getsentry/sentry-python/pull/3135/ but for more integrations. 
The other PR is already getting too big.


sentrivana (approved):
🚀

Issue #3142, opened 2024-06-07
Django ASGI based async views do not create Transactions


antonpirker:
If there is an async view in Django then a transaction should be created for that view. 

It seems that this is not the case right now: https://github.com/getsentry/sentry-python/blob/antonpirker/trace-origin-in-integrations/tests/integrations/django/asgi/test_asgi.py#L340

The view triggered in the linked test above should create a transaction.

There was a change in Django that we missed: https://github.com/getsentry/sentry-python/pull/3141

We need to make sure that async views in Django emit transactions:
- when run with Django channels (`channels.http.AsgiHandler` is monkey patched to start transactions)
- when running with Djangos own ASGI handler (`django.core.handlers.asgi.ASGIHandler` only available in Django 3.0+)

Make sure this works in old Django (<3.0) and also newer Djangos. When checking newer Djangos make sure it works when using Channels or Djangos built in ASGI handler.

szokeasaurusrex:
Looks like the problem is only for sync views on async applications

PR #3164, opened 2024-06-12
breakpoints/debug printing for Django ASGI investigation

szokeasaurusrex:
The problem (see #3142) is as follows: We expect there to be three events in the test, but right now (in master), the test verifies only that we have two events, with the transaction missing. So, we want to find out what happens with the transaction. 

The lack of transaction seems unique to the testing environment – when I [tried to reproduce](https://github.com/szokeasaurusrex/issue-reproductions/tree/main/getsentry/sentry-python/3142) with a similar setup outside of a testing environment, I saw transactions as expected in Sentry. Therefore, it appears that either no transaction is created during testing, or we somehow fail to capture the transaction in the TestTransport that we use for the `capture_events` fixture.

I have tried placing breakpoints in many different places, but have still struggled to figure out the root cause for the strange behavior. In this PR, I have included two breakpoints that cause unexpected behavior. The first breakpoint we have is one in the `__exit__` of a span; if the code is run as is with `tox -e py3.12-django-latest -- tests/integrations/django/asgi/test_asgi.py::test_trace_from_headers_if_performance_enabled`, we hit that breakpoint three times (perhaps, because we have spans as well as a transaction). Then, we will end up inside the second breakpoint in `capture_event`. Since this breakpoint is inside an `if is_transaction` block, we know that we are about to capture the transaction. If we step through the transport `capture_envelope` call, we see that we correctly add the transaction event to the events list. However, when we continue from this last breakpoint, the test fails with an asyncio `TimeoutError`. 

If we comment out the breakpoint in `Span.__exit__`, however, we do not hit the breakpoint in `capture_event` at all, suggesting that for whatever reason the transaction does not get captured in this case. We end up failing with the assertion error, since there are only two, rather than three, events in the test. No asyncio TimeoutError in this case.

I am not really sure yet what to make of all this strange behavior, but I guess asyncio somehow prevents the transaction from being captured in our tests

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -763,6 +763,9 @@ class _Client(BaseClient):
         if self.transport is None:
             return None
 
+        if is_transaction:
+            breakpoint()
+
         self.transport.capture_envelope(envelope)
 
         return event_id
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -310,6 +310,7 @@ class Span:
 
     def __exit__(self, ty, value, tb):
         # type: (Optional[Any], Optional[Any], Optional[Any]) -> None
+        breakpoint()
         if value is not None:
             self.set_status("internal_error")
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -210,6 +210,7 @@ def capture_events(monkeypatch):
         def append_event(envelope):
             for item in envelope:
                 if item.headers.get("type") in ("event", "transaction"):
+                    print(f"append_event: {item}")
                     events.append(item.payload.json)
             return old_capture_envelope(envelope)
 
--- a/tests/integrations/django/asgi/test_asgi.py
+++ b/tests/integrations/django/asgi/test_asgi.py
@@ -317,7 +317,7 @@ async def test_has_trace_if_performance_disabled(sentry_init, capture_events):
 
 
 @pytest.mark.asyncio
-@pytest.mark.forked
+# @pytest.mark.forked
 @pytest.mark.skipif(
     django.VERSION < (3, 1), reason="async views have been introduced in Django 3.1"
 )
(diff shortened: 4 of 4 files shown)


szokeasaurusrex:
I noticed the following in the test failure logs when running with both breakpoints commented out:

```
Task was destroyed but it is pending!
task: <Task cancelling name='Task-2' coro=<patch_django_asgi_handler_impl.<locals>.sentry_patched_asgi_handler() running at /Users/dszoke/Development/SDK/sentry-python/sentry_sdk/integrations/django/asgi.py:101> wait_for=<Task cancelling name='Task-4' coro=<ASGIHandler.listen_for_disconnect() running at /Users/dszoke/Development/SDK/sentry-python/.tox/py3.12-django-latest/lib/python3.12/site-packages/django/core/handlers/asgi.py:243> wait_for=<Future cancelled> cb=[Task.task_wakeup()]>>
Exception ignored in: <coroutine object patch_django_asgi_handler_impl.<locals>.sentry_patched_asgi_handler at 0x10454c150>
  File "/Users/dszoke/Development/SDK/sentry-python/sentry_sdk/integrations/django/asgi.py", line 101, in sentry_patched_asgi_handler
  File "/Users/dszoke/Development/SDK/sentry-python/sentry_sdk/integrations/asgi.py", line 144, in _run_asgi3
  File "/Users/dszoke/Development/SDK/sentry-python/sentry_sdk/integrations/asgi.py", line 163, in _run_app
  File "/Users/dszoke/.pyenv/versions/3.12.0/lib/python3.12/contextlib.py", line 155, in __exit__
  File "/Users/dszoke/Development/SDK/sentry-python/sentry_sdk/scope.py", line 1627, in isolation_scope
ValueError: <Token var=<ContextVar name='current_scope' default=None at 0x10237f4c0> at 0x104406240> was created in a different Context
Task was destroyed but it is pending!
task: <Task cancelling name='Task-4' coro=<ASGIHandler.listen_for_disconnect() running at /Users/dszoke/Development/SDK/sentry-python/.tox/py3.12-django-latest/lib/python3.12/site-packages/django/core/handlers/asgi.py:243> wait_for=<Future cancelled> cb=[Task.task_wakeup()]>
```

So, I guess the asyncio task is somehow getting cancelled before we can capture the transaction.

antonpirker:
Your guess was right @szokeasaurusrex the problem was simply that we need to `wait()` for the ASGI application to be finished to be able to check for transactions (otherwise the task is killed before the transaction could be sent out)

See: https://github.com/getsentry/sentry-python/pull/3180

PR #3180, opened 2024-06-18
Cleaning up ASGI tests for Django

### LINKED ISSUE #3142: Django ASGI based async views do not create Transactions

antonpirker:
Cleaning up the ASGI tests for Django. Making sure it is always `wait()`ed for the application to finish and also made the tests a bit more readable and removed some useless asserts.

Fixes #3142

### DIFF
--- a/tests/integrations/django/asgi/test_asgi.py
+++ b/tests/integrations/django/asgi/test_asgi.py
@@ -31,12 +31,17 @@ if django.VERSION >= (3, 0):
 @pytest.mark.asyncio
 @pytest.mark.forked
 async def test_basic(sentry_init, capture_events, application):
-    sentry_init(integrations=[DjangoIntegration()], send_default_pii=True)
+    sentry_init(
+        integrations=[DjangoIntegration()],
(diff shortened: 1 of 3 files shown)


sentrivana (approved):
Nice! Good cleanup too.

PR #3182, opened 2024-06-18
feat(pymongo): Add MongoDB collection span tag

0Calories:
Adds the MongoDB collection as a tag on pymongo query spans. The semantics are set to match what is provided by OpenTelemetry: https://opentelemetry.io/docs/specs/semconv/database/mongodb/


antonpirker (approved):
Looks good!

PR #3186, opened 2024-06-19
ref(pymongo): Change span operation from `db.query` to `db`

0Calories:
The pymongo integration is the only place where span ops were being set to `db.query`, which is inconsistent with other SDKs like the Node SDK.


antonpirker (approved):
Nice! Thanks for the improvement!

antonpirker:
You can merge this @0Calories ! Thanks!

0Calories:
@antonpirker Thanks! But it seems that I don't have permission to merge this in, all required checks are met and there's no option to merge 😕

antonpirker:
Oh ok, I will merge it.

PR #3189, opened 2024-06-20
ref(pymongo): Remove redundant command name in query description

0Calories:
The query command is already included as the first key within the command JSON, so query spans would end up looking like this:

image

Our Node SDK, which uses OTel under the hood now, formats queries differently; it doesn't include the command name at the front, example:

image

We also already include the command within the `db.operation` tag, so there's no confusion or loss of details here. This change will ensure consistent behaviour between our other SDKs!

### DIFF
--- a/sentry_sdk/integrations/pymongo.py
+++ b/sentry_sdk/integrations/pymongo.py
@@ -155,7 +155,7 @@ class CommandTracer(monitoring.CommandListener):
             if not should_send_default_pii():
                 command = _strip_pii(command)
 
-            query = "{} {}".format(event.command_name, command)
+            query = "{}".format(command)
             span = sentry_sdk.start_span(op=op, description=query)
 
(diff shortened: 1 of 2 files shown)


antonpirker:
But then when I look at the span tree, I do not know which command was run, right? So I always have to click into the span and find `db.operation` to know what command was used. Not sure if this is the best experience

antonpirker:
But if otel does it this way, then we probably need to update the span tree, to display `db.operation`.

antonpirker (approved):

antonpirker:
But this is something that "future us" need to do :-)

0Calories:
@antonpirker Thanks for the review! You won't always have to click on the span to see the operation, since in most cases, the operation is the first key in the JSON, which you can see in the span description. It seems like the operation isn't included only when it's an insert

PR #3192, opened 2024-06-21
fix(llm-monitoring): If there is an internal error, still return a value

colin-sentry:
There might be cases where there is an internal error and the function would return None


antonpirker (approved):
Nice!

PR #3201, opened 2024-06-25
Removed Hub from some integrations that still use it.

antonpirker:
Remove Hub usage in integrations.

Removing the hub in our tests is done in this PR:
https://github.com/getsentry/sentry-python/pull/3197


sentrivana (approved):
Nice!

PR #3206, opened 2024-06-27
build: Update tornado version in setup.py to match code check.

aclemons:
c06bf06a set the minimum version to 6, but setup.py was not updated to match.


sentrivana (approved):
Thank you @aclemons!

PR #3223, opened 2024-06-28
Implement new POTel span processor

sl0thentr0py:
* only acts on `on_end` instead of both `on_start/on_end` as before
* store children spans in a dict mapping `span_id -> children`
* new dict only stores otel span objects and no sentry transaction/span objects so we save a bit of useless memory allocation
* I'm not using our current `Transaction/Span` classes at all to build the event because when we add our APIs later, we'll need to rip these out and we also avoid having to deal with the `instrumenter` problem
* if we get a root span (without parent), we recursively walk the dict and find the children and package up the transaction event and send it 
  * I didn't do it like JS because I think this way is better
  *  they [group an array of `finished_spans`](https://github.com/getsentry/sentry-javascript/blob/7e298036a21a5658f3eb9ba184165178c48d7ef8/packages/opentelemetry/src/spanExporter.ts#L132) every time a root span ends and I think this uses more cpu than what I did
  * and the dict like I used it doesn't take more space than the array either
* if we get a span with a parent we just update the dict to find the span later
* moved the common `is_sentry_span` logic to utils


sl0thentr0py:
tested simple nested trace 

```python
import sentry_sdk
from opentelemetry import trace
from time import sleep

sentry_sdk.init(
    debug=True,
    traces_sample_rate=1.0,
    _experiments={"otel_powered_performance": True},
)

tracer = trace.get_tracer(__name__)

with tracer.start_as_current_span("request"):
    sleep(0.1)
    with tracer.start_as_current_span("db"):
        sleep(0.5)
        with tracer.start_as_current_span("redis"):
            sleep(0.2)
    with tracer.start_as_current_span("http"):
        sleep(1)
```

--- 

image

sentry_sdk/integrations/opentelemetry/potel_span_processor.py
```diff
@@ -22,15 +34,23 @@ def __new__(cls):
 
     def __init__(self):
         # type: () -> None
-        pass
+        self._children_spans = {}  # type: dict[int, List[ReadableSpan]]
```
sentrivana, line 37: Could we do `defaultdict(list)` here and then just append at l50? Personal preference but I find the api much more intuitive than `setdefault`

sentry_sdk/integrations/opentelemetry/potel_span_processor.py
```diff
@@ -42,3 +62,113 @@ def shutdown(self):
...
+            node_children = self._children_spans.pop(parent_span_id, [])
+            children.extend(node_children)
+            bfs_queue.extend(
+                [child.context.span_id for child in node_children if child.context]
+            )
+
+        return children
+
+    # we construct the event from scratch here
+    # and not use the current Transaction class for easier refactoring
```
sentrivana, line 102: This is cool, haven't thought of completely bypassing this. I'll have to think about the implications for the granular instrumenter if we isolate OTel and our instrumentation like this.

sentrivana (commented):
Cool ✨

sentry_sdk/integrations/opentelemetry/potel_span_processor.py
```diff
@@ -22,15 +34,23 @@ def __new__(cls):
...
-        pass
+        if is_sentry_span(span):
+            return
+
+        # TODO-neel-potel-remote only take parent if not remote
+        if span.parent:
+            self._children_spans.setdefault(span.parent.span_id, []).append(span)
+        else:
+            # if have a root span ending, we build a transaction and send it
+            self._flush_root_span(span)
```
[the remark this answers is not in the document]
sentrivana: Another option: Also use `on_start` to create some record of the span and then in `on_end`, if a transaction comes and we detect that it has unfinished child spans, wait for a grace period to give them time to finish?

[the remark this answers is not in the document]
sl0thentr0py: some heuristic cleanup logic will be done in a follow up PR yes, I haven't thought through exactly how we'll do it but JS just [has a cutoff logic](https://github.com/getsentry/sentry-javascript/blob/7e298036a21a5658f3eb9ba184165178c48d7ef8/packages/opentelemetry/src/spanExporter.ts#L177-L180) of [5 minutes](https://github.com/getsentry/sentry-javascript/blob/7e298036a21a5658f3eb9ba184165178c48d7ef8/packages/opentelemetry/src/spanExporter.ts#L36)

sentry_sdk/integrations/opentelemetry/potel_span_processor.py
sl0thentr0py: so for now I'm just creating a raw dict, but I'm thinking eventually of a `TypedDict` for `Span` too and the current `Span/Transaction` classes will basically be completely removed.

sl0thentr0py:
@sentrivana moved op/description/status logic to utils, but status mapping will be done separately. Maybe @szokeasaurusrex can look into making this logic as similar to JS as possible, but for now this is good enough for this PR.

https://github.com/getsentry/sentry-javascript/blob/master/packages/opentelemetry/src/utils/mapStatus.ts
https://github.com/getsentry/sentry-javascript/blob/master/packages/opentelemetry/src/utils/parseSpanDescription.ts

PR #3237, opened 2024-07-03
build: Remove `ipdb` from test requirements

rominf:
[ipdb](https://pypi.org/project/ipdb) is not used by testing suite. To avoid installing extra dependencies, remove it from requirements file. Developers who find ipdb helpful can install the package themselves.

The change was initially driven by the fact that `ipdb` was not installable on Fedora 41 because of its missing Python 3.13 support, which made the Sentry SDK build fail due to unsatisfied build requirements.

---

## General Notes

#### For maintainers

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -10,5 +10,4 @@ executing
 asttokens
 responses
 pysocks
-ipdb
 setuptools


requirements-devenv.txt
```diff
@@ -1,5 +1,6 @@
 -r requirements-linting.txt
 -r requirements-testing.txt
+ipdb
```
szokeasaurusrex, line 3: If we, in fact, don't use `ipdb` anywhere, then I don't think it makes sense to list it as a dependency here, either. Developers who find `ipdb` helpful can install the package themselves.

```suggestion
```

szokeasaurusrex:
Thank you for the contribution! I would suggest that we simply remove the `ipdb` dependency completely

rominf:
@szokeasaurusrex I agree! Done.

szokeasaurusrex (approved):
Thanks for making the change, looks good!

rominf:
@szokeasaurusrex I noticed your change in the PR title. I agree, this looks better. I can change commit message accordingly. Should I or you do not want extra CI runs?

szokeasaurusrex:
@rominf We usually use "Squash and merge" to merge PRs to `master`, and I am able to edit the commit message used in the merge. I was already planning to update the commit message accordingly, as that is the easiest way to fix the commit message (no CI rerun needed). 

But, thanks for checking!

PR #3240, opened 2024-07-03
ref(tests): Unhardcode integration list

rominf:
Benefits of unhardcoding integration list and disabling auto integrations:
1. It becomes possible to successfully run tests in environments where certain extra auto integrations get enabled.
2. There is no need to update hardcoded list when new default integrations are introduced.

The origin of this change is a patch that I had to apply while packaging for Fedora: https://src.fedoraproject.org/rpms/python-sentry-sdk/blob/30f00540cd30328d578dfb84625226dcc7f28969/f/python-sentry-sdk.spec#_52 (https://src.fedoraproject.org/rpms/python-sentry-sdk/blob/30f00540cd30328d578dfb84625226dcc7f28969/f/0002-tests.test_new_scopes_compat_event-compatibility.patch). The patch was necessary because all possible optional dependencies for optional integrations are pre-installed as system packages, so that as many tests as possible could be run. Unfortunately, the patch no longer applies to a new version of the Sentry SDK, so instead of redoing the patch in Fedora, I am submitting this PR.

---

## General Notes

#### For maintainers

### DIFF
--- a/tests/new_scopes_compat/test_new_scopes_compat_event.py
+++ b/tests/new_scopes_compat/test_new_scopes_compat_event.py
@@ -4,6 +4,7 @@ from unittest import mock
 
 import sentry_sdk
 from sentry_sdk.hub import Hub
+from sentry_sdk.integrations import iter_default_integrations
 from sentry_sdk.scrubber import EventScrubber, DEFAULT_DENYLIST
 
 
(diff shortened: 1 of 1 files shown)


szokeasaurusrex (approved):
Looks good, appreciate the contribution!

rominf:
GitHub told to update the branch.

image

I rebased it on `master`.

rominf:
@szokeasaurusrex It happened again:
image

What should I do?

szokeasaurusrex:
@rominf I can handle getting the PR merged from here

szokeasaurusrex (approved):

PR #3244, opened 2024-07-03
feat(tracing): Record lost spans in client reports

### LINKED ISSUE #3248: Improve `record_lost_event` parameter documentation
`data_category` and `item` are mutually exclusive; only one should be provided at a time. 

https://github.com/getsentry/sentry-python/blob/31efa62c90e5b88c6c15b55f6908a25133d65958/sentry_sdk/transport.py#L133-L142

szokeasaurusrex:
Resolves https://github.com/getsentry/sentry-python/issues/3229
Resolves #3248

**Note to reviewers:** I have attempted to split the commits up into logical units, so you may wish to review the changes commit-by-commit.

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -119,10 +119,8 @@ if TYPE_CHECKING:
         },
     )
 
-
 BAGGAGE_HEADER_NAME = "baggage"
 SENTRY_TRACE_HEADER_NAME = "sentry-trace"
-
 
 # Transaction source
 # see https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations
@@ -858,6 +856,8 @@ class Transaction(Span):
 
                 client.transport.record_lost_event(reason, data_category="transaction")
 
+                # Only one span (the transaction itself) is discarded, since we did not record any spans here.
+                client.transport.record_lost_event(reason, data_category="span")
             return None
 
         if not self.name:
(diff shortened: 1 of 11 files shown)


szokeasaurusrex:
Marked as draft, still need to test this

szokeasaurusrex:
> We need to keep track of the amount of spans prior before_send_transaction and any event processors and calculate the delta afterwards, so we can report eventually dropped spans as well.

@cleptric, does this mean that we also need to report dropped spans in the `before_send_transaction` even if the transaction gets sent to Sentry?

For example, in the situation below, do we report any dropped spans to Sentry?

Transaction diagram


And for clarity, in the following situation where the transaction gets dropped after `before_send`/`before_send_transaction`, e.g. in the `Transport`, how many spans do we report as dropped, 4 or 6?

Transaction diagram-2

cleptric:
The first example reports two dropped spans, the second example reports six dropped spans and one dropped transaction.

sentry_sdk/tracing.py
```diff
@@ -856,7 +854,8 @@ def finish(self, hub=None, end_timestamp=None):
                 else:
                     reason = "sample_rate"
 
-                client.transport.record_lost_event(reason, data_category="transaction")
+                # No spans are discarded here because they were never recorded in the first place
+                client.transport.record_lost_transaction(reason, span_count=0)
```
szokeasaurusrex, line 858: Just realized I still need to test this codepath

PR #3253, opened 2024-07-08
test: Fix non-idempotent test

szokeasaurusrex:
Fix `tests/test_basic.py::test_event_processor_drop_records_client_report` so that the test is idempotent on failure. Previously, the test was only idempotent on success; if the test failed, it would cause many other unrelated tests to fail with it.


antonpirker (approved):
Nice!

Issue #3256, opened 2024-07-09
Load Balancer with 2-way SSL authentication support


grammy-jiang:
### Problem Statement

Hi,

I am configuring Nginx as load balancer in front of Sentry and 2-way SSL authentication is enabled.

I can visit the web GUI as usual after I add the client cert and key to the browser. But the server can't receive the log messages, and I notice that in sentry-python it doesn't support two-way SSL authentication as a client - from the source code no client cert related options are passed to the urllib3.PoolManager:

* [sentry-python/sentry_sdk/transport.py at 69ecd87aa4539de03754af5afb4af4be53efd260 · getsentry/sentry-python](https://github.com/getsentry/sentry-python/blob/69ecd87aa4539de03754af5afb4af4be53efd260/sentry_sdk/transport.py#L542)
* [sentry-python/sentry_sdk/transport.py at 69ecd87aa4539de03754af5afb4af4be53efd260 · getsentry/sentry-python](https://github.com/getsentry/sentry-python/blob/69ecd87aa4539de03754af5afb4af4be53efd260/sentry_sdk/transport.py#L453)

The document of urllib3.PoolManager:

* [Advanced Usage - urllib3 1.26.3 documentation](https://urllib3.readthedocs.io/en/1.26.3/advanced-usage.html#client-certificates)

### Solution Brainstorm

From these lines of options defined, I think it can be configured in the same way:

* [sentry-python/sentry_sdk/transport.py at 69ecd87aa4539de03754af5afb4af4be53efd260 · getsentry/sentry-python](https://github.com/getsentry/sentry-python/blob/69ecd87aa4539de03754af5afb4af4be53efd260/sentry_sdk/transport.py#L477-L482)

   ```python
   options["ca_certs"] = (
       ca_certs  # User-provided bundle from the SDK init
       or os.environ.get("SSL_CERT_FILE")
       or os.environ.get("REQUESTS_CA_BUNDLE")
       or certifi.where()
   )
   ```

   Example:

   ```python
   options["cert_file"] = (
       cert_file  # User-provided bundle from the SDK init
       or os.environ.get("CLIENT_CERT_FILE")
   )
   options["key_file"] = (
       key_file  # User-provided bundle from the SDK init
       or os.environ.get("CLIENT_KEY_FILE")
   )
   ```

szokeasaurusrex:
Hi @grammy-jiang, thanks for the feature suggestion! This is something we can definitely look into adding.

[...]

PR #3258, opened 2024-07-09
Add the client cert and key support to HttpTransport

grammy-jiang:
Enable 2-way SSL authentication support in sentry-sdk #3256 

The document of [self-hosted Sentry](https://develop.sentry.dev/self-hosted/#productionalizing) strongly recommends using a load balancer in front of Sentry. And when the load balancer enables two-way SSL authentication, this pull request makes sentry-sdk allow to pass the client cert and key file to [urllib3.PoolManager](https://urllib3.readthedocs.io/en/stable/advanced-usage.html#client-certificates).

This pull request adds two new key arguments to [`sentry_sdk.init`](https://github.com/getsentry/sentry-python/blob/2.8.0/sentry_sdk/hub.py#L96-L106):

```python
import sentry_sdk

sentry_sdk.init(
    dsn="https://examplePublicKey@o0.ingest.sentry.io/0",
    ...,
    ca_certs="THE PATH OF ROOT CERTIFICATE",
    # The two new key arguments
    cert_file="THE PATH OF CERTIFICATE",
    key_file="THE PATH OF KEY",
)
```

---

## General Notes

#### For maintainers


szokeasaurusrex (changes requested):
Thanks for the contribution!

Before we can merge this, you also need to make sure to add the new options [here](https://github.com/getsentry/sentry-python/blob/ee84c81bd00ee9286cdc53f4c1980009e0297eb5/sentry_sdk/consts.py#L453), so they show up in the type hint for `sentry_sdk.init`. 

Also, please add tests for your change

sentry_sdk/consts.py
```diff
@@ -504,6 +504,8 @@ def __init__(
         debug=None,  # type: Optional[bool]
         attach_stacktrace=False,  # type: bool
         ca_certs=None,  # type: Optional[str]
+        cert_file=None,  # type: Optional[str]
```
sl0thentr0py, line 507: whenever you add options here, please append to the end of the list since these are also positional.
No one should ideally use them positionally that way but we have to be safe.

sl0thentr0py (changes requested):

sentry_sdk/consts.py
grammy-jiang: Thanks for your comment. `cert_file` and `key_file` are moved to the end of the arguments Now.

sl0thentr0py (approved):
ty!

sl0thentr0py:
@szokeasaurusrex needs to approve this too because his requests changes status is blocking the merge

szokeasaurusrex (approved):

PR #3260, opened 2024-07-09
ref(debug): Rename debug logging filter

szokeasaurusrex:
Previous name said that this filter was "hub-based," when the logic in reality is not related to hubs. So, we should rename the filter to something more sensible.


antonpirker (approved):
💯

PR #3261, opened 2024-07-09
Improved handling of span status

antonpirker:
Having the Sentry span statuses in a class and introducing `get_span_status_from_http_code()`. This will be used in our new OpenTelementry instrumentation. 

---

## General Notes

#### For maintainers

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -386,6 +386,32 @@ class SPANDATA:
     """
 
 
+class SPANSTATUS:
+    """
+    The status of a Sentry span.
+
+    See: https://develop.sentry.dev/sdk/event-payloads/contexts/#trace-context
+    """
+
+    ABORTED = "aborted"
+    ALREADY_EXISTS = "already_exists"
+    CANCELLED = "cancelled"
+    DATA_LOSS = "data_loss"
+    DEADLINE_EXCEEDED = "deadline_exceeded"
+    FAILED_PRECONDITION = "failed_precondition"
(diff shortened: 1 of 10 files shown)


sentry_sdk/consts.py
```diff
@@ -386,6 +386,32 @@ class SPANDATA:
     """
 
 
+class SPANSTATUS:
```
szokeasaurusrex, line 389: How about we make this an [enum](https://docs.python.org/3/library/enum.html#enum.Enum)?

sentry_sdk/tracing.py
```diff
@@ -151,6 +151,45 @@ class TransactionKwargs(SpanKwargs, total=False):
...
+
+    elif 500 <= http_status_code < 600:
+        if http_status_code == 504:
+            return SPANSTATUS.DEADLINE_EXCEEDED
+        elif http_status_code == 501:
+            return SPANSTATUS.UNIMPLEMENTED
+        elif http_status_code == 503:
+            return SPANSTATUS.UNAVAILABLE
+        else:
+            return SPANSTATUS.INTERNAL_ERROR
```
szokeasaurusrex, line 188: Would be nice to sort the if statements in increasing order by status code. Although, perhaps this would better belong in a separate PR

szokeasaurusrex (approved):
Looks good, please consider suggestions before merging

antonpirker:
Thanks for the quick review @szokeasaurusrex 
I will keep it as is for now and just merge, do not want to change to much right now, we can always improve later.

Issue #3265, opened 2024-07-10
Emit deprecation warnings for `Hub`-based APIs


szokeasaurusrex:
Per our documentation, all `Hub`-based APIs have been deprecated since 2.0.0, but we have not been emitting deprecation warnings yet, because we still use `Hub` APIs in a few places.

We should add deprecation warnings for all `Hub`-based APIs.

Before we can do this, we need to remove all remaining `Hub` API usages (besides those needed for backwards-compatibility) from the SDK.

szokeasaurusrex:
Following offline discussion with @antonpirker, we will only warn when:
  - Calling `Hub.__init__`
  - Accessing `Hub.current` or `Hub.main`

The idea here is that all other `Hub` APIs can only be accessed after obtaining a `Hub` object, which one can only do with one of the above methods. Once we have warned the user once that using the `Hub` is deprecated, we don't need to warn every other time they call a method on the hub.

PR #3267, opened 2024-07-10
ref(tracing): Remove `Hub` in `Transaction.finish`

szokeasaurusrex:
Rename `Transaction.finish` method's `hub` parameter to `scope` (in a backwards-compatible manner), and update the method so that it is using `Scope` API under the hood as much as possible.

Prerequisite for #3265

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -1205,8 +1205,17 @@ class NoOpSpan(Span):
         # type: () -> Any
         return {}
 
-    def finish(self, hub=None, end_timestamp=None):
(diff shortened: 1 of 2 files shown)


sentry_sdk/tracing.py
```diff
@@ -823,15 +834,57 @@ def containing_transaction(self):
...
+                stacklevel=3,
+            )
+
+            return scope_or_hub.scope
+
+        return scope_or_hub
+
+    def finish(
+        self,
+        scope=None,  # type: Optional[sentry_sdk.Scope]
```
antonpirker, line 872: The signature of `NoOpSpan.finish()` probably also needs an update.

antonpirker (approved):
Looks good. 

I left one comment, please check if `NoOpSpan` also needs an update.

PR #3270, opened 2024-07-10
ref(profiling): Deprecate `hub` in `Profile`

szokeasaurusrex:
Related to #3265

### DIFF
--- a/sentry_sdk/profiler/transaction_profiler.py
+++ b/sentry_sdk/profiler/transaction_profiler.py
@@ -33,6 +33,7 @@ import sys
 import threading
 import time
 import uuid
+import warnings
 from abc import ABC, abstractmethod
 from collections import deque
 
@@ -213,7 +214,6 @@ class Profile:
     ):
         # type: (...) -> None
         self.scheduler = _scheduler if scheduler is None else scheduler
-        self.hub = hub
 
         self.event_id = uuid.uuid4().hex  # type: str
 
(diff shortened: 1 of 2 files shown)


antonpirker (approved):
I guess this is fine!

sl0thentr0py (changes requested):
the working of the profiler is internal to the SDK so you are emitting a deprecation warning for our own code which makes no sense.
There is no way for the user to do something to not get this deprecation warning. Deprecation warnings are for _changes the user can make_.

szokeasaurusrex:
@sl0thentr0py none of our code accesses the `hub` attribute, as far as I can tell.

> the working of the profiler is internal to the SDK

Where do we defined that this `Profile` is for internal use only? Neither the module, class, nor the attribute are underscored, so I assumed that this attribute was public API, and therefore it would be a breaking change to simply remove the attribute, since a user's code might access it (not sure why, but hypothetically this could be the case).

If it is truly internal only, I would be fine with simply removing the `hub` attribute and/or (if it is also internal only) the `hub` parameter from the constructor. But if it is not, then we cannot just remove these and we need a deprecation warning.

sl0thentr0py (approved):
right ok, I don't think we wanted anyone to use the `Profile` but that's fine then

szokeasaurusrex:
> right ok, I don't think we wanted anyone to use the Profile but that's fine then

@sl0thentr0py, I think generally something we could do better in the SDK would be to limit our public API surface and maybe more clearly define what we actually consider to be public API. For instance, perhaps we could decide that everything that we don't explicitly re-export via `__all__` is considered an SDK implementation detail, or we can just get better about adding underscores to things that don't need to be public. Probably this is something that needs to wait for a major, but it would certainly make maintaining the SDK a lot easier if we had a smaller public API surface and therefore did not need to introduce quite as much backwards-compatibility code

PR #3276, opened 2024-07-10
ref(init): Move `sentry_sdk.init` out of `hub.py`

### LINKED ISSUE #3233: Move `sentry_sdk.init` implementation out of `hub.py`
Following the `hub` API deprecation, it would likely make more sense to move `sentry_sdk.init`'s implementation to `client.py` (or possibly `scope.py`).

szokeasaurusrex:
Now that the `Hub`-based API is deprecated, `sentry_sdk.init` should no longer be in `hub.py`. Since it is kind of its own thing, it makes sense to implement `init` in its own file.

Closes #3233

### DIFF
--- a/sentry_sdk/_init_implementation.py
+++ b/sentry_sdk/_init_implementation.py
@@ -1,0 +1,63 @@
+from typing import TYPE_CHECKING
+
+import sentry_sdk
+
+if TYPE_CHECKING:
+    from typing import Any, ContextManager, Optional
+
+    import sentry_sdk.consts
+
+
+class _InitGuard:
+    def __init__(self, client):
+        # type: (sentry_sdk.Client) -> None
+        self._client = client
+
+    def __enter__(self):
+        # type: () -> _InitGuard
+        return self
+
+    def __exit__(self, exc_type, exc_value, tb):
+        # type: (Any, Any, Any) -> None
+        c = self._client
+        if c is not None:
+            c.close()
+
+
+def _check_python_deprecations():
+    # type: () -> None
+    # Since we're likely to deprecate Python versions in the future, I'm keeping
+    # this handy function around. Use this to detect the Python version used and
+    # to output logger.warning()s if it's deprecated.
+    pass
+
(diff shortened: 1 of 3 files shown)


sentry_sdk/_init_implementation.py
```diff
@@ -0,0 +1,62 @@
...
+    rv = _InitGuard(client)
+    return rv
+
+
+if TYPE_CHECKING:
+    # Make mypy, PyCharm and other static analyzers think `init` is a type to
+    # have nicer autocompletion for params.
+    #
+    # Use `ClientConstructor` to define the argument types of `init` and
+    # `ContextManager[Any]` to tell static analyzers about the return type.
```
szokeasaurusrex, line 52: This comment is inaccurate, since we actually use `_InitGuard` in the class definition. However, since I am just copying this code over from `hub` for now, I think it is out of scope to fix here. Perhaps, we can consider fixing in a separate PR later.

sl0thentr0py: I also don't really know what the purpose of the `_InitGuard` is since we never tell people to do the following anywhere, so this is just vague unadvertised functionality.
```python
with sentry_sdk.init():
    pass
```

szokeasaurusrex: Indeed, perhaps we should explicitly deprecate this context manager functionality and remove in the next major, in that case? Or is this a potentially valuable feature we should advertise more?

sl0thentr0py: `init` is generally one per process anyway. In case people want DSN isolation, multiple clients is the way to go now. So I really don't see the use case for this. 
Let's deprecate and remove.

szokeasaurusrex: @sl0thentr0py, I opened an issue to make this change, since I think it belongs in a separate PR. Would appreciate an approval here if everything else looks good to you.

sl0thentr0py (approved):

PR #3280, opened 2024-07-11
feat(hub): Emit deprecation warnings from `Hub` API

### LINKED ISSUE #3265: Emit deprecation warnings for `Hub`-based APIs

szokeasaurusrex:
`sentry_sdk.Hub` has been deprecated since Sentry SDK version 2.0.0 per our docs; however, we waited with adding deprecation warnings because the SDK itself was still using `Hub` APIs until recently.

Since we no longer use `Hub` APIs in the SDK (except in `Hub` APIs which are themselves deprecated), we can now start emitting deprecation warnings.

Closes #3265


szokeasaurusrex:
When testing this change manually, I was able to see the deprecation warnings in console when calling Hub APIs from a simple Python script.

However, when I tried calling Hub APIs from a request handler (both in Flask and FastAPI), I did not see the deprecation warning in the console, unless I manually configured a warning filter to display deprecation warnings. 

Do you have any idea why this could be @sl0thentr0py? Is there anything we can do here to make sure server framework users get warned, or is this just a configuration option that the user would have to set?

szokeasaurusrex:
@sl0thentr0py I wanted to make this be raised as a `DeprecationWarning` and I was unsure whether this is possible to do with the `logger`, which is why I originally used `warnings`. 

However, I looked into `warnings` vs `logging` a bit further, and it seems we could emit a `DeprecationWarning` with `logging`, but based on [this table](https://docs.python.org/3/howto/logging.html#when-to-use-logging), it seems that using `warnings` is the preferred option for this situation.

Do you think using the `logger` would make the warnings show up any differently?

sl0thentr0py:
so far everyone else has just used `logger.warning` for deprecations and related stuff and I think it's better to just keep using those? Either way, just try it out and see if it makes a difference.

szokeasaurusrex:
@sl0thentr0py after further investigation, it looks like the reason why I was not seeing the `DeprecationWarning` when running with Flask is because the Flask development server (`flask run`) runs the server with `__name__ == "main"` instead of `__name__ == "__main__"` but does not adjust the [default warnings filter](https://docs.python.org/3/library/warnings.html#default-warning-filter), which ignores `DeprecationWarning`s which originate from a module other than `"__main__"`. The warnings are visible when running the server via an `app.run` call within the code, since in this case, `__name__ == "__main__"`. 

Furthermore, using the `logger` is not really an ideal solution here, since (at least from what I can tell) there is no way that I can suppress the `logger` message when we intentionally call a deprecated method. For example, when `Hub.current` calls the deprecated `Hub.__init__`, I want to suppress the `Hub.__init__` deprecation warning since we already warn for `Hub.current`. Or, when we initialize the `GLOBAL_HUB` at module import, I want to make sure we suppress the deprecation warning, and I am unaware of how this would be possible with the logger.

So in short, I think using `warnings` is the best solution here. If ever the warnings are not showing, that is due to how the user code has configured their warnings filter, and this is outside our control.

szokeasaurusrex:
Deprecation warnings can also be set to always be shown by setting the [`PYTHONDEVMODE` environment variable to `1`](https://docs.python.org/3/library/devmode.html#devmode). Then, even when running a Flask app with `flask run`, you would see the deprecation warnings

sl0thentr0py (approved):

PR #3281, opened 2024-07-11
test: Allow passing of PostgreSQL port

rominf:
Allow passing an arbitrary port via the `SENTPY_PYTHON_TEST_POSTGRES_PORT` environmental variable.

Fedora's RPM macro `%postgresql_tests_run` which starts PostgreSQL dynamically selects PostgreSQL port to prevent start failures when running multiple PostgreSQL servers on the same default port [1].

This issue is not specific to Fedora. In case there is some application running on the same machine with port `5432` opened, such as PostgreSQL instance with the default port, this will result in failure to start the PostgreSQL server, resulting in an inability to run these tests.

This change allows running these tests in environments where PostgreSQL has a non-default port and where other applications (including a PostgreSQL instance with the default port) have opened port `5432`, while at the same time keeping the old behavior as the default.

[1] The macro is provided by https://packages.fedoraproject.org/pkgs/postgresql15/postgresql15-test-rpm-macros/,
which packages https://github.com/devexp-db/postgresql-setup. Dynamic selection of the port was added in 2018:
https://github.com/devexp-db/postgresql-setup/pull/16, for the reasoningsee `NEWS` file changes: https://github.com/devexp-db/postgresql-setup/pull/16/files#diff-7ee66c4f1536ac84dc5bbff1b8312e2eef24b974b3e48a5c5c2bcfdf2eb8f3ce).

---

## General Notes

#### For maintainers

### DIFF
--- a/tests/integrations/asyncpg/test_asyncpg.py
+++ b/tests/integrations/asyncpg/test_asyncpg.py
@@ -13,7 +13,7 @@ import os
 
 
 PG_HOST = os.getenv("SENTRY_PYTHON_TEST_POSTGRES_HOST", "localhost")
-PG_PORT = 5432
+PG_PORT = int(os.getenv("SENTRY_PYTHON_TEST_POSTGRES_PORT", "5432"))
 PG_USER = os.getenv("SENTRY_PYTHON_TEST_POSTGRES_USER", "postgres")
 PG_PASSWORD = os.getenv("SENTRY_PYTHON_TEST_POSTGRES_PASSWORD", "sentry")
 PG_NAME = os.getenv("SENTRY_PYTHON_TEST_POSTGRES_NAME", "postgres")
--- a/tests/integrations/django/myapp/settings.py
+++ b/tests/integrations/django/myapp/settings.py
@@ -122,7 +122,7 @@ try:
     DATABASES["postgres"] = {
         "ENGINE": db_engine,
         "HOST": os.environ.get("SENTRY_PYTHON_TEST_POSTGRES_HOST", "localhost"),
-        "PORT": 5432,
+        "PORT": int(os.environ.get("SENTRY_PYTHON_TEST_POSTGRES_PORT", "5432")),
         "USER": os.environ.get("SENTRY_PYTHON_TEST_POSTGRES_USER", "postgres"),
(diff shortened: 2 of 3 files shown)


rominf:
Test failure in `Test Web Frameworks 2` looks like a flaky test: https://github.com/getsentry/sentry-python/actions/runs/9881406304/job/27310056793.

rominf:
Hi @szokeasaurusrex.

There is no doubt related tests successfully run on your CI without the change. This is not always true for running these tests in other environments, such as developer machines and other CI/CD servers.

While enabling PostgreSQL-dependent tests for Fedora's `python-sentry-sdk` package packaging Sentry SDK for Python, I discovered that RPM macro `%postgresql_tests_run` which starts PostgreSQL dynamically selects PostgreSQL port to prevent start failures when running multiple PostgreSQL servers on the same default port [1]. Therefore, I had to apply this change to the `python-sentry-sdk` package as a patch for running these tests [2].

This issue is not specific to Fedora. In case there is some application running on the same machine with port `5432` opened, such as PostgreSQL instance with the default port, this will result in failure to start the PostgreSQL server, resulting in an inability to run these tests.

Thus, this few-line change allows running these tests in environments where PostgreSQL has a non-default port and where other applications (including a PostgreSQL instance with the default port) have opened port `5432`, while at the same time keeping the old behavior as the default.

[1] The macro is provided by https://packages.fedoraproject.org/pkgs/postgresql15/postgresql15-test-rpm-macros/, which packages https://github.com/devexp-db/postgresql-setup. Dynamic selection of the port was added in 2018: https://github.com/devexp-db/postgresql-setup/pull/16 where you can find the reasoning (see `NEWS` file changes: https://github.com/devexp-db/postgresql-setup/pull/16/files#diff-7ee66c4f1536ac84dc5bbff1b8312e2eef24b974b3e48a5c5c2bcfdf2eb8f3ce).
[2] https://src.fedoraproject.org/fork/rominf/rpms/python-sentry-sdk/c/7194c68da701a2b2d76058ec8b48ca7c6b5098e5

szokeasaurusrex:
Fair enough @rominf, my main concern is that adding this environment variable will provide only limited usefulness, since it is not documented anywhere.

However, if this feature helps you, I can go ahead and merge it. It looks like we are already allowing other PostreSQL options to be configured via environment variable, so it is probably fine to allow the port to be set as well

szokeasaurusrex (approved):

PR #3283, opened 2024-07-12
docs(init): Fix `sentry_sdk.init` type hint

szokeasaurusrex:
The current type hint suggests that all the parameters can be passed as positional arguments, when this is not the case. Only the `dsn` can be passed as a positional argument; the rest must be passed as keyword arguments.

This PR makes the type hint reflect the reality of what parameters can be passed to `sentry_sdk.init`.

### DIFF
--- a/sentry_sdk/hub.py
+++ b/sentry_sdk/hub.py
@@ -89,7 +89,7 @@ def _init(*args, **kwargs):
 
     This takes the same arguments as the client constructor.
     """
-    client = Client(*args, **kwargs)  # type: ignore
+    client = Client(*args, **kwargs)
     Scope.get_global_scope().set_client(client)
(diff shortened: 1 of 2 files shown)


sl0thentr0py (approved):

sentry_sdk/hub.py
```diff
@@ -89,7 +89,7 @@ def _init(*args, **kwargs):
 
     This takes the same arguments as the client constructor.
     """
-    client = Client(*args, **kwargs)  # type: ignore
+    client = Client(*args, **kwargs)
```
szokeasaurusrex, line 92: Apparently my changes have made the `# type: ignore` here unnecessary

sl0thentr0py (approved):

PR #3290, opened 2024-07-12
feat(pymongo): Set MongoDB tags directly on span data

0Calories:
Another PR to iron out some inconsistencies between the Python and Node SDKs. OTel attaches MongoDB tags directly on the span data, rather than in the `tags` databag. This change is necessary so that Relay can parse MongoDB spans in Python for metrics extraction.

### Node SDK Spans:
Relevant MongoDB data is set directly as span data
image

### Python SDK Spans:
Relevant MongoDB data is set in the `tags` databag
image

### DIFF
--- a/sentry_sdk/integrations/pymongo.py
+++ b/sentry_sdk/integrations/pymongo.py
@@ -162,7 +162,7 @@ class CommandTracer(monitoring.CommandListener):
             )
 
             for tag, value in tags.items():
-                span.set_tag(tag, value)
+                span.set_data(tag, value)
 
             for key, value in data.items():
                 span.set_data(key, value)
--- a/sentry_sdk/integrations/pymongo.py
+++ b/sentry_sdk/integrations/pymongo.py
@@ -162,6 +162,7 @@ class CommandTracer(monitoring.CommandListener):
             )
 
             for tag, value in tags.items():
+                span.set_tag(tag, value)
                 span.set_data(tag, value)
 
             for key, value in data.items():
--- a/sentry_sdk/integrations/pymongo.py
+++ b/sentry_sdk/integrations/pymongo.py
@@ -162,7 +162,10 @@ class CommandTracer(monitoring.CommandListener):
             )
 
             for tag, value in tags.items():
+                # set the tag for backwards-compatibility.
+                # TODO: remove the set_tag call in the next major release!
                 span.set_tag(tag, value)
+
                 span.set_data(tag, value)
 
             for key, value in data.items():
--- a/tests/integrations/pymongo/test_pymongo.py
+++ b/tests/integrations/pymongo/test_pymongo.py
@@ -61,23 +61,23 @@ def test_transactions(sentry_init, capture_events, mongo_server, with_pii):
         assert span["data"][SPANDATA.SERVER_ADDRESS] == "localhost"
         assert span["data"][SPANDATA.SERVER_PORT] == mongo_server.port
         for field, value in common_tags.items():
-            assert span["tags"][field] == value
+            assert span["data"][field] == value
 
     assert find["op"] == "db"
     assert insert_success["op"] == "db"
     assert insert_fail["op"] == "db"
 
-    assert find["tags"]["db.operation"] == "find"
-    assert insert_success["tags"]["db.operation"] == "insert"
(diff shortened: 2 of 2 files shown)


0Calories:
@szokeasaurusrex That's a good point, but consistency is still more important so that we can extract metrics for multiple SDKs. I believe the Python SDK in the near future will be implementing OTel's SDK under the hood, just like the JS SDK, so this exact change will occur and break dashboards for users who rely on these tags anyways

szokeasaurusrex:
@0Calories I think this change will still require a major version bump if it is breaking. I understand that consistency matters, but semantic versioning dictates a major version bump for any breaking change.

We are in progress with implementing POTel in Python and intend to release a major when we do so, so I guess we can include this change then.

szokeasaurusrex:
I suppose the change could be made backwards-compatible by setting this data both as `tags` and as `data` on the span. The data would of course be redundant, and we would probably need to consider implications in terms of storage costs on the server, but then we could include this change in a minor without violating semantic versioning

0Calories:
That makes sense to me, I'll update this to include the data on both `tags` and `data`

tests/integrations/pymongo/test_pymongo.py
```diff
@@ -62,22 +62,23 @@ def test_transactions(sentry_init, capture_events, mongo_server, with_pii):
...
     assert find["op"] == "db"
     assert insert_success["op"] == "db"
     assert insert_fail["op"] == "db"
 
-    assert find["tags"]["db.operation"] == "find"
-    assert insert_success["tags"]["db.operation"] == "insert"
-    assert insert_fail["tags"]["db.operation"] == "insert"
+    assert find["data"]["db.operation"] == "find"
+    assert insert_success["data"]["db.operation"] == "insert"
+    assert insert_fail["data"]["db.operation"] == "insert"
```
szokeasaurusrex, line 73: Since we are going the backwards-compatible route, please keep the assertions that make sure we set the tags, as well

tests/integrations/pymongo/test_pymongo.py
```diff
@@ -62,22 +62,23 @@ def test_transactions(sentry_init, capture_events, mongo_server, with_pii):
...
     assert find["description"].startswith("{'find")
     assert insert_success["description"].startswith("{'insert")
     assert insert_fail["description"].startswith("{'insert")
 
-    assert find["tags"][SPANDATA.DB_MONGODB_COLLECTION] == "test_collection"
-    assert insert_success["tags"][SPANDATA.DB_MONGODB_COLLECTION] == "test_collection"
-    assert insert_fail["tags"][SPANDATA.DB_MONGODB_COLLECTION] == "erroneous"
+    assert find["data"][SPANDATA.DB_MONGODB_COLLECTION] == "test_collection"
+    assert insert_success["data"][SPANDATA.DB_MONGODB_COLLECTION] == "test_collection"
+    assert insert_fail["data"][SPANDATA.DB_MONGODB_COLLECTION] == "erroneous"
```
szokeasaurusrex, line 81: Here also

sentry_sdk/integrations/pymongo.py
```diff
@@ -163,6 +163,7 @@ def started(self, event):
 
             for tag, value in tags.items():
                 span.set_tag(tag, value)
```
szokeasaurusrex, line 165: We should include a comment to indicate that the `set_tag` call is included for backwards-compatibility. Something like the following could do:

```suggestion
                # set the tag for backwards-compatibility.
                # TODO: remove the set_tag call in the next major release!
                span.set_tag(tag, value)

```

szokeasaurusrex (commented):
Thanks for making the changes! Just a few more small thins, and then I think the PR will be ready for merging

sentry_sdk/integrations/pymongo.py
0Calories: Good point!

szokeasaurusrex (approved):
Looks good!

0Calories:
Awesome thanks @szokeasaurusrex! Could you also please merge the PR for me? It seems I don't have permission to do so on my own

szokeasaurusrex:
@0Calories the PR needs to be updated with the base branch before you can merge

PR #3295, opened 2024-07-16
ref: Deprecate `instrumenter`

szokeasaurusrex:


szokeasaurusrex:
I have gone with deprecating `instrumenter` instead of fully removing the parameter for the following reasons:

1. We only state that `instrumenter` is for internal use only in a few of the functions that take `instrumenter` as a parameter.
2. Even in some of the functions which explicitly state `instrumenter` is for internal use only, there are positional arguments after `instrumenter`. These positional arguments would be broken by simply removing `instrumenter`, since they would no longer be in the expected position.

PR #3374, opened 2024-07-30
Add span data to the transactions trace context

antonpirker:
Fixes https://github.com/getsentry/sentry-python/issues/3372

### DIFF
--- a/tests/tracing/test_misc.py
+++ b/tests/tracing/test_misc.py
@@ -11,6 +11,7 @@ from sentry_sdk.consts import MATCH_ALL
 from sentry_sdk.tracing import Span, Transaction
 from sentry_sdk.tracing_utils import should_propagate_trace
 from sentry_sdk.utils import Dsn
+from tests.conftest import ApproxDict
 
 
 def test_span_trimming(sentry_init, capture_events):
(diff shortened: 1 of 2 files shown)


tests/tracing/test_misc.py
```diff
@@ -60,6 +61,37 @@ def test_transaction_naming(sentry_init, capture_events):
...
+    sentry_init(traces_sample_rate=1.0)
+    events = capture_events()
+
+    with start_transaction(name="test-transaction"):
+        span_or_tx = sentry_sdk.get_current_span()
+        span_or_tx.set_data("foo", "bar")
+        with start_span(op="test-span") as span:
+            span.set_data("spanfoo", "spanbar")
+
+    transaction = events[0]
```
szokeasaurusrex, line 74: Do we also want to assert here whether there is exactly one event? With how the test is currently written, the test would pass if there are multiple events.

I am guessing we probably want something like the following:

```
(transaction,) = events
```

tests/tracing/test_misc.py
```diff
@@ -60,6 +61,37 @@ def test_transaction_naming(sentry_init, capture_events):
...
+    transaction_data = transaction["contexts"]["trace"]["data"]
+
+    assert "data" not in transaction.keys()
+    assert transaction_data == ApproxDict(
+        {
+            "foo": "bar",
+        }
+    )
```
szokeasaurusrex, line 82: [nitpick]
I would find the following easier to understand than using `ApproxDict`:

```suggestion
    assert transaction_data.items() >= { "foo": "bar" }.items()
```

Here, we explicitly compare that `transaction_data` at least includes the `"foo": "bar"` key-value pair. With the `ApproxDict`, I always have to go to the `ApproxDict` class definition to remind myself what an `ApproxDict` does. 

I would prefer to move away from using `ApproxDict`, and perhaps replace existing usages with comparing the `.items()`. If you agree, I can make an issue for this.

tests/tracing/test_misc.py
```diff
@@ -60,6 +61,37 @@ def test_transaction_naming(sentry_init, capture_events):
...
+
+    span = transaction["spans"][0]
```
szokeasaurusrex, line 84: Same here regarding asserting the number of spans: Is it important that we only have one span? If yes, we should assert that we have only one span. Otherwise, I think we should just make an order-independent assertion that `transaction["spans"]` includes a span that matches our expectations.

szokeasaurusrex (commented):
Looks good, but I have some suggestions for the test

tests/tracing/test_misc.py
antonpirker: done.

tests/tracing/test_misc.py
antonpirker: done.

szokeasaurusrex (approved):

PR #3386, opened 2024-07-31
feat(profiling): Add client sdk info to profile chunk

Zylphrex:
We want to attach the client sdk info for debugging purposes.


sentry_sdk/_types.py
```diff
@@ -20,11 +20,21 @@
...
 
+    SdkInfo = TypedDict(
+        "SdkInfo",
+        {
+            "name": str,
+            "version": str,
+            "packages": Sequence[Any],
+        },
+    )
+
```
szokeasaurusrex, line 37: A few nitpicks here, in order from most to least important:
  - We can likely make the `packages` type stricter, `Sequence[Mapping[str, str]]` seems like it will work (you will need to import `Mapping` from `collections.abc`).
  - I think the class-based `TypedDict` syntax might be cleaner here.
  - Given that SDK is usually capitalized, I would name this type as `SDKInfo` rather than `SdkInfo`.

```suggestion
    class SDKInfo(TypedDict):
        name: str
        version: str
        packages: Sequence[Mapping[str, str]]

```

tests/profiler/test_continuous_profiler.py
```diff
@@ -31,14 +32,29 @@ def experimental_options(mode=None, auto_start=None):
     }
 
 
+@pytest.fixture
+def sdk_info():
+    return {
+        "name": "sentry.python",  # SDK name will be overridden after integrations have been loaded with sentry_sdk.integrations.setup_integrations()
```
szokeasaurusrex, line 38: Is this comment accurate here? If not, let's remove it:

```suggestion
        "name": "sentry.python",
```

tests/profiler/test_continuous_profiler.py
```diff
@@ -31,14 +32,29 @@ def experimental_options(mode=None, auto_start=None):
     }
 
 
+@pytest.fixture
```
szokeasaurusrex, line 35: Why a fixture as opposed to a constant?

szokeasaurusrex (commented):
Looks mostly good, but I have some comments and questiosn

sentry_sdk/_types.py
Zylphrex: Can we actually use the class based `TypedDict` in the SDK? I thought we used this for compatibility reasons and all other instances of `TypedDict` in the SDK uses this syntax.

szokeasaurusrex: We use it already:

https://github.com/getsentry/sentry-python/blob/2c1e31c5390310ae696108aa135c055452600f43/sentry_sdk/tracing.py#L43

Perhaps what you said was true with Sentry SDK 1.x, but we can use class-based syntax now.

sentry_sdk/_types.py
```diff
@@ -20,11 +20,21 @@
     from typing import Mapping
     from typing import NotRequired
     from typing import Optional
+    from typing import Sequence
```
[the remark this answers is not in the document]
Zylphrex: Seems like it was deprecated in 3.9 and before that, subscripts were not supported by `collections.abc.Sequence`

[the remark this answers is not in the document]
szokeasaurusrex: I think that is fine, these type hints are all executed within `if TYPE_CHECKING` and `# type` comments, so it does not really matter that only 3.9 introduced subscripting. None of this code gets evaluated at runtime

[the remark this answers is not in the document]
szokeasaurusrex: fwiw we already subscript things in `collections.abc` in several places in the codebase

szokeasaurusrex (approved):
Nice!

PR #3388, opened 2024-08-01
ref(tests): remove `ApproxDict` #3380

### LINKED ISSUE #3380: (tests) Move away from `ApproxDict`
`conftest.py` defines an `ApproxDict`, which is a subclass of `dict`, whose equality method has been overridden such that, if `a` and `b` are dictionaries, then `a == ApproxDict(b)` will have the same truth value as `a.items() >= b.items()` (i.e. `a` is equal to `ApproxDict(b)` if `a`'s key-value pairs are a superset of `b`'s key-value pairs). 

Since `a.items() >= b.items()` is a much clearer assertion than `a == ApproxDict(b)`, let's replace usages of `ApproxDict` in our test suite with the more explicit assertion.

czyber:
This commit removes `ApproxDict` in favour of a more explicit assertion style.

Closes #3380

---

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -633,15 +633,3 @@ def patch_start_tracing_child(fake_transaction_is_none=False):
         "sentry_sdk.tracing_utils.get_current_span", return_value=fake_transaction
     ):
         yield fake_start_child
-
-
-class ApproxDict(dict):
-    def __eq__(self, other):
-        # For an ApproxDict to equal another dict, the other dict just needs to contain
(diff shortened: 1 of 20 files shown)


czyber:
Still have to fix some assertions (which werent picked up by the regex)

szokeasaurusrex:
Thanks for picking this up at @czyber! Happy to review once the PR is ready

szokeasaurusrex:
Closing because of https://github.com/getsentry/sentry-python/issues/3380#issuecomment-2535305931

PR #3400, opened 2024-08-06
Explicitly export symbols in subpackages

hartungstenio:
Without explicitly exporting those symbols, mypy strict linting fails on some projects.

This change solves this, and we don't need to ignore F401 (unused symbol) anymore on those files.

---


sentrivana (approved):
Thank you @hartungstenio!

Issue #3402, opened 2024-08-07
Remove `configure_scope`


szokeasaurusrex:
Remove `sentry_sdk.configure_scope` and any backwards compatibility tests testing this function. To be done in `potel-base` branch.

szokeasaurusrex:
Merged in `potel-base`

PR #3406, opened 2024-08-07
feat(api): Remove `sentry_sdk.configure_scope`

szokeasaurusrex:
Also, remove any tests for `sentry_sdk.configure_scope`.

Since Strawberry's deprecated [Sentry tracing extensions](https://strawberry.rocks/docs/extensions/sentry-tracing) import `sentry_sdk.configure_scope`, importing `strawberry.extensions.tracing.SentryTracingExtension` (or `SentryTracingExtensionSync`) will result in an unhandled exception. Therefore, these imports, and any functionality associated with them, have also been removed. This itself is not a breaking change, as it is necessitated by the removal of `sentry_sdk.configure_scope`.

BREAKING CHANGE: Remove `sentry_sdk.configure_scope`.

Closes: #3402

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -51,8 +51,6 @@ Client Management
 Managing Scope (advanced)
 =========================
 
-.. autofunction:: sentry_sdk.api.configure_scope
 .. autofunction:: sentry_sdk.api.push_scope
 
 .. autofunction:: sentry_sdk.api.new_scope
-
--- a/sentry_sdk/__init__.py
+++ b/sentry_sdk/__init__.py
@@ -19,7 +19,6 @@ __all__ = [  # noqa
     "capture_event",
     "capture_exception",
     "capture_message",
-    "configure_scope",
     "continue_trace",
     "flush",
     "get_baggage",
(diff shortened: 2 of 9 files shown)


antonpirker (approved):
Looking good. 

Strawberry will then not be compatible with the new SDK anymore, because they use configure_scope in their sentry extension:
https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/extensions/tracing/sentry.py

szokeasaurusrex:
> Strawberry will then not be compatible with the new SDK anymore, because they use configure_scope in their sentry extension:
> https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/extensions/tracing/sentry.py

I am going to see if there is anything we can do here about this, or perhaps we need to submit a PR upstream

szokeasaurusrex:
@antonpirker looks like only Strawberry's Sentry tracing extensions will be incompatible – everything else from Strawberry still works. I removed the imports of those extensions and any functionality based on them; that way, the tests should pass. I will also open an issue in Strawberry's repo to let them know we are dropping support for `configure_scope`.

antonpirker (approved):
looks good

PR #3419, opened 2024-08-08
ref(sessions): Deprecate hub-based `sessions.py` logic

### LINKED ISSUE #3417: Fully deprecate hub-based logic in `sessions.py`

szokeasaurusrex:
Make several changes to prepare for fully removing Hubs in the next major:
  - Deprecate the Hub-based `auto_session_tracking` function, replacing it with a new Scope-based function called `track_session`
  - Deprecate the scope-based `auto_session_tracking_scope` in favor of the new `track_session` function
  - Change usages of `auto_session_tracking_scope` to `track_sessions`. There are no usages of `auto_session_tracking` outside of tests.
  - Run all tests that were previously run against `auto_session_tracking` also against the new `track_session`. Previously, `auto_session_tracking_scope` was completely untested.

Fixes #3417

### DIFF
--- a/sentry_sdk/sessions.py
+++ b/sentry_sdk/sessions.py
@@ -47,8 +47,15 @@ def is_auto_session_tracking_enabled(hub=None):
 @contextmanager
 def auto_session_tracking(hub=None, session_mode="application"):
     # type: (Optional[sentry_sdk.Hub], str) -> Generator[None, None, None]
-    """Starts and stops a session automatically around a block."""
-    # TODO: add deprecation warning
+    """DEPRECATED: Use track_session instead
+    Starts and stops a session automatically around a block.
+    """
+    warnings.warn(
+        "This function is deprecated and will be removed in the next major release. "
+        "Use track_session instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
 
     if hub is None:
         hub = sentry_sdk.Hub.current
@@ -98,13 +105,28 @@ def _is_auto_session_tracking_enabled(scope):
 @contextmanager
 def auto_session_tracking_scope(scope, session_mode="application"):
     # type: (sentry_sdk.Scope, str) -> Generator[None, None, None]
-    """
+    """DEPRECATED: This function is a deprecated alias for track_session.
     Starts and stops a session automatically around a block.
-
-    TODO: This uses the new scopes. When the Hub is removed, the function
-    auto_session_tracking should be removed and this function
-    should be renamed to auto_session_tracking.
     """
+
+    warnings.warn(
+        "This function is a deprecated alias for track_session and will be removed in the next major release.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+
+    with track_session(scope, session_mode=session_mode):
+        yield
+
+
+@contextmanager
+def track_session(scope, session_mode="application"):
+    # type: (sentry_sdk.Scope, str) -> Generator[None, None, None]
+    """
+    Start a new session in the provided scope, assuming session tracking is enabled.
+    This is a no-op context manager if session tracking is not enabled.
+    """
+
     should_track = _is_auto_session_tracking_enabled(scope)
     if should_track:
         scope.start_session(session_mode=session_mode)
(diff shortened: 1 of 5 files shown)


sentry_sdk/sessions.py
```diff
@@ -19,32 +20,56 @@
...
     if should_track is None:
-        client_options = hub.client.options if hub.client else {}
+        client_options = sentry_sdk.get_client().options
         should_track = client_options.get("auto_session_tracking", False)
 
     return should_track
 
 
+@contextmanager
+def track_session(scope, session_mode="application"):
```
szokeasaurusrex, line 38: I originally tried to make the `auto_session_tracking` function use the scope based API by default but still be backwards compatible with hub-based API, but implementing this was going to be extremely overcomplicated.

I decided it would be much simpler to just implement a new function to replace `auto_session_tracking` instead, hence why I have proposed fully deprecating `auto_session_tracking` and `auto_session_tracking_scope` and replacing both with the new `track_session`.

sentry_sdk/sessions.py
```diff
@@ -19,32 +20,56 @@
     from typing import Union
 
 
-def is_auto_session_tracking_enabled(hub=None):
-    # type: (Optional[sentry_sdk.Hub]) -> Union[Any, bool, None]
-    """Utility function to find out if session tracking is enabled."""
-    # TODO: add deprecation warning
-
-    if hub is None:
-        hub = sentry_sdk.Hub.current
```
antonpirker, line 28: Imo to deprecate the hub based session tracking we could add a deprecation warning here if `hub is not None` and explain that in the next major they can not use the `hub` as a parameter in the next major, but can pass a `scope` parameter.

When we do the next major we do what is written in the comment in sessions.py:
```
TODO: This (is_auto_session_tracking_enabled_scope) uses the new scopes. When the Hub is removed, the function is_auto_session_tracking_enabled should be removed and this function should be renamed to is_auto_session_tracking_enabled.
```

szokeasaurusrex: Discussed offline. The `is_auto_session_tracking_enabled` deprecation is now in a separate PR (#3428). After clarifying the reasoning for introducing a new function was that it would allow us in our own code to avoid using deprecated APIs, @antonpirker agreed that introducing `track_sessions` was a good idea

tests/test_sessions.py
```diff
@@ -120,6 +195,35 @@ def test_no_thread_on_shutdown_no_errors(sentry_init):
...
+                except Exception:
+                    sentry_sdk.capture_exception()
+
+        with sentry_sdk.isolation_scope() as scope:
+            with track_session(scope, session_mode="request"):
+                pass
+
+        sentry_sdk.get_isolation_scope().start_session(session_mode="request")
+        sentry_sdk.get_isolation_scope().end_session()
+        sentry_sdk.flush()
```
antonpirker, line 216:
```suggestion
        sentry_sdk.flush()
        
        # If we reach this point without error, the test is successful.
```

antonpirker: Can you also add this comment to the next test? (can not suggest this change from the github ui)

tests/test_sessions.py
```diff
@@ -50,6 +50,48 @@ def test_aggregates(sentry_init, capture_envelopes):
...
+            except Exception:
+                sentry_sdk.capture_exception()
+
+    with sentry_sdk.isolation_scope() as scope:
+        with track_session(scope, session_mode="request"):
+            pass
+
+    sentry_sdk.get_isolation_scope().start_session(session_mode="request")
+    sentry_sdk.get_isolation_scope().end_session()
+    sentry_sdk.flush()
```
antonpirker, line 67: Tests are always also a kind of documentation on how to use the code. 

Could you add comments to the three sessions to mark what aggregate status is expected for each one of them?

Please add those comments to all similar tests in this file. This will help us in the future (or others) to understand what is going on.

tests/test_sessions.py
szokeasaurusrex: Accidentally merged before fixed, opened #3430 with this change

PR #3446, opened 2024-08-13
Remove `Hub` and related code for good.

antonpirker:
Finally it is the time to delete the Hub for good.

---


sentrivana (approved):
🎉 🎉 🎉

PR #3313, opened 2024-07-19
fix(tracing): Fix `add_query_source` with modules outside of project root

rominf:
Fix: https://github.com/getsentry/sentry-python/issues/3312

Previously, when packages added in `in_app_include` were installed to a location outside of the project root directory, span from those packages were not extended with OTel compatible source code information. Cases include running Python from virtualenv created outside of the project root directory or Python packages installed into the system using package managers. This resulted in an inconsistency: spans from the same project would be different, depending on the deployment method.

The first change moves `should_be_included` logic into the function to enable unit testing.

In the second change, the logic was slightly changed to avoid these discrepancies and conform to the requirements, described in the PR with a better setting of in-app in stack frames: https://github.com/getsentry/sentry-python/pull/1894#issue-1579192436.

---

## General Notes

Thank you for contributing to `sentry-python`!

Please add tests to validate your changes, and lint your code using `tox -e linters`.

Running the test suite on your PR might require maintainer approval. Some tests (AWS Lambda) additionally require a maintainer to add a special label to run and will fail if the label is not present.

#### For maintainers

Sensitive test suites require maintainer review to ensure that tests do not compromise our secrets. This review must be repeated after any code revisions.

Before running sensitive test suites, please carefully check the PR. Then, apply the `Trigger: tests using secrets` label. The label will be removed after any code changes to enforce our policy requiring maintainers to review all code revisions before running sensitive tests.

PR #3452, opened 2024-08-20
chore(tracing): Refactor `tracing_utils.py`

rominf:
Preparation for:
https://github.com/getsentry/sentry-python/pull/3313
Proposed in:
https://github.com/getsentry/sentry-python/pull/3313#discussion_r1704258749

Note that the `_module_in_list` function returns `False` if `name` is `None` or `items` are falsy, hence extra check before function call can be omitted to simplify code.

---


szokeasaurusrex (approved):
Thanks for the contribution! Code looks much cleaner this way.

I also made a further simplification to the `should_be_included` logic

PR #3460, opened 2024-08-26
Implement new continue_trace and make WSGI work

### LINKED ISSUE #3458: Make `continue_trace` work

sl0thentr0py:
The new `continue_trace` API will no longer return a `Transaction` entity. Instead, it will simply update the propagation context and run as a contextmanager.

(TODO) It will set a remote span on the OTEL context so that it can be picked up by `start_span` later.

closes #3458

---

working flask + sqlalchemy trace

image

### DIFF
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -1,4 +1,5 @@
 import inspect
+from contextlib import contextmanager
 
 from sentry_sdk import tracing_utils, Client
 from sentry_sdk._init_implementation import init
@@ -23,6 +24,7 @@ if TYPE_CHECKING:
     from typing import Callable
     from typing import TypeVar
     from typing import Union
+    from typing import Generator
 
     from typing_extensions import Unpack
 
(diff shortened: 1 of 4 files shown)


szokeasaurusrex (commented):
Have you considered any way we might be able to avoid making a breaking API change to `continue_trace` here?

I am thinking it might be possible to avoid this breaking change by somehow setting the `environ_or_headers` on the transaction, then when the transaction is started, it would be responsible for calling `PotelScope.continue_trace` when an `environ_or_headers` is present. Have you considered doing something like this, and if yes, why are we choosing to make a breaking change, instead?

Admittedly, my suggestion might be a bit harder to think through than what you are suggesting in this PR, but it would come with the huge upside of not having a breaking change and therefore not needing to update all our `continue_trace` calls (plus, less hassle for users when upgrading)

sentrivana (approved):

sl0thentr0py:
yes I have considered and decided. JS also broke this and it's much better semantics this way.

We might reconsider breaking changes together at the end when we are in an acceptable functional state but for now I want to focus on designing ideal semantics in the new world.

szokeasaurusrex:
> yes I have considered and decided. JS also broke this and it's much better semantics this way.
> 
> We might reconsider breaking changes together at the end when we are in an acceptable functional state but for now I want to focus on designing ideal semantics in the new world.

Fair enough, I guess this is fine for now. I agree that if we were designing this from scratch, it would probably make the most sense to do it your way, but I think we should really try to avoid a breaking change here if we can

szokeasaurusrex:
Why we are making the breaking API change: https://github.com/getsentry/sentry-python/pull/3486#pullrequestreview-2275224197

PR #3468, opened 2024-08-27
Fix "indented block" linter error

szokeasaurusrex:
Fix this strange [linter error](https://github.com/getsentry/sentry-python/actions/runs/10390581522/job/28771379618#step:4:73). Also, fix other mypy errors which mypy only surfaced after fixing the linked linter error, and fix typing in this file, which is broken by the fix for the original mypy error.

It might be easiest to review each of the three commits in this PR individually, since they each build on each other and have through commit messages.


sl0thentr0py (approved):

szokeasaurusrex:
Looks like linter is still failing, but now there are a bunch more failures that were not detected previously, because the linter failure I fixed was preventing further linting

sentrivana (approved):
Nice work, detective 🎖️ 

We can merge this even with the new failures -- no need to fix that in this PR

PR #3501, opened 2024-09-05
Potel Sampling

### LINKED ISSUE #3318: Sampling in POtel
Sample Opentelemetry spans according to the configuration options `traces_sampler` and `traces_sample_rate` also take incoming requests into account (upstreams sampling decision or if the upstream service has deferred the sampling decision)

Also think about: Sentry's sampling is based on transactions, but Otel does sampling on a span basis.

antonpirker:
Add a new SentrySampler that is used for sampling OpenTelemetry spans the Sentry way (using Sentrys `traces_sample_rate` and `traces_sampler` config options)

Fixes #3318


sentry_sdk/integrations/opentelemetry/sampler.py
```diff
@@ -0,0 +1,105 @@
...
+        client = sentry_sdk.get_client()
+
+        # No tracing enabled, thus no sampling
+        if not has_tracing_enabled(client.options):
+            return SamplingResult(Decision.DROP)
+
+        # Check if sampled=True was passed to start_transaction
+        # TODO-anton: Do we want to keep the start_transaction(sampled=True) thing?
+
+        parent_span = trace.get_current_span()
```
sl0thentr0py, line 55:
```suggestion
        parent_span = trace.get_current_span(parent_context)
```

usually `parent_context` is `None` so this falls back to your code 99% of the time, but there are use cases where it might be passed explicitly, so we need to use that

sentry_sdk/integrations/opentelemetry/sampler.py
```diff
@@ -0,0 +1,105 @@
...
+                "transaction_context": {
+                    "name": name,
+                },
+                "parent_sampled": parent_context.trace_flags.sampled,
+            }
+
+            sample_rate = client.options["traces_sampler"](sampling_context)
+
+        # Check if there is a parent with a sampling decision
+        if sample_rate is None:
```
sl0thentr0py, line 74: these need to if/elif to match with these semantics
```
        # we would have bailed already if neither `traces_sampler` nor
        # `traces_sample_rate` were defined, so one of these should work; prefer
        # the hook if so
        sample_rate = (
            client.options["traces_sampler"](sampling_context)
            if callable(client.options.get("traces_sampler"))
            else (
                # default inheritance behavior
                sampling_context["parent_sampled"]
                if sampling_context["parent_sampled"] is not None
                else client.options["traces_sample_rate"]
            )
        )
```
otherwise the behavior subtly changes, for instance when `traces_sampler` returns None, we do extra stuff now which shouldn't be done

sentry_sdk/integrations/opentelemetry/sampler.py
```diff
@@ -0,0 +1,105 @@
...
+            )
+            return SamplingResult(Decision.DROP)
+
+        # Roll the dice on sample rate
+        sampled = random() < float(sample_rate)
+
+        if sampled:
+            return SamplingResult(Decision.RECORD_AND_SAMPLE)
+        else:
+            parent_context.trace_state.update(SENTRY_TRACE_STATE_DROPPED, "true")
```
sl0thentr0py, line 101: why are you updating `parent_context.trace_state`, if I understand the otel sdk correctly, you need to pass the updated `trace_state` into the `SamplingResult` so that the children will have access to it

see
https://github.com/open-telemetry/opentelemetry-python/blob/794c51bb2d13c095e9078e2d0154b7147394798d/opentelemetry-sdk/src/opentelemetry/sdk/trace/sampling.py#L187

which is then picked up here
https://github.com/open-telemetry/opentelemetry-python/blob/794c51bb2d13c095e9078e2d0154b7147394798d/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L1160

sl0thentr0py (changes requested):
ty for taking care of this! left a few comments first

sentry_sdk/integrations/opentelemetry/sampler.py
antonpirker: oh! thanks for the pointer!

sentry_sdk/integrations/opentelemetry/sampler.py
antonpirker: done.

sentry_sdk/integrations/opentelemetry/sampler.py
antonpirker: done.

sentry_sdk/integrations/opentelemetry/sampler.py
antonpirker: fixed

sentry_sdk/integrations/opentelemetry/sampler.py
```diff
@@ -0,0 +1,117 @@
...
+        SENTRY_TRACE_STATE_DROPPED, "true"
+    )
+    return SamplingResult(
+        Decision.DROP,
+        trace_state=updated_trace_context,
+    )
+
+
+class SentrySampler(Sampler):
+    def should_sample(
```
sl0thentr0py, line 49: types need to be a comment annotation and not inline

sentry_sdk/integrations/opentelemetry/sampler.py
```diff
@@ -0,0 +1,117 @@
...
+            if parent_sampled is not None:
+                sample_rate = parent_sampled
+            else:
+                # Check if there is a traces_sample_rate
+                sample_rate = client.options.get("traces_sample_rate")
+
+        # If we still have no sample rate (but enable_tracing==True,
+        # because has_tracing_enabled returned True above), default to 1.0
+        if sample_rate is None:
+            sample_rate = 1.0
```
sl0thentr0py, line 99: this is not correct, `has_tracing_enabled` is a combination of `enable_tracing`, `traces_sample_rate` and `traces_sampler` and we convert `enable_tracing` to `traces_sample_rate` [here](https://github.com/getsentry/sentry-python/blob/ea2b2931c5ec816800d21307528111275525070a/sentry_sdk/client.py#L120-L121)

so please remove these 2 lines

sl0thentr0py (changes requested):

tests/integrations/opentelemetry/test_sampler.py
```diff
@@ -0,0 +1,351 @@
...
+
+def test_sampling_traces_sample_rate_50(init_sentry_with_potel, capture_envelopes):
+    init_sentry_with_potel(traces_sample_rate=0.5)
+
+    envelopes = capture_envelopes()
+
+    # Make sure random() always returns the same values
+    with mock.patch(
+        "sentry_sdk.integrations.opentelemetry.sampler.random",
+        side_effect=[0.7, 0.7, 0.7, 0.2, 0.2, 0.2],
```
sl0thentr0py, line 90: this one is not a good test, because we want the transaction to be complete and this is bypassing the `parent_sampled` logic testing for the child spans.

I would wrap each transaction individually in a `mock.patch` that only stubs the first root span random call, the children should then derive from the parent decision.

tests/integrations/opentelemetry/test_sampler.py
```diff
@@ -0,0 +1,351 @@
...
+
+        with sentry_sdk.start_span(description="request b"):  # drop
+            with sentry_sdk.start_span(description="cache b"):  # drop
+                with sentry_sdk.start_span(description="db b"):  # drop
+                    ...
+
+        with sentry_sdk.start_span(description="request c"):  # drop
+            with sentry_sdk.start_span(
+                description="cache a c"
+            ):  # keep (but trx dropped, so not collected)
```
sl0thentr0py, line 146: I don't like these mixed span drop/keep tests, the children span should just derive from the root span

sl0thentr0py (changes requested):

sentry_sdk/integrations/opentelemetry/sampler.py
antonpirker: done.

sentry_sdk/integrations/opentelemetry/sampler.py
antonpirker: done.

tests/integrations/opentelemetry/test_sampler.py
antonpirker: I changed them to make it easier readable. One Note: random() will be called on each span, because we retrieve the parent_sampled boolean and cast it to a float and use it as a sample rate. But you are right, the test looked clumsy :-)

tests/integrations/opentelemetry/test_sampler.py
antonpirker: When you have a `traces_sampler` it is run for every span and you can ignore the parent sampling decision in your custom `traces_sampler`. This is what I wanted to show here.

tests/integrations/opentelemetry/test_sampler.py
sl0thentr0py: ok yea it's better now, I just want to avoid stubbing the wrong thing because that kind of flow will never happen in reality. If `parent_sampled` is set, it will always be 1 and never 0.7 or 0.2

tests/integrations/opentelemetry/test_sampler.py
```diff
@@ -0,0 +1,351 @@
...
+
+tracer = trace.get_tracer(__name__)
+
+
+@pytest.fixture()
+def init_sentry_with_potel(sentry_init):
+    def wrapped_sentry_init(*args, **kwargs):
+        kwargs.update(
+            {
+                "_experiments": {"otel_powered_performance": True},
```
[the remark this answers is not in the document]
antonpirker: removed.

tests/integrations/opentelemetry/test_sampler.py
sl0thentr0py: ah ok, all good then

sl0thentr0py (approved):
perfect, ty!

PR #3503, opened 2024-09-05
Add feature flag tracking

cmanallen:
Related RFC: https://github.com/getsentry/rfcs/pull/140

TODO:

- [ ] Write a better PR description.
- [ ] Use an LRU cache instead of FIFO queue.
- [ ] Find the best place to append the flags to the error event.

Related to:
* https://github.com/getsentry/sentry/issues/59428

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -482,6 +482,9 @@ class _Client(BaseClient):
                     "event_processor", data_category="span", quantity=spans_delta
                 )
 
+        # TODO: Where should I put this?
+        event["contexts"]["flags"] = scope._flag_manager.serialize()
+
         if (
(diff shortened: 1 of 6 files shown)


sentry_sdk/client.py
```diff
@@ -482,6 +482,9 @@ def _prepare_event(
                     "event_processor", data_category="span", quantity=spans_delta
                 )
 
+        # TODO: Where should I put this?
+        event["contexts"]["flags"] = scope._flag_manager.serialize()
```
cmanallen, line 486: @antonpirker Could you help me fill in the gaps here?  Where should this live?  Also are scopes nested?  Would it be possible to merge the flags from a global scope and a thread scope here?

PR #3517, opened 2024-09-11
Remove 3.6 compat

sentrivana:
Remove compat for Python < 3.7


antonpirker (approved):
Its always a great feeling to delete code

PR #3519, opened 2024-09-11
Make sure op/name/description are set correctly on transactions/spans

antonpirker:
Make sure that `op`, `name`, `description` are set correctly on `Span`s and `Transaction`s.


sentry_sdk/tracing.py
```diff
@@ -1225,7 +1225,7 @@ def __init__(
                 start_timestamp = convert_to_otel_timestamp(start_timestamp)
 
             self._otel_span = tracer.start_span(
-                description or op or "", start_time=start_timestamp
+                name or description or op or "", start_time=start_timestamp
```
sl0thentr0py, line 1228: I actually don't know why we have all 3 here, could just be name? @sentrivana

antonpirker: Because in our api the `description` is the spans name: https://docs.sentry.io/platforms/python/tracing/instrumentation/custom-instrumentation/#using-a-context-manager

sl0thentr0py (approved):

sentrivana (approved):

PR #3523, opened 2024-09-11
Store Baggage object on PropagationContext instead of DSC hash

sl0thentr0py:


antonpirker (approved):
This looks good. (and also cleaner) 

Because the tests are broken atm we can not ensure if this changes behavior (Only form looking at it, I think it does not change any behavior)

sentrivana (approved):
lgtm

PR #3537, opened 2024-09-13
Fix timezones in breadcrumb sorting

antonpirker:
Just always use UTC.


antonpirker:
closing in favor of https://github.com/getsentry/sentry-python/pull/3537/files

PR #3545, opened 2024-09-17
ci: update actions/upload-artifact to v4 with merge

joshuarli:
Supersedes https://github.com/getsentry/sentry-python/pull/3381 using artifacts/merge to merge together multiple artifacts from different jobs into the single one named github.sha that craft expects.

Summarily, upload-artifact v3 is deprecated but v4 doesn't support mutating an artifact with the name name by uploading different filepaths to the same artifact. Because we need a single artifact "github.sha", we have to use actions/merge to create it. Alternatively craft could be modified but this is the easiest way forward and I like the idea of a unified artifact, it makes craft simpler.

ref: https://github.com/getsentry/craft/issues/552

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,7 @@ jobs:
     needs: [build_lambda_layer, docs]
     steps:
       - name: Merge Artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/merge@v4
         with:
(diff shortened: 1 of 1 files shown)


joshuarli:
need to fix this: `Warning: No files were found with the provided path: docs/_build/gh-pages.zip dist/*. No artifacts will be uploaded.`

sentrivana (approved):

PR #3550, opened 2024-09-20
ref(client): Improve `get_integration` typing

szokeasaurusrex:
Improve `get_integration` typing to make it clear that we return an `Optional[Integration]`. Further, add overloads to specify that when called with some integration type `I` (i.e. `I` is a subclass of `Integration`), then `get_integration` guarantees a return value of `Optional[I]`.

These changes should enhance type safety by explicitly guaranteeing the existing behavior of `get_integration`.

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -5,7 +5,7 @@ import socket
 from collections.abc import Mapping
 from datetime import datetime, timezone
 from importlib import import_module
-from typing import cast
+from typing import cast, overload
 
 from sentry_sdk._compat import PY37, check_uwsgi_thread_support
 from sentry_sdk.utils import (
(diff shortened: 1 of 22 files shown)


szokeasaurusrex:
@sentrivana yep still working on fixing mypy, will mark as draft until this is ready to be reviewed. I think the cleanest solution will be to replace `@ensure_integration_enabled` with the old manual checking that the integration is not `None` whenever we need to access the `Integration` object, because this allows `mypy` to recognize that the integration object cannot be `None`. 

We can still use `@ensure_integration_enabled` anywhere we don't need to access the `Integration` object

szokeasaurusrex:
The `mypy` errors led me to find a couple places where we were not checking that the integration was still enabled, so I think this change will potentially help us avoid making such mistakes in the future

sentrivana (approved):
Are we still using `ensure_integration_enabled` after this? If not, we can remove it.

szokeasaurusrex:
@sentrivana, yeah we are; I found ca. 100 usages after these changes. `@ensure_integration_enabled` is still helpful anytime we don't need access to the integration object within the patched function, which is quite often apparently. However, it is less useful when we need the integration object, since we anyways need to check whether the integration object is `None` in order to ensure type safety

PR #3557, opened 2024-09-23
Testing in CI

szokeasaurusrex:
Using this PR to debug a test failure in #3550 which is only occurring in CI


szokeasaurusrex:
no longer needed

PR #3558, opened 2024-09-23
ref(aiohttp): Make `DEFUALT_FAILED_REQUEST_STATUS_CODES` private

szokeasaurusrex:
There is no reason this constant should be part of the public API. Since no release has included this constant yet, making this constant private does not require a major version bump.


sentrivana (approved):
Honestly this would've been fine to leave as is -- I don't see any reason for anyone to use this outside of the SDK and it's not part of the advertised API. But I'm fine either way.

PR #3571, opened 2024-09-25
Fix record sql queries tests

antonpirker:
Use freezegun to create spans of a certain time duration. (avoids monkey patching the underlying otel span)


sentrivana (approved):
TIL about freezegun

PR #3572, opened 2024-09-25
allowing ASGI to use drf_request in DjangoRequestExtractor

PakawiNz:
since we already have patched a request object (both ASGI/WSGI) before arriving, we should move patched-using logic closer to where it actually being used. for minimize impact and allow ASGI functionality.

---


sentrivana (approved):
Thank you @PakawiNz! Makes sense to me.

PR #3580, opened 2024-09-27
fix(django): improve getting psycopg3 connection info

### LINKED ISSUE #3573: PostgreSQL getting connection info slowing down tracing
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.14.0

### Steps to Reproduce

1. Use sentry-sdk with Django integration.
2. Profile a view with doing several SQL queries.
3. Notice there is `expanduser` shown on top of “Slowest Application Functions” in Sentry.

### Expected Result

We don't invoke `expanduser` at all, so it should not be there. Definitely it should not be on top of the list, taking more time than template rendering.

### Actual Result

`expanduser` being identified by Sentry as the slowest thing on this view:

Image

Profile trace:

Image

This seems to be invoked for each `SQLCompiler.execute_sql` invocation, so for every SQL query.

It is invoked from here:

https://github.com/getsentry/sentry-python/blob/dce589ca49a8e0e2d4eda3839836de6d8188f17b/sentry_sdk/integrations/django/__init__.py#L705

The `expanduser` then comes from psycopg3 via `Path.home()`:

https://github.com/psycopg/psycopg/blob/af5bfa0ec2690571033e38205a98cfc8c6287134/psycopg/psycopg/_connection_info.py#L86

nijel:
Fetch the few needed parameters manually instead of relying on `get_parameters()` which adds visible overhead due to excluding default values for parameters.

Fixes #3573

---


nijel:
This might need additional changes, but I'd like to know if this is a viable direction to address https://github.com/getsentry/sentry-python/issues/3573.

I didn't investigate how `unix_socket` would be handled here (if at all).

szokeasaurusrex:
Thanks for the PR, @nijel! I just approved our test suit to run against it – let's see whether anything fails

szokeasaurusrex:
Hi @nijel, to be honest I am not super familiar with psycopg3 (someone else from the team wrote this part of the integration).

Could you please explain why you are only including the `dbname`, `host`, and `port` here? From [what I can tell](https://www.psycopg.org/psycopg3/docs/api/objects.html#psycopg.ConnectionInfo.get_parameters), `get_parameters` could return more information

nijel:
These are the only fields used in the code below (+ Unix socket which I didn't figure out how to obtain).

sentry_sdk/integrations/django/__init__.py
```diff
@@ -701,8 +701,14 @@ def _set_db_data(span, cursor_or_db):
...
-            # psycopg3
-            connection_params = cursor_or_db.connection.info.get_parameters()
+            # psycopg3, only extract needed params as get_parameters
+            # can be slow because of the additional logic to filter out default
+            # values
+            connection_params = {
+                "dbname": cursor_or_db.connection.info.dbname,
+                "host": cursor_or_db.connection.info.host,
+                "port": cursor_or_db.connection.info.port,
+            }
```
nijel, line 711:
```suggestion
            connection_params = {
                item.keyword.decode(): item.val.decode()
                for item in cursor_or_db.connection.info.pgconn.info
                if item.val is not None
            }
```

sentry_sdk/integrations/django/__init__.py
```diff
@@ -701,8 +701,14 @@ def _set_db_data(span, cursor_or_db):
         connection_params = cursor_or_db.connection.get_dsn_parameters()
     else:
         try:
-            # psycopg3
-            connection_params = cursor_or_db.connection.info.get_parameters()
+            # psycopg3, only extract needed params as get_parameters
+            # can be slow because of the additional logic to filter out default
+            # values
```
nijel, line 706:
```suggestion
            # psycopg3, extract current connection params as get_parameters
            # can be slow because of the additional logic to filter out default
            # values
```

nijel:
I've adjusted the code to expose all PostgreSQL connection parameters even when they are currently not used. I've also verified that `unix_socket` is never present with psycopg3, so the now code is completely consistent with get_parameters.

szokeasaurusrex (commented):
Can we go back to what you had in c862f55d4f72f4e637975e1d937bf180e2e79354? This `pgconn` stuff appears to be internal to psycopg3 because it is not super well documented from what I can tell, which makes me worried that this logic could break. 

Your original suggestion is fine with me; the only reason I was originally worried about it is because I did not look at the code outside the diff where we actually use the `connection_params`. As soon as you explained that we are only using some of the parameters, I realized I needed to expand that section, and now I see that you were correct that those are the only parameters we need.

Also, have you already tested this change? Did you notice a performance improvement?

szokeasaurusrex:
> The problem with the implementation in https://github.com/getsentry/sentry-python/commit/c862f55d4f72f4e637975e1d937bf180e2e79354 is that when using UNIX sockets, it returns part of the UNIX socket path as host (what might be a bug in psycopg3).

Could you share an example of how c862f55d4f72f4e637975e1d937bf180e2e79354 changes the `host`?

Would it be possible for us to trim this UNIX socket path out of the `host`? I would really prefer to avoid the `pgconn` stuff, as I could not find any documentation on it.

nijel:
`host` is `/var/run/postgresql` when obtained via `cursor_or_db.connection.info.host`. 

`cursor_or_db.connection.info.get_parameters()` returns in such case:

```
{'user': 'USER', 'dbname': 'DATABASE', 'client_encoding': 'UTF8', 'sslcertmode': 'allow'}
```

The new code returns the same values (+ some additional settings which are not used).

The `host` behaves this way because it calls `PQhost` and it is [documented](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQHOST) to return directory where socket is placed.

The alternate approach to the current patch would be to use `cursor_or_db.connection.info.host` as `host` only in case it does not start with `/`.

szokeasaurusrex:
@nijel Just making sure I understand correctly:

1. `cursor_or_db.connection.info.host` can return an absolute path for the host, but `cursor_or_db.connection.info.get_parameters()` cannot
2. If `cursor_or_db.connection.info.host` returns an absolute path (starting with `/`), then the `host` will be missing from `cursor_or_db.connection.info.get_parameters()`.
3. If `cursor_or_db.connection.info.host` returns something that is not an absolute path (not starting with `/`), then the `host` will be contained in `cursor_or_db.connection.info.get_parameters()` and will have the same value as `cursor_or_db.connection.info.host`.

Are all of those three points correct? If yes, then I think we should go with your suggestion, where we use https://github.com/getsentry/sentry-python/commit/c862f55d4f72f4e637975e1d937bf180e2e79354 but only add the `host` if it does not start with `/`:

> The alternate approach to the current patch would be to use `cursor_or_db.connection.info.host` as `host` only in case it does not start with `/`.

nijel:
Yes, all points are correct. I've adjusted the code accordingly.

szokeasaurusrex (approved):
Looks good.

Thanks for the contribution @nijel!

PR #3587, opened 2024-09-30
Remove `runtox.sh`

antonpirker:
The functionality of `runtox.sh` can be also done with `tox -f` so I will remove the shell script to simplify things.


antonpirker:
There is one problem with this: We group different frameworks together in one YAML file (like all database integrations) and those integrations need different Python versions to test on. The YAML has specified a union of all the Python verisons needed. 

When we run `tox -f` for a Python version and framework combination that is not specified in `tox.ini` then the tox command fails with the same exit code then it would when the test are run but failing. So we can not decide on that level if the Github actions failed or succeeded (meaning no tests to run, continue).

So I will park this PR for now.

antonpirker:
will not do this right now. Closing

PR #3592, opened 2024-10-01
Skip one of the Lambda tests

antonpirker:
AWS Lambda has changed something in their environment and now our tests can not capture events in the init phase of the Lambda function.

Will skip the test for now.


sentrivana (approved):

antonpirker:
AWS Lambda has changed something in their environment and now our tests can not capture events in the init phase of the Lambda function.

PR #3593, opened 2024-10-01
Fix Scope tests

antonpirker:
Fix the Scope related tests in Potel.


antonpirker:
In POTel this only works when there is this `with sentry_sdk.start_span()` in the test, because this is the point where our scopes are put on the OTel context.

This should also work without having a span. @sl0thentr0py said he will implement a switch to switch to Otel and in that case the Sentry scopes will be put on the OTel context during SDK init(). 

So I will leave this Issue here until it is time to implement that switch.

antonpirker:
This is now outdated. Will close.

PR #3599, opened 2024-10-02
Head SDK DSC population

### LINKED ISSUE #3479: Make DSC/baggage work with head SDK
figure out how to populate the baggage and update the trace_state

sl0thentr0py:
* Populates a DSC with correct values when we don't have an incoming trace.
* We rely on `trace_state.add` only adding new keys to the tracestate so these values will be populated in the first samplign decision on the root and just be propagated further.

Note that transaction name is missing here for now and will be dealt with separately as part of the transaction name PRs.

closes #3479


sentrivana (approved):

sl0thentr0py:
tested both head SDK and that child SDK setups continue to work

head SDK screenshot:

image

PR #3602, opened 2024-10-03
Enable back pressure monitor in Potel

antonpirker:
This enables back pressure management in POtel. 

It down samples in case of the transport not being healthy and also sends outcomes to Sentry.

Right now the sampling is done on span level, but should be done on transaction level, which is not possible right now, because we do not have a reference to the transaction (root span) in the sampler right now. This will be done in a separate PR.

The test that is updated in this PR still fails, because it needs the sample_rate to be somewhere on the Span, but this will be done in a separate PR.


sentry_sdk/integrations/opentelemetry/sampler.py
```diff
@@ -45,6 +45,18 @@ def dropped_result(span_context):
     # type: (SpanContext) -> SamplingResult
     trace_state = span_context.trace_state.update(SENTRY_TRACE_STATE_DROPPED, "true")
 
+    # Tell Sentry why we dropped the transaction/span
```
sl0thentr0py, line 48: to match the logic in the earlier tracing impl, we should rather do this at the end of the transaction flow, so I'd not do this here but in span processor `on_end` and only for the root span case. Here it will apply to every span and the numbers will be wrong.

sl0thentr0py (changes requested):

antonpirker:
If the transaction is not sampled, the `span_processor.on_end()` is not run, so we need to do this in the sampler. 
I have now updated the logic, to only emit the outcomes for transactions/root-spans.

sentry_sdk/integrations/opentelemetry/potel_span_processor.py
```diff
@@ -59,11 +59,12 @@ def on_end(self, span):
         if is_sentry_span(span):
             return
 
-        if span.parent and not span.parent.is_remote:
-            self._children_spans[span.parent.span_id].append(span)
-        else:
+        is_root_span = not (span.parent and not span.parent.is_remote)
```
sl0thentr0py, line 62:
```suggestion
        is_root_span = not span.parent or span.parent.is_remote
```

sentry_sdk/integrations/opentelemetry/sampler.py
```diff
@@ -54,6 +54,20 @@ def dropped_result(span_context, attributes, sample_rate=None):
     if sample_rate:
         trace_state = trace_state.add(TRACESTATE_SAMPLE_RATE_KEY, str(sample_rate))
 
+    is_root_span = not (span_context.is_valid and not span_context.is_remote)
```
sl0thentr0py, line 57: if `span_context` is not valid, why is it a root span?

sentry_sdk/integrations/opentelemetry/sampler.py
```diff
@@ -124,6 +138,11 @@ def should_sample(
                 # Check if there is a traces_sample_rate
                 sample_rate = client.options.get("traces_sample_rate")
 
+        # Down-sample in case of back pressure monitor says so
+        # TODO: this should only be done for transactions (aka root spans)
+        if client.monitor:
```
sl0thentr0py, line 143: move this below the `is_valid_sample_rate` block, just before the actual random call is made

sl0thentr0py (changes requested):

sentry_sdk/integrations/opentelemetry/potel_span_processor.py
antonpirker: its the same thing, but I can update (rule of de morgan or however it is called in english)

sentry_sdk/integrations/opentelemetry/sampler.py
antonpirker: makes sense

sentry_sdk/integrations/opentelemetry/sampler.py
antonpirker: Its the parent span context, updated the naming so it is not confusing.

sl0thentr0py (approved):

PR #3603, opened 2024-10-03
Fix type of sample_rate in DSC (and add explanatory tests)

### LINKED ISSUE #3601: Trace sample rate in DSC not a string
See also: https://github.com/getsentry/relay/issues/4094

The sample rate in the DSC must be a string according to the [spec](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#dsc-specification).

After a quick investigation with @antonpirker, it seems like the Python SDK just uses the `traces_sample_rate` from the option and propagates it as a float/integer: https://github.com/getsentry/sentry-python/blob/5080c7614bd1cfb315c5bf7d5e8833ca01ca9e41/sentry_sdk/tracing_utils.py#L534-L535

antonpirker:
In the DSC send in the envelope header for envelopes containing errors the type of `sample_rate` was `float` instead of the correct `str` type.

Fixes #3601

### DIFF
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -532,7 +532,7 @@ class Baggage:
             sentry_items["public_key"] = Dsn(options["dsn"]).public_key
 
         if options.get("traces_sample_rate"):
-            sentry_items["sample_rate"] = options["traces_sample_rate"]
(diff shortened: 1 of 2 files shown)


sentrivana (approved):
🥔 🥔 🥔

PR #3605, opened 2024-10-03
ref(bottle): Delete never-reached code

szokeasaurusrex:
The `prepared_callback` should never raise an `HTTPResponse` exception because `prepared_callback` is already decorated by Bottle using a `@route` decorator (or a decorator for the specific HTTP methods, e.g. `@get`). This decorated function never raises `HTTPResponse`, because the `@route` wrapper [captures any `HTTPResponse` exception and converts it into the return value](https://github.com/bottlepy/bottle/blob/cb36a7d83dc560e81dd131a365ee09db2f756a52/bottle.py#L2006-L2009). So, we do not need this code and should delete it.


antonpirker (approved):
looks good, thanks!

PR #3608, opened 2024-10-03
Maybe this pin is not required anymore

antonpirker:
Trying if we can remove this


antonpirker:
ok, it is still required.

PR #3615, opened 2024-10-05
Fix Anthropic integration when using tool calls

kwnath:
# Description

If you've initialized Sentry with Anthropic integration, streaming responses with [tool calls](https://docs.anthropic.com/en/docs/build-with-claude/tool-use) fail. 

The reason is quite simple, the API introduced a new data model for a `ContentBlockDeltaEvent`. This new refactor was made in `0.27.0` because json was needed to be streamed back (so arguments were returned as json for tool calls, I think this was the initial reason). This introduced a new data type [InputJsonDelta](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/raw_content_block_delta_event.py#L13) that had it's contents in `partial_json` instead of what is currently supported `text` ([reference](https://github.com/anthropics/anthropic-sdk-python/blob/v0.25.0/src/anthropic/types/content_block_delta_event.py#L12)).

And you'll find in the original integration we try to get the `text` if the type is a `content_block_delta`, this doesn't exist if you're doing any sort of tool calling with streaming. This doesn't fail however if you turn off streaming.

More info on reproduction steps [here](https://github.com/anthropics/anthropic-sdk-python/issues/658#issuecomment-2393687417)

## Side notes

- Backwards compatibility is probably not that important but I'd rather be safe here.
- I've only updated this for `ContentBlockDeltaEvent`, there are likely other incompatible areas but this is the most breaking one. (I can add a few follow up PR's)

### DIFF
--- a/sentry_sdk/integrations/anthropic.py
+++ b/sentry_sdk/integrations/anthropic.py
@@ -1,4 +1,5 @@
 from functools import wraps
+from typing import TYPE_CHECKING
 
 import sentry_sdk
 from sentry_sdk.ai.monitoring import record_token_usage
@@ -10,8 +11,6 @@ from sentry_sdk.utils import (
     event_from_exception,
     package_version,
 )
-
-from typing import TYPE_CHECKING
 
 try:
     from anthropic.resources import Messages
@@ -46,7 +45,10 @@ class AnthropicIntegration(Integration):
         if version < (0, 16):
             raise DidNotEnable("anthropic 0.16 or newer required.")
 
-        Messages.create = _wrap_message_create(Messages.create)
+        if version >= (0, 27, 0):
+            Messages.create = _wrap_message_create(Messages.create)
+        else:
+            Messages.create = _wrap_message_create_old(Messages.create)
 
 
 def _capture_exception(exc):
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/anthropic.py
```diff
@@ -46,7 +45,10 @@ def setup_once():
         if version < (0, 16):
             raise DidNotEnable("anthropic 0.16 or newer required.")
 
-        Messages.create = _wrap_message_create(Messages.create)
+        if version >= (0, 27, 0):
+            Messages.create = _wrap_message_create(Messages.create)
+        else:
+            Messages.create = _wrap_message_create_old(Messages.create)
```
kwnath, line 51: This is probably unnecessary but I'm not sure how the maintainers feel about separating major changes in Anthropic python sdk.

sentry_sdk/integrations/anthropic.py
```diff
@@ -75,6 +77,112 @@ def _calculate_token_usage(result, span):
...
+                                    usage = event.message.usage
+                                    input_tokens += usage.input_tokens
+                                    output_tokens += usage.output_tokens
+                                elif event.type == "content_block_start":
+                                    pass
+                                elif event.type == "content_block_delta":
+                                    if hasattr(event.delta, "text"):
+                                        content_blocks.append(event.delta.text)
+                                    elif hasattr(event.delta, "partial_json"):
+                                        content_blocks.append(event.delta.partial_json)
```
kwnath, line 153: Opted for checking whether the delta contained x key instead of checking the instance type. Importing the instances would be a bit tricky as we'll need to wrap it in a try block as `InputJsonDelta` doesn't exist for Anthropic versions <0.27.

kwnath:
@antonpirker is this something you might have some time to look at 😅

antonpirker:
Will have a look at on Thursday

sentry_sdk/integrations/anthropic.py
antonpirker: I took the liberty to commit to your branch. When we only handle the "text" responses then it works the same in old and new Anthropic so we do not need this if (and the _wrap_message_create_old function)

antonpirker:
Hey @kwnath !

Thanks again for your contribution! I finally had some time to look at your PR. I took the liberty to change some stuff to bring it over the finish line. When all the tests have passed, this is ready to be merged!

kwnath:
@antonpirker Looks like the codecov/patch is stuck 😅

antonpirker (approved):
Looks good!

antonpirker:
@kwnath your fix has been released in this version: https://pypi.org/project/sentry-sdk/2.17.0/

PR #3616, opened 2024-10-05
fix(integrations): Don't capture StopIteration in SentryWsgiMiddleware

cstranex:
Currently Sentry will emit an unhandled exception for StopIteration. However,
when running certain combinations of libraries with gunicorn and sync workers
(eg: python-socketio) the StopIteration exception is actually expected. To
workaround this issue, don't capture this exception when using this middleware
with gunicorn and just reraise it.

See discussion about StopIteration here: https://github.com/miguelgrinberg/python-engineio/discussions/335 and https://github.com/miguelgrinberg/flask-sock/issues/64

### DIFF
--- a/sentry_sdk/integrations/wsgi.py
+++ b/sentry_sdk/integrations/wsgi.py
@@ -134,6 +134,11 @@ class SentryWsgiMiddleware:
                                     _sentry_start_response, start_response, transaction
                                 ),
                             )
+                        except StopIteration:
+                            if "gunicorn" in environ.get("SERVER_SOFTWARE", ""):
+                                raise
+                            else:
+                                reraise(*_capture_exception())
                         except BaseException:
                             reraise(*_capture_exception())
         finally:
--- a/tests/integrations/wsgi/test_wsgi.py
+++ b/tests/integrations/wsgi/test_wsgi.py
@@ -2,7 +2,7 @@ from collections import Counter
 from unittest import mock
 
 import pytest
-from werkzeug.test import Client
+from werkzeug.test import Client, EnvironBuilder
 
 import sentry_sdk
 from sentry_sdk import capture_message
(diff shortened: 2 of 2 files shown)


cstranex:
@szokeasaurusrex Hmm, for the two libraries linked the `StopIteration` exception has a blank message so it will be difficult to know exactly where it comes from without asking the library maintainers to make some changes for us to be able to identify if it should be caught. I had a similar idea using the stack trace but I think that's maybe bit too brittle.

At least for the case I'm looking at we are opening websocket connections and the `StopIteration` gets issued to request gunicorn to close the socket when a client disconnects. If we are only dealing with this case then we can inspect the WSGI environ for an `Connection: Upgrade` header which should only present for websockets. If it is present then we can ignore the `StopIteration`, otherwise we capture it?

cstranex:
I suppose another way to handle it could be a way to specify uncaught exceptions that should be ignored either by a list or by a callback

szokeasaurusrex:
> I suppose another way to handle it could be a way to specify uncaught exceptions that should be ignored either by a list or by a callback

@cstranex If all you want to do is have these `StopIteration` exceptions be suppressed for your setup, you can simply use a custom [Using before_send](https://docs.sentry.io/platforms/python/configuration/filtering/#using-before-send) to filter out the `StopIteration` exceptions you wish to ignore.

If this is not a very widespread problem, and if this solution works for you, I think just having users who wish to ignore these exceptions configure a custom `before_send` would be the best solution. The logic to infer when we can safely ignore a `StopIteration` exception will likely always be a bit brittle, so having users configure the logic to their needs is maybe the most robust and least error prone option here.

cstranex:
Fair enough, I was thinking it is probably better to ignore it universally since it is somewhat expected for gunicorn to handle it, but I agree it is probably not common enough that it could just be filtered out on a per usage basis

PR #3623, opened 2024-10-08
feat(integrations): Add integration for qdrant

mxrcooo:
Adds an integration for Qdrant, supporting both REST and gRPC mode.
---

Qdrant is a vector-search database. Mentioned here: https://github.com/getsentry/sentry-python/discussions/3007#discussioncomment-9251889

Opening this as draft PR as communicated in the community Discord since there are still a few open questions from my side. Also, there's a lot of work (tests) to be done to finalize this.

1) The Qdrant service offers two APIs: REST and gRPC. They are almost the same in terms of data sent and regarding Qdrant SDK usage they only differ on the `prefer_grpc` parameter when creating the (Async)QdrantClient. While the data sent to the server is almost the same, they are still structurally different. Question: I currently handle this by simply using different `op` arguments (`db.qdrant.rest` and `db.qdrant.grpc` respectively) when creating the span. Is that fine or should they be merged? If yes, what about the description? They differ and imo it is not clean to merge them.

2) The HttpxIntegration captures the REST request caused by a Qdrant SDK call, leading to an almost duplicate span with less information than the one from our integration. QdrantIntegration offers the ability to mute this span which is done by [accessing the _span_recorder of the current transaction and removing subsequent span from our current span](https://github.com/mxrcooo/sentry-python/blob/d4a30b52700534b50e6d9f862af56a54e50f8ce5/sentry_sdk/integrations/qdrant/qdrant.py#L61-L104). This feels very hacky.. is there a better way? Should this be done at all? Same goes for the GRPCIntegration when using Qdrant in gRPC mode.

2) Any ideas on writing tests for this? Qdrant supports a `:memory:` option but the monkey patches do not apply in this case, since it doesn't simulate a server but instead does all operations locally (see https://github.com/qdrant/qdrant-client/tree/master/qdrant_client/local). Mocking the responses would work but would be extremely cumbersome as - I think - we'd have to write different mocks for every single endpoint to not break the Qdrant SDK when handling the response. This would also have to be done twice, once for REST and once for gRPC. I guess we could parse their docs and auto-generate mock responses?


sentry_sdk/integrations/qdrant/qdrant.py
```diff
@@ -0,0 +1,523 @@
...
+            self.args = None
+            self.kwargs = None
+
+    # Callable[anything] will return None
+    class _Callable:
+        def __getitem__(self, _):
+            return None
+
+    # Make instances
+    Callable = _Callable()
```
szokeasaurusrex, line 56: You can (and should) remove this, because you are using type comments.

```suggestion
```

sentry_sdk/integrations/qdrant/qdrant.py
```diff
@@ -0,0 +1,523 @@
...
+    _DISALLOWED_REST_FIELDS,
+    _DISALLOWED_PROTO_FIELDS,
+    _qdrant_trie,
+    _IDENTIFIER,
+)
+
+# Hack to get new Python features working in older versions
+# without introducing a hard dependency on `typing_extensions`
+# from: https://stackoverflow.com/a/71944042/300572
+# taken from sentry_sdk.integrations.grpc.__init__.py
```
szokeasaurusrex, line 26: Not needed, see comment below

```suggestion
```

sentry_sdk/integrations/qdrant/qdrant.py
```diff
@@ -0,0 +1,523 @@
...
+        Optional,
+        Sequence,
+        Generator,
+    )
+    from grpc import Channel, ClientCallDetails
+    from grpc.aio import UnaryUnaryCall
+    from grpc.aio import Channel as AsyncChannel
+    from google.protobuf.message import Message
+    from sentry_sdk.tracing import Span, Transaction, _SpanRecorder
+    from sentry_sdk.integrations.qdrant import QdrantIntegration
```
szokeasaurusrex, line 42: Are these only being used in type comments? If not, they should be imported outside the `if TYPE_CHECKING` block

sentry_sdk/integrations/qdrant/qdrant.py
```diff
@@ -0,0 +1,523 @@
...
+            return f(*args, **kwargs)
+
+        request = args[1]  # type: Request
+        with _prepare_span(request, integration.mute_children_http_spans):
+            return await f(*args, **kwargs)
+
+    return wrapper
+
+
+# taken from grpc integration
```
szokeasaurusrex, line 178: Why not just use the grpc integration (or use this code from the grpc integration) directly?

sentry_sdk/integrations/qdrant/qdrant.py
```diff
@@ -0,0 +1,523 @@
...
+
+            response = await continuation(updated_call_details, request)
+
+            status_code = await response.code()
+            span.set_data("code", status_code.name)
+
+            return response
+
+
+def _protobuf_to_dict(message, prefix=""):
```
szokeasaurusrex, line 353: Does GRPC not provide a built-in way to do this?

szokeasaurusrex (commented):
Left a few suggestions, where I think we could simplify the logic. I also replied to your questions on Discord

PR #3625, opened 2024-10-08
test(profiler): Fork flaky test

szokeasaurusrex:
This test is flaky; Codecov has even identified it as such. Let's try forking the test, maybe that will make it more stable.


antonpirker (approved):
lgtm

szokeasaurusrex:
Forking this test causes a seemingly unrelated test to fail. Marking as draft pending further investigation

PR #3637, opened 2024-10-09
Move `subprocess` breadcrumbs from `maybe_create_breadcrumbs_from_span` to integration.

antonpirker:
Move `subprocess` breadcrumbs from `maybe_create_breadcrumbs_from_span` into the stdlib integration and preserve the breadcrumb behavior in POTel.


sentrivana (approved):
💯

PR #3641, opened 2024-10-10
fix: set default Scope._transaction_info value to None

andrewsapw:
In [#1490](https://github.com/getsentry/sentry-python/pull/1490), the `._transaction_info` property was added to the `Scope` class. However, there are some issues with this property:

- It causes the error transaction_info: Discarded unknown attribute in older versions of the UI.
- There is inconsistency in how the field is handled. Although it is marked as Mapping[str, str], in some places (for example, [here](https://github.com/getsentry/sentry-python/pull/1490/files#diff-c935006d78056bc518dd6e898d31d972c130c63119df74edf122e2cf4a5733bcR388)), it is checked against None.

This PR sets the default value of `Scope._transaction_info` to `None`.


szokeasaurusrex:
@andrewsapw Could you please help me understand the problem you are experiencing?

> It causes the error transaction_info: Discarded unknown attribute in older versions of the UI.

Regarding this error, could you please share a screenshot of the error message (or a link to the page, if you are using Sentry's SaaS offering)? 

> There is inconsistency in how the field is handled. Although it is marked as Mapping[str, str], in some places (for example, [here](https://github.com/getsentry/sentry-python/pull/1490/files#diff-c935006d78056bc518dd6e898d31d972c130c63119df74edf122e2cf4a5733bcR388)), it is checked against None.

We can likely remove these checks against `None`, unless we can find somewhere where the field is set to `None`.

andrewsapw:
[szokeasaurusrex](https://github.com/szokeasaurusrex) of course! Screenshot of the error:
image

szokeasaurusrex:
@andrewsapw Could you please also share a code snippet that reproduces this issue? And what version of Sentry self-hosted are you using?

andrewsapw:
@szokeasaurusrex , example of code that reproduces this issue:

```python
sentry_sdk.capture_message("[INFO] Test .capture_message", level="info")
```

Version of Sentry: 22.6.0 e68da5c

PR #3665, opened 2024-10-16
fix(spotlight): More defensive Django spotlight middleware injection

BYK:
Turns out `settings.MIDDLEWARE` does not have to be a `list`. This causes issues as not all iterables support appending items to them. This PR leverages `itertools.chain` along with `type(settings.MIDDLEWARE)` to extend the middleware list while keeping its original type. It also adds a try-except block around the injection code to make sure it doesn't block anything further down in the unexpected case that it fails.

### DIFF
--- a/sentry_sdk/spotlight.py
+++ b/sentry_sdk/spotlight.py
@@ -5,6 +5,8 @@ import urllib.request
 import urllib.error
 import urllib3
 
+from itertools import chain
+
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -18,6 +20,7 @@ from sentry_sdk.envelope import Envelope
 
 
 DEFAULT_SPOTLIGHT_URL = "http://localhost:8969/stream"
+DJANGO_SPOTLIGHT_MIDDLEWARE_PATH = "sentry_sdk.spotlight.SpotlightMiddleware"
 
 
 class SpotlightClient:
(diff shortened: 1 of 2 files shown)


sentry_sdk/spotlight.py
```diff
@@ -112,9 +115,18 @@ def setup_spotlight(options):
     else:
         return None
 
-    if settings is not None and env_to_bool(
-        os.environ.get("SENTRY_SPOTLIGHT_ON_ERROR", "1")
+    if (
+        settings is not None
```
untitaker, line 119: can use `with capture_internal_exception`, we spray that liberally and it has the added benefit of being automatically disabled in sdk's own testsuite

sentry_sdk/spotlight.py
```diff
@@ -112,9 +115,18 @@ def setup_spotlight(options):
...
+        and settings.DEBUG
+        and env_to_bool(os.environ.get("SENTRY_SPOTLIGHT_ON_ERROR", "1"))
     ):
-        settings.MIDDLEWARE.append("sentry_sdk.spotlight.SpotlightMiddleware")
+        try:
+            middleware = settings.MIDDLEWARE
+            if DJANGO_SPOTLIGHT_MIDDLEWARE_PATH not in middleware:
+                settings.MIDDLEWARE = type(middleware)(
```
untitaker, line 126: i'd always coerce to list instead of this (but can't articulate why, it just "feels less magical" to me)

untitaker (approved):

sentry_sdk/spotlight.py
BYK: I was gonna do that but then remembered about fiddling with people's settings. I wanna be as hands-off as possible.

sentrivana (approved):

PR #3669, opened 2024-10-17
fix(profiling): Update active thread for asgi

Zylphrex:
Ensure the handling thread is set on the transaction for asgi transactions not just main thread.


antonpirker (approved):
Looks good. Thanks for updating the tests everywhere.

PR #3673, opened 2024-10-18
Add missing stack frames

### LINKED ISSUE #3646: Missing execution stack and output for handled errors
When capturing errors, the Sentry stack traces only contain the frames from when an error happens. It does not contain any frames from the beginning of the execution.

This screenshot shows that only lines 25 and 17 are tracked (see [1] for full code; [link](https://sentry-test.sentry.io/issues/5984059407/?project=6730869&query=is%3Aunresolved&referrer=issue-stream&sort=date&statsPeriod=1h&stream_index=1) to event). Lines 11 and 29 are missing:
Image

You can also notice that *all* breadcrumbs are missing [2]. 

You can check how I hacked a script [3] to give those extra missing frames by using `traceback.extract_stack()`; if you're curious.
[link to event](https://sentry-test.sentry.io/issues/5984061529/?project=6730869&query=is%3Aunresolved&referrer=issue-stream&sort=date&statsPeriod=24h&stream_index=0)
Image

[1]
```python
import os
import traceback

import sentry_sdk
from sentry_sdk import capture_exception

sentry_sdk.init(dsn=os.environ.get("SENTRY_DSN"))


def main():
    foo()  # line 11


def foo():
    try:
        print("Message before exception")
        bar()  # line 17
    except Exception:
        print("Message after capturing exception")
        traceback.print_exc()
        capture_exception()


def bar():
    raise Exception("This is a test exception")  # line 25


if __name__ == "__main__":
    main()  # line 29
```

[2] Output
```
Message before exception
Message after capturing exception
  File "/Users/armenzg/code/learning/scripts/foo.py", line 16, in foo
    bar()
  File "/Users/armenzg/code/learning/scripts/foo.py", line 23, in bar
    raise Exception("This is a test exception")
Exception: This is a test exception
```

[3]
```python
import os
import sys
import traceback

import sentry_sdk
from sentry_sdk import capture_exception

# For Python 3.11+
if sys.version_info >= (3, 11):
    from builtins import ExceptionGroup
else:
    # For Python versions before 3.11
    from exceptiongroup import ExceptionGroup

sentry_sdk.init(dsn=os.environ.get("SENTRY_DSN"))


class CustomException(Exception):
    def __init__(self, message, tb_frames):
        super().__init__(message)
        self.tb_frames = tb_frames

    def __str__(self):
        frames = []
        for f in self.tb_frames[:-1]:
            # File "sentry_capture_exception.py", line 34, in foo
            line1 = f'  File "{f[0]}", line {f[1]}, in {f[2]}'
            line2 = f"    {f[3]}"
            frames.append("\n".join([line1, line2]))
        tb_str = "\n".join(frames)
        # Sample:
        # Exception: This is a test exception
        #   File "sentry_capture_exception.py", line 40, in foo
        #     bar()
        #   File "sentry_capture_exception.py", line 58, in bar
        #     raise Exception("This is a test exception")
        return f"{self.__class__.__name__}: foo\n{tb_str}"


def main():
    foo()


def foo():
    try:
        bar()
    except Exception as original_error:
        stack_trace = traceback.extract_stack()
        stack_exception = CustomException("An error occurred", stack_trace)

        # Create an ExceptionGroup with the original error and traceback exception
        exception_group = ExceptionGroup(
            "Error occurred in foo()", [original_error, stack_exception]
        )

        # Capture the ExceptionGroup
        capture_exception(exception_group)


def bar():
    raise Exception("This is a test exception")


if __name__ == "__main__":
    main()
```

antonpirker:
Add a new `init()` option `add_full_stack` (default `False`), when set to `True` it will add all the missing frames from the beginning of the execution to the stack trace sent to Sentry. 

Also adds another option `max_stack_frames` (default `100`) to limit the number of frames sent. The limitation is only enforced when `add_full_stack=True` to not change behavior for existing users. 

Documentation update: https://github.com/getsentry/sentry-docs/pull/11844

Fixes #3646


sentry_sdk/utils.py
```diff
@@ -1096,6 +1137,45 @@ def exc_info_from_error(error):
...
+    # type: () -> List[Dict[str, Any]]
+    """
+    Returns a serialized representation of the full stack from the first frame that is not in sentry_sdk.
+    """
+    try:
+        # Raise an exception to capture the current stack
+        raise Exception
+    except:
+        # Get the current stack frame
+        _, _, tb = sys.exc_info()
```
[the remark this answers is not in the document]
antonpirker: Nice, thanks for the `sys._getframe()` hint!

sentry_sdk/utils.py
```diff
@@ -1096,6 +1119,73 @@ def exc_info_from_error(error):
...
+    """
+    stack_info = []
+
+    # Walk up the stack
+    frame = sys._getframe(1)  # type: Optional[FrameType]
+    while frame:
+        in_sdk = False
+        try:
+            if "sentry_sdk" in frame.f_code.co_filename:
+                in_sdk = True
```
asottile-sentry, line 1135: might be better to use f_globals['module'] as this is a pretty weak assertion currently

sentry_sdk/utils.py
```diff
@@ -1096,6 +1119,73 @@ def exc_info_from_error(error):
     return exc_info
 
 
+def get_full_stack():
```
asottile-sentry, line 1122: should this replace the code in the logging plugin which does similar?

asottile-sentry (approved):
while the implementation seems fine -- I'm not sure this is going to be super useful being a global option -- I would imagine this is going to be more useful on a callsite-by-callsite basis?

PR #3678, opened 2024-10-21
fix(HTTP2Transport): Only enable HTTP2 when DSN is HTTPS

BYK:


sentry_sdk/transport.py
```diff
@@ -783,10 +767,11 @@ def _request(
...
             return response
 
-        def _get_pool_options(self, ca_certs, cert_file=None, key_file=None):
-            # type: (Any, Any, Any) -> Dict[str, Any]
+        def _get_pool_options(self):
+            # type: (Self) -> Dict[str, Any]
             options = {
-                "http2": True,
+                "http2": self.parsed_dsn is not None
+                and self.parsed_dsn.scheme == "https",
```
BYK, line 774: This is the core change. All the rest is refactoring to make this easier to implement.

PR #3680, opened 2024-10-22
Fix breadcrumbs in redis

antonpirker:
This moves the creation of `redis` breadcrumbs from the `maybe_create_breadcrumbs_from_span` into the integrations. And as well fixes some span related tests by using `render_span_tree`.


sentrivana (approved):
Kudos for keeping this readable. I think this is a good level of abstraction.

PR #3693, opened 2024-10-28
build: Remove Pytest pin

### LINKED ISSUE #3035: Unpin pytest
### Problem Statement

As a Fedora packager I would like to have pytest to not have pin `< 7.0.0`, so that `python-sentry-sdk` could be tested in Fedora without workarounds. Also, `pytest==7.0.0` was [released](https://pypi.org/project/pytest/7.0.0/) on Feb 4, 2022.

All actual Fedora releases [contain](https://src.fedoraproject.org/rpms/pytest) `pytest>7.0.0` and installing an older version from PyPI during package build for testing is not possible. Thus, I [unpinned it](https://src.fedoraproject.org/rpms/python-sentry-sdk/blob/24c0bc5851e5d9832ee617f42bca59a084e607ac/f/python-sentry-sdk.spec#_241). Since some tests fail with `pytest>7.0.0` because of the [issue](https://github.com/pytest-dev/pytest-forked/issues/67) in `pytest-forked`, I had to find a way how to fix them. Luckily, thanks to [this investigation](https://github.com/pytest-dev/pytest-forked/issues/67#issuecomment-1964718720) there is a workaround: reorder the tests so that non-forked tests are not last in the file, see [patch](https://src.fedoraproject.org/rpms/python-sentry-sdk/blob/24c0bc5851e5d9832ee617f42bca59a084e607ac/f/0001-Reorder-forked-tests.patch).

### Solution Brainstorm

Of course, the best solution would be to fix the [issue](https://github.com/pytest-dev/pytest-forked/issues/67) in `pytest-forked`, however it looks like a hard problem.

My suggestions:
* Unpin pytest.
* Reorder the tests.
* Write a note in `CONTRIBUTING.md` about this quirk. Another option is to add a linter that would inform about the need to reorder tests.

szokeasaurusrex:
Remove Pytest pin. This requires ensuring that any test modules which contain both forked and non-forked tests do not have a forked test as the last test in the module. See [here](https://github.com/pytest-dev/pytest-forked/issues/67#issuecomment-1964718720).

Closes #3035

### DIFF
--- a/requirements-devenv.txt
+++ b/requirements-devenv.txt
@@ -1,5 +1,5 @@
 -r requirements-linting.txt
 -r requirements-testing.txt
 mockupdb # required by `pymongo` tests that are enabled by `pymongo` from linter requirements
-pytest<7.0.0 # https://github.com/pytest-dev/pytest/issues/9621; see tox.ini
+pytest
 pytest-asyncio
--- a/tests/integrations/celery/integration_tests/test_celery_beat_cron_monitoring.py
+++ b/tests/integrations/celery/integration_tests/test_celery_beat_cron_monitoring.py
@@ -151,3 +151,10 @@ def test_beat_task_crons_error(celery_init, capture_envelopes):
     assert check_in["type"] == "check_in"
     assert check_in["monitor_slug"] == "failure_from_beat"
     assert check_in["status"] == "error"
+
+
+def test_placeholder():
+    """Forked test cannot be last in a module.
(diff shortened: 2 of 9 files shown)


szokeasaurusrex:
All tests have already passed; CI is now rerunning because I rebased on master

szokeasaurusrex:
modifying the tests may not be needed, after all: https://github.com/getsentry/sentry-python/pull/3696

szokeasaurusrex:
This has been superseded by some other PRs incrementally removing the pin

PR #3699, opened 2024-10-28
ci: Run license compliance action on all PRs

szokeasaurusrex:
This action only is triggered on PRs to master, but the action is required. This becomes a problem when a PR is opened against a branch other than master (e.g. as part of a PR tree). When the parent branch is merged to master, the PR's base automatically changes to master, but this action does not get triggered. Instead, it blocks on "Expected" and can only be run by adding commits to the branch.

Running the action on PRs against any branch should fix this.

Also, add logic to cancel in-progress workflows on pull requests (logic taken from our other actions)

---

### DIFF
--- a/.github/workflows/enforce-license-compliance.yml
+++ b/.github/workflows/enforce-license-compliance.yml
@@ -8,10 +8,11 @@ on:
       - release/*
       - sentry-sdk-2.0
   pull_request:
-    branches:
-      - master
-      - main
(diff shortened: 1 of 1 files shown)


sl0thentr0py (approved):

szokeasaurusrex:
This PR is against a branch other than `master`, so it will test #3698 and itself

PR #3702, opened 2024-10-28
Fix broken sampler after backpressure change

sl0thentr0py:


sentry_sdk/integrations/opentelemetry/sampler.py
```diff
@@ -68,10 +68,11 @@ def dropped_result(parent_span_context, attributes, sample_rate=None):
         else:
             reason = "sample_rate"
 
-        client.transport.record_lost_event(reason, data_category="transaction")
+        if client.transport and has_tracing_enabled(client.options):
```
szokeasaurusrex, line 71: Should we explicitly check against None?

szokeasaurusrex (commented):
lgtm, but see nit

sentry_sdk/integrations/opentelemetry/sampler.py
sl0thentr0py: we do this everywhere it's fine

PR #3704, opened 2024-10-28
Make profiling work with potel

sl0thentr0py:
All profiling should work now except for the `profiles_sampler` config which will be handled separately.

image


szokeasaurusrex (approved):
Looks good! Do we need to add tests, or should existing tests already cover these changes?

sl0thentr0py:
these PRs are about fixing existing CI, not adding new tests

Issue #3707, opened 2024-10-28
RQ 2.0 breaks a test


szokeasaurusrex:
[This test](https://github.com/getsentry/sentry-python/blob/b6482f0a474847b1e65b5ec1a9575b929b7207c6/tests/integrations/rq/test_rq.py#L254-L267) has broken with the recent release of RQ 2.0. 

I spent some time trying to figure out how to fix the test, but did not make much progress, so I will disable the test in RQ 2.0 for now.

Investigate why the test is broken, and then re-enable.

PR #3708, opened 2024-10-28
test: Disable broken RQ test in newly-released RQ 2.0

szokeasaurusrex:
See #3707


tests/integrations/rq/test_rq.py
```diff
@@ -35,6 +35,7 @@ def _patch_rq_get_server_version(monkeypatch):
 
 
 def crashing_job(foo):
+    print("RUNNING CRASHING JOB")
```
sl0thentr0py, line 38: remove print

sl0thentr0py (approved):

PR #3710, opened 2024-10-28
Prefer python_multipart import over multipart

musicinmybrain:
See: https://github.com/Kludex/python-multipart/issues/16

See also releases 0.0.13 through 0.0.16 at
https://github.com/Kludex/python-multipart/releases.

This prefers the new `python_multipart` import name for https://pypi.org/project/python-multipart/ introduced in release 0.0.13 to resolve conflicts with https://pypi.org/project/multipart/. A fallback to the original `multipart` import name remains for compatibility with older versions.

Starlette itself now does something similar:

https://github.com/encode/starlette/compare/0.41.0...0.41.2 

---

I tried `tox -e linters`, `tox -e py3.13-starlette-latest`, `tox -e py3.13-starlette-0.36`, and `tox -e py3.13-starlette-0.32` locally. I don’t expect any trouble with the rest of the test matrix.

### DIFF
--- a/tests/integrations/starlette/test_starlette.py
+++ b/tests/integrations/starlette/test_starlette.py
@@ -13,16 +13,10 @@ import pytest
 
 from sentry_sdk import capture_message, get_baggage, get_traceparent
 from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
-with warnings.catch_warnings():
-    # Error on deprecation-type warnings related to python_multipart, which are
-    # probably about migrating from "import multipart" to "import
-    # python_multipart".
-    for w in (PendingDeprecationWarning, DeprecationWarning):
-        warnings.filterwarnings("error", ".*python_multipart.*", category=w)
-    from sentry_sdk.integrations.starlette import (
-        StarletteIntegration,
-        StarletteRequestExtractor,
-    )
+from sentry_sdk.integrations.starlette import (
+    StarletteIntegration,
+    StarletteRequestExtractor,
+)
 from sentry_sdk.utils import parse_version
 
 import starlette


musicinmybrain:
This avoids a `PendingDeprecationError` in the tests when using the `multipart` import name with `python-multipart` 0.0.13 or later.

sentry_sdk/integrations/starlette.py
```diff
@@ -65,9 +65,14 @@
 
 try:
     # Optional dependency of Starlette to parse form data.
-    import multipart  # type: ignore
+    try:
+        # python-multipart 0.0.13 and later
+        import python_multipart  # type: ignore
+    except ImportError:
+        # python-multipart 0.0.12 and later
```
szokeasaurusrex, line 72: Did you mean "earlier" here?

```suggestion
        # python-multipart 0.0.12 and earlier
```

szokeasaurusrex (commented):
Thanks for the contribution! This mostly looks good.

Two things though:
  - Would you be able to add a test to ensure we don't get the `PendingDeprecationWarning`?
  - Also small nit, I think it might be better to `import python_multipart as multipart`. This would minimize the diff, since we would not need to update existing usages of `multipart`. 

Let me know what you think; I can also take over the PR from here if you prefer.

sentry_sdk/integrations/starlette.py
musicinmybrain: Good catch. Thanks. I will fix this.

musicinmybrain:
>     * Would you be able to add a test to ensure we don't get the `PendingDeprecationWarning`?

I am not sure how to add a proper test since the `PendingDeprecationWarning` arises when `multipart` is imported, which happens when `sentry_sdk/integrations/starlette.py` is imported:

https://github.com/getsentry/sentry-python/blob/d48dc46823d3602ca899ecb2178cfe4b8267f89c/tests/integrations/starlette/test_starlette.py#L16-L19

And of course, a second import would use caches, and the warning would not be triggered again.

Maybe I can wrap the import with `catch_warnings()` and at least get `PendingDeprecationWarning` to cause an error in test collection, even if I can’t add a named test.

>     * Also small nit, I think it might be better to `import python_multipart as multipart`. This would minimize the diff, since we would not need to update existing usages of `multipart`.

Sure, that’s fine. I defaulted to using the new import name throughout because I thought it would be least confusing in the future when everyone has forgotten the old name, but there is no harm in doing it as you suggested.

I’m going to go ahead and force-push with the other changes while I think about testing.

musicinmybrain:
Added a commit that causes an error to occur in the tests when there is a `PendingDeprecationWarning` or `DeprecationWarning` from the Starlette integrations.

With just the first commit, `tox -e py3.13-starlette-latest` gives:

```
======================================================== ERRORS =========================================================
____________________________ ERROR collecting tests/integrations/starlette/test_starlette.py ____________________________
tests/integrations/starlette/test_starlette.py:22: in <module>
    from sentry_sdk.integrations.starlette import (
sentry_sdk/integrations/starlette.py:68: in <module>
    import multipart  # type: ignore
.tox/py3.13-starlette-latest/lib/python3.13/site-packages/multipart/__init__.py:19: in <module>
    warnings.warn("Please use `import python_multipart` instead.", PendingDeprecationWarning, stacklevel=2)
E   PendingDeprecationWarning: Please use `import python_multipart` instead.
---------------------------- generated xml file: /home/ben/src/forks/sentry-python/.junitxml ----------------------------
================================================ short test summary info ================================================
ERROR tests/integrations/starlette/test_starlette.py - PendingDeprecationWarning: Please use `import python_multipart` instead.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=================================================== 1 error in 0.32s ====================================================
```

With both commits:

```
=================================================== warnings summary ====================================================
tests/integrations/starlette/test_starlette.py::test_starletterequestextractor_json
tests/integrations/starlette/test_starlette.py::test_starletterequestextractor_form
tests/integrations/starlette/test_starlette.py::test_starletterequestextractor_body_consumed_twice
tests/integrations/starlette/test_starlette.py::test_starletterequestextractor_extract_request_info_too_big
tests/integrations/starlette/test_starlette.py::test_starletterequestextractor_extract_request_info
tests/integrations/starlette/test_starlette.py::test_starletterequestextractor_extract_request_info_no_pii
  /usr/lib64/python3.13/asyncio/events.py:89: RuntimeWarning: coroutine '_mock_receive' was never awaited
    self._context.run(self._callback, *self._args)
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.

tests/integrations/starlette/test_starlette.py::test_template_tracing_meta
  /home/ben/src/forks/sentry-python/.tox/py3.13-starlette-latest/lib/python3.13/site-packages/starlette/templating.py:161: DeprecationWarning: The `name` is not the first parameter anymore. The first parameter should be the `Request` instance.
  Replace `TemplateResponse(name, {"request": request})` by `TemplateResponse(request, name)`.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
---------------------------- generated xml file: /home/ben/src/forks/sentry-python/.junitxml ----------------------------
================================================== slowest 5 durations ==================================================
0.19s call     tests/integrations/starlette/test_starlette.py::test_transaction_style[/message-url-/message-route]
0.13s call     tests/integrations/starlette/test_starlette.py::test_transaction_http_method_custom
0.11s call     tests/integrations/starlette/test_starlette.py::test_catch_exceptions[/some_url-ZeroDivisionError-division by zero]
0.11s call     tests/integrations/starlette/test_starlette.py::test_configurable_status_codes_deprecated[failed_request_status_codes7-403-True]
0.11s call     tests/integrations/starlette/test_starlette.py::test_user_information_transaction
============================================ 67 passed, 7 warnings in 5.12s =============================================
<sys>:0: ResourceWarning: unclosed file <_io.BufferedReader name='/home/ben/src/forks/sentry-python/tests/integrations/starlette/photo.jpg'>
```

szokeasaurusrex:
@musicinmybrain I don't think we should test in this way, with the no warnings assertion outside of a test function. If it is non-trivial to add a test, let's just skip adding a test here.

szokeasaurusrex (approved):
Removed the test code, looks good now! Thanks again for the contribution.

PR #3717, opened 2024-10-30
feat: introduce rust_tracing integration

matt-codecov:
this PR introduces a new integration that allows traces to descend into code in Rust native extensions by hooking into Rust's popular `tracing` framework. it relies on the Rust native extension using [`pyo3-python-tracing-subscriber`](https://crates.io/crates/pyo3-python-tracing-subscriber), a crate i recently published under Sentry, to expose a way for the Python SDK to hook into `tracing`.

in this screenshot, the transaction was started in Python but the rest of the span tree reflects the structure and performance of a naive fibonacci generator in Rust.
Screenshot 2024-10-29 at 7 11 12 PM

i have [a `sentry-docs` branch with new docs for the integration](https://github.com/matt-codecov/sentry-docs/tree/matt/python-rust-tracing-integration) pretty much ready to go. just want to get this merged and then add a working example to `pyo3-python-tracing-subscriber` to link to first

### DIFF
--- a/sentry_sdk/integrations/rust_tracing.py
+++ b/sentry_sdk/integrations/rust_tracing.py
@@ -37,6 +37,7 @@ from typing import Any, Callable, Dict, Tuple, Optional
 import sentry_sdk
 from sentry_sdk.integrations import Integration
 from sentry_sdk.tracing import Span as SentrySpan
+from sentry_sdk.utils import SENSITIVE_DATA_SUBSTITUTE
 
 TraceState = Optional[Tuple[Optional[SentrySpan], SentrySpan]]
 
@@ -149,10 +150,12 @@ class RustTracingLayer:
             [Dict[str, Any]], EventTypeMapping
         ] = default_event_type_mapping,
         span_filter: Callable[[Dict[str, Any]], bool] = default_span_filter,
+        send_sensitive_data: bool = None,
     ):
         self.origin = origin
         self.event_type_mapping = event_type_mapping
         self.span_filter = span_filter
+        self.send_sensitive_data = send_sensitive_data
 
     def on_event(self, event: str, _span_state: TraceState) -> None:
         deserialized_event = json.loads(event)
@@ -221,9 +224,19 @@ class RustTracingLayer:
             return
         _parent_sentry_span, sentry_span = span_state
 
+        client_options = sentry_sdk.get_client().options
+        send_sensitive_data = (
+            client_options["send_default_pii"]
+            if self.send_sensitive_data is None
+            else self.send_sensitive_data
+        )
+
         deserialized_values = json.loads(values)
         for key, value in deserialized_values.items():
-            sentry_span.set_data(key, value)
+            if send_sensitive_data:
+                sentry_span.set_data(key, value)
+            else:
+                sentry_span.set_data(key, SENSITIVE_DATA_SUBSTITUTE)
 
 
 class RustTracingIntegration(Integration):
@@ -246,11 +259,13 @@ class RustTracingIntegration(Integration):
             [Dict[str, Any]], EventTypeMapping
         ] = default_event_type_mapping,
         span_filter: Callable[[Dict[str, Any]], bool] = default_span_filter,
+        send_sensitive_data: Optional[bool] = None,
     ):
         self.identifier = identifier
-
         origin = f"auto.function.rust_tracing.{identifier}"
-        self.tracing_layer = RustTracingLayer(origin, event_type_mapping, span_filter)
+        self.tracing_layer = RustTracingLayer(
+            origin, event_type_mapping, span_filter, send_sensitive_data
+        )
 
         initializer(self.tracing_layer)
 
--- a/sentry_sdk/integrations/rust_tracing.py
+++ b/sentry_sdk/integrations/rust_tracing.py
@@ -18,7 +18,7 @@ Usage in Python would then look like:
 sentry_sdk.init(
     dsn=sentry_dsn,
     integrations=[
-        SentryIntegrationFactory.create(
+        RustTracingIntegration(
             "demo_rust_extension",
             demo_rust_extension.initialize_tracing,
             event_type_mapping=event_type_mapping,
@@ -81,13 +81,13 @@ def extract_contexts(event: Dict[str, Any]) -> Dict[str, Any]:
         if field in metadata:
             location[field] = metadata[field]
     if len(location) > 0:
-        contexts["Rust Tracing Location"] = location
+        contexts["rust_tracing_location"] = location
 
     fields = {}
     for field in metadata.get("fields", []):
         fields[field] = event.get(field)
     if len(fields) > 0:
-        contexts["Rust Tracing Fields"] = fields
+        contexts["rust_tracing_fields"] = fields
 
     return contexts
 
@@ -189,7 +189,7 @@ class RustTracingLayer:
             sentry_span_name = "<unknown>"
 
         kwargs = {
-            "op": "native_extension",
+            "op": "function",
             "name": sentry_span_name,
             "origin": self.origin,
         }
@@ -249,7 +249,7 @@ class RustTracingIntegration(Integration):
     ):
         self.identifier = identifier
 
-        origin = f"auto.native_extension.{identifier}"
+        origin = f"auto.function.rust_tracing.{identifier}"
         self.tracing_layer = RustTracingLayer(origin, event_type_mapping, span_filter)
 
         initializer(self.tracing_layer)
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/rust_tracing.py
```diff
@@ -0,0 +1,259 @@
...
+    contexts = extract_contexts(event)
+
+    sentry_event = {
+        "logger": logger,
+        "level": level,
+        "message": message,
+        "contexts": contexts,
+    }  # type: sentry_sdk._types.Event
+
+    sentry_sdk.capture_event(sentry_event)
```
matt-codecov, line 107: this is a draft PR because this line is giving me some grief

in my example project, each span emits a debug-level event or two. if `event_type_mapping` is used to ignore them, the "naive fibonacci" transaction and its whole span tree show up like i'd expect in the "performance" and "traces" tabs. however, if the events are allowed to be logged, the transaction + span tree are lost and all i get are issues
Screenshot 2024-10-29 at 7 52 17 PM

if i click through one of the issues and look at the trace it came from, i see nonsense:
Screenshot 2024-10-29 at 7 53 14 PM

notice that the events know they come from the "naive fibonacci" transaction, but the transaction otherwise does not show up in the UI and is not useful

sl0thentr0py: * is the transaction sent at all from the SDK in the latter case? check with `sentry_sdk.init(debug=True)`
* what is the lifecycle like?  
* what is the actual behaviour you want for those debug-level events? do you want them to create issues or not?

when you say 
> each span emits a debug-level event 

is the span always going through `on_new_span` and those debug-level events through `on_event`?

sl0thentr0py: I think it might be wiser to map the debug events to breadcrumbs rather than making issues out of all of them.

matt-codecov: > is the transaction sent at all from the SDK in the latter case? check with sentry_sdk.init(debug=True)

i poked around with this and discovered i am simply logging too many events. thanks!

showing my work: "naive fibonacci" is a 2**n algorithm that should generate 2**n debug logs. i called it with n=10, so there should have been 2**10 events. but the transport queue only ever had suspiciously close to 100 events in it. that pointed me to
https://github.com/getsentry/sentry-python/blob/02d09346e6d070e03b828807d72485b6f23b2c11/sentry_sdk/consts.py#L80

so my transaction envelope was getting pushed off the back of the truck. if i call with n=6 instead, everything fits in the transport queue and arrives to sentry

> what is the lifecycle like? is the span always going through on_new_span and those debug-level events through on_event?

yes

> what is the actual behaviour you want for those debug-level events? do you want them to create issues or not? I think it might be wiser to map the debug events to breadcrumbs rather than making issues out of all of them.

the purpose of the project i am asking about was to demo all of the behavior of the integration:
- transaction/span tree, extra fields set on spans
- breadcrumbs
- issues/events, contexts set on them

so yes, a real-world project would want breadcrumbs instead (or nothing) for these particular logs. by default, the integration maps warn/info to breadcrumbs and completely ignores debug/trace, i just overrode that for sake of testing/demoing for this project

sentry_sdk/integrations/rust_tracing.py
```diff
@@ -0,0 +1,259 @@
...
+        identifier: str,
+        initializer: Callable[[RustTracingLayer], None],
+        event_type_mapping: Callable[
+            [Dict[str, Any]], EventTypeMapping
+        ] = default_event_type_mapping,
+        span_filter: Callable[[Dict[str, Any]], bool] = default_span_filter,
+    ):
+        self.identifier = identifier
+
+        origin = f"auto.native_extension.{identifier}"
```
antonpirker, line 252: To comply with the [trace origin docs](https://develop.sentry.dev/sdk/telemetry/traces/trace-origin/) the origin should be something like `f"auto.function.rust_tracing.{identifier}"`.

sentry_sdk/integrations/rust_tracing.py
```diff
@@ -0,0 +1,259 @@
...
+            sentry_span_name = message
+        elif module_path is not None and name is not None:
+            sentry_span_name = f"{module_path}::{name}"  # noqa: E231
+        elif name is not None:
+            sentry_span_name = name
+        else:
+            sentry_span_name = "<unknown>"
+
+        kwargs = {
+            "op": "native_extension",
```
antonpirker, line 192: Please set `op` to`"function"`.

sentry_sdk/integrations/rust_tracing.py
```diff
@@ -0,0 +1,259 @@
...
+        sentry_sdk.get_current_scope().span = parent_sentry_span
+
+    def on_record(self, span_id: str, values: str, span_state: TraceState) -> None:
+        if span_state is None:
+            return
+        _parent_sentry_span, sentry_span = span_state
+
+        deserialized_values = json.loads(values)
+        for key, value in deserialized_values.items():
+            sentry_span.set_data(key, value)
```
antonpirker, line 226: Could there be PII (Personally Identifiable Information) in that data? (Stuff like names, addresses, email addresses, phone numbers etc) If yes you should use `sentry_sdk.scope.should_send_default_pii()` to guard it.

sentry_sdk/integrations/rust_tracing.py
```diff
@@ -0,0 +1,259 @@
...
+def extract_contexts(event: Dict[str, Any]) -> Dict[str, Any]:
+    metadata = event.get("metadata", {})
+    contexts = {}
+
+    location = {}
+    for field in ["module_path", "file", "line"]:
+        if field in metadata:
+            location[field] = metadata[field]
+    if len(location) > 0:
+        contexts["Rust Tracing Location"] = location
```
antonpirker, line 84: Please use `"rust_tracing_location"` as key. (All other contexts have this format. The Sentry UI will then make them nice)

sentry_sdk/integrations/rust_tracing.py
```diff
@@ -0,0 +1,259 @@
...
+
+    fields = {}
+    for field in metadata.get("fields", []):
+        fields[field] = event.get(field)
+    if len(fields) > 0:
+        contexts["Rust Tracing Fields"] = fields
```
antonpirker, line 90: Please use `"rust_tracing_fields"` as key. (All other contexts have this format. The Sentry UI will then make them nice)

antonpirker (commented):
Looks very cool. I added some minor requests.

sentry_sdk/integrations/rust_tracing.py
```diff
@@ -0,0 +1,259 @@
...
+        .init();
+}
+```
+
+Usage in Python would then look like:
+```
+sentry_sdk.init(
+    dsn=sentry_dsn,
+    integrations=[
+        SentryIntegrationFactory.create(
```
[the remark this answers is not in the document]
matt-codecov: good catch

sentry_sdk/integrations/rust_tracing.py
matt-codecov: sure, can change. i chose this to follow suit with the Rust SDK https://github.com/getsentry/sentry-rust/blob/1b65b5c99af975496880e7325218479e0037d097/sentry-tracing/src/converters.rs#L199-L202

sentry_sdk/integrations/rust_tracing.py
matt-codecov: oh hmm, great callout

it looks like `should_send_default_pii()` (and [the Rust SDK's `send_default_pii`](https://github.com/getsentry/sentry-rust/blob/1b65b5c99af975496880e7325218479e0037d097/sentry-core/src/clientoptions.rs#L88)) are for information the integration knows is sensitive. this integration, by contrast, doesn't understand any of this data and would have to introspect or guess whether it's sensitive. so i don't think `should_send_default_pii()` is the mechanism to use here. also, this module is basically a Python port of [the Rust SDK's `tracing` integration](https://github.com/getsentry/sentry-rust/blob/master/sentry-tracing/src) and the Rust version has no particular logic around PII

as for what to do instead: is it enough to make the docs point users at the `before_send` and `before_send_transaction` hooks to provide their own PII cleanup behavior? or should this integration attempt to find obvious PII and filter it?

antonpirker: I have now added some sensitive data control. So you extension uses the send_default_pii from the sentry_sdk.init() call but also can set a new paramater `send_sensitive_data` parameter to override the inits settings. 

This way the user has good control of whether they want to send sensitive data.
In case they do not want to sent sensitive data, only the values are redacted. 

See my commit: https://github.com/getsentry/sentry-python/pull/3717/commits/c66c1a04a2422c2d33895d926dbfc8c3d14b46a8

antonpirker (approved):
After some rework, this integration is now ready to be merged!

PR #3721, opened 2024-10-31
test(gcp): Only run GCP tests when they should

szokeasaurusrex:
GCP tests have been running in our common test suite, including on Python versions other than 3.7 (the only version which supports the GCP integration), even though we have a separate `py3.7-gcp` tox environment for these tests. The tests take a long time (approx 1 minute+), so only executing in the appropriate `tox` environment should speed up CI time.


antonpirker (approved):
lgtm

PR #3722, opened 2024-10-31
test: Fix UTC assuming test

### LINKED ISSUE #3720: Test assumes system timezone is UTC
[This test](https://github.com/getsentry/sentry-python/blob/02d09346e6d070e03b828807d72485b6f23b2c11/tests/test_utils.py#L70-L73), added in https://github.com/getsentry/sentry-python/issues/3546, assumes that the system time is UTC. The test fails otherwise because Python interprets the timestamp as if it is using the system's local timezone, but we assert that it is interpreted in UTC.

BYK:
Fixes #3720.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -69,8 +69,8 @@ def _normalize_distribution_name(name):
         ),  # UTC time
         (
             "2021-01-01T00:00:00.000000",
-            datetime(2021, 1, 1, tzinfo=timezone.utc),
-        ),  # No TZ -- assume UTC
+            datetime(2021, 1, 1).astimezone(timezone.utc),
+        ),  # No TZ -- assume local but convert to UTC
         (
             "2021-01-01T00:00:00Z",
             datetime(2021, 1, 1, tzinfo=timezone.utc),


tests/test_utils.py
```diff
@@ -69,8 +69,8 @@ def _normalize_distribution_name(name):
         ),  # UTC time
         (
             "2021-01-01T00:00:00.000000",
-            datetime(2021, 1, 1, tzinfo=timezone.utc),
-        ),  # No TZ -- assume UTC
+            datetime(2021, 1, 1).astimezone(timezone.utc),
+        ),  # No TZ -- assume local but convert to UTC
```
asottile-sentry, line 73: given the comment I actually think the implementation is incorrect and should be something like:

```python
try:
   dt = datetime.fromisoformat(s)
except ...:
    ...
else:
   if dt.tzinfo is None:
       return dt.replace(tzinfo=UTC)
   else:
       return dt.astimezone(UTC)
```

BYK: I wrote that comment 😅 And after reading the ISO 8601 spec, I think I was mistaken: 

https://en.wikipedia.org/wiki/ISO_8601#Local_time_(unqualified)
>If no UTC relation information is given with a time representation, the time is assumed to be in local time.

Now there's the obvious warning this being unsafe in the next sentence but I'm split between being spec-compliant or "doing the right thing" which may confuse people if they rely on the spec.

What's your take @asottile-sentry?

asottile-sentry: I suspect it won't matter as most should run their servers in UTC anyway -- so I guess we can go with the current implementation

it seems to only be used to sort breadcrumbs in the event payload (is that even necessary? surely relay or the frontend should sort those right? do we even need this function and the caller?)

asottile-sentry (approved):
maybe we can just delete this code?

tests/test_utils.py
BYK: >surely relay or the frontend should sort those right?

Agree with this but maybe we're trying to save resources? It also seems like the side effect of converting a string to a python date might be something we are after here. That said that python date will then be converted back into an ISO 8601 timestamp so 🤷🏻 

Will ask ingest folks.

PR #3728, opened 2024-11-04
fix(httpx): Prevent Sentry baggage duplication

### LINKED ISSUE #3709: Duplicated values in Sentry Baggage header cause eventual 431 "Request header fields too large" error on HTTP fetch
**Issue:**
While polling an external API endpoint, the `baggage` header grows each iteration of the poll with `sentry-trace_id`, `sentry-environment`, and `sentry-release` repeating in each loop. After a while (if the poll does not resolve) the header eventually grows so big it errors with 431 "Request header fields too large" error on HTTP fetch.

**Similar Issues**
- https://github.com/getsentry/sentry-python/issues/2916
    - Similar issue, but our case is not using celery
- https://github.com/getsentry/sentry-python/issues/3526
    - Similar issue, but our case is not using Azure API Management
    - Failed to recognize that the reason they were too large was due to the headers being added multiple times.
- https://github.com/getsentry/sentry-javascript/issues/12350
    - Same issue, but this example is from javascript and our case is in python (using sentry-python)

**SDK Version:**
sentry-sdk==2.13.0

**Example:**
```
from httpx import AsyncClient


while is_still_processing:
  async with AsyncClient() as http_client:
      http_request = http_client.build_request(
          method="GET",
          url=MY_URL,
          content=MY_BODY,
          headers={"content-type":  "application/json"}
      )
  
    http_response = await AsyncClient.send(request=http_request)
    http_response.raise_for_status()

    is_still_processing = is_successful_response(http_response)
    if not is_still_processing():
        return http_response.content
    await asyncio.sleep(0.5)
```

In the above example, the `baggage` header in the request grows with each iteration. 

**Note**
The `trace_propagation_targets` is available to bypass the issue, but it doesn't fix the root cause.

**The javascript lib issue was fixed by using .set() instead of .append(), so the Python lib might be resolved in a similar manner?**

szokeasaurusrex:
Sentry baggage will get added to an HTTPX request multiple times if the same request is repeated. To prevent this from occurring, we can strip any existing Sentry baggage before adding Sentry baggage to the request.

Fixes #3709

### DIFF
--- a/sentry_sdk/integrations/httpx.py
+++ b/sentry_sdk/integrations/httpx.py
@@ -2,7 +2,7 @@ import sentry_sdk
 from sentry_sdk.consts import OP, SPANDATA
 from sentry_sdk.integrations import Integration, DidNotEnable
 from sentry_sdk.tracing import BAGGAGE_HEADER_NAME
-from sentry_sdk.tracing_utils import should_propagate_trace
+from sentry_sdk.tracing_utils import Baggage, should_propagate_trace
 from sentry_sdk.utils import (
     SENSITIVE_DATA_SUBSTITUTE,
     capture_internal_exceptions,
@@ -14,6 +14,7 @@ from sentry_sdk.utils import (
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
+    from collections.abc import MutableMapping
     from typing import Any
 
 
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/httpx.py
```diff
@@ -148,3 +147,21 @@ async def send(self, request, **kwargs):
...
+    they are stripped out first.
+    """
+    existing_baggage = headers.get(BAGGAGE_HEADER_NAME, "")
+    stripped_existing_baggage = Baggage.strip_sentry_baggage(existing_baggage)
+
+    seperator = "," if len(stripped_existing_baggage) > 0 else ""
+
+    headers[BAGGAGE_HEADER_NAME] = (
+        stripped_existing_baggage + seperator + sentry_baggage
+    )
```
sentrivana, line 167:
```suggestion
    separator = "," if len(stripped_existing_baggage) > 0 else ""

    headers[BAGGAGE_HEADER_NAME] = (
        stripped_existing_baggage + separator + sentry_baggage
    )
```

sentrivana (commented):
Found a typo. Overall looks good to me. Can we also add an httpx test specifically for the scenario where we were accumulating baggage?

sentrivana (approved):

PR #3729, opened 2024-11-05
ref(init): Deprecate `sentry_sdk.init` context manager

### LINKED ISSUE #3282: Deprecate `init` context manager
See below:

> `init` is generally one per process anyway. In case people want DSN isolation, multiple clients is the way to go now. So I really don't see the use case for this. 
> Let's deprecate and remove.

_Originally posted by @sl0thentr0py in https://github.com/getsentry/sentry-python/pull/3276#discussion_r1675776684_

szokeasaurusrex:
It is possible to use the return value of `sentry_sdk.init` as a context manager; however, this functionality has not been maintained for a long time, and it does not seem to be documented anywhere.

So, we are deprecating this functionality, and we will remove it in the next major release.

Closes #3282

### DIFF
--- a/sentry_sdk/_init_implementation.py
+++ b/sentry_sdk/_init_implementation.py
@@ -1,3 +1,5 @@
+import warnings
+
 from typing import TYPE_CHECKING
 
 import sentry_sdk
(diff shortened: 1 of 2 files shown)


antonpirker (approved):
Cool, thanks!

PR #3731, opened 2024-11-05
Fix Cohere tests

antonpirker:
Fix tests of Cohere integration to work with POtel.


sentrivana (approved):
We can remove the `_serialize_span_attribute` now, else looks good

PR #3732, opened 2024-11-05
Moved `_serialize_span_attribute` into `set_attribute`

antonpirker:
So we never forget to serialize the span attribute.


sentrivana (approved):
🙏🏻

PR #3738, opened 2024-11-05
Fix OpenAI tests in POTel

antonpirker:


tests/integrations/openai/test_openai.py
```diff
@@ -330,6 +330,7 @@ async def test_streaming_chat_completion_async(
         pass  # if tiktoken is not installed, we can't guarantee token usage will be calculated properly
 
 
+@pytest.mark.forked
```
antonpirker, line 333: We need to fork tests that capture exceptions. (Somehow capturing exceptions messes up the next sentry_sdk.init() and capturing events/exeptions in following tests)

sentrivana (approved):

Issue #3741, opened 2024-11-05
Arq integration fails when worker setting is dict


saber-solooki:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.17.0

### Steps to Reproduce

According to the arq [implementation](https://github.com/python-arq/arq/blob/main/arq/typing.py#L47) it is a common case to define Wroker Settings either as class (the protocol for that is [WorkerSettingsBase](https://github.com/python-arq/arq/blob/main/arq/typing.py#L39-L43)) or dict. We have defined many worker settings in our services which all of them are based on dict and I noticed that I can't catch the exception in the sentry console although I could see "POST /api/1811479/envelope/ HTTP/11" 200" message in debug mode.

So to make it more clear you can test it like this:
Script to enqueue a jon:
```
import asyncio

import sentry_sdk
from arq import create_pool
from arq.connections import RedisSettings


async def main():
    sentry_sdk.init(dsn='DSN_SAMPLE, traces_sample_rate=1.0, profiles_sample_rate=1.0)
    redis = await create_pool(RedisSettings())
    await redis.enqueue_job("add_numbers", a=1, b=2)

asyncio.run(main())

```
And the worker sample:
```
import sentry_sdk

sentry_sdk.init(dsn='DSN_SAMPLE', traces_sample_rate=1.0, profiles_sample_rate=1.0)

async def add_numbers(ctx, a, b):
    1/0 # raises an error!
    return a + b

# This one works
class WorkerSettings:
    functions = [add_numbers]

# This one doesn't work
WorkerSettings = {
    "functions": [add_numbers]
}
```

PS:
I've already prepared [a fix PR](https://github.com/getsentry/sentry-python/pull/3742). So please review 🙏 

### Expected Result

To see exceptions in the sentry console when an exception happens in arq job

### Actual Result

I can't catch any exceptions in the sentry console

PR #3742, opened 2024-11-05
Fix(Arq): fix integration with Worker settings as a dict

saber-solooki:
Fix https://github.com/getsentry/sentry-python/issues/3741

### DIFF
--- a/tests/integrations/arq/test_arq.py
+++ b/tests/integrations/arq/test_arq.py
@@ -1,4 +1,6 @@
 import asyncio
+from typing import Any
+
 import pytest
 
 from sentry_sdk import get_client, start_transaction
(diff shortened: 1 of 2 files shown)


tests/integrations/arq/test_arq.py
```diff
@@ -83,14 +85,65 @@ class WorkerSettings:
...
+
+    return inner
+
+
 @pytest.mark.asyncio
-async def test_job_result(init_arq):
+@pytest.mark.parametrize(
+    "init_arq_settings", ["init_arq", "init_arq_with_dict_settings"]
+)
+async def test_job_result(init_arq_settings: str, request):
```
sl0thentr0py, line 138:
```suggestion
async def test_job_result(init_arq_settings, request):
```

we don't have inline mypy types, for reasons

sl0thentr0py (changes requested):

PR #3744, opened 2024-11-06
WIP Potel base run all tests

antonpirker:
Do not stop when running one test suit fails, but continue. 

---


antonpirker:
not needed anymore.

PR #3748, opened 2024-11-07
Add `only_if_parent` option to POTelSpan and use it in integrations

sl0thentr0py:
If this option is on, we will only create a new underlying otel span if there's an active valid parent, otherwise we will just return an invalid `NonRecordingSpan` (`INVALID_SPAN`).

All internal integration child `start_span` calls have been modified so that now we will only create spans if there is an active root span (transaction) active.


antonpirker (approved):
Thanks for this! Lookin' good.

PR #3749, opened 2024-11-07
Fix starlette tests

sl0thentr0py:
mainly some middleware span reordering


tests/integrations/starlette/test_starlette.py
```diff
@@ -636,7 +636,7 @@ def test_user_information_transaction_no_pii(sentry_init, capture_events):
     assert "user" not in transaction_event
 
 
-def test_middleware_spans(sentry_init, capture_events):
+def test_middleware_spans(sentry_init, capture_events, render_span_tree):
```
antonpirker, line 639: This fixture is not needed, right? (also a second time below)

antonpirker (approved):
lgtm

PR #3762, opened 2024-11-11
Fix rq tests

antonpirker:
One of the tests involves traces_sampler, have not fixed that one.

### DIFF
--- a/tests/integrations/rq/test_rq.py
+++ b/tests/integrations/rq/test_rq.py
@@ -5,7 +5,6 @@ import rq
 from fakeredis import FakeStrictRedis
 
 import sentry_sdk
-from sentry_sdk import start_transaction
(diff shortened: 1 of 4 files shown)


tests/integrations/rq/test_rq.py
```diff
@@ -159,18 +163,17 @@ def test_tracing_enabled(
     queue = rq.Queue(connection=FakeStrictRedis())
     worker = rq.SimpleWorker([queue], connection=queue.connection)
 
-    with start_transaction(op="rq transaction") as transaction:
-        queue.enqueue(crashing_job, foo=None)
-        worker.work(burst=True)
```
antonpirker, line 164: This creates a transaction inside a transaction. This has undefined behavior so I removed it from the test.

sentrivana (approved):

PR #3772, opened 2024-11-13
Fix Celery tests in POTel

antonpirker:


tests/integrations/celery/test_celery.py
```diff
@@ -211,25 +211,21 @@ def dummy_task(x, y):
...
-            "description": "dummy_task",
-            "data": ApproxDict(),
-        }.items()
-    )
+    assert execution_event["spans"][0] == ApproxDict({
+        "trace_id": str(span.trace_id),
+        "same_process_as_parent": True,
+        "op": "queue.process",
+        "description": "dummy_task",
+    })
```
antonpirker, line 228: This just makes the diff of failing tests way easier to read

sentry_sdk/tracing.py
```diff
@@ -1582,6 +1582,9 @@ def is_success(self):
 
     def finish(self, end_timestamp=None):
         # type: (Optional[Union[float, datetime]]) -> Optional[str]
+        if hasattr(self._otel_span, "status") and self._otel_span.status.status_code == StatusCode.UNSET:
```
sl0thentr0py, line 1585: let's do this in `__exit__` above since we are already setting `INTERNAL_ERROR` there.

sl0thentr0py (changes requested):

sentry_sdk/tracing.py
antonpirker: But if people do not use the contextmanager, the status will not be set.

sl0thentr0py: already doing this in another branch so you can just remove this and pull later

antonpirker: ok. removing it.

sl0thentr0py (approved):

PR #3778, opened 2024-11-14
Bring all span data under pii control in rust_tracing integration

antonpirker:
Before only data that was set in `on_record` was protected by our PII controls. Now also the span data that is add in `on_new_span` is prtected by PII controls. 

---


matt-codecov (changes requested):
could you rename `send_sensitive_data` to `include_tracing_fields`? i worry users will not know what we could possibly mean by "sensitive data" because this is mostly data they explicitly want `tracing` to record. `include_tracing_fields` is a more accurate name

antonpirker:
Closing in favor of https://github.com/getsentry/sentry-python/pull/3780

PR #3780, opened 2024-11-14
fix: include_tracing_fields arg to control unvetted data in rust_tracing integration

matt-codecov:
implemented my suggestion from https://github.com/getsentry/sentry-python/pull/3778 so that if anton agrees we can merge this before the weekend

this PR renames the `send_sensitive_data` flag to `include_tracing_fields`. the data in question is generally data the user expects `tracing` to record or data they explicitly passed into a log statement to be recorded, so if we call it "sensitive" they may think we are referring to something else

also, this PR applies the same condition to both `on_record()` and `on_new_span()`. both callbacks set the same fields, so they should either both be redacted or both be allowed. previously only `on_record()` had the condition applied.

---


antonpirker (approved):
Yeah! This looks nice!

PR #3788, opened 2024-11-15
Fix test_breadcrumbs

sl0thentr0py:


sentry_sdk/tracing.py
```diff
@@ -1253,8 +1253,10 @@ def __init__(
 
                 # Prepopulate some attrs so that they're accessible in traces_sampler
                 attributes = attributes or {}
-                attributes[SentrySpanAttribute.OP] = op
-                attributes[SentrySpanAttribute.SOURCE] = source
+                if op is not None:
+                    attributes[SentrySpanAttribute.OP] = op
```
sl0thentr0py, line 1257: throws
```
Invalid type NoneType for attribute 'sentry.op' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```
if none

tests/test_breadcrumbs.py
```diff
@@ -37,41 +36,25 @@ def test_breadcrumbs(sentry_init, capture_events):
                     message="breadcrumb2", **add_breadcrumbs_kwargs
                 )
 
-                # Spans that create breadcrumbs automatically
-                with sentry_sdk.start_span(name="span3", op=OP.DB_REDIS) as span3:
```
sl0thentr0py, line 41: these were outdated @antonpirker since we don't auto-create breadcrumbs from spans now

PR #3795, opened 2024-11-18
fix: Trim long stacktraces

### LINKED ISSUE #2764: SDK stuck in processing RecursionError:maximum recursion depth exceeded events
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.40.5

### Steps to Reproduce

One Sentry user reported an issue where the SDK gets in am apparently endless loop while processing a RecursionError. 

I was able to reproduce it with a simple FastAPI project and setting recursion depth limit to 1M:

```
import resource, sys
sys.setrecursionlimit(10**6)

from fastapi import FastAPI
import logging
import sentry_sdk

sentry_sdk.init(
    dsn="....",
    debug=True,
)

def fibonacci_of(n):
    return fibonacci_of(n)

app = FastAPI()

@app.get("/")
        fibonacci_of(50)
        return {"OH dear"}
```

The code above creates an infinite recursion, which will eventually trigger a `RecursionError:maximum recursion depth exceeded events` error that is captured by the SDK. 

In a sync environment the SDK will hold the app while processing the event. I suspect it is during stack trace processing, as in this case the stack trace will have 1M frames. 

Original report can be found in this [internal ticket.](https://sentry.zendesk.com/agent/tickets/114546)

### Expected Result

The SDK is able to quickly process the event and attempt to send it, or stop processing in case it takes long freeing the app. 

### Actual Result

The SDK gets in an seemingly infinite loop.

szokeasaurusrex:
Only process and send the first 300 stack frames for an error event. This prevents the SDK from causing the program to hang when the SDK captures an error with a very large number of stack frames. Relay anyways [trims events to 250 stack frames](https://github.com/getsentry/relay/blob/aae36669414a1f7c6ef68d5226cb2e96a28f7667/relay-event-normalization/src/trimming.rs#L286), so it is pointless for us to process much beyond that number.

Fixes #2764

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -82,6 +82,18 @@ DEFAULT_QUEUE_SIZE = 100
 DEFAULT_MAX_BREADCRUMBS = 100
 MATCH_ALL = r".*"
 
+MAX_STACK_FRAMES = 300  # type: int
+"""The maximum number of stack frames to capture on error events.
+
+Relay enforces a limit of 250 frames on the stack trace. See
+https://github.com/getsentry/relay/blob/aae36669414a1f7c6ef68d5226cb2e96a28f7667/relay-event-normalization/src/trimming.rs#L286.
+We use a slightly higher value here; Relay will trim any frames
+beyond its limit.
+
+This constant is for internal use only, and it may be changed or removed
+at any time.
+"""
+
 FALSE_VALUES = [
     "false",
     "no",
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -1,4 +1,5 @@
 import base64
+import itertools
 import json
 import linecache
 import logging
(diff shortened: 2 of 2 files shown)


szokeasaurusrex (commented):
@jan-auer, is there any way we can or should indicate to Relay that we have trimmed the stacktrace? @dav1dde was saying that this might be possible to do via the `_meta` field.

sl0thentr0py:
please make sure that this is trimming from the correct side of the backtrace since there's a lot of weird reverses in the pipeline from SDK -> ingest -> final persisting

szokeasaurusrex:
@sl0thentr0py looks like actually this is trimming from the wrong end of the backtrace (the final calls will get trimmed, when in fact, they are probably the most interesting to the user). Thanks for the good catch!

The problem is that I don't think there is a straightforward way to iterate the stack trace in the reverse order. Iterating through the entire stack trace to get the last 300 (or other limit) entries, when the stack trace is very large, is not a good idea though, since the slowness causing https://github.com/getsentry/sentry-python/issues/2764 seems to stem from the iteration.

So, perhaps we can set a larger stack trace limit (e.g. 1000 frames), and if this is exceeded, we would just drop the stack trace completely from the event? What do you think?

We should also perhaps confirm which end of the stack trace Relay is trimming. Likely we want to be doing the same thing as them.

jan-auer:
> is there any way we can or should indicate to Relay that we have trimmed the stacktrace?

Yes, this can be done through the `len` field on the corresponding meta entry for the frames field in the stack trace. For example, in an exception's frames, this would be set on meta for `exception.values[0].stacktrace.frames`.

szokeasaurusrex:
Closed, as discussed in the above comments, this solution is not ideal.

PR #3802, opened 2024-11-18
Fix deferred tracing decisions for twp

sl0thentr0py:
* fixes the case when we get a header like `trace_id-span_id-` without an explicit sampled flag

Also update a bunch of potel tests to current behavior


sentrivana (approved):
makes sense to me

PR #3803, opened 2024-11-19
fix(integrations): Check retries_left before capturing exception

### LINKED ISSUE #3707: RQ 2.0 breaks a test

malkovro:
Since https://github.com/rq/rq/pull/1964 the job status is set to Failed before the handler decides whether to capture or not the exception while `handle_job_failure ` has not yet been called so the job is not yet re-scheduled leading to all exceptions getting captured in RQ version >= 2.0.

Related to #1076
Fixes #3707

### DIFF
--- a/sentry_sdk/integrations/rq.py
+++ b/sentry_sdk/integrations/rq.py
@@ -90,9 +90,9 @@ class RqIntegration(Integration):
 
         def sentry_patched_handle_exception(self, job, *exc_info, **kwargs):
             # type: (Worker, Any, *Any, **Any) -> Any
-            # Note, the order of the `or` here is important,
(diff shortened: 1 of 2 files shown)


sentrivana (approved):
Thank you @malkovro!

PR #3817, opened 2024-11-22
Fix rust tracing

sl0thentr0py:


sentry_sdk/integrations/rust_tracing.py
```diff
@@ -205,48 +203,35 @@ def on_new_span(self, attrs: str, span_id: str) -> TraceState:
...
-            "origin": self.origin,
-        }
-
-        scope = sentry_sdk.get_current_scope()
-        parent_sentry_span = scope.span
-        if parent_sentry_span:
-            sentry_span = parent_sentry_span.start_child(**kwargs)
-        else:
-            sentry_span = scope.start_span(**kwargs)
+        span = sentry_sdk.start_span(
```
sl0thentr0py, line 206: @matt-codecov @antonpirker I simplified this part since we don't want users to do scope stuff in integrations, it is anti-patterny

matt-codecov: from a quick look at the code, it looks like `__enter__()` will still update the current span and keep track of the parent so that `__exit__()` can restore it, right? if so then i am fine with this. not qualified to stamp, and also CI is unhappy

sl0thentr0py: yes exactly, span management should be done in the core and will just work
ci is unhappy because this is a dev branch

sl0thentr0py:
this is a very large refactor of the SDK to use otel under the hood for the next major. And yes it's fine that the rust side is not otel.

You can ignore all the other stuff tbh, I just pinged you to take a look at the changes because I was changing your integration.

sl0thentr0py:
possibly! we can investigate that once we ship this and if it's feature compatible, we can remove it.

sentrivana (approved):
lgtm

PR #3820, opened 2024-11-22
Fix httplib

sl0thentr0py:


sentry_sdk/integrations/opentelemetry/utils.py
```diff
@@ -161,36 +161,43 @@ def span_data_for_http_method(span):
...
-        url = cast("Optional[str]", url)
-
-        if url:
-            parsed_url = urlparse(url)
-            url = "{}://{}{}".format(
-                parsed_url.scheme, parsed_url.netloc, parsed_url.path
-            )
-            description = f"{http_method} {url}"
+    # TODO-neel-potel remove description completely
+    description = span_attributes.get(
```
sl0thentr0py, line 179: this logic is a mess, I will remove description completely later. 
Also this function messes up order of precedence of op and name/desc if explicitly set.

tests/integrations/stdlib/test_httplib.py
```diff
@@ -131,40 +156,28 @@ def test_httplib_misuse(sentry_init, capture_events, request):
     )
 
 
-def test_outgoing_trace_headers(sentry_init, monkeypatch, capture_envelopes):
-    # HTTPSConnection.send is passed a string containing (among other things)
-    # the headers on the request. Mock it so we can check the headers, and also
-    # so it doesn't try to actually talk to the internet.
-    mock_send = mock.Mock()
```
sl0thentr0py, line 138: removed the mock in favor of the fixture above

PR #3831, opened 2024-11-29
Dont use threading.Timer when finishing long running Transactions

antonpirker:
After thinking about high volume servers, creating a thread on each request is probably a lot of overhead. 

So this is now a version, to check the duration of the transaction when a span is closed.


sl0thentr0py:
@antonpirker this needs to rebased after the revert

antonpirker:
As we are moving to Potel, we will not merge this, because the performance part of the SDK will be done by OpenTelementry instrumentations in the future. So this is not needed anymore.

PR #3849, opened 2024-12-04
Fix scope client reference on copy/fork

sl0thentr0py:


tests/test_scope.py
```diff
@@ -23,6 +23,7 @@
     use_scope,
     use_isolation_scope,
 )
+from tests.conftest import ApproxDict
 
 
 SLOTS_NOT_COPIED = {"client"}
```
antonpirker, line 29: We should also remove code related to SLOTS_NOT_COPIED

antonpirker (changes requested):
There is related code that has to be changed as well. (see other comment)

sl0thentr0py:
closing this in favor of one against master
https://github.com/getsentry/sentry-python/pull/3857

PR #3850, opened 2024-12-04
Move scope context init outside integration

sl0thentr0py:
The scope context management init cannot be inside the integration since that would not work when we use `default_integrations=False` and other such cases.


sentrivana (approved):
👍🏻

PR #3851, opened 2024-12-04
Fix threading by carrying forward span reference in `use_scope`

sl0thentr0py:
Since the otel context span reference is the source of truth for the current active span, we need to explicitly pass the span reference on the scope through when we use `use_scope` since we are changing context variables in the other thread.

Also,

* remove deprecated `propagate_hub` completely
* fixes some typing in the original scope
* adds more types to the `contextvars_context` manager

### DIFF
--- a/sentry_sdk/integrations/threading.py
+++ b/sentry_sdk/integrations/threading.py
@@ -7,7 +7,6 @@ from sentry_sdk.integrations import Integration
 from sentry_sdk.utils import (
     event_from_exception,
     capture_internal_exceptions,
-    logger,
     reraise,
 )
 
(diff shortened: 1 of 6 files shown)


sentry_sdk/integrations/threading.py
```diff
@@ -99,7 +86,8 @@ def _run_old_run_func():
                 with sentry_sdk.use_scope(current_scope_to_use):
                     return _run_old_run_func()
         else:
-            return _run_old_run_func()
+            with sentry_sdk.isolation_scope():
```
sl0thentr0py, line 89: I think the old logic here was actually wrong but kinda worked because of some fallback code in the scope implementation.

Semantically, if we don't explicitly propagate the scopes, we should ALWAYS have created a new isolation scope for the new thread.

sentrivana (approved):
makes sense to me

Issue #3300, opened 2024-07-17
Insights: enable Django cache by default


bcoe:
### Problem Statement

Cache spans are turned off by default for Django cache due to performance issues described in https://github.com/getsentry/sentry-python/issues/2122

Recent load testing showed that cache spans did not introduce significant overhead, granted in a clean room environment, https://github.com/getsentry/sentry-python/issues/3058

## Proposal

We should turn on cache spans by default, as it will allow more sentry users to get value out of [cache insights](https://docs.sentry.io/product/insights/caches/).

However, before doing so, we should find a suitably high traffic deployment to validate the findings in #3058.

### Solution Brainstorm

Are there any other ways we could detect that emitting cache spans is creating too much load on a system and disable cache spans, rather than defaulting `cache_spans=False`, i.e., disable cache spans if a certain volume of traffic is hit, rather than by default:

```
high_volume_cache_spans=False
```

antonpirker:
We can dogfood cache spans on our `sentry` installation and see how much overview this creates. If we are comfortable turning cache spans on, than I guess setting it to true by default is OK.

BYK:
@antonpirker you wanna enable it over at `getsentry/sentry` then? I'll approve it :)

BYK:
https://github.com/getsentry/sentry/pull/82645 is up. If all goes well, we can enable this for everyone by default.

antonpirker:
@BYK I have done some more local load testing (see https://github.com/getsentry/sentry-python/issues/3907) to see how much overhead is added using the newest Python SDK.

Issue #2122, opened 2023-05-17
Django cache_spans cause excessive overhead


sl0thentr0py:
We noticed a drastic uptick in latency for a few key endpoints that utilize the cache (redis) starting on May 8th. After looking through what we deployed we noticed an update to `sentry-sdk` from `1.21.1` to `1.22.1`. After looking at the change log the "add some overhead to your server" comment stood out and so we reverted this upgrade yesterday and the endpoint normalized.  ![Here is a screenshot of the endpoint's latency](https://github.com/getsentry/sentry-python/assets/5312952/07a46f9c-a1e8-4651-bfcb-be9c5cd1f19f)

This is our sentry init if it helps:

```
sentry.init(
        before_send=event_filter,
        dsn=settings.SENTRY_DSN if get_environment() in ['production', 'staging'] else '',
        environment=get_environment(),
        integrations=[CeleryIntegration(), DjangoIntegration(), RedisIntegration(), SqlalchemyIntegration()],
        release=get_release(),
        send_default_pii=True,
    )
```

_Originally posted by @dvprz in https://github.com/getsentry/sentry-python/issues/2116#issuecomment-1550101647_

jarekwg:
This knocked out our servers, causing a severe outage across many customers.
We had to roll back our latest release as it took a while to diagnose that the CPU shredding was a result of bumping sentry.
The irony of having an error monitoring framework causing a major outage....
Would it be possible to mark releases as beta or something when they're introducing this level of risk?

HazAT:
@jarekwg you are entirely correct, and we apologize that this happened.
We already had an internal conversation about making sure we roll out changes similar to this with more care.

antonpirker:
We added a parameter `cache_spans` to `init()` that defaults to `False`. See the [documentation](https://docs.sentry.io/platforms/python/integrations/django/#options). This resolves this issue.

PR #3855, opened 2024-12-05
fix(django): Fix errors when instrumenting Django cache

BYK:
I was testing Spotlight with Sentry and realized things started to get slow and crashy. It looks like sometimes `args` is just an empty array on cache's `_instruments_call` causing lots of exceptions being thrown. This patch fixes that with explicit length checks and also adds a note for the missing instrumentation for `get_or_set` method. This might be related to #2122 and #3300.


antonpirker (approved):
Nice, thanks!

PR #3860, opened 2024-12-05
feat(flags): Add integration for custom tracking of flag evaluations

aliu39:
Follow-up to
- https://github.com/getsentry/sentry-python/pull/3648

1. Adds an integration for buffering feature flags manually with an API (add_feature_flag), and automatically capturing them on error events.

2. Adds improved test coverage to launchdarkly and openfeature by testing end to end with `capture_events` fixture, rather than asserting the flags are stored to scope.

### DIFF
--- a/sentry_sdk/_lru_cache.py
+++ b/sentry_sdk/_lru_cache.py
@@ -62,7 +62,7 @@ Agreement.
 
 """
 
-from copy import copy
+from copy import copy, deepcopy
 
 SENTINEL = object()
 
@@ -92,10 +92,13 @@ class LRUCache:
         self.hits = self.misses = 0
 
     def __copy__(self):
+        """
+        Cache keys and values are shallow copied.
+        """
         cache = LRUCache(self.max_size)
         cache.full = self.full
         cache.cache = copy(self.cache)
-        cache.root = copy(self.root)
+        cache.root = deepcopy(self.root)
         return cache
 
     def set(self, key, value):
--- a/sentry_sdk/_lru_cache.py
+++ b/sentry_sdk/_lru_cache.py
@@ -92,9 +92,6 @@ class LRUCache:
         self.hits = self.misses = 0
 
     def __copy__(self):
-        """
-        Cache keys and values are shallow copied.
-        """
         cache = LRUCache(self.max_size)
         cache.full = self.full
         cache.cache = copy(self.cache)
--- a/sentry_sdk/_lru_cache.py
+++ b/sentry_sdk/_lru_cache.py
@@ -170,7 +170,15 @@ class LRUCache:
     def get_all(self):
         nodes = []
         node = self.root[NEXT]
-        while node is not self.root:
+
+        # To ensure the loop always terminates we iterate to the maximum
+        # size of the LRU cache.
+        for _ in range(self.max_size):
+            # The cache may not be full. We exit early if we've wrapped
+            # around to the head.
+            if node is self.root:
+                break
             nodes.append((node[KEY], node[VALUE]))
             node = node[NEXT]
+
         return nodes
--- a/sentry_sdk/integrations/featureflags.py
+++ b/sentry_sdk/integrations/featureflags.py
@@ -1,0 +1,18 @@
+from sentry_sdk.flag_utils import flag_error_processor
+
+import sentry_sdk
+from sentry_sdk.integrations import Integration
+
+
+class FeatureFlagsIntegration(Integration):
+    identifier = "featureflags"
+
+    @staticmethod
+    def setup_once():
+        scope = sentry_sdk.get_current_scope()
+        scope.add_error_processor(flag_error_processor)
+
+    @staticmethod
+    def set_flag(flag: str, result: bool):
+        flags = sentry_sdk.get_current_scope().flags
+        flags.set(flag, result)
--- a/sentry_sdk/integrations/featureflags.py
+++ b/sentry_sdk/integrations/featureflags.py
@@ -29,11 +29,13 @@ class FeatureFlagsIntegration(Integration):
 
     @staticmethod
     def setup_once():
+        # type: () -> None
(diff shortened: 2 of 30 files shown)


sentry_sdk/integrations/featureflags.py
```diff
@@ -0,0 +1,18 @@
+from sentry_sdk.flag_utils import flag_error_processor
+
+import sentry_sdk
+from sentry_sdk.integrations import Integration
+
+
+class FeatureFlagsIntegration(Integration):
```
cmanallen, line 7: This should probably be called "GenericFeatureFlagIntegration".

tests/integrations/featureflags/test_featureflags.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def test_featureflags_integration(sentry_init, capture_events):
+    _processed_integrations.discard(
+        FeatureFlagsIntegration.identifier
+    )  # force reinstall
+    sentry_init(integrations=[FeatureFlagsIntegration()])
+    flags_integration = sentry_sdk.get_client().get_integration(FeatureFlagsIntegration)
+
+    flags_integration.set_flag("hello", False)
```
cmanallen, line 16: Is this the sort of API we want to design?  Our customers will either need to know a lot about the SDK or we will need to educate them.  It might be better to expose a function in the generic feature flags module which hides this behavior.

tests/test_lru_cache.py
```diff
@@ -58,3 +60,20 @@ def test_cache_get_all():
...
+    copied = copy(cache)
+    cache.set(2, 2)
+    cache.set(3, 3)
+    assert copied.get_all() == [(0, 0), (1, 1)]
+    assert cache.get_all() == [(1, 1), (2, 2), (3, 3)]
+
+    copied = copy(cache)
+    cache.get(1)
+    assert copied.get_all() == [(1, 1), (2, 2), (3, 3)]
+    assert cache.get_all() == [(2, 2), (3, 3), (1, 1)]
```
cmanallen, line 79: Is this coverage still relevant?

sentry_sdk/_lru_cache.py
```diff
@@ -92,10 +92,13 @@ def __init__(self, max_size):
...
 
     def __copy__(self):
+        """
+        Cache keys and values are shallow copied.
+        """
         cache = LRUCache(self.max_size)
         cache.full = self.full
         cache.cache = copy(self.cache)
-        cache.root = copy(self.root)
+        cache.root = deepcopy(self.root)
```
cmanallen, line 101: Great catch!

cmanallen (changes requested):

sentry_sdk/integrations/featureflags.py
aliu39: Should it be consistent with JS? There we have `featureFlagsIntegration`

tests/integrations/featureflags/test_featureflags.py
aliu39: Was trying to be consistent with JS here too. Lucas also noted on that PR this feels kinda jank as an API

tests/test_lru_cache.py
aliu39: Yes, we never tested this copy functionality

sentry_sdk/integrations/featureflags.py
cmanallen: It should be consistent between both but "FeatureFlagsIntegration" is too vague.  A user may think they need to install this to use LD or OF.

cmanallen: "Custom" or "Generic" communicates what the integration is for.

tests/integrations/featureflags/test_featureflags.py
cmanallen: Let's put a function in the integration module that abstracts this away.

aliu39: Ok yeah definitely. Just remembered the reason we do it like this in JS is bundle size concerns

sentry_sdk/integrations/featureflags.py
billyvg: I don't think it's needed - we rarely have integrations that depend on each other (replay and replay canvas being the one exception I know of). Also, users are likely to stumble upon these integrations from docs where they'll be copy and pasting anyway.  

That being said, I care less about the specific name, and more about it being consistent across SDKs

aliu39: Decided on slack to go with just FeatureFlags

cmanallen (approved):

tests/integrations/featureflags/test_featureflags.py
```diff
@@ -0,0 +1,142 @@
...
+import sentry_sdk
+from sentry_sdk.integrations import _processed_integrations, _installed_integrations
+from sentry_sdk.integrations.featureflags import (
+    FeatureFlagsIntegration,
+    add_feature_flag,
+)
+
+
+@pytest.fixture
+def uninstall_integration():
```
aliu39, line 15: I found using the `reset_integrations` fixture in `conftest.py` caused issues with the dedupe integration

antonpirker (approved):
Looks OK

antonpirker:
We also need to add docs for this, can you prepare a PR in the docs repo?

aliu39:
Sounds good! I've opened https://github.com/getsentry/sentry-docs/pull/12152, will do a last check in the morning to make sure it's up to date!

aliu39:
Do you know anything about the failing ubuntu tests @antonpirker ?

antonpirker:
> Do you know anything about the failing ubuntu tests @antonpirker ?

In Python 3.6 there is probably no `asyncio.run()`. 

Can you fix this somehow, or skip this test in Python 3.6 (then this integration will only be supported form Python 3.7 and up)

PR #3861, opened 2024-12-05
fix(LRUCache): copy using deepcopy, and ensure the get_all function always terminates

cmanallen:
@aliu39 discovered that under certain circumstances a process can get stuck in an infinite loop.  Andrew fixed this by using `deepcopy` which prevents the infinite loop and fixes a bug where the LRU returns incorrect results.  Additionally, I've added a terminating loop in case there are any future bugs we've missed.

Closes: https://github.com/getsentry/sentry-python/issues/3862

Out of precaution, we disabled flagpole evaluation tracking Sentry while we wait for this to be merged.

### DIFF
--- a/sentry_sdk/_lru_cache.py
+++ b/sentry_sdk/_lru_cache.py
@@ -62,7 +62,7 @@ Agreement.
 
 """
 
-from copy import copy
+from copy import copy, deepcopy
 
 SENTINEL = object()
 
@@ -95,7 +95,7 @@ class LRUCache:
         cache = LRUCache(self.max_size)
         cache.full = self.full
         cache.cache = copy(self.cache)
-        cache.root = copy(self.root)
+        cache.root = deepcopy(self.root)
         return cache
 
     def set(self, key, value):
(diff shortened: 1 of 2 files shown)


aliu39 (approved):

aliu39:
We found out the process will hang if you copy a LRUCache, then call get_all() on the **original**. It works as expected if you call get_all() on clones. Will add some coverage in FF tests for this case (fork a scope then read the flags on the original scope)

aliu39:
Because of this, we had to [disable](https://github.com/getsentry/sentry/pull/81765) flagpole evaluation tracking in sentry.
We want to quickly get this out in the next release so
1. we can restore flagpole tracking for sentry employees.
2. fix bugs for users of the [SDK integration](https://docs.sentry.io/platforms/python/feature-flags/), which released to beta today.

antonpirker:
This has been released with version `2.19.2`: https://github.com/getsentry/sentry-python/blob/master/CHANGELOG.md#2192

PR #3879, opened 2024-12-18
Fix AWS tests in POTel

antonpirker:
Make AWS Lambda tests in POTel.


sentrivana (approved):

tests/integrations/aws_lambda/test_aws.py
```diff
@@ -487,6 +490,15 @@ def test_handler(event, context):
         ),
         (b"[]", False, 1),
     ],
+    ids=[
```
sl0thentr0py, line 493: don't understand what this is for

sl0thentr0py (approved):

tests/integrations/aws_lambda/test_aws.py
antonpirker: those are just some identifiers for the pytest output. so it is easier to see which ones are failing.

Issue #3886, opened 2024-12-20
LRUCache.__copy__ is still unsound


cmanallen:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

latest git revision

### Steps to Reproduce

Mutable references are copied -- not the underlying value.  This is visible here: https://github.com/getsentry/sentry-python/blob/f6281f557fe62c847a0aca95eb666129e893cf32/sentry_sdk/_lru_cache.py#L26 Where we copy the _data attribute and not deepcopy.

```python
class MutableObject:
    def __init__(self):
        self.count = 0

cache1 = LRUCache(max_size=2)
cache1.set(1, MutableObject())

cache2 = cache1.__copy__()
node = cache2.get(1)
node.count += 1

assert cache1.get(1).count == 0  # fail
assert cache2.get(1).count == 1
```

We should not be manually implementing `__copy__` at all.  This is a general helper class and should not assume copy semantics.  I lay out this case in [this comment](https://github.com/getsentry/sentry-python/pull/3883#discussion_r1892132287) and [this comment](https://github.com/getsentry/sentry-python/pull/3883#discussion_r1892146702).

### Expected Result

Mutable objects do not pollute copied caches.

### Actual Result

Mutable objects do pollute copied caches.

PR #3887, opened 2024-12-20
fix: Fix cache pollution when copying mutable references

cmanallen:
- Removes manual overrides of copy behavior and leaves it up to the caller.
    - E.g. a future use case may require a non-deepcopy. If we override copy they would have to remove the dunder copy, update every implementation which relies copy, before finally creating their own copy implementation.
- Deepcopies the flag buffer.
    - Though we do not cache mutable references yet we may soon and so this foot gun should be removed from possibility.
- Removes "copy" test coverage from `test_lru_cache.py`. We're no longer assuming copy usage and leave it up to the caller.
    - The existing test in `tests/test_scope.py` covers the cache pollution case [originally mentioned here](https://github.com/getsentry/sentry-python/issues/3852).
    - The mutable cache pollution case is not covered because we do not currently cache mutable objects.

In general a generic class should assume as few implementation details as possible.  If we leave the existing copy method someone may assume copy semantics and rely on it in a way that is inappropriate.

Closes: https://github.com/getsentry/sentry-python/issues/3886

### DIFF
--- a/sentry_sdk/_lru_cache.py
+++ b/sentry_sdk/_lru_cache.py
@@ -17,15 +17,6 @@ class LRUCache:
         self.hits = self.misses = 0
         self.full = False
 
-    def __copy__(self):
-        # type: () -> LRUCache
-        new = LRUCache(max_size=self.max_size)
-        new.hits = self.hits
-        new.misses = self.misses
-        new.full = self.full
-        new._data = self._data.copy()
-        return new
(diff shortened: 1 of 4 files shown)


antonpirker (approved):
Sounds good to me. Not changing the copy behavior and leaving it for the caller to do the right thing makes sense.

PR #3888, opened 2024-12-21
feat(flags): add Unleash feature flagging integration

aliu39:
Adds an integration for tracking flag evaluations from [Unleash](https://www.getunleash.io/) customers. 

Implementation
Unleash has no native support for evaluation hooks/listeners, unless the user opts in for each flag. Therefore we decided to patch the `is_enabled` and `get_variant` methods on the `UnleashClient` class. The methods are wrapped and the only side effect is writing to Sentry scope, so users shouldn't see any change in behavior. 

We patch one `UnleashClient` instance instead of the whole class. The reasons for this are described in
- https://github.com/getsentry/sentry-python/pull/3895

It's also safer to not modify the unleash import.

References
- https://develop.sentry.dev/sdk/expected-features/#feature-flags
- https://docs.getunleash.io/reference/sdks/python for methods we're patching/wrapping

Documentation PR
- https://github.com/getsentry/sentry-docs/pull/12216


antonpirker:
Hey @aliu39 ! Thanks for the integration. This will be reviewed after the holiday season.

aliu39:
Todo:
- [x] verify w manual test
- [x] threading and asyncio tests
- [x] determine if [variant payload](https://docs.getunleash.io/reference/feature-toggle-variants#variant-payload) should factor into tracking (no, track every feature)
- [ ] docs w/ sample verification code

cmanallen (changes requested):
Blocking this.  We're going to talk about how we handle this going forward and may close this.

cmanallen (approved):
Okay we're all good. We're going to keep this setup.

antonpirker (approved):
Looks good to me. 
Nice integrations, and GREAT test suite! Thanks for that!

Issue #2696, opened 2024-02-01
Transaction Source Enum


szokeasaurusrex:
The [`TRANSACTION_SOURCE_*` constants defined in `tracing.py`](https://github.com/getsentry/sentry-python/blob/c8e917205d77e143c68c0cda4df877167d402830/sentry_sdk/tracing.py#L40) would likely make more sense to be defined as enums. We can then redefine the functions that accept as an argument a transaction source (for example, [`scope.set_transaction_name`](https://github.com/getsentry/sentry-python/blob/c8e917205d77e143c68c0cda4df877167d402830/sentry_sdk/scope.py#L499)) to accept the Transaction Source enum type.

PR #3889, opened 2024-12-22
ref(tracing): Move `TRANSACTION_SOURCE_*` constants to `Enum`

mgaligniana:
Fixes GH-2696


mgaligniana:
Linter is failing because of:

image

Should I fix them in a different commit?

antonpirker (approved):
nice! thanks for the cleanup

antonpirker:
> Linter is failing because of:
> 
> image
> 
> Should I fix them in a different commit?

yes, this is because of a new release of mypy. I will fix this.

antonpirker:
This breaks a lot of tests, because the full enum ends up in the envelope that is sent to Sentry and not the `.value` of the enum. Can you please fix this?

antonpirker:
I just made it so one does not have to always use the .value, so the enum is easier to use and error due to missing .value are not happening. otherwise it is a really cool change and I will merge it today

mgaligniana:
Thank you @antonpirker !!

sentry_sdk/tracing.py
```diff
@@ -131,30 +131,33 @@ class TransactionKwargs(SpanKwargs, total=False):
...
-    URL = "url"
     ROUTE = "route"
-    VIEW = "view"
-    COMPONENT = "component"
     TASK = "task"
+    URL = "url"
+    VIEW = "view"
+
+    def __str__(self):
+        return self.value
```
mgaligniana: I didn't see this! Thanks for the explanation!

PR #3890, opened 2024-12-22
ref(flags): register LD hook in setup instead of init, and don't check for initialization

aliu39:
Refactors setup code and removes an error case from the [LaunchDarkly integration](https://docs.sentry.io/platforms/python/integrations/launchdarkly/)

> register LD hook in setup instead of init

IMO we should only hook into LD once, when this integration is setup by `sentry_sdk.init`. It's rare, but this class could be instantiated more than once, from user error, tests, etc. It's reasonable to expect Sentry `init` is only called once in prod.

> don't check for initialization

We shouldn't fail if LDClient hasn't initialized yet. Initialization is the async process of connecting to the LD server, and may take a moment. With the setup code from our docs, we configure LD right before calling Sentry `init`, and don't wait for initialization to complete. **This could lead to a lot of errors / failure to start your app if you have a slow internet connection. It's ultimately the user's responsibility to ensure LD is initialized before evaluating flags**. Even if the client's never initialized, flags will default to False, so nothing's breaking on our end.

### DIFF
--- a/sentry_sdk/integrations/launchdarkly.py
+++ b/sentry_sdk/integrations/launchdarkly.py
@@ -35,7 +35,6 @@ class LaunchDarklyIntegration(Integration):
         # type: () -> None
         try:
             client = LaunchDarklyIntegration._ld_client or ldclient.get()
-            print("got client")
         except Exception as exc:
             raise DidNotEnable("Error getting LaunchDarkly client. " + repr(exc))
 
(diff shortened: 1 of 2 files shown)


antonpirker (approved):
looks good.

cmanallen:
@aliu39 This needs to be reverted.  This is not the correct way to manage this behavior.  Regardless, where is the documentation for this change?  A notion doc or anything to describe why this is important?

cmanallen:
Reverted here: https://github.com/getsentry/sentry-python/pull/3900.  There are too many holes in this.  How would a user track multiple clients?  Multiple integrations?  Does that work?  There are no tests for it.  Let's discuss on Monday.

PR #3895, opened 2024-12-23
ref(flags): allow OpenFeature integration to track a single client

aliu39:
All provider SDKs so far have the concept of a "client" class, which is used to connect to the server and query flags. Since clients can query multiple projects/environments by specifying a `context`, it is rare for an app to use >1 client. 

I discussed with @billyvg and we want to stay consistent for all FF integrations, and focus on tracking one client. This means users should pass a client to the Integration, and we'll register hooks on that client, rather than globally.

Update 12/26:
For backwards compatibility, I think it's best to
- keep LD as-is: optionally takes a client. If one is not given, then hook into the global singleton.
- mirror this in OF, optionally taking in a client.

Updated setup code can be seen at: https://github.com/getsentry/sentry-docs/pull/12222/files

### DIFF
--- a/sentry_sdk/integrations/launchdarkly.py
+++ b/sentry_sdk/integrations/launchdarkly.py
@@ -4,39 +4,37 @@ import sentry_sdk
 from sentry_sdk.integrations import DidNotEnable, Integration
 from sentry_sdk.flag_utils import flag_error_processor
 
+if TYPE_CHECKING:
+    from typing import Any, Optional
+
 try:
-    import ldclient
     from ldclient.hook import Hook, Metadata
 
     if TYPE_CHECKING:
         from ldclient import LDClient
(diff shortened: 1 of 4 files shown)


aliu39:
For backwards compatibility, I think it's best to
- keep LD as-is (optionally takes in a client, if not given then hooks into global singleton)
- make OF similar to LD, taking in an _optional_ client.

cmanallen:
> I discussed with @billyvg and we want to stay consistent for all FF integrations, and focus on tracking one client. This means users should pass a client to the Integration, and we'll register hooks on that client, rather than globally.

This needs to be discussed as a team and documented.  Writing code is the last step not the first.  I'm closing this for now.  Not to stop us from working on this forever but to prevent someone from merging code accidentally.  We can re-open once we understand the implications of the decisions we're making.

PR #3898, opened 2024-12-31
Allow OTEL instrumenter to work even with no DSN.

jwhitaker-gridcog:
Allow OTEL instrumenter to work even with no DSN.

Context: we are writing tests for our Sentry integration itself, setting up a client like

```py
events = []
c = sentry_sdk.Client(dsn=None, instrumenter='otel', transport=events.append)
```

, and verifying that events contains spans from OpenTelemetry such as

```py
with tracer.start_as_current_span('fn'):
   raise Exception('boo')

assert events = [...]
```

This currently doesn't work because the Sentry OTEL integration has some hacks poking into the client and hard-returning if DSN is None.

I conjecture the hard-return can be removed - the check-DSN-to-avoid-sentry-otel-loops logic still works even without the DSN check in on_start().


alexander-alderman-webb:
Going to close this because the `OpenTelemetryIntegration` is now deprecated.
I unfortunately do not have context for why your PR wasn't reviewed.

PR #3900, opened 2025-01-03
Revert "ref(flags): register LD hook in setup instead of init, and don't check for initialization"

cmanallen:
Mutating a class attribute on `__init__` violates encapsulation and will lead to strange errors.  We need to rethink how we want to implement this before we merge any code. 

Reverts getsentry/sentry-python#3890

### DIFF
--- a/sentry_sdk/integrations/launchdarkly.py
+++ b/sentry_sdk/integrations/launchdarkly.py
@@ -20,7 +20,6 @@ except ImportError:
 
 class LaunchDarklyIntegration(Integration):
     identifier = "launchdarkly"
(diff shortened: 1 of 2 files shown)


aliu39 (approved):

aliu39:
> `ERROR tests/integrations/sanic/test_sanic.py - TypeError: 'type' object is not subscriptable`

There's a similar error at https://github.com/getsentry/sentry-python/actions/runs/12600825778/job/35120696927?pr=3888

antonpirker:
Yes, Sanic is sometimes flaky

PR #3902, opened 2025-01-06
ref(flags): underscore featureflags module and identifier

aliu39:
Follow-up to https://github.com/getsentry/sentry-python/pull/3860, which hasn't been released yet

### DIFF
--- a/sentry_sdk/integrations/feature_flags.py
+++ b/sentry_sdk/integrations/feature_flags.py
@@ -16,7 +16,7 @@ class FeatureFlagsIntegration(Integration):
     @example
     ```
     import sentry_sdk
-    from sentry_sdk.integrations.featureflags import FeatureFlagsIntegration, add_feature_flag
+    from sentry_sdk.integrations.feature_flags import FeatureFlagsIntegration, add_feature_flag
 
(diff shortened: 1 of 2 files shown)


cmanallen (approved):

antonpirker (changes requested):
This is a breaking change and needs a major update for the SDK. 

Is this something we MUST have or it is just a nice too have? 
I would not want to make this breaking change only for cosmetic reasons.

antonpirker:
@aliu39 and @cmanallen the `FeatureFlagsIntegration` is not documented yet so I guess it is still in beta or some kind of stealth mode, right? 
If it is not yet used by a lot of users, we probably can rename it without a major update.

antonpirker (approved):
In our [feature flag documentation](https://docs.sentry.io/platforms/python/feature-flags/) we state that feature flag tracking is a beta feature. 

Also `FeatureFlagsIntegration` is not yet documented at all. 

So renaming this is OK in a minor release.

cmanallen:
@antonpirker The `featureflags` modules hasn't been released yet so unless someone is running from master no one has ever used the integration.  That's why we felt it was safe.  Thanks for merging this! :)

PR #3903, opened 2025-01-06
fix: preserve ARQ enqueue_job __kwdefaults__ after patching

danmr:
Problem:
- The ARQ integration loses `__kwdefaults__` when patching the `enqueue_job` method
- This causes issues when filtering kwargs in custom implementations

Solution:
- Save original `__kwdefaults__` before patching
- Restore them in the patched function

---


sentrivana (approved):
Thanks @danmr.

PR #3914, opened 2025-01-10
feat(flags): remove Unleash get_variant patching code

aliu39:
Follow-up to https://github.com/getsentry/sentry-python/pull/3888

The original PR patched 2 methods used for evaluating feature flags, `is_enabled` (simple toggle on/off) and `get_variant` (returns a dict of metadata, see https://docs.getunleash.io/reference/sdks/python#getting-a-variant).

We want to remove all `get_variant` code since we only support boolean flag evals atm. It seems like the main usecase for variants is reading payloads (non-bool) for A/B/multivariate testing. This could lead to a lot of extraneous flags, so until it is requested and/or we support non-bool values, let's not patch this method.


cmanallen:
@aliu39 How are people using `get_variant` in practice?  Does it make sense to send a `true`, `false` value to Sentry if what the user is actually working with strings `a`, `b`, or `c`?

aliu39:
> @aliu39 How are people using `get_variant` in practice? Does it make sense to send a `true`, `false` value to Sentry if what the user is actually working with strings `a`, `b`, or `c`?

Discussed on a call, also looping in the FFs slack channel

antonpirker (approved):
less code is good!

PR #3916, opened 2025-01-10
Setup scopes before client

sl0thentr0py:
The newly added feature flags add an error processor to `current_scope` in their `setup_once`. This is actually an antipattern and shouldn't be encouraged.
Either way, this PR sets up the scopes first since the integrations get setup in the `Client` and require that scope to be correct.


antonpirker (approved):
lgtm

PR #3917, opened 2025-01-10
Moved adding of `flags` context into Scope

antonpirker:
Using an error_processor to read data from the scope to add to the event is an anti-pattern.  Moving this into `Scope.apply_to_event()`. 

This PR:
- moves code that adds flags to an event from an error processor into the `Scope` class 
- moves `add_feature_flag()` function from `sentry_sdk.integrations.feature_flags` into `sentry_sdk.feature_flags`

This PR needs an update to this documentation PR: https://github.com/getsentry/sentry-docs/pull/12300


sentry_sdk/scope.py
```diff
@@ -1378,6 +1378,9 @@ def _apply_contexts_to_event(self, event, hint, options):
             else:
                 contexts["trace"] = self.get_trace_context()
 
+        # Add "flags" context
+        contexts.setdefault("flags", {}).update({"values": self.flags.get()})
```
sl0thentr0py, line 1382: this needs to be done only if the event is an error event

sl0thentr0py (changes requested):

sentry_sdk/scope.py
antonpirker: are flags not allowed on transaction or check ins? (I know before the change they where only added to errors)

sl0thentr0py: the docstring added says 
https://github.com/getsentry/sentry-python/blob/fa241c3425e446878f173407fd7358f38d8bd529/sentry_sdk/integrations/feature_flags.py#L9

antonpirker: ok, will change to error only

sentry_sdk/scope.py
```diff
@@ -1378,6 +1378,11 @@ def _apply_contexts_to_event(self, event, hint, options):
             else:
                 contexts["trace"] = self.get_trace_context()
 
+        # Add "flags" context
+        flags = self.flags.get()
+        if len(flags) > 0:
+            contexts.setdefault("flags", {}).update({"values": flags})
+
```
cmanallen, line 1385: Up to this point, we've appended feature flags to the context if the integration was enabled regardless of if there were 0 flags or not.  This meant in Sentry we could determine if the user had enabled the integration and show a different UI state.  If we remove this its a breaking change.  Its manageable but it is still a breaking change.

aliu39: > This meant in Sentry we could determine if the user had enabled the integration and show a different UI state.

I think it's alright, because after @michellewzhang merges her work on the `has_flags` project flag, we won't rely on having the context anymore.

michellewzhang: the PR to close out the `has_flags` (implements it on the FE) is here: https://github.com/getsentry/sentry/pull/83282

sentry_sdk/scope.py
antonpirker: Done. Only added to error events again.

sentry_sdk/scope.py
antonpirker: I would like to add the "flags" context only if there are flags. With the frontend change mentioned above this is a OK solution. 

Is it fine with you @cmanallen to only add the "flags" context when there are actually flags on the scope?

antonpirker:
@cmanallen with this refactoring the `FeatureFlagsIntegration` is deleted, because it is not needed anymore. And `add_feature_flag()` would be living in `sentry_sdk.feature_flags` (instead of `sentry_sdk.integrations.feature_flags`). 

I think this is a cleaner approach. 
Is this is OK with you, or would that break users code or something? (I guess this would also be fine, because it is still in beta)

sentry_sdk/scope.py
antonpirker: @cmanallen confirmed in a private message, that only adding flags context when there are flags is OK and works with the backend and the UI.

sl0thentr0py (approved):

cmanallen (approved):

PR #3934, opened 2025-01-14
Deprecate `set_measurement()` API.

### LINKED ISSUE #3074: Deprecate `set_measurement` API
We want to deprecate custom measurements. So the public API should output a deprecation warning.

antonpirker:
Deprecate `set_measurement()`. This will be replaced by `set_data()` which internally is using the Otel `set_attribute()`. 

Fixes #3074


antonpirker:
As this kind of in sleep mode right now, we keep this PR open but will not merge it right now.

See: https://github.com/getsentry/projects/issues/73

antonpirker:
Picking this up again, Daniel G said we should deprecated now and remove in 3.0.

sentrivana (approved):

PR #3943, opened 2025-01-17
Fix error logging during AWS init

itsbjoern:
Hi I noticed that the AWS integration fails to capture errors during the init phase (at least in python 3.8 and above environments).

It appears tests for this were disabled after a change in AWS' own runtime environment: https://github.com/getsentry/sentry-python/pull/3592

I was able to hunt down a change from a few months ago where it seems like string serialisation of the json payload was disabled and instead the `post_init_error` is invoked directly with the json payload

https://github.com/aws/aws-lambda-python-runtime-interface-client/commit/a37a43a48bc151c211ad72a6556044aa62b2c671#diff-4513a869520b19ae4e30058106d7c3b5ddbb79216b5e9bd922d83389fb86c603R483

This breaks and causes an error internally when trying to parse the string back into json, and the error is actually swallowed because of `with capture_internal_exceptions():`.

I did a bit of trialling in the test suite and it seems like python3.8 still is invoked with a string, whereas 3.9+ is invoked with JSON. But because I'm not 100% sure on the behaviour I just opted to check the arg type instead. I couldn't find any better documentation on this behaviour than the link above which just seems to be a continuous stream of updates to the main AWS python image.


itsbjoern:
Hey @antonpirker, really sorry to ping you like this but just wondering if there is any other action required from me here. Happy to patch this in our own codebase for now, but I'd love to help out where I can to see this fixed upstream.

antonpirker (approved):
Very cool @itsbjoern, thanks for your work!

As this is still guarded by our capture_internal_exceptions and the re-enabled test case is also happy, I think it is safe to merge this!

PR #3987, opened 2025-01-21
chore(profiling): Change continuous profile sampling frequency and bu…

Zylphrex:
…ffer size

This lowers the sampling frequency of continous profiles to 21Hz and increases the buffer size to 1 minute.


sentrivana (approved):

Zylphrex:
Holding off on merging this until I confirm a few things.

Zylphrex:
Only bumping the buffer size to 1min and keeping the sampling frequency at 101hz. Reducing the sampling frequency on the SDK will cause pains that are difficult to undo.

PR #4008, opened 2025-01-30
styles(black): Run black on _queue.py

Zylphrex:
Seems new with black==25.1.0


Zylphrex:
nvm, gotta fix everything together

PR #4034, opened 2025-02-10
fix(flags): Fix bug where concurrent accesses to the flags property could raise a RunTime error

cmanallen:
Fixes: https://sentry.sentry.io/issues/6286673308/?project=1

### DIFF
--- a/sentry_sdk/feature_flags.py
+++ b/sentry_sdk/feature_flags.py
@@ -1,5 +1,7 @@
+import copy
 import sentry_sdk
 from sentry_sdk._lru_cache import LRUCache
+from threading import Lock
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


cmanallen:
@antonpirker Strawberry integration seems broken for unrelated reasons.

aliu39 (approved):

antoniobg:
Apologies if this isn’t the right place to ask. We started [a discussion](https://github.com/getsentry/sentry-python/discussions/4101) about this fix but haven't received a response.

> We’ve been testing the new feature flag functionality (currently in beta) and ran into an issue that appears to have been addressed by https://github.com/getsentry/sentry-python/pull/4034
> The fix introduces a lock to prevent concurrency issues, but we were wondering why FlagBuffered is shared across contexts rather than being thread-local. Wouldn’t it make more sense for it to be per-context to avoid the need for locking?
> Thanks in advance for your help!

cmanallen:
@antoniobg This is a fine place.  Discussions don't send Slack notifications so apologies I missed it.  Thanks for raising.  I'll respond on the discussion topic to keep everything in one place.

PR #4074, opened 2025-02-19
fix(Litestar): Apply `failed_request_status_codes` to exceptions raised in middleware

vrslev:
This is a fix for #4021: exceptions raised in middleware were sent without taking into account `failed_request_status_codes` value.

See the test case for an example.

---

### DIFF
--- a/tests/integrations/litestar/test_litestar.py
+++ b/tests/integrations/litestar/test_litestar.py
@@ -488,3 +488,6 @@ def test_catch_non_http_exceptions_in_middleware(
         pass
 
     assert len(events) == 1
+    event_exception = events[0]["exception"]["values"][0]
+    assert event_exception["type"] == "RuntimeError"
+    assert event_exception["value"] == "Too Hot"


vrslev:
Relates to #3134

vrslev:
@sentrivana @antonpirker @untitaker can you review this please? 🙂

sentry_sdk/integrations/asgi.py
```diff
@@ -57,17 +57,6 @@
 TRANSACTION_STYLE_VALUES = ("endpoint", "url")
 
 
-def _capture_exception(exc, mechanism_type="asgi"):
```
szokeasaurusrex, line 60: I would prefer to keep this function the same, as a standalone function instead of a method. I don't think it has to be modified to implement your changes

sentry_sdk/integrations/asgi.py
```diff
@@ -145,6 +134,23 @@ def __init__(
...
+            client_options=sentry_sdk.get_client().options,
+            mechanism={"type": self.mechanism_type, "handled": False},
+        )
+        sentry_sdk.capture_event(event, hint=hint)
+
+    def _capture_lifespan_exception(self, exc):
+        # type: (Exception) -> None
+        return self._capture_exception(exc)
+
+    def _capture_request_exception(self, exc):
```
szokeasaurusrex, line 150: I am a bit unsure of the difference between these two functions. Please add docstrings to both to clarify the distinction

tests/integrations/litestar/test_litestar.py
```diff
@@ -16,7 +16,6 @@
 from litestar.middleware.rate_limit import RateLimitConfig
 from litestar.middleware.session.server_side import ServerSideSessionConfig
 from litestar.testing import TestClient
-
```
szokeasaurusrex, line 19: [nit] please add this line back

sentry_sdk/integrations/litestar.py
```diff
@@ -87,6 +87,10 @@ def __init__(self, app, span_origin=LitestarIntegration.origin):
             span_origin=span_origin,
         )
 
+    def _capture_request_exception(self, exc):
+        # type: (Exception) -> None
+        """Ignore exceptions for requests here: we catch them in Litestar.after_exception handler."""
```
szokeasaurusrex, line 92: I am slightly unsure why this function is needed/how it relates to what you are trying to explain in this PR. Could you please explain in a bit more detail?

Also, I think you need a `pass` here:

```suggestion
    def _capture_request_exception(self, exc):
        # type: (Exception) -> None
        """Ignore exceptions for requests here: we catch them in Litestar.after_exception handler."""
        pass
```

szokeasaurusrex (commented):
Thank you for the contribution! I need some extra clarification on some items and also have some suggestions, please see the comments I left

sentry_sdk/integrations/litestar.py
vrslev: Currently, Litestar integration catches exceptions raised in middleware twice:

1. In the custom ASGI handler.
2. In the `Litestar.after_exception` handler, which is added in `patch_app_init`.

The custom ASGI handler doesn't filter exceptions using `failed_request_status_codes`, so all exceptions (including those not filtered by the expected status code) appear in Sentry.

Interestingly, this behavior doesn't apply to exceptions raised in request handlers because Litestar prepends the built-in exception handling middleware.

---

This leads us to the conclusion that the ASGI handler only catches exceptions raised in middleware or lifespan handlers. It's still valuable to catch exceptions from lifespan handlers.

In this case, we patch `_capture_request_exception` to prevent catching exceptions raised in middleware.

sentry_sdk/integrations/litestar.py
```diff
@@ -87,6 +87,15 @@ def __init__(self, app, span_origin=LitestarIntegration.origin):
             span_origin=span_origin,
         )
 
+    def _capture_request_exception(self, exc):
+        # type: (Exception) -> None
+        """Avoid catching exceptions from request handlers.
+
+        Those exceptions are already han in Litestar.after_exception handler.
```
szokeasaurusrex:
```suggestion
        Those exceptions are already handled in Litestar.after_exception handler.
```

vrslev: Done!

vrslev:
Hey @szokeasaurusrex, I've added the test you asked for :)

tests/integrations/litestar/test_litestar.py
```diff
@@ -427,3 +427,64 @@ async def error() -> None:
...
+
+    app = Litestar([error], middleware=[create_raising_middleware])
+    client = TestClient(app)
+
+    try:
+        client.get("/error")
+    except RuntimeError:
+        pass
+
+    assert len(events) == 1
```
szokeasaurusrex, line 490: I'd consider also asserting that the event is as expected (i.e. it is a runtime error with the message "Too Hot")

szokeasaurusrex (approved):
lgtm, but I'd like a secondary review from someone else on the team before we merge

szokeasaurusrex (approved):

tests/integrations/litestar/test_litestar.py
vrslev: Makes sense. I’ll add it tomorrow 👍

vrslev: Done!

sentrivana (approved):
Thank you @vrslev and sorry for the wait!

PR #4076, opened 2025-02-19
Update tox setup for AWS Lambda integration

antonpirker:
Because we changed our AWS Lambda test suite to not create real Lambda functions but rather run a Lambda environment locally, we do not need all the code necessary for handling the authentication to AWS in the test suite. 

This PR removes all the AWS auth related code and als moves the AWS Lambda testsuite again into the "Cloud" group of test suites, because it is a "normal" test suite again.

The PR that includes the change to have a local Lambda environment is here: #3988


sentrivana (approved):
#feelsgood

There is couple more `aws-*` scripts in `scripts/`, do we still need any of those?

antonpirker:
Those scripts are still needed if someone is doing manual testing (creating a Lambda function by hand and wanting to add the Layer to it and stuff like this)

I cleaned the `scripts/` directory up a bit (and removed some other Lambda function that is not necessary anymore.

PR #4080, opened 2025-02-20
fix(scopes): Add more typing info to Scope.update_from_kwargs's "contexts"

Perlence:
The original type hint could be understood as a one-level `dict` of `str` to `Any`, when in fact, it's a two-level dict.


antonpirker (approved):
Looks good! Thanks for the contribution @Perlence!

PR #4099, opened 2025-02-25
security(gha): fix potential for shell injection

mdtro:
Running these workflows is gated pretty well, but this mitigates the potential for a script injection attack by passing the input to an intermediary environment variable first.

See https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#example-of-a-script-injection-attack for more details.


antonpirker (approved):
Very nice! Thank's @mdtro for discovering and fixing this!

PR #4100, opened 2025-02-25
Add `init()` parameters to ApiDocs.

antonpirker:
Copied the text from docs.sentry.io and added it to the ApiDocs. (some parameters are undocumented, it seems)


sentrivana (commented):
Should we maybe just link to the docs page?

I'm afraid this will get out of sync.

antonpirker:
The code should be single source of truth. In the long run the docs should be generated from the Apidocs in code. (and also this way we can improve how our SDK behaves in IDEs) And also this will make it easier for LLMs to understand how to use our SDK which in the long run will help all those people that will just ask LLMs to integrate Sentry in their project :)

sentrivana (approved):
Ok then, let's try it out and see how it goes. :)

PR #4115, opened 2025-03-03
Fixed bug when `cron_jobs` is set to `None` in arq integration

### LINKED ISSUE #3827: ARQ Integration can create an TypeError with cron_jobs
### How do you use Sentry?

Self-hosted/on-premise

### Version

2.19.0

### Steps to Reproduce

When working with the ARQ integration, if you set `cron_jobs` to `None` in your settings the sentry-sdk will crash with the following:

```
| File "/app/.venv/lib/python3.12/site-packages/sentry_sdk/integrations/arq.py", line 209, in _sentry_create_worker                                                                                                        │
│     for cron_job in settings_cls["cron_jobs"]                                                                                                                                                                                                   │
│                     ~~~~~~~~~~~~^^^^^^^^^^^^^
| TypeError: 'NoneType' object is not iterable
```

This was introduced in #3742 and the fix could simply be to check if `cron_jobs`  is present (like now) and not `None`.

### Expected Result

Sentry SDK should initialize correctly and integrate with ARQ

### Actual Result

```
| File "/app/.venv/lib/python3.12/site-packages/sentry_sdk/integrations/arq.py", line 209, in _sentry_create_worker                                                                                                        │
│     for cron_job in settings_cls["cron_jobs"]                                                                                                                                                                                                   │
│                     ~~~~~~~~~~~~^^^^^^^^^^^^^
| TypeError: 'NoneType' object is not iterable
```

antonpirker:
Handle `None` values in arq configuration gracefully. 

Fixes #3827

Unfortunately our test setup for arq does not allow for testing `None` values in the configuration because they are changed to `[]` in the arq init code in the test suite [here](https://github.com/getsentry/sentry-python/blob/master/tests/integrations/arq/test_arq.py#L50-L51) and [here](https://github.com/getsentry/sentry-python/blob/master/tests/integrations/arq/test_arq.py#L95-L96). 

This is why I have not added additional tests for this bug fix.

### DIFF
--- a/sentry_sdk/integrations/arq.py
+++ b/sentry_sdk/integrations/arq.py
@@ -199,7 +199,8 @@ def patch_create_worker():
         if isinstance(settings_cls, dict):
             if "functions" in settings_cls:
                 settings_cls["functions"] = [
-                    _get_arq_function(func) for func in settings_cls.get("functions", [])
+                    _get_arq_function(func)
+                    for func in settings_cls.get("functions", [])
                 ]
             if "cron_jobs" in settings_cls:
                 settings_cls["cron_jobs"] = [
--- a/sentry_sdk/integrations/arq.py
+++ b/sentry_sdk/integrations/arq.py
@@ -204,7 +204,7 @@ def patch_create_worker():
             if "cron_jobs" in settings_cls:
                 settings_cls["cron_jobs"] = [
                     _get_arq_cron_job(cron_job)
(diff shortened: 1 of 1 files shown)


szokeasaurusrex (approved):
Maybe we could make a test, which does not use the `init_arq` test fixture, to test this? But if it is too complicated, let's not block the bug fix on a lack of a test.

PR #4117, opened 2025-03-04
Better handling of exception groups

### LINKED ISSUE #3913: If Exception.__context__ is an ExceptionGroup, Sentry doesn't capture the ExceptionGroup's children
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.29.2

### Steps to Reproduce

Here's a toy example that demonstrates the problem:
```python
try:
    try:
        raise ExceptionGroup("group", [ValueError("child1"), ValueError("child2")])
    finally:
        raise TypeError("bar")
except BaseException:
    sentry_sdk.capture_exception()
```

`sentry_sdk.capture_exception()` receives the `TypeError`. `TypeError.__context__` points to the `ExceptionGroup`; and then `ExceptionGroup.exceptions` points to `[ValueError("child1"), ValueError("child2")]`.

### Expected Result

The Sentry event should contain all four exceptions: the `TypeError`, the `ExceptionGroup`, and the sub-exceptions `ValueError("child1")` and `ValueError("child2")`.

Additionally, the event should record that the `ExceptionGroup` is `TypeError.__context__` as opposed to `TypeError.__cause__`.

Note that the `sentry_sdk.utils.exceptions_from_error()` function actually does this correctly, but it isn't getting invoked in this case. Here's a screenshot of how it looks if I patch the Sentry SDK to always use `sentry_sdk.utils.exceptions_from_error()`:

Image

### Actual Result

When the `TypeError` is passed into `sentry_sdk.utils.exceptions_from_error_tuple()`, it checks whether the `TypeError` is an `ExceptionGroup` or not: https://github.com/getsentry/sentry-python/blob/2.19.2/sentry_sdk/utils.py#L968-L1009 Because the `TypeError` is not an `ExceptionGroup`, it calls `walk_exception_chain()`, which examines `TypeError.__context__` to find the `ExceptionGroup`; but it doesn't recurse into the sub-exceptions `ValueError("child1")` and `ValueError("child2")`. So the Sentry event only contains the `TypeError` and the `ExceptionGroup`:

Image

Also, the "Related Exceptions" tree in the above screenshot appears to be nonsensical.

I tested this with SDK version 1.29.2, but the bug appears to still be present in the latest version of the code.

antonpirker:
Properly handle grouped and chained exceptions. The test case in the linked issue illustrates that some ExceptionGroups have been handled in a wrong way. 

Updated some tests, because now that those are handled correctly all the mechanism types except for the root exception are set to "chained" like described in the RFC: https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md#interpretation

Because this will change the grouping of exiting Sentry Issues containing ExceptionGroups, it is safer to release this fix in the next major and make sure that we describe the change in behavior in the changelog. (Note: The grouping in the Ariadne issues will not change because those are not ExceptionGroups and only updating the `mechanism.type` does not change the grouping)

Fixes #3913


sentry_sdk/utils.py
```diff
@@ -697,7 +697,8 @@ def single_exception_from_error_tuple(
     is_root_exception = exception_id == 0
     if not is_root_exception and parent_id is not None:
         exception_value["mechanism"]["parent_id"] = parent_id
-        exception_value["mechanism"]["type"] = "chained"
+        if exception_id is not None and exception_id > 0:
```
szokeasaurusrex, line 700: [question] Given that we are already checking above that `is_root_exception` is `False`, which implies `exception_id != 0` (which, I believe is usually equivalent to `exception_id > 0` assuming it is not None since the ID should be nonnegative), can you explain the purpose of adding this conditional check?

sentry_sdk/utils.py
```diff
@@ -897,36 +898,17 @@ def exceptions_from_error_tuple(
     # type: (...) -> List[Dict[str, Any]]
     exc_type, exc_value, tb = exc_info
 
-    is_exception_group = BaseExceptionGroup is not None and isinstance(
-        exc_value, BaseExceptionGroup
+    _, exceptions = exceptions_from_error(
```
szokeasaurusrex, line 901: [question] does the `exceptions_from_error` function already internally handle the non-exception group case?

From naively looking at the diff, it seems that the code for handling exceptions which are not exception groups has been deleted, so I am just wondering where this logic is handled now.

szokeasaurusrex (commented):
@antonpirker I think I am generally lacking context in this part of the codebase, so I am struggling to understand how exactly this PR changes the exception grouping behavior. 

A walkthrough or technical explanation of how the changes work might help. Although, I suppose it is not urgent, since we will wait for a major release, anyways.

sentrivana (commented):
I took a look too and have the same questions as Daniel.

OT: we can change the base of this to `potel-base` so that it gets released with the next major.

sentry_sdk/utils.py
antonpirker: True. The if is probably useless.

sentry_sdk/utils.py
antonpirker: I have now updated the code to make it easier to grasp what is happening. (The diff is very hard to read, just checkout the file)

The function `exceptions_from_error` handles all the cases, single exceptions, chained exceptions and exception groups. It walks through the complete tree of exceptions. 

(We can in another PR also think about removing `walk_exception_chain` everywhere and just use `exceptions_from_error` for everything. But this is outside of the scope of this PR)

antonpirker:
I am closing this in favor of this PR that will be merged into potel-base:
https://github.com/getsentry/sentry-python/pull/4164

PR #4120, opened 2025-03-06
Added timeout to HTTP requests in CloudResourceContextIntegration

### LINKED ISSUE #2376: CloudResourceContextIntegration freezing in ECS Container
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

1.31.0

### Steps to Reproduce

Configure the CloudResourceContextIntegration without any parameters but run in an ECS container on AWS - which is running on EC2 hardware.

My suggestion is to either make the IMDS address configurable or do some detection to see if you are on ECS. There is an IMDS for ECS you could use.

At the very least make the http have a timeout so that things are not totally locked up.

### Expected Result

Not freeze

### Actual Result

Infinite freeze at https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/cloud_resource_context.py#L74

antonpirker:
The URL that works in EC2 does not work in ECS, this can lead to the HTTP request getting stuck. 

Fixes #2376

### DIFF
--- a/sentry_sdk/integrations/cloud_resource_context.py
+++ b/sentry_sdk/integrations/cloud_resource_context.py
@@ -86,7 +86,9 @@ class CloudResourceContextIntegration(Integration):
             return True
 
         except urllib3.exceptions.TimeoutError:
-            logger.debug("AWS metadata service timed out after %s seconds", HTTP_TIMEOUT)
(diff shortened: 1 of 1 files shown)


sentrivana (approved):
lgtm

PR #4121, opened 2025-03-06
Coerce None values into strings in logentry params.

### LINKED ISSUE #3660: Logging message is unrendered on a webhook call
### How do you use Sentry?

Self-hosted/on-premise

### Version

2.9.0

### Steps to Reproduce

We updated self hosted sentry to version 24.8.0 and after that logging integration do not render logging message. So we have a template in a sentry web-ui and webhook call instead of rendered message. For exampe:
```python
logger.error("task for news source %s returned empty result %s", news_source, result)
logger.error(f"[debug] task for news source {news_source} returned empty result {result}")
```
This leads to 2 messages in sentry:
- `task for news source %s returned empty result %s`
- `[debug] task for news source binance_api_changelog_spot returned empty result None`

So, only force-rendered via f-string message is correct. Not sure if it is relative to sentry python sdk, or self-hosted sentry itself. Previous version 23.x.x was correct for this. More to say, some messages are rendered correctly. For example with `%d`. In the example above `news_source` is `StrEnum` and `result` is `None`.

### Expected Result

-

### Actual Result

-

antonpirker:
Nice rendering of log messages containing parameters that are `None` values does not work. There we coerce `None` values into strings to have nicer messages in Sentry UI.

Fixes #3660

### DIFF
--- a/sentry_sdk/integrations/logging.py
+++ b/sentry_sdk/integrations/logging.py
@@ -248,7 +248,9 @@ class EventHandler(_BaseHandler):
         else:
             event["logentry"] = {
                 "message": to_string(record.msg),
-                "params": record.args,
+                "params": (
+                    tuple(to_string(arg) for arg in record.args) if record.args else ()
+                ),
             }
 
         event["extra"] = self._extra_from_record(record)


sentrivana (approved):
Nice.

Some tests also need updating, it seems. It'd also be great to have a test case for the `None` case specifically.

PR #4136, opened 2025-03-17
Remove `test-integrations-aws.yml`.

antonpirker:
AWS tests are run in `test-integrations-cloud.yml`, so this file is obsolete


antonpirker:
not needed anymore.

PR #4159, opened 2025-03-18
Using RLock to prevent deadlocks when using sentry and ddtrace at the same time.

antonpirker:
Refs #3753

### DIFF
--- a/sentry_sdk/worker.py
+++ b/sentry_sdk/worker.py
@@ -21,7 +21,7 @@ class BackgroundWorker:
     def __init__(self, queue_size=DEFAULT_QUEUE_SIZE):
         # type: (int) -> None
         self._queue = Queue(queue_size)  # type: Queue
-        self._lock = threading.Lock()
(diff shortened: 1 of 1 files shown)


sl0thentr0py:
what is the actual problem and how does this fix it?

antonpirker:
This is an experiment. 

The problem described in https://github.com/getsentry/sentry-python/issues/3753 is roughly like this:
- DataDog is monitoring lock operations
- Sentry's worker thread needs to acquire a lock
- When lock operations occur, DataDog tries to log information
- Logging triggers Sentry's logging integration
- Sentry tries to capture this event
- Capturing an event requires Sentry's worker thread
- The worker thread needs to acquire a lock...

By changing to RLock the given reproduction does not run into a deadlock anymore. But I am not 100% sure that this does not lead to other problems. (I think it does not, but I am not sure)

antonpirker:
I wanted to see if any of our test start failing if using RLock.

PR #4165, opened 2025-03-19
Fixed flaky test

antonpirker:
The URL www.squirrelchasers.com is actually existing, so we should not access it in our tests. Hope this make the test more stable.


sentrivana (approved):
😄

PR #4175, opened 2025-03-20
fix(integrations/dramatiq): use set_transaction_name

timdrijvers:
The Dramatiq integration is using a deprecated method to set the scope's transaction name, use set_transaction_name instead.

"Assigning to scope.transaction directly is deprecated: use scope.set_transaction_name() instead."


sentrivana (approved):
Thank you @timdrijvers!

PR #4178, opened 2025-03-20
feat(profiling): add platform header to the chunk item-type in the envelope

viglia:
We need to send the platform as part of the headers in the chunk item-type as this is the header that relay is checking to manage rate limiting.

### DIFF
--- a/sentry_sdk/envelope.py
+++ b/sentry_sdk/envelope.py
@@ -82,7 +82,7 @@ class Envelope:
             Item(
                 payload=PayloadRef(json=profile_chunk),
                 type="profile_chunk",
-                headers={"platform": profile_chunk.get("platform")},
+                headers={"platform": profile_chunk.get("platform", "python")},
             )
         )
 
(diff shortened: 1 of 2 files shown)


sentry_sdk/envelope.py
```diff
@@ -79,7 +79,11 @@ def add_profile_chunk(
     ):
         # type: (...) -> None
         self.add_item(
-            Item(payload=PayloadRef(json=profile_chunk), type="profile_chunk")
+            Item(
+                payload=PayloadRef(json=profile_chunk),
+                type="profile_chunk",
+                headers={"platform": profile_chunk.get("platform")},
```
phacops, line 85: I wouldn't mind to avoid any potential `None` value here. Can we hardcode the same value we set in `profile_chunk`? Or only add the headers if we have a key `platform`?

viglia: @phacops I could throw-in an extra check if you wish but just fyi: if `profile_chunk.get("platform")` returns `None` it means the platform is not set at the payload either and therefore the whole profile would be rejected as it does not comply to the Relay defined schema (platform is a mandatory field).

phacops: True. I'm just allergic to `null` values.

phacops (approved):

sentrivana (approved):

PR #4181, opened 2025-03-21
ci: Move `mypy` config into `pyproject.toml`

antonpirker:
First step to consolidate configuration into `pyproject.toml`.


pyproject.toml
```diff
@@ -20,4 +20,136 @@ omit = [
...
+module = "huggingface_hub.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "arq.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "grpc.*"
+ignore_missing_imports = true
```
szokeasaurusrex, line 155:
```suggestion
ignore_missing_imports = true

```

Newline is missing ([why this is somewhat important](https://sourabhkr.medium.com/why-should-text-files-end-with-a-newline-c2e08d4896b8#:~:text=In%20the%20UNIX%20world%2C%20a,to%20unexpected%20behavior%20or%20errors.)). I believe there's a way to configure vscode to automatically add final newline on save.

pyproject.toml
```diff
@@ -20,4 +20,136 @@ omit = [
...
+[tool.mypy]
+allow_redefinition = true
+check_untyped_defs = true
+disallow_any_generics = true
+disallow_incomplete_defs = true
+disallow_subclassing_any = true
+disallow_untyped_decorators = true
+disallow_untyped_defs = true
+no_implicit_optional = true
+python_version = "3.11"
```
szokeasaurusrex, line 34: Out of scope for this PR, but perhaps we should bump this?

szokeasaurusrex (approved):
See comments before merge. Also, if there is an issue associated with this PR, might help to link them for context (I was kinda missing context when reviewing this)

sentrivana (approved):

PR #4186, opened 2025-03-21
fix: Always set _spotlight_url

BYK:
The conditional early exit in `SpotlightMiddleware` may cause attribute access errors when trying to check if `_spotlight_url` is set or not. This patch sets it to `None` explicitly at class level.


BYK:
Test fail looks unrelated?

sentrivana (approved):

PR #4204, opened 2025-03-27
Update Ubuntu in Github test runners

antonpirker:
Because `ubuntu-20.04` will be removed on April 1st.

### DIFF
--- a/.github/workflows/test-integrations-ai.yml
+++ b/.github/workflows/test-integrations-ai.yml
@@ -34,7 +34,7 @@ jobs:
         # new versions of hosted runners on Github Actions
         # ubuntu-20.04 is the last version that supported python3.6
         # see https://github.com/actions/setup-python/issues/544#issuecomment-1332535877
-        os: [ubuntu-20.04]
+        os: [ubuntu-22.04]
     steps:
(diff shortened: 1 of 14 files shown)


tests/integrations/celery/integration_tests/test_celery_beat_cron_monitoring.py
```diff
@@ -52,6 +53,7 @@ def inner(propagate_traces=True, monitor_beat_tasks=False, **kwargs):
     return inner
 
 
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="Requires Python 3.7+")
```
antonpirker, line 56: Those integration tests (that are more for documentation) are hanging in Python 3.6 when using the Python Docker container. I just deactivated them, because we will remove 3.6 support soon anyhow.

antonpirker:
I tried to upgrade to 24.04 but then we need to use Docker for Python 3.6 and 3.7 (we need docker because Ubuntu does not have the old python installation packages anymore.)
And when switching to 24.04 some other things broke, and after an hour or two trying to fix this I just reverted and made 22.04 work.

PR #4217, opened 2025-03-31
Fixed some tests

antonpirker:
- function transports have been removed. removing the related test
- `profiles_sample_rate` is not experimental anymore
- `continuous_profiling_mode` has been removed so only the top level `profiler_mode` exists. (so the if is not necessary anymore)

### DIFF
--- a/tests/profiler/test_continuous_profiler.py
+++ b/tests/profiler/test_continuous_profiler.py
@@ -24,25 +24,16 @@ except ImportError:
 requires_gevent = pytest.mark.skipif(gevent is None, reason="gevent not enabled")
 
 
-def get_client_options(use_top_level_profiler_mode):
+def get_client_options():
     def client_options(
         mode=None, auto_start=None, profile_session_sample_rate=1.0, lifecycle="manual"
     ):
-        if use_top_level_profiler_mode:
-            return {
-                "profile_lifecycle": lifecycle,
-                "profiler_mode": mode,
-                "profile_session_sample_rate": profile_session_sample_rate,
-                "_experiments": {
-                    "continuous_profiling_auto_start": auto_start,
-                },
-            }
         return {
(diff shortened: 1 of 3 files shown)


antonpirker:
@sentrivana this PR makes the last three tests green. :-)

tests/profiler/test_continuous_profiler.py
```diff
@@ -83,8 +73,7 @@ def test_continuous_profiler_invalid_mode(mode, make_options, teardown_profiling
 @pytest.mark.parametrize(
     "make_options",
     [
-        pytest.param(get_client_options(True), id="non-experiment"),
-        pytest.param(get_client_options(False), id="experiment"),
+        pytest.param(get_client_options()),
```
szokeasaurusrex, line 76: [question] I am not familiar with `pytest.param`. Do we still need the parameterization though? It looks like there is only one item we parameterize by, so unless `pytest.param` somehow expands to more parameterizations, it seems we could get rid of the `pytest.mark.parametrize`

tests/test_client.py
```diff
@@ -378,13 +378,6 @@ def test_socks_proxy(testcase, http2):
     )
 
 
-def test_simple_transport(sentry_init):
```
szokeasaurusrex, line 381: Do we still support creating a custom transport by subclassing the `Transport` class? If yes, we should add a test for that functionality unless we have one already.

szokeasaurusrex (approved):
lgtm, see comments before merging

tests/test_client.py
antonpirker: we have a test_transport_option so this is covered.

tests/profiler/test_continuous_profiler.py
antonpirker: not worth the effort

PR #4219, opened 2025-03-31
feat: Do not spam sentry_sdk.warnings logger w/ Spotlight

BYK:
Sometimes one may have Spotlight turned on in the SDK but not have the sidecar running or reachable. In that case we spam the console with every event as they fail to reach Spotlight. This patch limits the fail warnings to 3: the first 2 are actual errors and the final one is a note about shutting up.


antonpirker (approved):
looks good

Issue #3482, opened 2024-08-30
Make sure span / transaction statuses work for all cases


sl0thentr0py:

sentrivana:
Will do this in scope of the manual testing

PR #4221, opened 2025-04-01
Add deprecation message to `Scope.transaction`

antonpirker:
This is the first part of #3482


szokeasaurusrex (commented):
The warning looks good, but it seems like we still use `Scope.transaction` several places throughout the codebase. 

I believe we should first remove those usages, then add the deprecation warning afterwards. Unless, is there some reason why we should instead do this change first?

antonpirker:
Yes you are right. Then it is not worth the effort. Closing this.

PR #4222, opened 2025-04-01
Deprecate `Span.set_data`

antonpirker:
This is the first part of #4102


szokeasaurusrex (commented):
Same comment as #4221; it looks like we use `Span.set_data` in many places – I think we should remove those usages first to avoid potentially causing users to encounter this deprecation warning

antonpirker:
Then the deprecation message will only happen in 3.0. So I am closing this.

PR #4224, opened 2025-04-01
Debug output from Sentry logs should always be `debug` level.

antonpirker:
Prevent emitting too many log messages.


szokeasaurusrex (approved):
Looks good! Although maybe the higher-severity messages (e.g. `warn`, `error`, and `fatal`) should get mapped to the `info` level?

PR #4226, opened 2025-04-01
fix: Handle JSONDecodeError gracefully in StarletteRequestExtractor

moodix:
Previously, when encountering malformed JSON in request bodies, the json() method would raise a JSONDecodeError. This change updates the method to catch the exception and return None instead, providing more consistent behavior and preventing unexpected crashes.

Added a test case to verify this error handling behavior.


sentrivana (approved):
Thanks @moodix!

moodix:
@sentrivana Looks like the workflow might need approval

PR #4227, opened 2025-04-01
fix(ourlogs): Use repr instead of json for message and arguments

colin-sentry:
Currently if you do something like

```
    python_logger = logging.Logger("test-logger")
    python_logger.error(Exception("test exc"))
```

It will error, because Exception is not JSON serializable.


antonpirker (approved):
Looks good. 

I added some more asserts to the test just to show what is actually sent. And we have a `safe_repr()` that handles error when calling repr() on objects. Just to be safe.

Issue #4225, opened 2025-04-01
Move the sentry.logging methods out of the _experimental namespace


colin-sentry:
Also need to move the Log type out of private _types into public types

Issue #4155, opened 2025-03-18
API should be `.warning()` instead of `.warn()`


antonpirker:
Change the API in the existing proof of concept to have the same API as the Python `logging` module.

Issue #4152, opened 2025-03-18
Batch send log items


antonpirker:
In the current PoC each log statement triggers the sending of an envelope. This should be changed. 

We should wait for X seconds or Y log items being collected until we create one envelope containing all the available log items and send them to Sentry.

Should work similar as it works/worked for `metrics` or `sessions`. 

We should also think about having just one thread that collects different kind of data items and sends them to Sentry. Because we already create some threads in the SDK and we sometimes already hit the limits of environments in regard of maximum numbers of threads allowed (in AWS Lambda this is already a problem)

PR #4229, opened 2025-04-01
fix(ourlogs): Add a class which batches groups of logs together.

colin-sentry:
Currently, sentry logs create a new envelope per-log, which is inefficient.

This changes the behavior to batch a large chunk of logs to be sent all at once.

Fixes https://github.com/getsentry/sentry-python/issues/4155
Fixes https://github.com/getsentry/sentry-python/issues/4225
Fixes https://github.com/getsentry/sentry-python/issues/4152

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*   @getsentry/owners-python-sdk
+*   @getsentry/team-web-sdk-backend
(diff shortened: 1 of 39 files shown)


antonpirker (approved):
After testing this PR locally, and some discussion outside of Github with Colin, and also because of the fact it is heavily based on the MetricsAggregator this is good to be merged.

PR #4232, opened 2025-04-02
fix(ai): Do not consume anthropic streaming stop

colin-sentry:
The old functionality wouldn't re-emit the `stop` message for streaming Anthropic calls.


antonpirker (approved):
Thanks!

PR #4256, opened 2025-04-08
Add multiplexed Transport

milanboers:
Adds a `make_multiplexed_transport` that sends all envelopes to multiple Sentry DSNs (backends or projects).

This is more or less a port of `makeMultiplexedTransport` [in the Javascript SDK](https://github.com/getsentry/sentry-javascript/blob/30591d4f536a65418577bef8255f0ea959b497c5/packages/core/src/transports/multiplexed.ts#L90) (relevant [issue](https://github.com/getsentry/sentry-javascript/pull/7926)).

It's a common ask from customers, and there seem to be several issues about it on github/online. Would you consider adding it to the Python SDK, like it is in the Javascript SDK?

- https://github.com/getsentry/sentry-python/discussions/1484
- https://github.com/getsentry/sentry-java/issues/620
- https://forum.sentry.io/t/java-sdk-multiple-dsns/12715
- https://stackoverflow.com/questions/58718196/how-to-have-multiple-global-dsn-in-sentry-sdk-in-python

(The Javascript SDK also allows routing events based on the event content. This seems more useful for the microfrontends use case, but it can be added)

Usage:
```python
sentry_sdk.init(
    dsn=dsn1,
    transport=make_multiplexed_transport([dsn1, dsn2]),
)
```

Opened a similar PR for Java: https://github.com/getsentry/sentry-java/pull/4323

### DIFF
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -878,6 +878,82 @@ class _FunctionTransport(Transport):
             self.capture_event(event)
 
 
+def make_multiplexed_transport(dsns):
+    # type: (List[str]) -> Type[Transport]
+    class MultiplexedTransport(Transport):
+        def __init__(self, options):
+            # type: (Self, Optional[Dict[str, Any]]) -> None
+            super().__init__(options)
+            self.transports = list(
+                filter(
+                    None,
+                    [
+                        make_transport(
+                            {**(options or {}), "dsn": dsn, "transport": None}
+                        )
+                        for dsn in dsns
+                    ],
+                )
+            )
+
+        @staticmethod
+        def _override_dsn(envelope, dsn):
+            # type: (Envelope, Optional[Dsn]) -> Envelope
+            if (
+                dsn is None
+                or "trace" not in envelope.headers
+                or "public_key" not in envelope.headers["trace"]
+            ):
+                return envelope
+            return Envelope(
+                {
+                    **envelope.headers,
+                    "trace": {
+                        **envelope.headers["trace"],
+                        "public_key": dsn.public_key,
+                    },
+                },
+                envelope.items,
+            )
+
+        def capture_envelope(self, envelope):
+            # type: (Self, Envelope) -> None
+            for transport in self.transports:
+                transport.capture_envelope(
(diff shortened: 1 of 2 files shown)


sentry_sdk/transport.py
```diff
@@ -878,6 +878,82 @@ def capture_envelope(self, envelope: Envelope) -> None:
...
+                        make_transport(
+                            {**(options or {}), "dsn": dsn, "transport": None}
+                        )
+                        for dsn in dsns
+                    ],
+                )
+            )
+
+        @staticmethod
+        def _override_dsn(envelope, dsn):
```
milanboers, line 900: this is a bit ugly and the JS SDK doesn't do it, but maybe nice to have

szokeasaurusrex:
Hey @milanboers, we appreciate that you took the time to open this PR; however, after discussing with the team, we have decided to close this PR since we already support other ways of getting events sent to multiple DSNs.

First of all, by passing a custom transport to `sentry_sdk.init` you can simply use the multiplexed transport that you implemented in this PR in order to send your events to multiple DSNs. This is essentially what you are suggesting in this PR, but instead of adding the multiplexed transport to the SDK, you would just add it to your application code. You could even release `make_multiplexed_transport` in a separate library if you really want to. You can see some other [example custom transports on this discussion](https://github.com/getsentry/sentry-python/discussions/1484); you could also post yours there if you would like.

Another approach would be to use [Sentry Mirror](https://github.com/getsentry/sentry-mirror). Sentry Mirror is a proxy that provides a DSN you can send events to; the Sentry Mirror then forwards the events it receives at that DSN to multiple outgoing DSNs. By using Sentry Mirror, you would avoid having to implement a custom transport.

I hope that those options solve your use case. If you have any further questions, or if we can help you in any way, please feel free too reach out to us, either by opening an issue on GitHub or by [pinging us on Discord](https://discord.com/invite/sentry).

___________________

P.S. In the future if you have a feature request, I would strongly recommend opening an issue rather than opening a PR. That way, you avoid wasting your time writing a PR if we decide against adding the feature to the SDK.

PR #4257, opened 2025-04-09
feat(logs): Add server.address to logs

AbhiPrasad:
Docs: https://develop-docs-git-abhi-logs-sdk-developer-documentation.sentry.dev/sdk/telemetry/logs/#default-attributes

> [BACKEND SDKS ONLY] `server.address`: The address of the server that sent the log. Equivalent to server_name we attach to errors and transactions.

`server.address` convention docs: https://getsentry.github.io/sentry-conventions/generated/attributes/server.html#serveraddress

resolves https://linear.app/getsentry/issue/LOGS-33

### DIFF
--- a/tests/test_logs.py
+++ b/tests/test_logs.py
@@ -11,6 +11,7 @@ from sentry_sdk import get_client
 from sentry_sdk.envelope import Envelope
 from sentry_sdk.integrations.logging import LoggingIntegration
 from sentry_sdk.types import Log
+from sentry_sdk.consts import SPANDATA
 
 minimum_python_37 = pytest.mark.skipif(
     sys.version_info < (3, 7), reason="Asyncio tests need Python >= 3.7"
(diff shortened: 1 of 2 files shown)


tests/test_logs.py
```diff
@@ -184,6 +185,7 @@ def test_logs_attributes(sentry_init, capture_envelopes):
     assert logs[0]["attributes"]["sentry.environment"] == "production"
     assert "sentry.release" in logs[0]["attributes"]
     assert logs[0]["attributes"]["sentry.message.parameters.my_var"] == "some value"
+    assert logs[0]["attributes"][SPANDATA.SERVER_ADDRESS]
```
szokeasaurusrex, line 188: [question] why do we only assert that this value is truthy, and not verify the specific value? Is it impossible to predict the value?

szokeasaurusrex (approved):
Nice!

tests/test_logs.py
AbhiPrasad: I didn't know what CI would set this as, or if it would change between CI environments, so I figured this was the safest option to ensure stable tests.

We can change this if we know what the stable value is going to be.

szokeasaurusrex: I think if you set `server_name` in `sentry_sdk.init` it would use that name

PR #4273, opened 2025-04-10
ref(scope): Properly type `Scope.root_span`

szokeasaurusrex:
Currently, this property has type `Any`, but it can now be changed to `Optional[Span]`

Depends on:
  - #4263


sentrivana (approved):
nice!

PR #4278, opened 2025-04-10
test(tracing): Simplify static/classmethod tracing tests

szokeasaurusrex:
These tests were causing flakes where the mock method was being called more than once. The tests were also difficult to understand.

This change removes the need for mocking (hopefully increasing test stability) and also should hopefully make it easier to understand what these tests are meant to be checking


szokeasaurusrex:
ah looks like there are some interactions between the tests here

sentrivana (approved):

PR #4280, opened 2025-04-10
feat(spans): Record flag evaluations as span attributes

cmanallen:
Flags evaluated within a span are appended to the span as attributes.

### DIFF
--- a/sentry_sdk/integrations/unleash.py
+++ b/sentry_sdk/integrations/unleash.py
@@ -1,7 +1,7 @@
 from functools import wraps
 from typing import Any
 
-import sentry_sdk
+from sentry_sdk.feature_flags import add_feature_flag
 from sentry_sdk.integrations import Integration, DidNotEnable
 
 try:
@@ -24,17 +24,9 @@ class UnleashIntegration(Integration):
             # type: (UnleashClient, str, *Any, **Any) -> Any
             enabled = old_is_enabled(self, feature, *args, **kwargs)
 
-            # Errors support.
-            #
             # We have no way of knowing what type of unleash feature this is, so we have to treat
             # it as a boolean / toggle feature.
-            flags = sentry_sdk.get_current_scope().flags
-            flags.set(feature, enabled)
-
-            # Spans support.
-            span = sentry_sdk.get_current_span()
-            if span:
-                span.set_data(f"flag.evaluation.{feature}", enabled)
+            add_feature_flag(feature, enabled)
 
             return enabled
 
--- a/sentry_sdk/integrations/unleash.py
+++ b/sentry_sdk/integrations/unleash.py
@@ -24,10 +24,18 @@ class UnleashIntegration(Integration):
             # type: (UnleashClient, str, *Any, **Any) -> Any
             enabled = old_is_enabled(self, feature, *args, **kwargs)
 
+            # Errors support.
+            #
             # We have no way of knowing what type of unleash feature this is, so we have to treat
             # it as a boolean / toggle feature.
             flags = sentry_sdk.get_current_scope().flags
             flags.set(feature, enabled)
+
+            # Spans support.
+            span = sentry_sdk.get_current_span()
+            print(span)
+            if span:
+                span.set_data(f"flag.{feature}", enabled)
 
             return enabled
 
--- a/sentry_sdk/integrations/unleash.py
+++ b/sentry_sdk/integrations/unleash.py
@@ -34,7 +34,7 @@ class UnleashIntegration(Integration):
             # Spans support.
             span = sentry_sdk.get_current_span()
             if span:
-                span.set_data(f"flag.{feature}", enabled)
+                span.set_data(f"flag.evaluation.{feature}", enabled)
 
             return enabled
 
--- a/tests/integrations/launchdarkly/test_launchdarkly.py
+++ b/tests/integrations/launchdarkly/test_launchdarkly.py
@@ -241,5 +241,5 @@ def test_launchdarkly_span_integration(
 
     (event,) = events
     assert event["spans"][0]["data"] == ApproxDict(
-        {"flag.hello": True, "flag.other": False}
+        {"flag.evaluation.hello": True, "flag.evaluation.other": False}
     )
--- a/tests/integrations/launchdarkly/test_launchdarkly.py
+++ b/tests/integrations/launchdarkly/test_launchdarkly.py
@@ -215,7 +215,6 @@ def test_launchdarkly_span_integration(
 ):
     td = TestData.data_source()
     td.update(td.flag("hello").variation_for_all(True))
-    td.update(td.flag("world").variation_for_all(True))
     # Disable background requests as we aren't using a server.
     config = Config(
         "sdk-key", update_processor_class=td, diagnostic_opt_out=True, send_events=False
--- a/tests/integrations/launchdarkly/test_launchdarkly.py
+++ b/tests/integrations/launchdarkly/test_launchdarkly.py
@@ -12,6 +12,8 @@ from ldclient.integrations.test_data import TestData
 import sentry_sdk
 from sentry_sdk.integrations import DidNotEnable
 from sentry_sdk.integrations.launchdarkly import LaunchDarklyIntegration
+from sentry_sdk import start_span, start_transaction
+from tests.conftest import ApproxDict
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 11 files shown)


indragiek:
@cmanallen We don't cap the # of span attributes today but we will likely cap a) the total size of the event b) the number of attributes or c) both - we haven't made a decision on this yet.

sentry_sdk/integrations/unleash.py
```diff
@@ -24,11 +24,19 @@ def sentry_is_enabled(self, feature, *args, **kwargs):
...
+            # Errors support.
+            #
             # We have no way of knowing what type of unleash feature this is, so we have to treat
             # it as a boolean / toggle feature.
             flags = sentry_sdk.get_current_scope().flags
             flags.set(feature, enabled)
 
+            # Spans support.
+            span = sentry_sdk.get_current_span()
+            print(span)
```
szokeasaurusrex, line 36: I'm guessing you added this for debugging

```suggestion
```

tests/integrations/launchdarkly/test_launchdarkly.py
```diff
@@ -202,3 +204,43 @@ def test_launchdarkly_integration_did_not_enable(monkeypatch):
...
+):
+    td = TestData.data_source()
+    td.update(td.flag("hello").variation_for_all(True))
+    td.update(td.flag("world").variation_for_all(True))
+    # Disable background requests as we aren't using a server.
+    config = Config(
+        "sdk-key", update_processor_class=td, diagnostic_opt_out=True, send_events=False
+    )
+
+    uninstall_integration(LaunchDarklyIntegration.identifier)
```
szokeasaurusrex, line 224: [question] why are we uninstalling the integration here? My guess is this would reduce the possibility of interactions between tests, but I this is my first time seeing this pattern in our tests so just want to check

tests/integrations/launchdarkly/test_launchdarkly.py
```diff
@@ -202,3 +204,43 @@ def test_launchdarkly_integration_did_not_enable(monkeypatch):
...
+    if use_global_client:
+        ldclient.set_config(config)
+        sentry_init(traces_sample_rate=1, integrations=[LaunchDarklyIntegration()])
```
szokeasaurusrex, line 227: `traces_sample_rate` is typed as `Optional[float]`

https://github.com/getsentry/sentry-python/blob/ef0dc37eb666bbd7e2d68f67d1ca62d10c640b4f/sentry_sdk/consts.py#L528

So please pass `1.0` instead of just `1`

```suggestion
        sentry_init(traces_sample_rate=1.0, integrations=[LaunchDarklyIntegration()])
```

tests/integrations/launchdarkly/test_launchdarkly.py
```diff
@@ -202,3 +204,43 @@ def test_launchdarkly_integration_did_not_enable(monkeypatch):
...
+        client = ldclient.get()
+    else:
+        client = LDClient(config=config)
+        sentry_init(
+            traces_sample_rate=1,
```
szokeasaurusrex, line 232:
```suggestion
            traces_sample_rate=1.0,
```

tests/integrations/launchdarkly/test_launchdarkly.py
```diff
@@ -202,3 +204,43 @@ def test_launchdarkly_integration_did_not_enable(monkeypatch):
...
+    events = capture_events()
+
+    with start_transaction(name="hi"):
+        with start_span(op="foo", name="bar"):
+            client.variation("hello", Context.create("my-org", "organization"), False)
+            client.variation("other", Context.create("my-org", "organization"), False)
+
+    (event,) = events
+    assert event["spans"][0]["data"] == ApproxDict(
+        {"flag.hello": True, "flag.other": False}
```
szokeasaurusrex, line 245: [question] maybe I am missing something or just don't understand how `LaunchDarkly` works, but I am struggling to see why these two flags have different values. It seems like the calls involving both are equivalent, so I am unsure then how they get to have different values. Where does the difference come from?

tests/integrations/statsig/test_statsig.py
```diff
@@ -181,3 +183,21 @@ def test_wrapper_attributes(sentry_init, uninstall_integration):
 
     # Clean up
     statsig.check_gate = original_check_gate
+
+
+def test_statsig_span_integration(sentry_init, capture_events, uninstall_integration):
+    uninstall_integration(StatsigIntegration.identifier)
+
+    with mock_statsig({"hello": True}):
+        sentry_init(traces_sample_rate=1, integrations=[StatsigIntegration()])
```
szokeasaurusrex, line 192:
```suggestion
        sentry_init(traces_sample_rate=1.0, integrations=[StatsigIntegration()])
```

tests/integrations/unleash/test_unleash.py
```diff
@@ -164,3 +166,21 @@ def test_wrapper_attributes(sentry_init, uninstall_integration):
         # Mock clients methods have not lost their qualified names after decoration.
         assert client.is_enabled.__name__ == "is_enabled"
         assert client.is_enabled.__qualname__ == original_is_enabled.__qualname__
+
+
+def test_unleash_span_integration(sentry_init, capture_events, uninstall_integration):
+    uninstall_integration(UnleashIntegration.identifier)
+
+    with mock_unleash_client():
+        sentry_init(traces_sample_rate=1, integrations=[UnleashIntegration()])
```
szokeasaurusrex, line 175:
```suggestion
        sentry_init(traces_sample_rate=1.0, integrations=[UnleashIntegration()])
```

tests/integrations/launchdarkly/test_launchdarkly.py
szokeasaurusrex: Ah, never mind. Now I see that above, you set the flags `"hello"` and `"world"` both to `True`. You did not set `"other"` at all, so I guess that is why it ends up being `False`. Is that correct?

Why are we setting the `"world"` flag at all, if we don't end up asserting on it? I think a code comment to describe why we set `"world"` but then assert on `"other"` might help me and anyone else who looks at this code in the future to understand what is going on here.

tests/integrations/unleash/test_unleash.py
```diff
@@ -164,3 +166,21 @@ def test_wrapper_attributes(sentry_init, uninstall_integration):
...
+        client = UnleashClient()  # type: ignore[arg-type]
+        with start_transaction(name="hi"):
+            with start_span(op="foo", name="bar"):
+                client.is_enabled("hello")
+                client.is_enabled("other")
+
+    (event,) = events
+    assert event["spans"][0]["data"] == ApproxDict(
+        {"flag.hello": True, "flag.other": False}
+    )
```
szokeasaurusrex, line 186: Are you intentionally checking the `"other"` flag here?

From what I can tell, the `mock_unleash_client` is defining a `"hello"` and `"world"` flag:

https://github.com/getsentry/sentry-python/blob/ef0dc37eb666bbd7e2d68f67d1ca62d10c640b4f/tests/integrations/unleash/testutils.py#L36-L45

Most of the other tests (besides `LaunchDarkly`, which I also commented on) test against the `"hello"` and `"world"` flags, what is the reason for checking `"other"` here instead of `"world"`?

szokeasaurusrex (commented):
Generally lgtm, although I have some questions on some items which were unclear to me.

Also, prior to merging these changes, we should settle on semantic conventions for feature flag span data keys and add them to our [develop docs](https://develop.sentry.dev/sdk/telemetry/traces/span-data-conventions/) and/or the [semantic convention docs](https://getsentry.github.io/sentry-conventions/generated/attributes/). I discussed the span conventions with @AbhiPrasad, it sounds like he has some ideas about how we should name span data keys, so you should discuss with him.

tests/integrations/launchdarkly/test_launchdarkly.py
szokeasaurusrex: Interesting, perhaps I just haven't touched this part of the codebase – if that's the pattern you can disregard my comment

tests/integrations/launchdarkly/test_launchdarkly.py
cmanallen: Removed the world variation.

tests/integrations/unleash/test_unleash.py
cmanallen: https://github.com/getsentry/sentry-python/blob/b391602d2a98d900070f0497073828b1d48084ec/tests/integrations/unleash/testutils.py#L45

Testing for missing keys.  Default value is `False`.

cmanallen:
Now capping the number of flags to `10`.  No particular reason for that default.  We can make it larger or smaller.

cmanallen:
@szokeasaurusrex https://github.com/getsentry/sentry-conventions/pull/42

AbhiPrasad (approved):
Exciting!

PR #4292, opened 2025-04-14
feat(logging): Add formatted message to log events

szokeasaurusrex:
Send the formatted log event to Sentry in the [`formatted` field](https://develop.sentry.dev/sdk/data-model/event-payloads/message/).

This builds on #4291, providing a more robust fix for #3660.


sentrivana:
LGTM. If you haven't yet, please double check that the example from https://github.com/getsentry/sentry-python/issues/3660 works with this fix and also that nothing changes in Sentry wrt to how log messages are shown (except for the buggy case from https://github.com/getsentry/sentry-python/issues/3660).

sentrivana (approved):
^ see comment above

szokeasaurusrex:
@sentrivana good idea, I will check the example from #3660 before merging.

I believe this change might affect how issues are displayed in Sentry. However, this should not be a breaking change, since [we group issues based on the `"message"` field](https://getsentry.github.io/relay/relay_event_schema/protocol/struct.LogEntry.html#structfield.message), so even if the issue is displayed differently, the grouping should be unaffected.

szokeasaurusrex:
I tested a log with a `None` value, similar to #3660. The log [shows as expected in Sentry](https://sentry-sdks.sentry.io/issues/6542999912/events/84fb31da59f34dd085fa0124f69b5784/).

image

PR #4317, opened 2025-04-17
chore(ourlogs): Use new transport

colin-sentry:
We've added a more efficient transport for logs handling, use that.

Solves LOGS-60


sentrivana (approved):
LGTM!

PR #4319, opened 2025-04-22
tests: Move aiohttp under toxgen

sentrivana:
Depends on https://github.com/getsentry/sentry-python/pull/4323


antonpirker (approved):
as the tests are passing, I think this is good.

PR #4362, opened 2025-05-05
Revert `traces_sample_rate` default change

antonpirker:
The default of `traces_sample_rate` should be `None`. This was a misunderstanding. With this default tracing without performance (twp) works, and this is the behavior that we like to have. 

Original change is in this PR: https://github.com/getsentry/sentry-python/pull/4335

See also this for better understanding: https://develop.sentry.dev/sdk/telemetry/traces/trace-propagation-cheat-sheet/


szokeasaurusrex (commented):
@antonpirker can you link the PR that this PR is reverting for context?

antonpirker:
that's a duplicate of https://github.com/getsentry/sentry-python/pull/4335

PR #4366, opened 2025-05-06
feat: Allow configuring `keep_alive` via environment variable

### LINKED ISSUE #4354: Allow setting keep_alive via environment variable.
### Problem Statement

Currently it is not possible to configure keep_alive using an environment variable. This can be an issue when users that use AWS layer that can only be configured using environment variable encounter connection issues like this https://github.com/getsentry/sentry-python/issues/4311 where this option can help. 

### Solution Brainstorm

Add support for an environment viable, such as SENTRY_KEEP_ALIVE, to set keep alive to true.

szokeasaurusrex:
This commit enables the `keep_alive` option to be set via the `SENTRY_KEEP_ALIVE` environment variable. When both the environment variable and the argument are provided, the argument takes precedence.

Closes #4354

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -142,6 +142,11 @@ def _get_options(*args, **kwargs):
         )
         rv["socket_options"] = None
 
+    if rv["keep_alive"] is None:
+        rv["keep_alive"] = (
+            env_to_bool(os.environ.get("SENTRY_KEEP_ALIVE"), strict=True) or False
+        )
+
     if rv["enable_tracing"] is not None:
         warnings.warn(
             "The `enable_tracing` parameter is deprecated. Please use `traces_sample_rate` instead.",
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1,3 +1,4 @@
+import contextlib
 import os
 import json
 import subprocess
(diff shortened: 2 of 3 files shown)


tests/test_client.py
```diff
@@ -1496,3 +1497,66 @@ def test_dropped_transaction(sentry_init, capture_record_lost_event_calls, test_
...
+    ("env_value", "arg_value", "expected_value"),
+    [
+        (None, None, False),  # default
+        ("0", None, False),  # env var false
+        ("1", None, True),  # env var true
+        (None, False, False),  # arg false
+        (None, True, True),  # arg true
+        # Argument overrides environment variable
+        ("0", True, True),  # env false, arg true
+        ("1", False, False),  # env true, arg false
```
szokeasaurusrex, line 1545: Do you all think this order of precedence is reasonable? Or, should we not guarantee any specific order of precedence (i.e. if both are set to different values, behavior is undefined)?

antonpirker (approved):
Nice!

tests/test_client.py
szokeasaurusrex: I suppose the middle option would be to remove the tests asserting order and not provide any specific guarantee on precedence for now.

That way we could change the precedence later even in a minor release

sentry_sdk/client.py
```diff
@@ -142,6 +142,11 @@ def _get_options(*args, **kwargs):
         )
         rv["socket_options"] = None
 
+    if rv["keep_alive"] is None:
+        rv["keep_alive"] = (
+            env_to_bool(os.environ.get("SENTRY_KEEP_ALIVE"), strict=True) or False
```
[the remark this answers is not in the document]
szokeasaurusrex: I'm not sure we should make this change because `env_to_bool(os.environ.get("SENTRY_KEEP_ALIVE"), strict=True) or False` and `env_to_bool(os.environ.get("SENTRY_KEEP_ALIVE", "False"), strict=True)` are not equivalent.

For example, if `SENTRY_KEEP_ALIVE` is set to an invalid value like `"invalid"`, then the current code with `or False` evaluates to `False`, whereas your proposal would evaluate to `None`. Since `keep_alive` should have a boolean value, I think the current option is preferable.

If anything, I would consider changing the `SENTRY_DEBUG` handling to match what I have implemented here (although that would need a separate PR)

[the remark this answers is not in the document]
antonpirker: The important part is that all env vars are handled in the same way.
Its fine if you change the others, as long as all are behaving the same.

tests/test_client.py
antonpirker: But we have an order right now: Options set when calling `init()` are overruling env vars.

And I think we can not change this in a minor release, because it can alter behavior for users that have a init option and an env var set.

the tests are good, because if we change the precedence by accident, the tests will fail.

antonpirker (approved):
see my comments, otherwise looks good!

tests/test_client.py
szokeasaurusrex: sounds good, then we stay with what we currently have

sentry_sdk/client.py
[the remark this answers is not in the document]
szokeasaurusrex: Okay, I made a PR to update the handling for `SENTRY_DEBUG` to match (#4400)

PR #4388, opened 2025-05-14
tests(logs): avoid failures when running with integrations enabled

rominf:
When (at least) one of integrations is enabled (because some dependencies
are installed in the environment), `sentry.sdk.name` is changed from
`sentry.python` to `sentry.python.[FIRST_ENABLED_INTEGRATION]` which
makes `test_logs_attributes` fail. Prevent failure by relaxing the check.

This change is beneficial not only for packaging (this patch was required
for packaging for Fedora), but also for running tests with `pytest`
directly.

See also: https://github.com/getsentry/sentry-python/pull/4316

---


antonpirker (approved):
Looks good! Thanks a lot!

PR #4390, opened 2025-05-14
Allow multiple craft configs in the repo

antonpirker:
Take craft config from the merge target (for having a alpha release package)

See: https://develop.sentry.dev/sdk/processes/releases/#optional-using-multiple-craft-configs-in-a-repo


antonpirker:
not needed anymore.

PR #4396, opened 2025-05-16
fix(logs): send `severity_text`: `warn` instead of `warning`

lcian:
We need to send `warn` as the `severity_text` for warning level logs, not `warning`.
If we send `warning` it will not be recognized by the backend and the severity will show up as `UNKNOWN` in the frontend.
See screenshot for comparison of after and before.

Screenshot 2025-05-16 at 18 24 50


szokeasaurusrex (approved):
Nice!

lcian:
It seems that previously I updated something that effected the `logging` integration (not when sending logs, but when sending events/breadcrumbs).
That code actually doesn't need changes because we correctly use `_python_level_to_otel` to set the `severity_text` when sending logs.

PR #4400, opened 2025-05-19
fix: Handle invalid `SENTRY_DEBUG` values properly

szokeasaurusrex:
Previously, if the environment variable `SENTRY_DEBUG` would have an invalid value, such as "invalid," we would set `rv["debug"]` to `None`. However, since this value should be a boolean, it should instead be set to `False`.

This change ensures that `rv["debug"]` always is a boolean. Where we previously would have set the value to `None`, we now set it to `False`. Other behavior remains unchanged.

Ref https://github.com/getsentry/sentry-python/pull/4366/files#r2075294825

Depends on:
  - #4366


antonpirker (approved):
thanks for fixing!

PR #4423, opened 2025-05-28
feat(logs): Add user attributes to logs

### LINKED ISSUE #4422: Add user attributes to Python logs
ref [https://develop.sentry.dev/sdk/telemetry/logs/#user-attributes](https://develop.sentry.dev/sdk/telemetry/logs/#user-attributes)

If `sendDefaultPii` is set to `true`, attach user information from the scope as is detailed in the spec to all logs.

szokeasaurusrex:
Example log with user attributes: https://sentry-sdks.sentry.io/explore/logs/trace/362b3309f7f9471eb504af407bf51932/?logsSortBys=-timestamp&project=4505239425777664&source=logs&tab=logs&timestamp=1748448946

Closes #4422

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -927,6 +927,21 @@ class _Client(BaseClient):
             elif propagation_context is not None:
                 log["trace_id"] = propagation_context.trace_id
 
+        # The user, if present, is always set on the isolation scope.
+        if self.should_send_default_pii() and isolation_scope._user is not None:
+            for log_attribute, user_attribute in (
+                ("user.id", "id"),
+                ("user.name", "username"),
+                ("user.email", "email"),
+            ):
+                if (
+                    user_attribute in isolation_scope._user
+                    and log_attribute not in log["attributes"]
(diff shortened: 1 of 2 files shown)


sentry_sdk/client.py
```diff
@@ -927,6 +927,20 @@ def _capture_experimental_log(self, current_scope, log):
             elif propagation_context is not None:
                 log["trace_id"] = propagation_context.trace_id
 
+        if self.should_send_default_pii() and isolation_scope._user is not None:
```
AbhiPrasad, line 930: shouldn't we also check the global and current scope?

szokeasaurusrex: It should not be necessary. User should always only be set on the isolation scope:

https://github.com/getsentry/sentry-python/blob/747f0f3bb08f4f21e41c6202a5c72e8b0236b761/sentry_sdk/api.py#L332-L334

I checked the (few) places we manually call `set_user` on a scope, these also use the isolation scope only.

PR #4428, opened 2025-05-29
fix: fix ARQ integration error (#4427)

### LINKED ISSUE #4427: ARQ Integration: Cron Jobs (cron_jobs) Handling Error
### How do you use Sentry?

Self-hosted/on-premise

### Version

2.29.1

### Steps to Reproduce

ARQ's worker (WorkerSettings) can have None as cron_jobs (ref. https://github.com/python-arq/arq/blob/main/arq/typing.py#L41)

But this section doesn't take it into consideration: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/arq.py#L215-L218.

Thus the following code raises an error:

```py
import sentry_sdk
from arq import func
from arq.connections import RedisSettings
from sentry_sdk.integrations.arq import ArqIntegration

REDIS_SETTINGS = RedisSettings()


sentry_sdk.init(
    "...",
    integrations=[ArqIntegration()],
)


async def task(_):
    return 1 + 1


class WorkerSettings:
    functions = [func(task)]  # type: ignore
    cron_jobs = None
```

```
...
  File "/path/to/.venv/lib/python3.11/site-packages/sentry_sdk/integrations/arq.py", line 218, in _sentry_create_worker
    settings_cls.cron_jobs = [
                             ^
Sentry is attempting to send 2 pending events
Waiting up to 2 seconds
Press Ctrl-C to quit
```

### Expected Result

ARQ integartion can handle cron_jobs = None without the above error.

### Actual Result

Please check "Steps to reproduce" section.

ninoseki:
Fix #4427 by setting an empty list if `settings_cls.cron_jobs` is None.

### DIFF
--- a/sentry_sdk/_types.py
+++ b/sentry_sdk/_types.py
@@ -170,7 +170,7 @@ if TYPE_CHECKING:
             "contexts": dict[str, dict[str, object]],
             "dist": str,
             "duration": Optional[float],
-            "environment": str,
+            "environment": Optional[str],
             "errors": list[dict[str, Any]],  # TODO: We can expand on this type
             "event_id": str,
             "exception": dict[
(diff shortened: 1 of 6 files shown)


sentrivana (approved):
Thanks for reporting and fixing @ninoseki!

PR #4436, opened 2025-06-03
Fix Spark in CI

sentrivana:


sentrivana:
Closing, will do this in https://github.com/getsentry/sentry-python/pull/4435

Issue #4327, opened 2025-04-25
Memory leak in AWS Fargate


InterstellarStella:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.14.0

### Steps to Reproduce

This issue is created on behalf of the customer.

We are using Sentry Python SDK in our FastAPI application, and we have discovered that it might be causing some memory leaks in our web server where a lot of memory is retained in primary memory and not being released. We also saw an open issue on GitHub where Logger.exception causes memory leaks.

So we are kind of hoping to understand if we can resolve this and stick to Sentry because this is causing issues on our server.

Further information is provided in [the shadow ticket](https://getsentry.atlassian.net/browse/FEEDBACK-2407?atlOrigin=eyJpIjoiNDNjNmMwZjI1MjBmNDA5NmI3YzNmOTkwMTdkYmI0ODEiLCJwIjoiaiJ9).

### Expected Result

No memory leaks.

### Actual Result

Using the Sentry SDK causes memory leaks. Graphs are provided in [the shadow ticket](https://getsentry.atlassian.net/browse/FEEDBACK-2407?atlOrigin=eyJpIjoiNDNjNmMwZjI1MjBmNDA5NmI3YzNmOTkwMTdkYmI0ODEiLCJwIjoiaiJ9).

InterstellarStella:
https://sentry.zendesk.com/agent/tickets/150418

szokeasaurusrex:
Hello! Thank you for reaching out. 

Could you please try the following and let us know whether it fixes the problem?

1. First, please try upgrading to the latest version of the Sentry SDK, in case we might have already fixed the underlying issue.
2. Second, if upgrading does not fix the problem, please try disabling the `DedupeIntegration`. This was [suggested as a workaround](https://github.com/getsentry/sentry-python/issues/3165#issuecomment-2173231686) in the `logging.exception` memory leak issue (#3165). If disabling `DedupeIntegration` fixes the memory leak, then we can close this issue as a duplicate of #3165 

If neither of those suggestions help, please let me know, and we can continue debugging from there.

tushar-wispr:
@szokeasaurusrex Thanks for this, will let you know if this fixes it.

tushar-wispr:
Hey Guys
It seems like disabling the integration solves the memory leak issue but we do have a lot more events on sentry.

szokeasaurusrex:
[...]

PR #4446, opened 2025-06-05
Store a hash of the exc instead of full exc in DedupeIntegration

antonpirker:
Users reported that the `DedupeIntegration` can use up a lot of memory, because it keeps a full exception in memory for checking if it has seen this exception already.

Depending on the code those exception object can be big because they also include the traceback and local variables (which can be huge).

Idea is now to not save the whole exception but just a hash of the important parts of the exception to decide if we have seen this exception again. 

fixes https://github.com/getsentry/sentry-python/issues/3165 
fixes https://github.com/getsentry/sentry-python/issues/4327

### DIFF
--- a/sentry_sdk/integrations/dedupe.py
+++ b/sentry_sdk/integrations/dedupe.py
@@ -19,6 +19,33 @@ class DedupeIntegration(Integration):
         self._last_seen = ContextVar("last-seen")
 
     @staticmethod
+    def _get_exception_hash(exc):
+        # type: (Exception) -> int
+        """
+        Create a memory-efficient hash for an exception.
+
+        Instead of storing the entire exception object, we store just enough
+        information to identify it uniquely. This avoids keeping the traceback
+        and local variables in memory.
+        """
+        # Get the exception type name and message
+        exc_type = type(exc).__name__
+        exc_message = str(exc)
+
+        # Get the first frame of the traceback if it exists
+        if hasattr(exc, "__traceback__") and exc.__traceback__:
+            frame = exc.__traceback__.tb_frame
+            filename = frame.f_code.co_filename
+            lineno = frame.f_lineno
+            func_name = frame.f_code.co_name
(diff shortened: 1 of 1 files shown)


sl0thentr0py:
@antonpirker can't you just `id(exc)`?

antonpirker:
i thought about that too. but garbage collection moves objects around. so id(exc) could be different when we safe it and when we compare it...

sentry_sdk/integrations/dedupe.py
```diff
@@ -18,6 +18,35 @@ def __init__(self):
...
+            while tb:
+                frame = tb.tb_frame
+                filename = frame.f_code.co_filename
+                lineno = tb.tb_lineno
+                func_name = frame.f_code.co_name
+                stacktrace.append((filename, lineno, func_name))
+                tb = tb.tb_next
+
+        # Create a tuple of the essential information and hash it
+        return hash((exc_type, exc_message, tuple(stacktrace)))
```
sl0thentr0py: if we're doing this, we don't need to walk the exception again, we already have everything in the event, so we can simply do
```suggestion
        return hash(iter_event_stacktraces(event))
```
this way, we don't even need the exception reference to be kept alive.

sl0thentr0py (changes requested):

sentry_sdk/integrations/dedupe.py
sl0thentr0py: JS is even considering just the first exception in the list
https://github.com/getsentry/sentry-javascript/blob/61940fc22052e9005043f77c7355db1742d1d78c/packages/core/src/integrations/dedupe.ts#L178-L180

antonpirker:
The `id()` of an object is [guaranteed to never change](https://docs.python.org/3/library/functions.html#id) during the life cycle of the object. So taking it.

antonpirker:
I will close this in favor of https://github.com/getsentry/sentry-python/pull/4809

Reason: Python reuses memory addresses after garbage collection. Making this path unreliable.

PR #4449, opened 2025-06-05
feat(scope): Remove `client` parameter

szokeasaurusrex:
We don't use the `client` parameter in the `Scope`'s constructor, perhaps we can remove it and simplify the API. It is still possible to set the client with `Scope.set_client`.

BREAKING CHANGE: `sentry_sdk.Scope` no longer has a `client` parameter.

---


szokeasaurusrex (commented):
This is just a small thing I noticed, feel free to close the PR if y'all think it is a bad idea

sl0thentr0py (approved):

PR #4465, opened 2025-06-15
🎨 Save a few CPU cycles by not generating error_type_name and error_full_name if not needed

coderanger:
No functionality difference but saves a nanosecond or two.

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -685,13 +685,13 @@ class _Client(BaseClient):
             return False
 
         error = exc_info[0]
-        error_type_name = get_type_name(exc_info[0])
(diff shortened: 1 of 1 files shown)


sentrivana (changes requested):
Hey @coderanger, thanks for the PR. As is, this doesn't look like an improvement to me, or at least only in specific scenarios. With your change, we don't always stringify the exception type and name anymore, but if someone's `ignore_errors` contains multiple strings and we enter the `for` loop, we'll end up doing the stringification multiple times, on the same original exception, which kind of defeats the point of making this more efficient.

coderanger:
Fair :)

Issue #4443, opened 2025-06-04
Langchain Integration adds redundant callbacks


mshavliuk:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.27.0

### Steps to Reproduce

When I configure the `SentryLangchainCallback` instance for `invoke` call I expect Sentry langchain integration not to create another instance of the same callback.

To reproduce the issue run code with breakpoints inside [`_wrap_configure`](https://github.com/getsentry/sentry-python/blob/c2d5a768a780375e75a6f24beb1358f92eb5766f/sentry_sdk/integrations/langchain.py#L415C5-L415C20):

```python
import os
import sentry_sdk
from langchain_aws import ChatBedrockConverse
from langchain_core.runnables import RunnableConfig
from sentry_sdk.integrations.langchain import LangchainIntegration, SentryLangchainCallback

sentry_sdk.init(
    dsn=os.environ["SENTRY_DSN"],
    environment="localhost",
    traces_sample_rate=1.0,
    integrations=[LangchainIntegration()],
)

llm = ChatBedrockConverse(
    region_name="eu-west-1",
    aws_access_key_id=os.environ["AWS_ACCESS_KEY_ID"],
    aws_secret_access_key=os.environ["AWS_SECRET_ACCESS_KEY"],
    model="eu.amazon.nova-lite-v1:0",
    # intentionally invalid config to trigger error
    max_tokens=1,
    additional_model_request_fields={"inferenceConfig": {"topK": 0}},
)

config = RunnableConfig(callbacks=[SentryLangchainCallback(max_span_map_size=100, include_prompts=True)])
llm.invoke('hello', config)
```

Installed dependencies:
```
annotated-types==0.7.0
anyio==4.9.0
boto3==1.38.22
botocore==1.38.22
certifi==2025.4.26
charset-normalizer==3.4.2
h11==0.16.0
httpcore==1.0.9
httpx==0.28.1
idna==3.10
jmespath==1.0.1
jsonpatch==1.33
jsonpointer==3.0.0
langchain-aws==0.2.18
langchain-core==0.3.55
langsmith==0.3.42
numpy==2.2.6
orjson==3.10.18
packaging==24.2
pydantic==2.11.5
pydantic_core==2.33.2
python-dateutil==2.9.0.post0
PyYAML==6.0.2
requests==2.32.3
requests-toolbelt==1.0.0
s3transfer==0.13.0
sentry-sdk==2.27.0
six==1.17.0
[...]

PR #4476, opened 2025-06-16
fix(langchain): Make `span_map` an instance variable

szokeasaurusrex:
`span_map` should be an instance variable; otherwise, separate instances of the `SentryLangchainCallback` share the same `span_map` object, which is clearly not intended here.

Also, remove the `max_span_map_size` class variable, it is always set on the instance, and so not needed.

Ref #4443

### DIFF
--- a/sentry_sdk/integrations/langchain.py
+++ b/sentry_sdk/integrations/langchain.py
@@ -89,12 +89,9 @@ class WatchedSpan:
 class SentryLangchainCallback(BaseCallbackHandler):  # type: ignore[misc]
     """Base callback handler that can be used to handle callbacks from langchain."""
 
-    span_map = OrderedDict()  # type: OrderedDict[UUID, WatchedSpan]
-
-    max_span_map_size = 0
-
     def __init__(self, max_span_map_size, include_prompts, tiktoken_encoding_name=None):
         # type: (int, bool, Optional[str]) -> None
+        self.span_map = OrderedDict()  # type: OrderedDict[UUID, WatchedSpan]
         self.max_span_map_size = max_span_map_size
         self.include_prompts = include_prompts
 
--- a/tests/integrations/langchain/test_langchain.py
+++ b/tests/integrations/langchain/test_langchain.py
@@ -416,3 +416,15 @@ def test_manual_callback_no_duplication(sentry_init):
 
     # Verify the callback ID matches our manual callback
     assert id(manual_callback) in tracked_callback_instances
+
+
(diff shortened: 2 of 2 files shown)


sentrivana (commented):
Q: Was this an actual problem or is this more of a nice-to-have small refactor? As I understand from the other PR, we're actively avoiding having more than one `SentryLangchainCallback` around at a time.

I think the change itself is fine but we don't need tests for an implementation detail like this. We also don't test that `max_span_map_size`, `include_prompts`, etc. are not shared between instances. (And we also shouldn't test that, because it's not behavior worth testing, and the tests would just add clutter without any signal.)

szokeasaurusrex:
> Q: Was this an actual problem or is this more of a nice-to-have small refactor? As I understand from the other PR, we're actively avoiding having more than one SentryLangchainCallback around at a time.

@sentrivana Yes, so the problem occurred when attempting to reproduce #4443. I was confused why, despite having two callbacks, the data in Sentry was not duplicated. I assumed that perhaps the `DedupeIntegration` was dropping stuff, but when I enabled `debug=True` in the SDK init, I got the following error, instead of a notice that the `DedupeIntegration` dropped the extra span:

```
 [sentry] ERROR: Internal error in sentry_sdk
  File "/Users/dszoke/Development/SDK/sentry-python/sentry_sdk/integrations/langchain.py", line 267, in on_llm_end
    span_data = self.span_map[run_id]
                ~~~~~~~~~~~~~^^^^^^^^
KeyError: UUID('2a2c72c8-66f1-4607-b4d2-bd6e93125c39')
```

The reason, of course, is that the first callback which ends deletes the span from the `span_map`, so that when the duplicate callback is called, that span is gone.

Of course, you are right that we try to only have one callback per LLM request in #4485. But, what happens if we have concurrent requests to LLMs, or some other more complex setup, where having two callbacks would be expected? I am unsure how realistic such setups are, but if they do occur, I am pretty confident that we don't want the span map to be shared between the instances.

sentrivana:
Not against the change itself, seems reasonable in any case.

I was more pondering if this is behavior that needs testing (rather than a small internal refactor), but if you believe it is, I'd say `test_span_map_is_instance_variable` is sufficient. I'd remove `test_span_map_not_class_attribute` as it doesn't add any value -- we should test behavior, not implementation details, and the behavior in this case is covered by the first test already.

szokeasaurusrex:
sounds reasonable, the first test is definitely more important anyhow

PR #4479, opened 2025-06-16
ref(langchain): Greatly simplify `_wrap_configure`

szokeasaurusrex:
Resolving #4443 requires some changes to this method, but the current `args`/`kwargs` business makes the method difficult to reason through.

This PR simplifies the logic by listing out the parameters we need to access, so we don't need to access them through `args` and `kwargs`.

We also cut down on the amount of branching and the amount of variables (`new_callbacks` vs `existing_callbacks`).

Behavior does not change in this PR; we will introduce behavior changes in a subsequent PR

---


sentrivana (approved):
Soo much better! 🥔

PR #4529, opened 2025-06-26
Migrate typing for integrations - part 1

sl0thentr0py:


sentry_sdk/integrations/__init__.py
```diff
@@ -276,13 +281,12 @@ class Integration(ABC):
     install = None
     """Legacy method, do not implement."""
 
-    identifier = None  # type: str
+    identifier: str
```
antonpirker, line 284: this used to be `None`, so should this not be `Optinal[str]`?

sentry_sdk/integrations/_wsgi_common.py
```diff
@@ -228,8 +218,7 @@ def _filter_headers(headers):
     }
 
 
-def _request_headers_to_span_attributes(headers):
-    # type: (dict[str, str]) -> dict[str, str]
+def _request_headers_to_span_attributes(headers: dict[str, str]) -> dict[str, str]:
```
antonpirker, line 221: Sometimes it is `dict` sometimes `Dict`, we should make it consistent.

sentry_sdk/integrations/anthropic.py
```diff
@@ -68,8 +65,7 @@ def _calculate_token_usage(result, span):
     record_token_usage(span, input_tokens, output_tokens, total_tokens)
 
 
-def _get_responses(content):
-    # type: (list[Any]) -> list[dict[str, Any]]
+def _get_responses(content: list[Any]) -> list[dict[str, Any]]:
```
antonpirker, line 68: Should we use `typing.Dict` and `typing.List` everywhere until we drop Python 3.8 and we can use `dict` and `list` etc everywhere?

probablyskela: Why not use `from __future__ import annotations` (which is already used in the codebase) and use `dict` and `list` everywhere? The only downside of `from __future__ import annotations` as I understand is that it does not evaluate annotated types into the `__annotation__` attribute of the object, which doesn't seem to be used anywhere anyway.

sentry_sdk/integrations/__init__.py
sl0thentr0py: this is an ABC, and all subsequent inherits set the identifier and all code assumes a str

sentry_sdk/integrations/_wsgi_common.py
sl0thentr0py: different task

sentry_sdk/integrations/anthropic.py
sl0thentr0py: we will clean this separately

PR #4535, opened 2025-06-30
chore: Reexport module `profiler`

zen-xu:
The example provided by sentry causes pylance to report `"profiler" is not a known attribute of module "sentry_sdk"`

image

### DIFF
--- a/sentry_sdk/__init__.py
+++ b/sentry_sdk/__init__.py
@@ -1,4 +1,4 @@
-from sentry_sdk import integrations, logger, profiler
+from sentry_sdk import profiler
 from sentry_sdk.scope import Scope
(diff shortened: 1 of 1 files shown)


antonpirker:
Hey @zen-xu thanks for fixing this!
Why is the `from sentry_sdk import integrations, logger` in your PR? Is this on purpose?

sentry_sdk/__init__.py
```diff
@@ -1,17 +1,18 @@
+from sentry_sdk import integrations, logger, profiler
```
antonpirker:
```suggestion
from sentry_sdk import profiler
```

antonpirker:
@zen-xu i took the liberty to update your PR.

PR #4552, opened 2025-07-07
test(loguru): Remove hardcoded line number in test_just_log

srothh:
Use inspect to dynamically compute the correct line number instead of hard coding it.

Fixes GH-4454
https://github.com/getsentry/sentry-python/issues/4454

Changed the loguru test_just_log test to use inspection logic for the line number instead of hard coding it.

### DIFF
--- a/tests/integrations/loguru/test_loguru.py
+++ b/tests/integrations/loguru/test_loguru.py
@@ -1,4 +1,5 @@
 from unittest.mock import MagicMock, patch
+import inspect
 
 import pytest
 from loguru import logger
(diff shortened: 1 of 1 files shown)


tests/integrations/loguru/test_loguru.py
```diff
@@ -54,12 +55,17 @@ def test_just_log(
     )
     events = capture_events()
 
+    frame = inspect.currentframe()
+    assert frame is not None
+    log_line_number = frame.f_lineno + 1
```
szokeasaurusrex: Why is it plus 1 here?

tests/integrations/loguru/test_loguru.py
```diff
@@ -72,7 +73,8 @@ def test_just_log(
             (breadcrumb,) = breadcrumbs
             assert breadcrumb["level"] == expected_sentry_level
             assert breadcrumb["category"] == "tests.integrations.loguru.test_loguru"
-            assert breadcrumb["message"][23:] == formatted_message
+            # Check that the message matches the expected pattern with a line number
+            assert re.search(expected_pattern, breadcrumb["message"][23:])
```
szokeasaurusrex: I believe we want to use [`fullmatch`](https://docs.python.org/3/library/re.html#re.fullmatch) rather than `search`. With `search`, the assertion passes if the string contains something matching the regex, but here, we instead want to ensure that the entire string matches

szokeasaurusrex (approved):

PR #4497, opened 2025-06-18
fix(profiling): Ensure profiler thread exits when needed

Zylphrex:
The soft exit wasn't properly shutting down the thread if another profiler started up too quickly. This ensures it is reused if possible but is properly shutdown if needed.

Specifically, the shutdown allowed the profiler 1 cycle before actually shutting down. If another profiler is started during this cycle, it's possible the old profiler never shuts down. Resulting in multiple profilers running.

Fixes #4489

PR #4553, opened 2025-07-07
fix(profiling): Fix race condition while processing profiles

### LINKED ISSUE #4542: IndexError: pop from an empty deque (continuous profiling?)
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.30.0

### Steps to Reproduce

We enabled continuous profiling recently and have been seeing this error since then. We can't reliably reproduce, it seems to happen randomly.

### Expected Result

No weird IndexErrors

### Actual Result

Image

szokeasaurusrex:
If multiple threads happen to be processing profiles, we could encounter a race condition, that would lead to more profiles being processed than are available in the `new_profiles` deque. To avoid this, we can break upon encountering an `IndexError`.

This line appears to be the most likely cause of #4542 – in any case, handling an `IndexError` is not harmful

Fixes #4542

---

### DIFF
--- a/sentry_sdk/profiler/continuous_profiler.py
+++ b/sentry_sdk/profiler/continuous_profiler.py
@@ -370,7 +370,13 @@ class ContinuousScheduler:
                 # that are starting profiles are not blocked until it
                 # can acquire the lock.
                 for _ in range(new_profiles):
-                    self.active_profiles.add(self.new_profiles.popleft())
+                    try:
+                        profile = self.new_profiles.popleft()
+                    except IndexError:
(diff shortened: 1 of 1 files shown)


sl0thentr0py:
@Zylphrex already [fixed a similar issue here](https://github.com/getsentry/sentry-python/pull/4497) and was released in https://github.com/getsentry/sentry-python/releases/tag/2.31.0. 

The report in #4542 is on 2.30.0 so I believe this is not needed, I would ask them to test on latest version again.

szokeasaurusrex:
True, nice catch @sl0thentr0py!

PR #4562, opened 2025-07-08
fix(sessions): clean up threading on kill+flush

bukzor:
This follows the precedent set in transport.flush, and helps me with my project
to assert thread hygeine in sentry test suite (related: DI-1008), but also makes
the system much more deterministic for everyone.

This will cause shutdown to be quite slow until #4561 goes in. So I'd reject
this if you reject that.


bukzor:
This needs test coverage, apparently. Any hints? I'll do the work.

sentry_sdk/sessions.py
```diff
@@ -191,6 +192,26 @@ def flush(self):
...
+                return
+            if self._thread:  # typing
+                self._thread.join()
+                self._thread = None
+                self._thread_for_pid = None
+
+    def _thread_stopping(self):
+        # type: (...) -> bool
+        return (
+            not self._running
```
sl0thentr0py: there's no more `self._running` so will wait till you rebase on the other change

sentry_sdk/sessions.py
```diff
@@ -190,14 +191,34 @@ def flush(self):
         if len(envelope.items) > 0:
             self.capture_func(envelope)
 
+        # hygiene: deterministically clean up any stopping thread
+        if not self._should_join_thread():
```
sl0thentr0py, line 195: why are two of these required? isn't the one under the lock sufficient?

sentry_sdk/sessions.py
```diff
@@ -190,14 +191,34 @@ def flush(self):
...
+            return
+        with self._thread_lock:
+            if not self._should_join_thread():
+                return
+            if self._thread:  # typing
+                self._thread.join()
```
sl0thentr0py, line 201: sorry but I actually don't understand this change at all properly

the `flush` function is called by the thread in the loop, so joining that thread from within `flush` makes no sense, it should be done from the calling thread?

untitaker: yeah this might deadlock if flush is called from within the thread. the goal was to make it so that `client.close()` blocks until the thread is actually gone. client also calls flush directly, and in that call, blocking here makes sense.

proposal: let's block in `kill()` until the thread is gone, or make `flush` take a kwarg like `flush(join_thread=True)`

bukzor:
Closing — not planning to continue this work.

PR #4570, opened 2025-07-10
Simplify status mapping

sl0thentr0py:


tests/opentelemetry/test_utils.py
```diff
@@ -219,31 +219,6 @@ def test_span_data_for_db_query():
 @pytest.mark.parametrize(
     "kind, status, attributes, expected",
     [
-        (
-            SpanKind.CLIENT,
-            None,  # None means unknown error
```
sl0thentr0py, line 224: a `ReadableSpan` can never have a `None` status, this is just bad mocking.

Issue #3454, opened 2024-08-23
Add tracing to `DramatiqIntegration`


ramonsaraiva:
### How do you use Sentry?

Self-hosted/on-premise

### Version

2.13.0

### Steps to Reproduce

1. Follow the instructions on https://docs.sentry.io/platforms/python/integrations/dramatiq/ for using the `DramatiqIntegration`
2. Enable tracing and profiling

Additional context:
I know https://github.com/getsentry/sentry-python/pull/3397 was recently released and this is most likely happening because that integration is only capturing events from exceptions, but the documentation still states that it is just a matter of enabling tracing or profiling.

Would be amazing to have tasks instrumented automatically, but maybe a temporary warning in the dramatiq integration docs to make it explicit that it doesn't work without manual instrumentation.

### Expected Result

Dramatiq actors tracing and profiling appear in performance and profiling sections of Sentry without manual instrumentation

### Actual Result

Dramatiq actors tracing and profiling **do not** appear in performance and profiling sections of Sentry

szokeasaurusrex:
Hi @ramonsaraiva, it indeed appears that the `DramatiqIntegration` does not support tracing/profiling; the docs page is incorrect. 

I opened a [separate issue](https://github.com/getsentry/sentry-docs/issues/11143) to make sure we update our docs page.

Additionally, I renamed this issue to "Add tracing to `DramatiqIntegration`" and will place it on our backlog. We are busy with larger projects at the moment, so it might take us some time to get around to adding tracing to the Dramatiq integration. If you (or anyone else from the community) would like, please feel free to contribute a PR to add this functionality. We are happy to help with reviews.

szokeasaurusrex:
Hey @wsantos, the separate issue that you are referencing was the issue to update our docs to state that `Dramatic` does not support tracing, and I have already completed this.

Are you instead wondering about how to add tracing to Dramatiq, which would solve this issue? If so, we have some [SDK Development Docs](https://develop.sentry.dev/sdk/), but I don't think these cover how to write a tracing integration (perhaps this is something we can add). My recommendation is to take a look at the source code for other integrations which do have tracing and base your implementation on that. You can feel free to reach out to us if you get stuck anywhere

szokeasaurusrex:
> Correct, I'd like to know how to add it, I'm going to try it if I can find some spare time.

[...]

PR #4571, opened 2025-07-10
feat(integrations): Add tracing to DramatiqIntegration

Igreh:
Adds tracing support to DramatiqIntegration #3454


sentry_sdk/integrations/dramatiq.py
```diff
@@ -85,20 +89,27 @@ class SentryMiddleware(Middleware):  # type: ignore[misc]
     DramatiqIntegration.
     """
 
+    # type: contextvars.ContextVar[Transaction]
+    _transaction = contextvars.ContextVar("_transaction", default=None)
```
szokeasaurusrex: Why are we using a `ContextVar` here, instead of just a simple instance variable on the `SentryMiddleware`?

Igreh:
> Hi @Igreh, I took a quick glance at your PR – at a high level this seems reasonable; however, before I do a full review, I would appreciate if you could please add some tests for these changes, since we need tests before we can consider merging this.

Hi!
Thanks for reply. I'm not very familiar about what should be tested exactly for now  (transaction was successfully started? or what objects/props must be created? and etc ) but I will take a look and try to do my best =)
If you have any advice - would be very helpful

Edit:
Or I just need to cover one missed line?

sentry_sdk/integrations/dramatiq.py
Igreh: To isolate variable between threads:
1. Dramatiq can be started in multithreading mode:
```dramatiq --processes 1 --threads 5 ...```
2. And I was inspired by official middleware:
[CurrentMessageMiddleware](https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/middleware/current_message.py#L43-L57)

I will try recheck if it's actually needed

**Comment by sentrivana** [comment]
Hey @Igreh, it'd be good to have "end-to-end" (i.e., not unit tests) for the different tracing code paths. Stuff like:
- the happy path (if everything works fine, you get the transaction and it has the data you expect it to have)
- how it behaves if there's an error (a transaction still gets sent, and it has the correct status, etc.; maybe also check if the user's program still behaves as expected -- the exception bubbles up if it should bubble up, etc.)
- trace propagation (i.e., the transaction picks up propagation headers/vars if present) -- I think this might still be missing in your PR

...and/or whatever you consider important so that when the test suite is run and ends up green, you'd have a reasonable amount of confidence that tracing in the Dramatiq integration is working.

You can have a look at some of [these similar integrations](https://docs.sentry.io/platforms/python/integrations/#data-processing) and their test suites (e.g. [arq](https://github.com/getsentry/sentry-python/blob/master/tests/integrations/arq/test_arq.py), [huey](https://github.com/getsentry/sentry-python/blob/master/tests/integrations/huey/test_huey.py)) for inspiration.

Thanks for working on this! Let us know if you need any help.

**Comment by Igreh** [comment]
Hi @sentrivana !
Thanks for your reply. It was very helpful. 
I believe I was able to solve propagation issue (see last commit)
Also, I've changed the way dramatiq_task_id is stored - now as tag (like in celery). It's actually very convenient to find transaction by ID.
New tests were added (huey integration I used for inspiration).
One thing I can't understand for now: mechanism.type of error event is "logging" instead of "dramatiq".  https://github.com/getsentry/sentry-python/pull/4571/files#diff-ed170393871d90a02937ffecacd1b736805e6ea3c43e5549fd0df9bab883709eR82
Disabling LoggingIntegration fixes it (and actually it's the way we use sentry) but I suppose the general solution must work on defaults)
Probably, I have to dig deeper to dramatiq source code to understand what's happening... Not sure but I will keep investigating

File: sentry_sdk/integrations/dramatiq.py (hunk 1 above)
Comment (Igreh) [reply sentry_sdk/integrations/dramatiq.py]: Confirmed. 
One `SentryMiddleware` instance will be shared between threads. So, isolating is necessary.
Previous (actually current) realisation adds _scope_manager to message, but I think ContextVar is much cleaner solution

**Comment by Igreh** [comment]
Hi @sentrivana !
Things are more complicated than I expected... =)

I'm using (and testing) my current integration on real project and 
we're experiencing different side effects, but I'm curious about one thing for now:
Retried transactions must be connected by the same trace? Is any sentry agreement for such issue?
Retrying in terms of Dramatiq means requeuing message and it means baggage is propagated also. Bug or feature ? )

Failed task with 3 retries might look like this:
screen_2025-07-21_13-55-15

**Comment by sentrivana** [comment]
Hey again 👋🏻 Thanks for your time on this!

> One thing I can't understand for now: mechanism.type of error event is "logging" instead of "dramatiq". https://github.com/getsentry/sentry-python/pull/4571/files#diff-ed170393871d90a02937ffecacd1b736805e6ea3c43e5549fd0df9bab883709eR82
> Disabling LoggingIntegration fixes it (and actually it's the way we use sentry) but I suppose the general solution must work on defaults)

What you're seeing is likely dramatiq logging the error when it happens, and our `LoggingIntegration` picking it up. In case that happens, our deduplication logic (in `DedupeIntegration`) drops the second (in this case better) event coming directly from the integration.

What you can do is add an `ignore_logger("<logger-name>")` to the integration, see for example [here](https://github.com/getsentry/sentry-python/blob/24790ebb2728e03a2044b0a877220b4823cc6418/sentry_sdk/integrations/strawberry.py#L58) where we were solving the same issue for strawberry-graphql. This will tell the `LoggingIntegration` to skip capturing events from that specific logger, so the actual events coming from the specialized integration should get through.

(I know you're not working on the error part of the integration in this PR but it's still a good improvement so feel free to include it.)

> I'm using (and testing) my current integration on real project and
> we're experiencing different side effects, but I'm curious about one thing for now:
> Retried transactions must be connected by the same trace? Is any sentry agreement for such issue?
> Retrying in terms of Dramatiq means requeuing message and it means baggage is propagated also. Bug or feature ? )

Not sure off the top of my head how this works in our other task queue integrations, but my hunch is that retries should be under one trace.

**Comment by Igreh** [comment]
Good day everyone!
Sorry for long delay but now I'm back and going to complete this task)

Last patches are aimed to solve (I hope) all current issues:
- py36 support (ContextVar). aiocontextvars as new deps for tox.
- typing issues: incorrect import (type R was absent in earlier versions of dramatiq)
- fix test (ignore logger)

> Not sure off the top of my head how this works in our other task queue integrations, but my hunch is that retries should be under one trace.

I decided to not include retries under one trace. As mentioned in commit: "Start new trace in case of retrying task. Such decision makes life easier to investigate failed transactions in complicated traces.". 
As example: one transaction may include 2 or more execution of the same task. If one or more failed and dramatiq starts retrying them then trace will include huge amount of task calls. Understanding what should have happened and what actually happened becomes an extremely difficult task. The sequence of calls (at least visually) is disrupted. It's really easier to find all retries by dramatiq_message_id (which is now became tag)

Let me know if anything needs to be improved or reworked. Thanks in advance!

**Comment by sentrivana** [comment]
Hey @Igreh, thanks for coming back to the PR and addressing the feedback! CI looks great, I'll give the PR a look this week.

File: sentry_sdk/integrations/dramatiq.py [hunk 2]
```diff
@@ -85,50 +107,83 @@ class SentryMiddleware(Middleware):  # type: ignore[misc]
...
+        if "retries" in message.options:
+            # start new trace in case of retrying
+            sentry_headers = {}
+
+        transaction = continue_trace(
+            sentry_headers,
+            name=message.actor_name,
+            op=OP.QUEUE_TASK_DRAMATIQ,
+            source=TransactionSource.TASK,
+            # origin=DramatiqIntegration.origin,
```
sentrivana: The `origin` looks good, any reason it's commented out?

sentry_sdk/integrations/dramatiq.py
```diff
@@ -85,50 +107,83 @@ class SentryMiddleware(Middleware):  # type: ignore[misc]
...
         if integration is None:
             return
 
-        message._scope_manager = sentry_sdk.new_scope()
-        message._scope_manager.__enter__()
-
         scope = sentry_sdk.get_current_scope()
-        scope.set_transaction_name(message.actor_name)
-        scope.set_extra("dramatiq_message_id", message.message_id)
+        scope.set_tag("dramatiq_message_id", message.message_id)
```
sentrivana: The SDK shouldn't set tags on its own (we still do this in a couple places, but are avoiding it for new code)

sentry_sdk/integrations/dramatiq.py
```diff
@@ -85,50 +107,83 @@ class SentryMiddleware(Middleware):  # type: ignore[misc]
     DramatiqIntegration.
     """
 
+    _transaction = ContextVar("_transaction")
```
sentrivana: I think I already may have asked this in an earlier iteration, but do we need the ContextVar here? The scopes themselves are stored in ContextVars and ideally they should govern the reference to the current transaction (this should already be working as long as the transaction/span is correctly set on the scope, which it is as long as you use `start_span/transaction` context manager or just `__enter__` directly)

sentrivana (commented):
Hey @Igreh, thanks for bearing with me. Left a few comments. :)

sentry_sdk/integrations/dramatiq.py
Igreh: No reasons. Will fix it. Thanks

sentry_sdk/integrations/dramatiq.py
Igreh: I'm a little bit confused here. What do you mean exactly by "SDK shouldn't set tags"?
Am I do it wrong now or you wanna say setting tags is only allowed for client's code and no tags allowed in SDK?
I definitely want to set dramatiq_message_id for each task like it happens in celery integration:
image

message_id is extremely helpful in investigating task failures

sentry_sdk/integrations/dramatiq.py
Igreh: It makes sense. 
Looks like I really needn't store transaction separately and I can get it directly from scope:
```python
transaction = sentry_sdk.get_current_scope().transaction
```
Right? =)

Igreh:
> Hey @Igreh, thanks for bearing with me. Left a few comments. :)

Hi! Thanks for review - super helpful as usual)
Left replies with some more questions. 
Waiting for you replies to prepare final patch.

sentry_sdk/integrations/dramatiq.py
sentrivana: Exactly, that'd be the idea.

sentry_sdk/integrations/dramatiq.py
sentrivana: > setting tags is only allowed for client's code and no tags allowed in SDK?

This. Going forward, we're not setting tags in the SDK. You can still set `dramatiq_message_id` with e.g. `span.set_data()` and it should be searchable.

sentrivana:
@Igreh Left some feedback, but we're pretty much almost there :) Thanks for your continued work on this!

Looks like the config in `scripts/populate_tox/config.py` has a merge conflict now because I've added a new option to dramatiq there -- should be enough to just accept both changes from your branch and from master. In case CI complains about anything `populate_tox` related I'll take care of it.

Igreh:
> @Igreh Left some feedback, but we're pretty much almost there :) Thanks for your continued work on this!
> 
> Looks like the config in `scripts/populate_tox/config.py` has a merge conflict now because I've added a new option to dramatiq there -- should be enough to just accept both changes from your branch and from master. In case CI complains about anything `populate_tox` related I'll take care of it.

@sentrivana 
I've pushed corrections according to your replies. Getting rid of contextvar has significantly simplified the code and no tox related issues left) Thanks! 
The only thing I'm upset for is removal of "dramatiq_message_id" tag from integration. 
Yes, I can do it in my project code, but I will have to do it manually for each dramatiq task. So, I'll have to find a workaround. Some sort of middleware should help - another dramatiq middleware or just a func decorator.  Will figure it out later =)

sentry_sdk/integrations/dramatiq.py
```diff
@@ -85,50 +100,79 @@ class SentryMiddleware(Middleware):  # type: ignore[misc]
     DramatiqIntegration.
     """
 
+    def before_enqueue(self, broker, message, delay):
+        # type: (Broker, Message[R], int) -> None
+        message.options["sentry_headers"] = {
```
sentrivana: Nit: I don't think we're really in risk of collisions here, but let's maybe call this `_sentry_headers` just to denote it's something internal
```suggestion
        message.options["_sentry_headers"] = {
```

sentry_sdk/integrations/dramatiq.py
```diff
@@ -85,50 +100,79 @@ class SentryMiddleware(Middleware):  # type: ignore[misc]
...
-        message._scope_manager = sentry_sdk.new_scope()
-        message._scope_manager.__enter__()
-
         scope = sentry_sdk.get_current_scope()
-        scope.set_transaction_name(message.actor_name)
-        scope.set_extra("dramatiq_message_id", message.message_id)
+        scope.clear_breadcrumbs()
         scope.add_event_processor(_make_message_event_processor(message, integration))
 
+        sentry_headers = message.options.get("sentry_headers") or {}
```
sentrivana: If you go with my renaming suggestion from above:
```suggestion
        sentry_headers = message.options.get("_sentry_headers") or {}
```

sentry_sdk/integrations/dramatiq.py
```diff
@@ -85,50 +100,79 @@ class SentryMiddleware(Middleware):  # type: ignore[misc]
...
+            op=OP.QUEUE_TASK_DRAMATIQ,
+            source=TransactionSource.TASK,
+        )
+        transaction.__enter__()
+
     def after_process_message(self, broker, message, *, result=None, exception=None):
-        # type: (Broker, Message, Any, Optional[Any], Optional[Exception]) -> None
-        integration = sentry_sdk.get_client().get_integration(DramatiqIntegration)
-        if integration is None:
-            return
```
sentrivana, line 106: Please keep this -- it's a fallback in case the integration/the SDK/the client has been deactivated in the meantime. In that case we shouldn't do anything.

We should have this also in the new `before_enqueue`.

sentry_sdk/integrations/dramatiq.py
```diff
@@ -85,50 +100,79 @@ class SentryMiddleware(Middleware):  # type: ignore[misc]
...
+
     def before_process_message(self, broker, message):
-        # type: (Broker, Message) -> None
+        # type: (Broker, Message[R]) -> None
         integration = sentry_sdk.get_client().get_integration(DramatiqIntegration)
         if integration is None:
             return
 
-        message._scope_manager = sentry_sdk.new_scope()
-        message._scope_manager.__enter__()
```
sentrivana, line 95: We still need some sort of scope management in order to make sure the data we collect about tasks is isolated.

The general rule of thumb is: if you start a transaction, you should start it in a new isolation scope. See for example [huey](https://github.com/getsentry/sentry-python/blob/b838765160305aacc9b1ea0f42d6ca2e090c4b2e/sentry_sdk/integrations/huey.py#L150).

So we should start an isolation scope right after the initial `if integration is None: return` check with
```python
scope = sentry_sdk.isolation_scope()
message._scope_manager = scope
scope.__enter__()
```
Everything that we do on the `scope` later in the function can stay, but it should be done on the isolation scope, not current scope as before.

And finally, we need to `__exit__` the saved scope in `after_process_message` with `message._scope_manager.__exit__(None, None, None)`.

sentry_sdk/integrations/dramatiq.py
```diff
@@ -85,50 +100,79 @@ class SentryMiddleware(Middleware):  # type: ignore[misc]
...
-
         scope = sentry_sdk.get_current_scope()
-        scope.set_transaction_name(message.actor_name)
-        scope.set_extra("dramatiq_message_id", message.message_id)
```
sentrivana, line 99: Let's keep this, since removing it would be a breaking change.

sentry_sdk/integrations/dramatiq.py
Igreh: fixed

sentry_sdk/integrations/dramatiq.py
Igreh: fixed

sentry_sdk/integrations/dramatiq.py
Igreh: Done, but I believe it's unnecessary due to this: https://github.com/getsentry/sentry-python/pull/4571/files#diff-2722e3fe31f13ffc24072313765f1fc89f0f0721154b7ca072bb46b1f9573f5bR84
Middleware won't be in use if integration is not enabled

sentry_sdk/integrations/dramatiq.py
Igreh: Done.
but please recheck it)

sentry_sdk/integrations/dramatiq.py
Igreh: Transaction name is set here https://github.com/getsentry/sentry-python/pull/4571/files#diff-2722e3fe31f13ffc24072313765f1fc89f0f0721154b7ca072bb46b1f9573f5bR135 with the same name (message.actor_name)
and this line:
`scope.set_extra("dramatiq_message_id", message.message_id)` was reverted as long as corresponding test - test_that_dramatiq_message_id_is_set_as_extra

Igreh:
> Heyo @Igreh, I'm back with the last round.
> 
> I've tried it out and it works great! The code looks so much nicer without the custom `ContextVar`, too. 🙏🏻 I've noticed a couple things in my last round of review, most of them minor, and [one that we need to fix](https://github.com/getsentry/sentry-python/pull/4571/files#r2398154709) (but should be easy since the logic was mostly there before, just needs adapting a little to use a different kind of scope).
> 
> LMK if you still have time to make the changes, if not, I'm also happy to take this over the finish line. As always, if anything is unclear, just let me know.

Hi @sentrivana!
one more final of final patch)
Fixed everything you mentioned. Your turn to find something more to fix)

sentrivana (approved):
Ty @Igreh! Let's get this merged! :)

Issue #4568, opened 2025-07-10
Restructure Transport class hierarchy


srothh:
Restructure class hierarchy in transport by extracting common logic for sync/async into a shared superclass and create a subclass containing the current sync specific code.

PR #4572, opened 2025-07-11
Add experimental async transport

srothh:
Includes:

* this PR: new shared base class for common sync/async logic
* https://github.com/getsentry/sentry-python/pull/4580
* https://github.com/getsentry/sentry-python/pull/4591
* https://github.com/getsentry/sentry-python/pull/4614
* https://github.com/getsentry/sentry-python/pull/4615
* https://github.com/getsentry/sentry-python/pull/4700

Fixes GH-4568

### DIFF
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -163,7 +163,7 @@ def _parse_rate_limits(
 
 
 class HttpTransportCore(Transport):
-    """The base HTTP transport."""
+    """Shared base class for sync and async transports."""
 
     TIMEOUT = 30  # seconds
 
@@ -392,6 +392,8 @@ class HttpTransportCore(Transport):
     def _prepare_envelope(
         self: Self, envelope: Envelope
     ) -> Optional[Tuple[Envelope, io.BytesIO, Dict[str, str]]]:
+
+        # remove all items from the envelope which are over quota
         new_items = []
         for item in envelope.items:
             if self._check_disabled(item.data_category):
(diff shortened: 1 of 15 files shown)


sentry_sdk/transport.py
```diff
@@ -162,7 +162,7 @@ def _parse_rate_limits(
             continue
 
 
-class BaseHttpTransport(Transport):
+class HttpTransportCore(Transport):
     """The base HTTP transport."""
```
antonpirker: Please also update the docstring

sentry_sdk/transport.py
```diff
@@ -398,9 +389,9 @@ def _is_worker_full(self: Self) -> bool:
     def is_healthy(self: Self) -> bool:
         return not (self._is_worker_full() or self._is_rate_limited())
 
-    def _send_envelope(self: Self, envelope: Envelope) -> None:
-
-        # remove all items from the envelope which are over quota
+    def _prepare_envelope(
+        self: Self, envelope: Envelope
+    ) -> Optional[Tuple[Envelope, io.BytesIO, Dict[str, str]]]:
         new_items = []
```
antonpirker, line 417: please keep the `# remove all items from the envelope which are over quota` comment

sentry_sdk/transport.py
```diff
@@ -506,6 +494,54 @@ def _request(
...
+        if _prepared_envelope is None:
+            return None
+        envelope, body, headers = _prepared_envelope
+        self._send_request(
+            body.getvalue(),
+            headers=headers,
+            endpoint_type=EndpointType.ENVELOPE,
+            envelope=envelope,
+        )
+        return None
```
antonpirker: Just a personal taste but I would change this to only have one `return None` so it is easier to read. 
```suggestion
        _prepared_envelope = self._prepare_envelope(envelope)
        if _prepared_envelope is not None:
            envelope, body, headers = _prepared_envelope
            self._send_request(
                body.getvalue(),
                headers=headers,
                endpoint_type=EndpointType.ENVELOPE,
                envelope=envelope,
            )
        return None
```

tests/test_transport.py
```diff
@@ -641,3 +641,38 @@ def test_record_lost_event_transaction_item(capturing_server, make_client, span_
         "reason": "test",
         "quantity": span_count + 1,
     } in discarded_events
+
```
antonpirker, line 730: Please add a test that tests those untested lines here: https://app.codecov.io/gh/getsentry/sentry-python/pull/4572/blob/sentry_sdk/transport.py#L292

srothh: Done!

sentry_sdk/transport.py
srothh: Agree, this looks nicer. Thanks!

antonpirker (approved):
Looks good.

sl0thentr0py:
okay, so now that we have 2 base/core classes, it's a bit more confusing than before what the contracts of each entity actually are.

I want us to define more clearly and explicitly what is
* shared across sync and async 
* only sync 
  * only http2
* only async

but I think it's clearer if we merge all the changes in first and then do some shuffling. We will do that with mixin classes.
https://adamj.eu/tech/2025/05/01/python-type-hints-mixin-classes/

Issue #4538, opened 2025-06-30
Remove as many forked tests as possible


sl0thentr0py:

alexander-alderman-webb:
Many of the django tests relied on forked tests because of how asgiref's async-sync bridge and `concurrent.futures` interact.

Now we patch `concurrent.futures` we can remove the `forked`.

See https://github.com/getsentry/sentry-python/pull/4770

PR #4577, opened 2025-07-11
Remove forked from test_transport, separate gevent tests and generalize capturing_server to be module level

sl0thentr0py:
* Move `CapturingServer` to `conftest` for reuse
* Make `capturing_server` in original non-forked transport tests to be module level for faster tests
* Move `gevent` based transport tests to separate file since they still need to be forked because of the global monkeypatch 

As a result, `test_transport` now takes 13 seconds instead of almost 3 minutes.

BEFORE

```
 235 passed, 384 skipped in 173.78s (0:02:53)
```

AFTER
```
235 passed in 12.69s 
```

part of #4538

### DIFF
--- a/tests/test_gevent.py
+++ b/tests/test_gevent.py
@@ -1,0 +1,118 @@
+import logging
+import pickle
+from datetime import datetime, timezone
+
+import sentry_sdk
+from sentry_sdk._compat import PY37, PY38
+
+import pytest
+from tests.conftest import CapturingServer
+
+pytest.importorskip("gevent")
+
+
+@pytest.fixture(scope="module")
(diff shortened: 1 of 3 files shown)


tests/test_gevent.py
sl0thentr0py, line 1: ultimately, I want to isolate all gevent tests here so they don't leak into the rest of the codebase

sl0thentr0py:
as a side, I also really tried to make the gevent test not forked, but the shutdown in `pytest-localserver` doesn't seem to be compatible with gevent and throws ... we can revisit this later once all gevent stuff is isolated

sentrivana (approved):
😍

PR #4591, opened 2025-07-17
Add async task background worker

srothh:
Add a new implementation of the transport background worker based on an async task. This worker mostly mirrors the same functionality as the thread-based worker, with the exception that it exposes a non-blocking async flush. Furthermore, the worker itself is not thread-safe and should be called using [run_coroutine_threadsafe](https://docs.python.org/3/library/asyncio-task.html#asyncio.run_coroutine_threadsafe) or similar when called from another thread.

### DIFF
--- a/sentry_sdk/worker.py
+++ b/sentry_sdk/worker.py
@@ -3,6 +3,7 @@ from abc import ABC, abstractmethod
 import os
 import threading
 import asyncio
+import inspect
 
 from time import sleep, time
 from sentry_sdk._queue import Queue, FullError
@@ -255,7 +256,12 @@ class AsyncWorker(Worker):
         while True:
             callback = await self._queue.get()
             try:
-                callback()
+                if inspect.iscoroutinefunction(callback):
+                    # Callback is an async coroutine, need to await it
+                    await callback()
+                else:
+                    # Callback is a sync function, need to call it
+                    callback()
             except Exception:
                 logger.error("Failed processing job", exc_info=True)
             finally:
--- a/sentry_sdk/worker.py
+++ b/sentry_sdk/worker.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 import os
 import threading
+import asyncio
 
 from time import sleep, time
 from sentry_sdk._queue import Queue, FullError
@@ -169,3 +170,94 @@ class BackgroundWorker(Worker):
             finally:
                 self._queue.task_done()
             sleep(0)
+
+
+class AsyncWorker(Worker):
+    def __init__(self, queue_size: int = DEFAULT_QUEUE_SIZE) -> None:
+        self._queue: asyncio.Queue = asyncio.Queue(queue_size)
+        self._task: Optional[asyncio.Task] = None
+        # Event loop needs to remain in the same process
+        self._task_for_pid: Optional[int] = None
+        self._loop: Optional[asyncio.AbstractEventLoop] = None
+
+    @property
+    def is_alive(self) -> bool:
+        if self._task_for_pid != os.getpid():
+            return False
+        if not self._task or not self._loop:
+            return False
+        return self._loop.is_running() and not self._task.done()
+
+    def kill(self) -> None:
+        if self._task:
+            self._task.cancel()
+            self._task = None
+            self._task_for_pid = None
+
+    def start(self) -> None:
+        if not self.is_alive:
+            try:
+                self._loop = asyncio.get_running_loop()
+                self._task = self._loop.create_task(self._target())
+                self._task_for_pid = os.getpid()
+            except RuntimeError:
+                # There is no event loop running
+                self._loop = None
+                self._task = None
+                self._task_for_pid = None
+
+    def full(self) -> bool:
+        return self._queue.full()
+
+    def _ensure_task(self) -> None:
+        if not self.is_alive:
+            self.start()
+
+    async def _wait_flush(self, timeout: float, callback: Optional[Any] = None) -> None:
+        if not self._loop or not self._loop.is_running():
+            return
+
+        initial_timeout = min(0.1, timeout)
+
+        # Timeout on the join
+        try:
+            await asyncio.wait_for(self._queue.join(), timeout=initial_timeout)
+        except asyncio.TimeoutError:
+            pending = self._queue.qsize() + 1
+            logger.debug("%d event(s) pending on flush", pending)
+            if callback is not None:
+                callback(pending, timeout)
+
+            try:
+                remaining_timeout = timeout - initial_timeout
+                await asyncio.wait_for(self._queue.join(), timeout=remaining_timeout)
+            except asyncio.TimeoutError:
+                pending = self._queue.qsize() + 1
+                logger.error("flush timed out, dropped %s events", pending)
+
+    async def flush(self, timeout: float, callback: Optional[Any] = None) -> None:
+        logger.debug("background worker got flush request")
+        if self.is_alive and timeout > 0.0:
+            await self._wait_flush(timeout, callback)
+        logger.debug("background worker flushed")
+
+    def submit(self, callback: Callable[[], None]) -> bool:
+        self._ensure_task()
+
+        try:
+            self._queue.put_nowait(callback)
+            return True
+        except asyncio.QueueFull:
+            return False
+
+    async def _target(self) -> None:
+        while True:
+            callback = await self._queue.get()
+            try:
+                callback()
+            except Exception:
+                logger.error("Failed processing job", exc_info=True)
+            finally:
+                self._queue.task_done()
+            # Yield to let the event loop run other tasks
+            await asyncio.sleep(0)
--- a/sentry_sdk/worker.py
+++ b/sentry_sdk/worker.py
@@ -36,7 +36,7 @@ class Worker(ABC):
         pass
 
     def flush(
-        self, timeout: float, callback: Optional[Callable[[int, float], None]] = None
+        self, timeout: float, callback: Optional[Callable[[int, float], Any]] = None
     ) -> None:
         """
         Flush the worker.
@@ -52,7 +52,7 @@ class Worker(ABC):
         pass
 
     @abstractmethod
-    def submit(self, callback: Callable[[], None]) -> bool:
+    def submit(self, callback: Callable[[], Any]) -> bool:
         """
         Schedule a callback to be executed by the worker.
 
@@ -151,7 +151,7 @@ class BackgroundWorker(Worker):
                 pending = self._queue.qsize() + 1
                 logger.error("flush timed out, dropped %s events", pending)
 
-    def submit(self, callback: Callable[[], None]) -> bool:
+    def submit(self, callback: Callable[[], Any]) -> bool:
         self._ensure_thread()
         try:
             self._queue.put_nowait(callback)
@@ -243,7 +243,7 @@ class AsyncWorker(Worker):
             await self._wait_flush(timeout, callback)
         logger.debug("background worker flushed")
 
-    def submit(self, callback: Callable[[], None]) -> bool:
+    def submit(self, callback: Callable[[], Any]) -> bool:
         self._ensure_task()
 
         try:
--- a/sentry_sdk/worker.py
+++ b/sentry_sdk/worker.py
@@ -248,7 +248,7 @@ class AsyncWorker(Worker):
                 pending = self._queue.qsize() + 1
                 logger.error("flush timed out, dropped %s events", pending)
 
-    async def flush(self, timeout: float, callback: Optional[Any] = None) -> None:
+    async def flush_async(self, timeout: float, callback: Optional[Any] = None) -> None:
         logger.debug("background worker got flush request")
         if self.is_alive and timeout > 0.0:
             await self._wait_flush(timeout, callback)


sentry_sdk/worker.py
```diff
@@ -170,3 +172,128 @@ def _target(self) -> None:
...
+        if not self._task or not self._loop:
+            return False
+        return self._loop.is_running() and not self._task.done()
+
+    def kill(self) -> None:
+        if self._task:
+            try:
+                self._queue.put_nowait(_TERMINATOR)
+            except asyncio.QueueFull:
+                logger.debug("async worker queue full, kill failed")
```
antonpirker: Could we use the `full()` method from below? This way we would only have one way to check if the queue is full.

sentry_sdk/worker.py
```diff
@@ -170,3 +172,128 @@ def _target(self) -> None:
...
+                callback(pending, timeout)
+
+            try:
+                remaining_timeout = timeout - initial_timeout
+                await asyncio.wait_for(self._queue.join(), timeout=remaining_timeout)
+            except asyncio.TimeoutError:
+                pending = self._queue.qsize() + 1
+                logger.error("flush timed out, dropped %s events", pending)
+
+    async def flush_async(self, timeout: float, callback: Optional[Any] = None) -> None:
```
antonpirker: Why is this called `flush_async` and not `flush`? Shouldnt there be a `flush()` that maybe calls `flush_async` so all the workers can be used the same way?

sentry_sdk/worker.py
```diff
@@ -170,3 +172,128 @@ def _target(self) -> None:
...
+        logger.debug("background worker got flush request")
+        if self.is_alive and timeout > 0.0:
```
antonpirker: I know this is also in the `BackgroundWorker` implementation, but do you know why a timeout of `0.0` should not flush the worker?

sentry_sdk/worker.py
```diff
@@ -170,3 +172,128 @@ def _target(self) -> None:
...
+            callback = await self._queue.get()
+            if callback is _TERMINATOR:
+                self._queue.task_done()
+                break
+            # Firing tasks instead of awaiting them allows for concurrent requests
+            task = asyncio.create_task(self._process_callback(callback))
+            # Create a strong reference to the task so it can be cancelled on kill
+            # and does not get garbage collected while running
+            self._active_tasks.add(task)
+            task.add_done_callback(self._on_task_complete)
```
antonpirker, line 302: Maybe we should add the done callback before adding it to _active_tasks. (Just a gut feeling that the task might be finishing before the callback is in place..)

sentry_sdk/worker.py
```diff
@@ -170,3 +172,128 @@ def _target(self) -> None:
...
+        if not self._loop or not self._loop.is_running():
+            return
+
+        initial_timeout = min(0.1, timeout)
+
+        # Timeout on the join
+        try:
+            await asyncio.wait_for(self._queue.join(), timeout=initial_timeout)
+        except asyncio.TimeoutError:
+            pending = self._queue.qsize() + 1
```
antonpirker: Because there could be multiple tasks in `self._active_tasks` maybe the more correct version would be `self.queue.qsize() + len(self._active_tasks)`?

sentry_sdk/worker.py
```diff
@@ -170,3 +172,128 @@ def _target(self) -> None:
...
+            self._task_for_pid = None
+
+    def start(self) -> None:
+        if not self.is_alive:
+            try:
+                self._loop = asyncio.get_running_loop()
+                self._task = self._loop.create_task(self._target())
+                self._task_for_pid = os.getpid()
+            except RuntimeError:
+                # There is no event loop running
```
antonpirker, line 237: I think we should at least log a warning here, so we do not swallow failures silently.

sentry_sdk/worker.py
```diff
@@ -170,3 +172,128 @@ def _target(self) -> None:
...
+            # Yield to let the event loop run other tasks
+            await asyncio.sleep(0)
+
+    async def _process_callback(self, callback: Callable[[], Any]) -> None:
+        if inspect.iscoroutinefunction(callback):
+            # Callback is an async coroutine, need to await it
+            await callback()
+        else:
+            # Callback is a sync function, need to call it
+            callback()
```
antonpirker: this will block the event loop. Maybe we should do `await asyncio.to_thread(callback)` here? @sl0thentr0py what is your take on this? Blocking or creating threads?

sentry_sdk/worker.py
```diff
@@ -170,3 +172,128 @@ def _target(self) -> None:
             finally:
                 self._queue.task_done()
             sleep(0)
+
+
+class AsyncWorker(Worker):
+    def __init__(self, queue_size: int = DEFAULT_QUEUE_SIZE) -> None:
+        self._queue: asyncio.Queue[Any] = asyncio.Queue(queue_size)
```
antonpirker: I think in older Pythons (3.7-3.9) it can be a problem if a Queue is initialized when there is no current event loop. I would not init it here, but in the `start()` after we got the event loop. something like:
```python
# in start()
self._loop = asyncio.get_running_loop()
if self._queue is None:
    self._queue = asyncio.Queue(self._queue_size)
```

sentry_sdk/worker.py
```diff
@@ -170,3 +172,128 @@ def _target(self) -> None:
...
+        initial_timeout = min(0.1, timeout)
+
+        # Timeout on the join
+        try:
+            await asyncio.wait_for(self._queue.join(), timeout=initial_timeout)
+        except asyncio.TimeoutError:
+            pending = self._queue.qsize() + 1
+            logger.debug("%d event(s) pending on flush", pending)
+            if callback is not None:
+                callback(pending, timeout)
```
antonpirker, line 265: This will block the entire event loop. I guess the callback for the AsyncCorker should always be a coroutine that can be awaited.

sentry_sdk/worker.py
```diff
@@ -170,3 +172,128 @@ def _target(self) -> None:
...
+                self._queue.task_done()
+                break
+            # Firing tasks instead of awaiting them allows for concurrent requests
+            task = asyncio.create_task(self._process_callback(callback))
+            # Create a strong reference to the task so it can be cancelled on kill
+            # and does not get garbage collected while running
+            self._active_tasks.add(task)
+            task.add_done_callback(self._on_task_complete)
+            # Yield to let the event loop run other tasks
+            await asyncio.sleep(0)
```
antonpirker, line 304: I think this is not necessary, because the `callback = await self._queue.get()` in the loop also gives up control.

sentry_sdk/worker.py
sl0thentr0py: just make a fully async version first, do we really need to mix sync and async? we can patch that on later if really necessary. Please keep it simple for now.

sentry_sdk/worker.py
srothh: I did it this way because the threaded worker does the same, but I think there should be no functional difference, so yes it is probably nicer!

sentry_sdk/worker.py
srothh: This is how I had it initially. However, when integrating this with the SDK there is an issue:

The client exposes a synchronous flush/close method, which in close expects flush to fully complete before shutting down the transport. If the worker/transport used the synchronous flush method, they could only create an async flush task (which is necessary to avoid deadlocking/use the async queue) in a fire-and-forget way. This means that there is not really a way to properly order the flush with the shutdown in the client, unless the client itself already spawns an async flush task using the transport.

I could rename it back to flush and keep it a coroutine, I just thought this way it is less confusing.

Update: I went back to this, and realised I can just make the flush in the worker already return the task, then pass it through the transport  to the client. This makes the flush non-blocking if it is not awaited, but preserves both the close logic for the client and the interface specification

sentry_sdk/worker.py
srothh: I am not really sure. But the client falls back to a "shutdown_timeout" option if the timeout is not set. Maybe setting shutdown_timeout to 0.0 is supposed to immediately quit the application on a call to close ?

sentry_sdk/worker.py
srothh: I actually oriented myself on the [official documentation](https://docs.python.org/3/library/asyncio-task.html#creating-tasks) for the add_done_callback for this. However, if I understand it correctly I do not think it matters, as the task can only start once the event loop yields control from the current function, which should only happen on await/return. If it is more readable, I can change it however.

sentry_sdk/worker.py
srothh: So pending should be not just the number still waiting in the queue, but the number of non-completed tasks in general ? I think this makes sense, thanks!

sentry_sdk/worker.py
srothh: Thank you, I was not aware of this !

sentry_sdk/worker.py
srothh: From what I can tell, this functionality is used only by the atexit of the SDK, which from what I can tell only uses this with synchronous functions currently. If the blocking is a problem, `run_in_executor` could be used ? But as this only happens once on exit currently, I am not sure if this is an issue.

sentry_sdk/worker.py
srothh: After checking again, I agree that this is not needed. I confused this with the flush callbacks when I initially explored the SDK, which are synchronous functions. Those are however, processed in the flush method, and I believe for this there should only be the async requests currently. Thanks for catching that!

Update: I think this might be needed after all. I changed it, and noticed my tests breaking. The transport actually submits _flush_client_reports to the worker as well, which currently is a sync function. It could technically be made an async callback, but there is no async component inside of it.

sentry_sdk/worker.py
srothh: I am not 100% sure about this, but if there is instantly items in the queue, does the await also give up control? If not, it might have the same issue, but I am not sure

sentry_sdk/worker.py
srothh: Actually, I thought about why the threaded worker uses this again and think the current version might be the better way. If this was replaced by a full call, this would no longer be atomic and potentially introduce race conditions between checking for fullness and inserting something into the queue.

sentry_sdk/worker.py
```diff
@@ -186,3 +187,127 @@
...
+
+    async def _process_callback(self, callback: Callable[[], Any]) -> None:
+        # Callback is an async coroutine, need to await it
+        await callback()
```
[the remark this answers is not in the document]
srothh: Discussed previously

sentry_sdk/worker.py
srothh: Added

sentry_sdk/worker.py
srothh: Fixed

sentry_sdk/worker.py
srothh: So I added this check to start, but was not sure if I should remove the initialization in init entirely (as that would add some extra checks in case the queue is optional), as I am not sure what the behaviour is in that problem; I tried replicating it with python 3.9 but I did not get any issues when instantiating the queue outside of the event loop

sentry_sdk/worker.py
```diff
@@ -186,3 +187,128 @@ def _target(self) -> None:
...
+            self._loop = None
+            self._task = None
+            self._task_for_pid = None
+
+    def start(self) -> None:
+        if not self.is_alive:
+            try:
+                self._loop = asyncio.get_running_loop()
+                if self._queue is None:
+                    self._queue = asyncio.Queue(maxsize=self._queue_size)
```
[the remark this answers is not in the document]
srothh: Intentional, discussed before

sentry_sdk/worker.py
srothh: Fixed, also added None checks for the queue to satisfy mypy with the queue as Optional in init now.

sentry_sdk/worker.py
```diff
@@ -186,3 +187,128 @@
             finally:
                 self._queue.task_done()
             sleep(0)
+
+
+class AsyncWorker(Worker):
+    def __init__(self, queue_size: int = DEFAULT_QUEUE_SIZE) -> None:
+        self._queue: asyncio.Queue[Any] = asyncio.Queue(maxsize=queue_size)
```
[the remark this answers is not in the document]
srothh: fixed

sentry_sdk/worker.py
antonpirker: Good point! Then it is ok.

sentry_sdk/worker.py
antonpirker: ok, makes sense.

sentry_sdk/worker.py
antonpirker: ok, then lets keep it like this

sentry_sdk/worker.py
antonpirker: ok lets keep it then. it does not hurt so it is fine.

sentry_sdk/worker.py
antonpirker: ok

sentry_sdk/worker.py
antonpirker: ok, then keep it so the behavior is the same as `BackgroundWorker`s behavior.

sentry_sdk/worker.py
antonpirker: ok

sentry_sdk/worker.py
antonpirker: ok thanks for the explanation. i guess it is fine then.

antonpirker (approved):

srothh:
CI lint issue is in an unrelated file, not sure why it is popping up.

sentry_sdk/worker.py
```diff
@@ -186,3 +187,134 @@ def _target(self) -> None:
             finally:
                 self._queue.task_done()
             sleep(0)
+
+
+class AsyncWorker(Worker):
+    def __init__(self, queue_size: int = DEFAULT_QUEUE_SIZE) -> None:
+        self._queue: Optional[asyncio.Queue[Any]] = None
```
sl0thentr0py, line 194: is there a reason you set queue to None and instantiated later? I would slightly prefer having an empty queue already and getting rid of all the `queue is None` checks.

srothh: https://github.com/getsentry/sentry-python/pull/4591#discussion_r2228712724
I had it this way initially, but added this on feedback from anton

srothh: But I am not sure if this is actually an issue, as an event loop should be running already at this point

sentry_sdk/worker.py
```diff
@@ -186,3 +187,134 @@ def _target(self) -> None:
...
+    async def _target(self) -> None:
+        if self._queue is None:
+            return
+        while True:
+            callback = await self._queue.get()
+            if callback is _TERMINATOR:
+                self._queue.task_done()
+                break
+            # Firing tasks instead of awaiting them allows for concurrent requests
+            task = asyncio.create_task(self._process_callback(callback))
```
sl0thentr0py, line 298: this one is `asyncio.create_task` while all the others use `self._loop` directly. Is that intentional?

sentry_sdk/worker.py
sl0thentr0py: hmm I see.. ok leave it for now then. I just prefer less branches in general.

sentry_sdk/worker.py
srothh: _target is the function actually running in the task, whereas other parts where tasks get instantiated are called from the transport. This means that running it directly with asyncio will reference the correct loop that is executing target. I believe using self._loop here could introduce a race, as kill() resets the loop reference to None, which can happen before the task _submit reaches the TERMINATOR

srothh: (for this reason also the other functions introduce an is_alive() check)

sl0thentr0py (approved):
very nice work, will merge this

one question I need to think about is:
* if we want multiple concurrent requests at the same time at all
* if we do, a way to maybe control the number of requests at the same time (but without adding another layer of queuing complexity)

PR #4590, opened 2025-07-17
Remove explicit __del__'s in threaded classes

sl0thentr0py:
The changes in #4577 introduced a bit of flakiness on pre-3.10 due to a weird interaction of `capsys`, `stderr` logging and our object lifecycles.

In this PR, I'm removing all the explicit `__del__` `kill`s since we [call them all explicitly in `client.close`](https://github.com/getsentry/sentry-python/blob/09c2e32cc7a618e49f5d8ae59e22d8b12f253687/sentry_sdk/client.py#L1001-L1021) anyway and that's already cleaner. Having logic in `__del__` causes non-deterministic GC behavior, especially with threaded code.

Stacktrace is linked in a comment below where you can see the `transport.__del__` method is causing the weird `reentrant` logging bug to `stderr`.

PR #4593, opened 2025-07-17
test: Remove `test_installed_modules`

szokeasaurusrex:
The test `test_installed_modules` appears to not be all that useful.

The test exists to verify the behavior of the [`_generate_installed_modules` function](https://github.com/getsentry/sentry-python/blob/9b66f3b51502ca600554c711bc3f599c18f8f18b/sentry_sdk/utils.py#L1689). However, all the test does is essentially check the output of `_generate_installed_modules` against a refactored version of the function call itself.

In short, in its current form, the test appears to not make too much sense. As the test recently started failing, let's just delete it.

---

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -30,7 +30,6 @@ from sentry_sdk.utils import (
     serialize_frame,
     is_sentry_url,
     _get_installed_modules,
-    _generate_installed_modules,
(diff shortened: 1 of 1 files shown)


sl0thentr0py:
gevent flaky test will be fixed in https://github.com/getsentry/sentry-python/pull/4590

PR #4595, opened 2025-07-17
Setup opentelemetry patch only once

sl0thentr0py:
All our integrations follow the `setup_once` logic while the otel patching was being done for each `Client` again leading to recursive patching. 
This was revealed by removing the transport test forks.


szokeasaurusrex (approved):
lgtm, but looks like it still could be possible for the opentelemetry readable span to be patched mutiple times in a multithreaded context if there is a race condition.

If it is possible that multiple threads would attempt the patch, we should make the code thread safe

Issue #4582, opened 2025-07-14
Create Async transport subclass


srothh:
Create a subclass of HttpTransportCore, implementing async transport using httpcore and an async-task based background worker.

PR #4614, opened 2025-07-23
Add async transport

srothh:
Add async implementation of the abstract Transport class. This transport utilizes the async task worker as well as the httpcore async functionality. 

Thread Safety: As capture_envelope is registered by the client as a callback for several background threads in the sdk, which are not running the event loop, capture_envelope in the transport is made to be thread safe and allow for execution on the event loop from other threads. The same is currently not the case for flush, as there does not seem to be a usage from background threads, however if necessary, it can also be added.

HTTP2 support: Currently not activated, but from the looks of the [httpcore docs](https://www.encode.io/httpcore/http2/) it should be as simple as setting the http2 in the init of the pool to true. This likely makes sense to support, as HTTP2 shows great performance improvements with concurrent requests.

Kill: The kill method is sync, but the pool needs to be closed asynchronously. Currently, this is done by launching a task. However, the task cannot be awaited in sync code without deadlocking, therefore kill followed by an immediate loop shutdown could technically lead to resource leakage. Therefore, I decided to make kill optionally return the async task, so it can be awaited if called from an async context.

Note also that parts of the code are very similar to the HTTP2 integration, as they both use the httpcore library. Maybe in a later PR there could be a shared superclass to avoid code duplication?

GH-4582

### DIFF
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -768,8 +768,10 @@ class AsyncHttpTransport(HttpTransportCore):
         for task in self.background_tasks:
             task.cancel()
         self.background_tasks.clear()
-
-        self._loop.create_task(self._pool.aclose())  # type: ignore
+        try:
+            self._loop.create_task(self._pool.aclose())  # type: ignore
+        except RuntimeError:
+            logger.warning("Event loop not running, aborting kill.")
 
 
 class HttpTransport(BaseHttpTransport):
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -1029,6 +1029,8 @@ def make_transport(options: Dict[str, Any]) -> Optional[Transport]:
             # No event loop running, fall back to sync transport
             logger.warning("No event loop running, falling back to sync transport.")
             transport_cls = Http2Transport if use_http2_transport else HttpTransport
+    else:
+        transport_cls = Http2Transport if use_http2_transport else HttpTransport
 
     if isinstance(ref_transport, Transport):
         return ref_transport
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -1021,11 +1021,14 @@ def make_transport(options: Dict[str, Any]) -> Optional[Transport]:
     use_http2_transport = options.get("_experiments", {}).get("transport_http2", False)
     use_async_transport = options.get("_experiments", {}).get("transport_async", False)
     # By default, we use the http transport class
-    if use_async_transport and asyncio.get_running_loop() is not None:
-        transport_cls: Type[Transport] = AsyncHttpTransport
-    else:
-        use_http2 = use_http2_transport
-        transport_cls = Http2Transport if use_http2 else HttpTransport
+    if use_async_transport:
+        try:
+            asyncio.get_running_loop()
+            transport_cls: Type[Transport] = AsyncHttpTransport
+        except RuntimeError:
+            # No event loop running, fall back to sync transport
+            logger.warning("No event loop running, falling back to sync transport.")
+            transport_cls = Http2Transport if use_http2_transport else HttpTransport
 
     if isinstance(ref_transport, Transport):
         return ref_transport
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -779,7 +779,9 @@ class AsyncHttpTransport(HttpTransportCore):
             task.cancel()
         self.background_tasks.clear()
         try:
-            self._loop.create_task(self._pool.aclose())  # type: ignore
+            task = self._loop.create_task(self._pool.aclose())  # type: ignore
+            self.background_tasks.add(task)
+            task.add_done_callback(lambda t: self.background_tasks.discard(t))
         except RuntimeError:
             logger.warning("Event loop not running, aborting kill.")
 
--- a/sentry_sdk/transport.py
+++ b/sentry_sdk/transport.py
@@ -6,6 +6,7 @@ import gzip
 import socket
 import ssl
 import time
+import asyncio
 from datetime import datetime, timedelta, timezone
 from collections import defaultdict
 from urllib.request import getproxies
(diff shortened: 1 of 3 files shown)


sentry_sdk/transport.py
```diff
@@ -569,6 +571,217 @@ def flush(
...
+    def kill(self: Self) -> None:
+
+        logger.debug("Killing HTTP transport")
+        self._worker.kill()
+        for task in self.background_tasks:
+            task.cancel()
+        self.background_tasks.clear()
+        try:
+            self._loop.create_task(self._pool.aclose())  # type: ignore
+        except RuntimeError:
```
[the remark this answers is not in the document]
srothh: GC problem: I added it just to make sure.
Background thread problem: From what I can tell, no background threads in the SDK call kill on the transport. As for whether this needs to be threadsafe for the possibility of a call from a background thread running the SDK, I am not sure. But I decided to keep it this way to minimize complexity in the initial version of the async transport.

sentry_sdk/transport.py
```diff
@@ -569,6 +571,219 @@ def flush(
...
+            method,
+            self._auth.get_api_url(endpoint_type),
+            content=body,
+            headers=headers,  # type: ignore
+        )
+
+    def _flush_client_reports(self: Self, force: bool = False) -> None:
+        client_report = self._fetch_pending_client_report(force=force, interval=60)
+        if client_report is not None:
+            self.capture_envelope(Envelope(items=[client_report]))
```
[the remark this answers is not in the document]
srothh: 1. Fixed
2. I do not think this is true - _flush_client_reports only calls capture_envelope if a client report is ready, which is not the case more than once in the call. This simply mirrors the logic in the synchronous case.

sentry_sdk/transport.py
```diff
@@ -569,6 +571,219 @@
...
+        for task in self.background_tasks:
+            task.cancel()
+        self.background_tasks.clear()
+        try:
+            task = self._loop.create_task(self._pool.aclose())  # type: ignore
+            self.background_tasks.add(task)
+            task.add_done_callback(lambda t: self.background_tasks.discard(t))
+        except RuntimeError:
+            logger.warning("Event loop not running, aborting kill.")
+
```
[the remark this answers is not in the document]
srothh: Addressed before

sentry_sdk/transport.py
```diff
@@ -569,6 +571,227 @@ def flush(
...
+            self.background_tasks.add(task)
+            task.add_done_callback(self.background_tasks.discard)
+        except RuntimeError:
+            # We are in a background thread, not running an event loop,
+            # have to launch the task on the loop in a threadsafe way.
+            if self._loop and self._loop.is_running():
+                asyncio.run_coroutine_threadsafe(
+                    self._capture_envelope(envelope),
+                    self._loop,
+                )
```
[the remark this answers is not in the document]
srothh: While I think these cases can technically happen, I think for the sake of keeping it simple for the first version this (especially the first section) is maybe not necessary

sentry_sdk/transport.py
```diff
@@ -569,6 +575,227 @@ def flush(
...
+        self._worker.kill()
+        for task in self.background_tasks:
+            task.cancel()
+        self.background_tasks.clear()
+        try:
+            # Return the pool cleanup task so caller can await it if needed
+            return self._loop.create_task(self._pool.aclose())  # type: ignore
+        except RuntimeError:
+            logger.warning("Event loop not running, aborting kill.")
+            return None
```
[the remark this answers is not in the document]
srothh: Intentional, can change the superclass return if necessary.

sentry_sdk/transport.py
```diff
@@ -571,6 +577,228 @@ def flush(
...
+                self.record_lost_event("queue_overflow", item=item)
+
+    def capture_envelope(self: Self, envelope: Envelope) -> None:
+        # Synchronous entry point
+        try:
+            asyncio.get_running_loop()
+            # We are on the main thread running the event loop
+            task = asyncio.create_task(self._capture_envelope(envelope))
+            self.background_tasks.add(task)
+            task.add_done_callback(self.background_tasks.discard)
```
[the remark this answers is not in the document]
srothh: While this is true, I think this is maybe not necessary for an initial version - the capture_envelope currenlty is designed to work with SDK threads, none of which run their own event loop.

sentry_sdk/transport.py
```diff
@@ -571,6 +577,228 @@ def flush(
...
+        self: Self,
+        timeout: float,
+        callback: Optional[Callable[[int, float], None]] = None,
+    ) -> Optional[asyncio.Task[None]]:
+        logger.debug("Flushing HTTP transport")
+
+        if timeout > 0:
+            self._worker.submit(lambda: self._flush_client_reports(force=True))
+            return self._worker.flush(timeout, callback)  # type: ignore[func-returns-value]
+        return None
```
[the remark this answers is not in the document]
srothh: I did it this way because the ABC definition already existed and worked for the current sync transport implementations, therefore I thought this was the better approach. I do not think this should ever cause errors, as simply None will be returned and the return Type is marked as optional anyway. Also, the worker for the Async Transport cannot be the threaded BackgroundWorker. This would break everything.

sentry_sdk/transport.py
```diff
@@ -571,6 +590,241 @@ def flush(
...
+                    asyncio.run_coroutine_threadsafe(
+                        self._capture_envelope(envelope),
+                        self.loop,
+                    )
+                else:
+                    # The event loop is no longer running
+                    logger.warning("Async Transport is not running in an event loop.")
+                    self.on_dropped_event("no_async_context")
+                    for item in envelope.items:
+                        self.record_lost_event("no_async_context", item=item)
```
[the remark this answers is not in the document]
srothh: Already mentioned in a previous comment

sentry_sdk/transport.py
```diff
@@ -571,6 +586,241 @@ def flush(
...
+                self.background_tasks.add(task)
+                task.add_done_callback(self.background_tasks.discard)
+            except RuntimeError:
+                # We are in a background thread, not running an event loop,
+                # have to launch the task on the loop in a threadsafe way.
+                if self.loop and self.loop.is_running():
+                    asyncio.run_coroutine_threadsafe(
+                        self._capture_envelope(envelope),
+                        self.loop,
+                    )
```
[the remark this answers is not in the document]
srothh: Already mentioned in a previous comment

sentry_sdk/transport.py
```diff
@@ -17,18 +18,27 @@
...
 
-    HTTP2_ENABLED = True
+    HTTP2_ENABLED = httpcore is not None
 except ImportError:
     HTTP2_ENABLED = False
 
+try:
+    ASYNC_TRANSPORT_ENABLED = httpcore is not None
+except ImportError:
+    ASYNC_TRANSPORT_ENABLED = False
```
antonpirker, line 34: I think there can not be an `ImportError` here. It is just setting a var.

sentry_sdk/transport.py
```diff
@@ -224,9 +234,18 @@ def __init__(self: Self, options: Dict[str, Any]) -> None:
         elif self._compression_algo == "br":
             self._compression_level = 4
 
-    def _create_worker(self: Self, options: Dict[str, Any]) -> Worker:
-        # For now, we only support the threaded sync background worker.
-        return BackgroundWorker(queue_size=options["transport_queue_size"])
+    def _create_worker(self, options: dict[str, Any]) -> Worker:
+        async_enabled = options.get("_experiments", {}).get("transport_async", False)
```
antonpirker: I think `transport_async` should also be added to the `Experiments` TypedDict in `sentry_sdk/consts.py`.

sentry_sdk/transport.py
```diff
@@ -571,6 +590,241 @@ def flush(
...
+            if _prepared_envelope is None:
+                return None
+            envelope, body, headers = _prepared_envelope
+            await self._send_request(
+                body.getvalue(),
+                headers=headers,
+                endpoint_type=EndpointType.ENVELOPE,
+                envelope=envelope,
+            )
+            return None
```
antonpirker: would also implement it with just one `return None` like in `BaseHttpTransport`

sentry_sdk/transport.py
```diff
@@ -571,6 +590,241 @@ def flush(
...
+                    "timeout": {
+                        "pool": self.TIMEOUT,
+                        "connect": self.TIMEOUT,
+                        "write": self.TIMEOUT,
+                        "read": self.TIMEOUT,
+                    }
+                },
+            )
+
+        def _flush_client_reports(self: Self, force: bool = False) -> None:
```
antonpirker: are there plans to make this async in the future?

sentry_sdk/transport.py
antonpirker: We have a official list of known loss reasons here: https://develop.sentry.dev/sdk/telemetry/client-reports/#envelope-item-payload (in `on_dropped_event` we can use whatever, because this is just a callback for the user, but in `record_lost_event` we need to pick one)

sentry_sdk/transport.py
srothh: You are right, I added the import only in the next branch because I added the dependency there, so I should probably have left this up in the next PR, but it is used there

sentry_sdk/transport.py
srothh: This is also in the next PR integrate-async-transport

sentry_sdk/transport.py
srothh: Yes, thank you! I will change it

sentry_sdk/transport.py
srothh: I left it sync currently because it itself does not have an async component. But making it async would allow the entire request path to be async without changing functionality (and might also make it so the worker only needs to handle async callbacks), so it makes sense I think to make it an async function

sentry_sdk/transport.py
srothh: Thanks for pointing that out! So it should probably be `internal_sdk_error` ?

sentry_sdk/transport.py
```diff
@@ -571,6 +590,240 @@ def flush(
...
+            self: Self,
+            timeout: float,
+            callback: Optional[Callable[[int, float], None]] = None,
+        ) -> Optional[asyncio.Task[None]]:
+            logger.debug("Flushing HTTP transport")
+
+            if timeout > 0:
+                self._worker.submit(lambda: self._flush_client_reports(force=True))
+                return self._worker.flush(timeout, callback)  # type: ignore[func-returns-value]
+            return None
```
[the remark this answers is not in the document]
srothh: This calls submit, which is a sync function. The actual function _is_ awaited in the workers' _target function

srothh:
The lint error is in an unrelated file, not sure why it is happening

sentry_sdk/transport.py
srothh: I changed it to use `internal_sdk_error`

sentry_sdk/transport.py
srothh: Changed this to be an async coroutine; this also meant that I could go back to the worker and simplify the callbacks to only handle async without breaking it !

sentry_sdk/transport.py
srothh: Fixed!

sentry_sdk/transport.py
antonpirker: i guess this is ok. no reason really fits 100%. and as this should be an edge case. this is fine.

antonpirker (approved):
Looks good now! (the linting just fails because openfeature released something yesterday that breaks our linting. ignore this)

sentry_sdk/transport.py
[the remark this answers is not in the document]
srothh: The import throwing the error happens in the next PR, https://github.com/getsentry/sentry-python/pull/4615, apologies for the slightly unclean separation here

sentry_sdk/transport.py
```diff
@@ -17,18 +18,27 @@
...
+try:
     import h2  # noqa: F401
 
-    HTTP2_ENABLED = True
+    HTTP2_ENABLED = httpcore is not None
 except ImportError:
     HTTP2_ENABLED = False
 
+try:
+    ASYNC_TRANSPORT_ENABLED = httpcore is not None
```
sl0thentr0py, line 32: we need a new entry in `setup.py` similar to 
https://github.com/getsentry/sentry-python/blob/9171c5d213de7af02928426e74bb637782e8a5d1/setup.py#L62

sentry_sdk/transport.py
```diff
@@ -17,18 +18,27 @@
...
+except ImportError:
```
sl0thentr0py, line 33: we are not importing anything again, so this `except` is not needed

sentry_sdk/transport.py
```diff
@@ -224,9 +234,18 @@ def __init__(self: Self, options: Dict[str, Any]) -> None:
         elif self._compression_algo == "br":
             self._compression_level = 4
 
-    def _create_worker(self: Self, options: Dict[str, Any]) -> Worker:
-        # For now, we only support the threaded sync background worker.
-        return BackgroundWorker(queue_size=options["transport_queue_size"])
+    def _create_worker(self, options: dict[str, Any]) -> Worker:
```
sl0thentr0py, line 237: `_create_worker` in the shared class should be `NotImplemented` while the respective sync and async classes should implement only their own version.

sentry_sdk/transport.py
```diff
@@ -816,11 +1069,18 @@ def make_transport(options: Dict[str, Any]) -> Optional[Transport]:
     ref_transport = options["transport"]
 
     use_http2_transport = options.get("_experiments", {}).get("transport_http2", False)
-
+    use_async_transport = options.get("_experiments", {}).get("transport_async", False)
     # By default, we use the http transport class
-    transport_cls: Type[Transport] = (
-        Http2Transport if use_http2_transport else HttpTransport
-    )
+    if use_async_transport:
```
sl0thentr0py: nit on writing good nested/multiple conditionals, each branch should only be responsible for its own logic, so a bit cleaner the other way around

```python
    transport_cls = Http2Transport if use_http2_transport else HttpTransport
    if use_async_transport:
        try:
            asyncio.get_running_loop()
            transport_cls: Type[Transport] = AsyncHttpTransport
        except RuntimeError:
            # No event loop running, fall back to sync transport
            logger.warning("No event loop running, falling back to sync transport.")

```

sentry_sdk/transport.py
```diff
@@ -571,6 +590,240 @@ def flush(
             self._worker.flush(timeout, callback)
 
 
+if not ASYNC_TRANSPORT_ENABLED:
+    # Sorry, no AsyncHttpTransport for you
+    class AsyncHttpTransport(BaseHttpTransport):
+        def __init__(self: Self, options: Dict[str, Any]) -> None:
+            super().__init__(options)
+            logger.warning(
+                "You tried to use AsyncHttpTransport but don't have httpcore[asyncio] installed. Falling back to sync transport."
```
sl0thentr0py: let's gather all the required warnings in one place, so move this logger statement and check to `make_transport` below and then this can just become
```python
AsyncHttpTransport = BaseHttpTransport
```

sl0thentr0py (changes requested):
cosmetic review first, will review logic now

sentry_sdk/transport.py
srothh: https://github.com/getsentry/sentry-python/pull/4614#discussion_r2244791693 apologies for the slightly unclean separation in this case

sentry_sdk/transport.py
srothh: Done!

sentry_sdk/transport.py
srothh: Refactored it :)

sentry_sdk/transport.py
srothh: I named it async for now and also picked the 1.* dependency, as at that point the asyncio support became an extension in httpcore.

sentry_sdk/transport.py
```diff
@@ -571,6 +583,226 @@
...
+            self._worker.kill()
+            for task in self.background_tasks:
+                task.cancel()
+            self.background_tasks.clear()
+            try:
+                # Return the pool cleanup task so caller can await it if needed
+                return self.loop.create_task(self._pool.aclose())  # type: ignore
+            except RuntimeError:
+                logger.warning("Event loop not running, aborting kill.")
+                return None
```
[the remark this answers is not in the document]
srothh: I will remove the background task set, it should not be necessary anymore I think

sentry_sdk/transport.py
```diff
@@ -571,6 +583,222 @@ def flush(
...
+                    "timeout": {
+                        "pool": self.TIMEOUT,
+                        "connect": self.TIMEOUT,
+                        "write": self.TIMEOUT,
+                        "read": self.TIMEOUT,
+                    }
+                },
+            )
+
+        async def _flush_client_reports(self: Self, force: bool = False) -> None:
```
sl0thentr0py, line 668: think this does not need to be `async def` anymore and can be moved to shared?

sl0thentr0py: merging it in, we will take care of this when we do the mixin

sl0thentr0py (approved):

sentry_sdk/transport.py
srothh: This is async because otherwise the worker process_callback needs the case distinction between coroutines and sync functions.

See:

https://github.com/getsentry/sentry-python/pull/4591#discussion_r2228681154
https://github.com/getsentry/sentry-python/pull/4614#discussion_r2244830971

Issue #4601, opened 2025-07-21
Integrate async transport with SDK


srothh:
Integrate async transport to work with the current client and provide options to select it.

PR #4615, opened 2025-07-23
Integrate async transport with SDK

srothh:
Integrate the async transport with the rest of the SDK. Provide a new experimental option `transport_async` that enables the async transport if an event loop is running. Otherwise, fall back to the sync transport.

Furthermore, adapt the client to work with the async transport. To this end, flush and close were changed to be non blocking and awaitable in an async context to avoid deadlocks, however close enforces a completed flush before shutdown. As there are to my knowledge no background threads running flush/close, these methods are currently not thread-safe/loop-aware for async, which can be changed if necessary.

Atexit issue: The atexit integration used by the SDK runs after the event loop has already closed if asyncio.run() is used. This makes it impossible for the async flush to happen, as atexit calls client.close(), but a loop is no longer present. I attempted to apply [this fix](https://discuss.python.org/t/atexit-for-asyncio/13911) by patching the loop close in the asyncio integration, but I am unsure if I did it correctly/put it in the correct spot, or if this is a good idea. From my SDK test however, it seems to fix the flush issue. Note also that this will apparently be patched in Python 3.14, as per the discussion in the linked thread.

As a final note, I added event loop checking. Whenever the event loop is used, the transport/client catch RuntimeErrors, which would be thrown in case the event loop was already shut down. I am not sure if this is a case we need to consider, but I added it for now because I did not want the transport to potentially throw RuntimeError if the event loop is shutdown during a program. If this should be left out currently for simplicity, I can remove it again.

I added the [httpcore[asyncio] ](https://www.encode.io/httpcore/async/) dependency to requirements-testing, as it is needed for the async httpcore functionality.

GH-4601

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -995,9 +995,6 @@ class _Client(BaseClient):
         :param callback: Is invoked with the number of pending events and the configured timeout.
         """
         if self.transport is not None:
-            if timeout is None:
-                timeout = self.options["shutdown_timeout"]
-
             if isinstance(self.transport, AsyncHttpTransport) and hasattr(
                 self.transport, "loop"
             ):
@@ -1009,27 +1006,37 @@ class _Client(BaseClient):
                     logger.warning("Event loop not running, aborting flush.")
                     return None
             else:
-                self.session_flusher.flush()
-
-                if self.log_batcher is not None:
-                    self.log_batcher.flush()
-
-                self.transport.flush(timeout=timeout, callback=callback)
+                self._flush_sync(timeout, callback)
         return None
+
+    def _flush_sync(
+        self, timeout: Optional[float], callback: Optional[Callable[[int, float], None]]
+    ) -> None:
+        """Synchronous flush implementation."""
+        if timeout is None:
+            timeout = self.options["shutdown_timeout"]
+
+        self._flush_components()
+        if self.transport is not None:
+            self.transport.flush(timeout=timeout, callback=callback)
 
     async def _flush_async(
         self, timeout: Optional[float], callback: Optional[Callable[[int, float], None]]
     ) -> None:
-
+        """Asynchronous flush implementation."""
         if timeout is None:
             timeout = self.options["shutdown_timeout"]
 
+        self._flush_components()
+        if self.transport is not None:
+            flush_task = self.transport.flush(timeout=timeout, callback=callback)  # type: ignore
+            if flush_task is not None:
+                await flush_task
+
+    def _flush_components(self) -> None:
         self.session_flusher.flush()
         if self.log_batcher is not None:
             self.log_batcher.flush()
-        flush_task = self.transport.flush(timeout=timeout, callback=callback)  # type: ignore
-        if flush_task is not None:
-            await flush_task
 
     def __enter__(self) -> _Client:
         return self
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -982,7 +982,9 @@ class _Client(BaseClient):
                 return
             await self.flush_async(timeout=timeout, callback=callback)
             self._close_components()
-            await self.transport.kill()
+            kill_task = self.transport.kill()  # type: ignore
+            if kill_task is not None:
+                await kill_task
             self.transport = None
 
     def flush(
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -3,7 +3,6 @@ import os
 import uuid
 import random
 import socket
-import asyncio
 from collections.abc import Mapping
 from datetime import datetime, timezone
 from importlib import import_module
@@ -923,14 +922,6 @@ class _Client(BaseClient):
             raise ValueError("Integration has no name")
 
         return self.integrations.get(integration_name)
-
-    def _close_transport(self) -> Optional[asyncio.Task[None]]:
-        """Close transport and return cleanup task if any."""
-        if self.transport is not None:
-            cleanup_task = self.transport.kill()  # type: ignore
-            self.transport = None
-            return cleanup_task
-        return None
 
     def _close_components(self) -> None:
         """Kill all client components in the correct order."""
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -918,7 +918,7 @@ class _Client(BaseClient):
     def _close_transport(self) -> Optional[asyncio.Task[None]]:
         """Close transport and return cleanup task if any."""
         if self.transport is not None:
-            cleanup_task = self.transport.kill()
+            cleanup_task = self.transport.kill()  # type: ignore
             self.transport = None
             return cleanup_task
         return None
@@ -963,7 +963,9 @@ class _Client(BaseClient):
                         _flush_and_close(timeout, callback)
                     )
                 except RuntimeError:
+                    # Shutdown the components anyway
                     self._close_components()
+                    self._close_transport()
                     logger.warning("Event loop not running, aborting close.")
                     return None
                 # Enforce flush before shutdown
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -939,6 +939,7 @@ class _Client(BaseClient):
         async def _flush_and_close(
             timeout: Optional[float], callback: Optional[Callable[[int, float], None]]
         ) -> None:
+
             await self._flush_async(timeout=timeout, callback=callback)
             self._close_components()
 
@@ -996,6 +997,9 @@ class _Client(BaseClient):
     async def _flush_async(
         self, timeout: Optional[float], callback: Optional[Callable[[int, float], None]]
     ) -> None:
+
+        if timeout is None:
+            timeout = self.options["shutdown_timeout"]
 
         self.session_flusher.flush()
         if self.log_batcher is not None:
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -1023,8 +1023,6 @@ class _Client(BaseClient):
         self.session_flusher.flush()
         if self.log_batcher is not None:
             self.log_batcher.flush()
-
-        # For async transport, flush() returns a Task that needs to be awaited
         flush_task = self.transport.flush(timeout=timeout, callback=callback)  # type: ignore
         if flush_task is not None:
             await flush_task
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -1023,7 +1023,11 @@ class _Client(BaseClient):
         self.session_flusher.flush()
         if self.log_batcher is not None:
             self.log_batcher.flush()
-        await self.transport.flush_async(timeout=timeout, callback=callback)  # type: ignore
+
+        # For async transport, flush() returns a Task that needs to be awaited
+        flush_task = self.transport.flush(timeout=timeout, callback=callback)  # type: ignore
+        if flush_task is not None:
+            await flush_task
 
     def __enter__(self) -> _Client:
         return self
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -950,7 +950,7 @@ class _Client(BaseClient):
             else:
                 self.flush(timeout=timeout, callback=callback)
                 self._close_components()
-                return None
+        return None
 
     def flush(
         self,
@@ -967,6 +967,7 @@ class _Client(BaseClient):
         if self.transport is not None:
             if timeout is None:
                 timeout = self.options["shutdown_timeout"]
+            assert timeout is not None  # shutdown_timeout should never be None
 
             self.session_flusher.flush()
 
@@ -979,12 +980,12 @@ class _Client(BaseClient):
                 )
             else:
                 self.transport.flush(timeout=timeout, callback=callback)
-
-            return None
+        return None
 
     async def _flush_async(
-        self, timeout: float, callback: Optional[Callable[[int, float], None]]
+        self, timeout: Optional[float], callback: Optional[Callable[[int, float], None]]
     ) -> None:
+
         self.session_flusher.flush()
         if self.log_batcher is not None:
             self.log_batcher.flush()
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -915,6 +915,14 @@ class _Client(BaseClient):
 
         return self.integrations.get(integration_name)
 
+    def _close_transport(self) -> Optional[asyncio.Task[None]]:
+        """Close transport and return cleanup task if any."""
+        if self.transport is not None:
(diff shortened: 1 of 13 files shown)


sentry_sdk/client.py
```diff
@@ -952,12 +976,35 @@ def flush(
...
+                    logger.warning("Event loop not running, aborting flush.")
+                    return None
+            else:
+                self.session_flusher.flush()
 
             if self.log_batcher is not None:
                 self.log_batcher.flush()
 
-            self.transport.flush(timeout=timeout, callback=callback)
+                self.transport.flush(timeout=timeout, callback=callback)
```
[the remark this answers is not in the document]
srothh: Fixed

sentry_sdk/client.py
```diff
@@ -952,12 +975,36 @@ def flush(
...
+                try:
+                    return self.transport.loop.create_task(
+                        self._flush_async(timeout, callback)
+                    )
+                except RuntimeError:
+                    logger.warning("Event loop not running, aborting flush.")
+                    return None
+            else:
+                self.transport.flush(timeout=timeout, callback=callback)
+        return None
```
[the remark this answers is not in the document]
srothh: Fixed. The context manager issue I think is fine, as if anything an async context manager would need to be added to use the client in an async context.

sentry_sdk/client.py
```diff
@@ -914,36 +915,58 @@ def get_integration(
...
-            self.transport = None
+                try:
+                    flush_task = self.transport.loop.create_task(
+                        _flush_and_close(timeout, callback)
+                    )
+                except RuntimeError:
+                    logger.warning("Event loop not running, aborting close.")
+                    return None
+                # Enforce flush before shutdown
+                return flush_task
```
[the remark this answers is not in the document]
srothh: Fixed

sentry_sdk/integrations/asyncio.py
```diff
@@ -124,3 +125,46 @@ class AsyncioIntegration(Integration):
...
+            def _patched_close() -> None:
+                try:
+                    loop.run_until_complete(_flush())
+                finally:
+                    orig_close()
+
+            loop.close = _patched_close  # type: ignore
+            loop._sentry_flush_patched = True  # type: ignore
+
+        _patch_loop_close()
```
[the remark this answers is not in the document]
srothh: First issue: If init is called outside of an event loop, the transport falls back to the synchronous version anyway, so this should not be a problem.

Second issue: This is the same way the python discussion thread as well as [asyncio-atexit](https://github.com/minrk/asyncio-atexit/blob/main/asyncio_atexit.py) handle it, so it seems to be the standard way of circumventing this issue. From my quick manual test, it also did not lead to problems.

srothh:
> ### Bug: Async Cleanup Fails in Sync Context
> The `__exit__` method calls `self.close()` but does not await the `asyncio.Task` returned when `AsyncHttpTransport` is used. Since `__exit__` is synchronous, this prevents the async flush and cleanup operations from completing, leading to potential data loss or resource leaks when the client is used as a context manager.
> 
> `sentry_sdk/client.py#L1011-L1014`
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjM1YjgxZGE1LTQ0MDctNGVkMC1hMzNkLTgzOTQ4OWY2N2YyMiIsImVuY3J5cHRpb25LZXkiOiJENjNvRDZCcXNXeEc2LVFudkJSQ1FqLVF0YmhDOUdaamJLYWcyajNFbzNJIiwiYnJhbmNoIjoic3JvdGhoL2FzeW5jLXRyYW5zcG9ydC1pbnRlZ3JhdGlvbiJ9LCJpYXQiOjE3NTM0Mzg3NDgsImV4cCI6MTc1NDA0MzU0OH0.J6p47EGu6EdKZSb82nbLiiRKAKYDmdYV-weGC-JKRrcVHGdtkx63Uur3-mS-VGDRGVkDJgdwcWxcsPCEKgwUrqeT7_SybsajXF0ZYDSvHpbHlss99enAxvPp_AXXjD0bB0S_LeCjlIEejdENHH4axZcGfc-tuzrlRLWFwqV6pXHaHOkLBNLEg6AH0LbxkXzqsBrFq3xbOI_WxwCiJxhtBddcCbyrTQPTzAOED0NY6X7QqoYU7RfdDlu6U6wXTQsIwYqpBOMT-dNvnMm6zQbFwnSRNRB2ksbrHl5Np-KUDGCUvqW5dQewf5uYTuaqYx0vv2aDJ3UsuO7pp89p2mX8tQ) • [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjM1YjgxZGE1LTQ0MDctNGVkMC1hMzNkLTgzOTQ4OWY2N2YyMiIsImVuY3J5cHRpb25LZXkiOiJENjNvRDZCcXNXeEc2LVFudkJSQ1FqLVF0YmhDOUdaamJLYWcyajNFbzNJIiwiYnJhbmNoIjoic3JvdGhoL2FzeW5jLXRyYW5zcG9ydC1pbnRlZ3JhdGlvbiIsInJlcG9Pd25lciI6ImdldHNlbnRyeSIsInJlcG9OYW1lIjoic2VudHJ5LXB5dGhvbiIsInByTnVtYmVyIjo0NjE1LCJjb21taXRTaGEiOiI2YmRjMWZmMWJiNzhhMWMwN2U5ZjYxZjlkMzRjNGUzYzJkMzJmODU2In0sImlhdCI6MTc1MzQzODc0OCwiZXhwIjoxNzU0MDQzNTQ4fQ.ZuOxS2Z-ZQvZePny89Q4dSM9zidbZh3cx-4q6Gky37hTyZ9uWVm2hR8svxEyHJ03f1CZhGrLg6SK1EcNGeUQ9InmPWv8cVxAo4WswQ49pWat1YjkYcc0YzYV3Ne07S0scnneVZmlz1mqdLJBmbDfCRsbSxudVJTx3N-NNN2P3F_UpgkJ1S-u-nly_Hf6qiulwFOUQYM2j-V3DQ2dlKwVEEyB_kpFzcE-G3Nhyx7cH5imPg3EtcYJFy35nBscU2Q6MXq1hgzHHDZt7vGnk3Fd99Cll9K_YfyUP_NuviMpC8SFzW62Et6Qx2ENLVaeQ9iZwqVx07Hq4BNgh3CgCAjP0A)

I think this is not a problem. If this should work, I think it would need to implement a seperate async context manager.

sentry_sdk/client.py
```diff
@@ -952,12 +976,35 @@ def flush(
...
+        self, timeout: Optional[float], callback: Optional[Callable[[int, float], None]]
+    ) -> None:
+
+        if timeout is None:
+            timeout = self.options["shutdown_timeout"]
+
+        self.session_flusher.flush()
+        if self.log_batcher is not None:
+            self.log_batcher.flush()
+        await self.transport.flush_async(timeout=timeout, callback=callback)  # type: ignore
```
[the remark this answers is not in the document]
srothh: The close method ensures _close_components only ever runs after flushing has concluded.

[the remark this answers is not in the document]
srothh: 3) fixed, rest is adressed already

sentry_sdk/client.py
```diff
@@ -914,36 +915,73 @@ def get_integration(
...
-            self.transport.kill()
-            self.transport = None
+                try:
+                    flush_task = self.transport.loop.create_task(
+                        _flush_and_close(timeout, callback)
+                    )
+                except RuntimeError:
+                    self._close_components()
+                    logger.warning("Event loop not running, aborting close.")
+                    return None
```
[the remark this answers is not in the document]
srothh: Fixed

sentry_sdk/client.py
```diff
@@ -955,12 +997,39 @@ def flush(
...
+                except RuntimeError:
+                    logger.warning("Event loop not running, aborting flush.")
+                    return None
+            else:
+                self.session_flusher.flush()
+
+                if self.log_batcher is not None:
+                    self.log_batcher.flush()
+
+                self.transport.flush(timeout=timeout, callback=callback)
```
antonpirker: this made me think: is the session_flusher not flushed whene there is the async transport in use? (only later I saw that it is flushed in _flush_async.)

can we have a _flush_sync, so that it looks the same in the async and sync case and it easier to see that the same things are happening in async and sync?

sentry_sdk/consts.py
```diff
@@ -75,6 +75,7 @@ class CompressionAlgo(Enum):
             "transport_compression_algo": Optional[CompressionAlgo],
             "transport_num_pools": Optional[int],
             "transport_http2": Optional[bool],
+            "transport_async": Optional[bool],
```
antonpirker, line 78: 👍 so you can ignore my comment in the previous PR! :)

sentry_sdk/integrations/asyncio.py
```diff
@@ -124,3 +125,46 @@ class AsyncioIntegration(Integration):
     @staticmethod
     def setup_once() -> None:
         patch_asyncio()
+
+        def _patch_loop_close() -> None:
```
antonpirker: can we move this outside to the same level as `patch_asyncio`?

sentry_sdk/integrations/asyncio.py
```diff
@@ -124,3 +125,46 @@ class AsyncioIntegration(Integration):
...
+
+            if getattr(loop, "_sentry_flush_patched", False):
+                return
+
+            async def _flush() -> None:
+                client = sentry_sdk.get_client()
+                if not client:
+                    return
+                try:
+
```
antonpirker:
```suggestion

                try:
```

sentry_sdk/integrations/asyncio.py
```diff
@@ -124,3 +125,46 @@ class AsyncioIntegration(Integration):
...
+                    if not isinstance(client.transport, AsyncHttpTransport):
+                        return
+
+                    t = client.close()  # type: ignore
```
antonpirker: please choose a more descriptive name than `t`.

antonpirker (changes requested):
Some questsions

sentry_sdk/client.py
srothh: I think this is a great idea to make it more readable, I was anyways not quite happy with how there were differences like that in the branches! Thank you !

sentry_sdk/integrations/asyncio.py
srothh: I named it task

sentry_sdk/integrations/asyncio.py
srothh: Yes, fixed that! Also added some tests for this, as I noticed they were not included in the testsuite I made

srothh:
I am not sure why CI Lint is failing. The issue is in a completely unrelated file that the PR did not change

sentry_sdk/client.py
srothh: And refactored!

antonpirker: cool!

antonpirker (approved):
looks great now. 

ignore the linting problem, we need to fix this in master and potel-base branches.

sentry_sdk/client.py
```diff
@@ -917,51 +918,126 @@ def get_integration(
...
-            self.transport = None
+                try:
+                    flush_task = self.transport.loop.create_task(
+                        _flush_and_close(timeout, callback)
+                    )
+                except RuntimeError:
+                    # Shutdown the components anyway
+                    self._close_components()
+                    self._close_transport()
+                    logger.warning("Event loop not running, aborting close.")
```
sl0thentr0py:
```suggestion
                    logger.debug("Event loop not running, aborting close.")
```

sentry_sdk/client.py
```diff
@@ -917,51 +918,126 @@ def get_integration(
...
-            self.session_flusher.flush()
+            if isinstance(self.transport, AsyncHttpTransport) and hasattr(
+                self.transport, "loop"
+            ):
+                try:
+                    return self.transport.loop.create_task(
+                        self._flush_async(timeout, callback)
+                    )
+                except RuntimeError:
+                    logger.warning("Event loop not running, aborting flush.")
```
sl0thentr0py:
```suggestion
                    logger.debug("Event loop not running, aborting flush.")
```

sentry_sdk/client.py
srothh: I think these logging statements can be removed entirely with the sync/async seperation, as the async versions are coroutines and therefore do not need to rely on spawning a task, meaning they will not run into this error in the client.

sentry_sdk/client.py
srothh: See above

sentry_sdk/client.py
```diff
@@ -927,19 +950,43 @@ def close(
...
-
-            self.session_flusher.kill()
-
-            if self.log_batcher is not None:
-                self.log_batcher.kill()
-
-            if self.monitor:
-                self.monitor.kill()
-
+            self._close_components()
```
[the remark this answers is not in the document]
srothh: There is no way around this when separating the functions into two. If we were to handle the async case in close() itself, we would have the exact same situation as before.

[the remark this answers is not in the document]
srothh: However I do think an async context manager makes sense as a further addition.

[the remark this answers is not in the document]
sl0thentr0py: this is fine for now

sentry_sdk/client.py
```diff
@@ -917,6 +924,22 @@ def get_integration(
 
         return self.integrations.get(integration_name)
 
+    def _close_transport(self) -> Optional[asyncio.Task[None]]:
```
sl0thentr0py: this is no longer used

sl0thentr0py (approved):
one unused method, then we merge!

sentry_sdk/client.py
srothh: Thanks, missed that one. Is removed now

PR #4619, opened 2025-07-24
Fix `huggingface_hub` CI tests.

antonpirker:
HuggingFace has changed its default model used in `InferenceClient` and the new model breaks our CI. Change to use a non-existent model.


tests/integrations/huggingface_hub/test_huggingface_hub.py
```diff
@@ -15,8 +15,10 @@
 def mock_client_post(client, post_mock):
     # huggingface-hub==0.28.0 deprecates the `post` method
     # so patch `_inner_post` instead
-    client.post = post_mock
-    client._inner_post = post_mock
+    if hasattr(client, "post"):
```
antonpirker, line 18: This is just a small improvement that i did on the fly. Because in `huggingface-hub>=0.28.0` there is no `.post` attribute and it would be created in the tests.

PR #4638, opened 2025-07-29
How an upgraded `@trace` decorator might look like. 

antonpirker:


antonpirker:
this poc is not needed anymore.

PR #4648, opened 2025-07-30
feat(tracing): Improve `@trace` decorator. 

antonpirker:
Update the `@trace` decorator and make it more powerful. It accepts now the following parameters: `op`, `name`, `attributes`.

Example usage:
```python
import sentry_sdk
from sentry_sdk.consts import OP

# Simple usage (like before)
@sentry_sdk.trace
def process_data():
    # Function implementation
    pass

# With custom parameters
@sentry_sdk.trace(
    op=OP.DB_QUERY,
    name="Get user data",
    attributes={"postgres": True}
)
def make_db_query(sql):
    # Function implementation
    pass
```

This creates better DX for our users.

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -5,7 +5,7 @@ from datetime import datetime, timedelta, timezone
 from enum import Enum
 
 import sentry_sdk
-from sentry_sdk.consts import INSTRUMENTER, SPANSTATUS, SPANDATA, SpanTemplate
+from sentry_sdk.consts import INSTRUMENTER, SPANSTATUS, SPANDATA, SPANTEMPLATE
 from sentry_sdk.profiler.continuous_profiler import get_profiler_id
 from sentry_sdk.tracing_utils import create_span_decorator
 from sentry_sdk.utils import (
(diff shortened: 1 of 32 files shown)


sentry_sdk/tracing_utils.py
```diff
@@ -896,6 +830,92 @@ def _sample_rand_range(parent_sampled, sample_rate):
...
+            sync_wrapper.__signature__ = inspect.signature(f)  # type: ignore[attr-defined]
+        except Exception:
+            pass
+
+        if inspect.iscoroutinefunction(f):
+            return async_wrapper
+        else:
+            return sync_wrapper
+
+    return span_decorator
```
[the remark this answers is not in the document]
antonpirker: fixed.

sentry_sdk/tracing_utils.py
```diff
@@ -896,6 +830,96 @@ def _sample_rand_range(parent_sampled, sample_rate):
...
+                logger.debug(
+                    "Cannot create a child span for %s. "
+                    "Please start a Sentry transaction before calling this function.",
+                    qualname_from_function(f),
+                )
+                return await f(*args, **kwargs)
+
+            start_span_func = (
+                current_span.start_child if current_span else sentry_sdk.start_span
+            )
```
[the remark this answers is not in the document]
antonpirker: fixed.

sentry_sdk/tracing.py
```diff
@@ -1340,43 +1341,81 @@ def _set_initial_sampling_decision(self, sampling_context):
 if TYPE_CHECKING:
 
     @overload
-    def trace(func=None):
-        # type: (None) -> Callable[[Callable[P, R]], Callable[P, R]]
+    def trace(func=None, *, op=None, name=None, attributes=None):
+        # type: (Optional[Callable[P, R]], Optional[str], Optional[str], Optional[dict[str, Any]]) -> Callable[[Callable[P, R]], Callable[P, R]]
```
szokeasaurusrex: I'm pretty sure we want it like this, as I don't think it makes sense to pass the parameters and also the funciton.

```suggestion
    def trace(*, op=None, name=None, attributes=None):
        # type: (Optional[str], Optional[str], Optional[dict[str, Any]]) -> Callable[[Callable[P, R]], Callable[P, R]]
```

sentry_sdk/tracing.py
```diff
@@ -1340,43 +1341,81 @@ def _set_initial_sampling_decision(self, sampling_context):
...
         pass
 
     @overload
-    def trace(func):
-        # type: (Callable[P, R]) -> Callable[P, R]
+    def trace(func, *, op=None, name=None, attributes=None):
+        # type: (Callable[P, R], Optional[str], Optional[str], Optional[dict[str, Any]]) -> Callable[P, R]
```
szokeasaurusrex: For the same reason, I think we want this overload to only have the function

```suggestion
    def trace(func):
        # type: (Callable[P, R]) -> Callable[P, R]
```

sentry_sdk/tracing_utils.py
```diff
@@ -896,6 +830,88 @@ def _sample_rand_range(parent_sampled, sample_rate):
         return sample_rate, 1.0
 
 
+def create_span_decorator(op=None, name=None, attributes=None):
```
szokeasaurusrex, line 773: Could you please move this function above `get_current_span`? I think it might simplify the diff here

sentry_sdk/tracing_utils.py
```diff
@@ -896,6 +830,88 @@ def _sample_rand_range(parent_sampled, sample_rate):
...
+    # type: (Optional[str], Optional[str], Optional[dict[str, Any]]) -> Any
+    """
+    Create a span decorator that can wrap both sync and async functions.
+
+    :param op: The operation type for the span.
+    :param name: The name of the span.
+    :param attributes: Additional attributes to set on the span.
```
szokeasaurusrex, line 780: Do we intend to add more parameters in the future? How were these ones selected?

szokeasaurusrex (commented):
Added a few small suggestions, can do a more in depth review once they're addressed

sentry_sdk/tracing_utils.py
antonpirker: there will be a `template` parameter and thats it.

PR #4666, opened 2025-08-04
Let `set_data` accept a `dict`.

antonpirker:
This allows users to set multiple `span.data` attributes at once. 
In 3.x this then will become `set_attribute` and `set_attributes` (plural form).

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -598,9 +598,24 @@ class Span:
         # type: (str, Any) -> None
         self._tags[key] = value
 
-    def set_data(self, key, value):
-        # type: (str, Any) -> None
-        self._data[key] = value
+    def set_data(self, key=None, value=None):
+        # type: (Optional[Union[str, Dict[str, Any]]], Optional[Any]) -> None
+        """Set data on the span.
+        Can be called in two ways:
+        - set_data(key, value) - sets a single key-value pair
+        - set_data({"key": "value"}) - sets multiple key-value pairs from a dict
+        """
+        if key is None:
+            return
+
+        if isinstance(key, dict):
+            # Dictionary calling pattern: set_data({"key": "value"})
+            for k, v in key.items():
+                self._data[k] = v
+
+        elif isinstance(key, str):
+            # Traditional calling pattern: set_data(key, value)
+            self._data[key] = value
 
     def set_flag(self, flag, result):
         # type: (str, bool) -> None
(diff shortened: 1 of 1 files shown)


sentry_sdk/tracing.py
```diff
@@ -598,9 +598,24 @@ def set_tag(self, key, value):
...
+            return
+
+        if isinstance(key, dict):
+            # Dictionary calling pattern: set_data({"key": "value"})
+            for k, v in key.items():
+                self._data[k] = v
+
+        elif isinstance(key, str):
+            # Traditional calling pattern: set_data(key, value)
+            self._data[key] = value
```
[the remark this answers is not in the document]
szokeasaurusrex: I think these are also valid concerns and potential arguments in favor of creating a separate API for setting data via a dict

antonpirker:
True @szokeasaurusrex i have now changed it to have a `update_data()` function instead. Similar to pythons own `dict.update()`.

tests/tracing/test_misc.py
```diff
@@ -509,3 +509,34 @@ def test_transaction_not_started_warning(sentry_init):
         "The transaction will not be sent to Sentry. To fix, start the transaction by"
         "passing it to sentry_sdk.start_transaction."
     )
+
+
+def test_span_set_data(sentry_init, capture_events):
```
szokeasaurusrex: I would also add `update_data` to the name of the test. And, perhaps add a test which only checks `update_data` in isolation

sentry_sdk/tracing.py
```diff
@@ -602,6 +602,10 @@ def set_data(self, key, value):
         # type: (str, Any) -> None
         self._data[key] = value
 
+    def update_data(self, data):
+        # type: (Dict[str, Any]) -> None
+        self._data.update(data)
```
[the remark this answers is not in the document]
antonpirker: fixed.

tests/tracing/test_misc.py
antonpirker: done

pyproject.toml
```diff
@@ -66,6 +66,7 @@ strict_optional = true
 warn_redundant_casts = true
 warn_unused_configs = true
 warn_unused_ignores = true
+exclude = ["tests/"]
```
szokeasaurusrex: Why are we introducing this exclude line here? I am struggling to understand why it is relevant to the rest of the PR

tests/tracing/test_misc.py
```diff
@@ -509,3 +509,34 @@ def test_transaction_not_started_warning(sentry_init):
...
+    (event,) = events
+    span = event["spans"][0]
+
+    assert span["data"] == {
+        "key0": "value0",
+        "key1": "updated-value1",
+        "key2": "value2",
+        "key3": "value3",
+        "thread.id": mock.ANY,
+        "thread.name": mock.ANY,
```
szokeasaurusrex, line 541: Cool, TIL about `mock.ANY`

szokeasaurusrex (approved):
Nice, thanks!

pyproject.toml
antonpirker: ups

PR #4667, opened 2025-08-04
ref(clickhouse): List `send_data` parameters

szokeasaurusrex:
Explicitly list the `send_data` parameters in the wrapped function. The parameters are coming from [here](https://github.com/mymarilyn/clickhouse-driver/blob/8a4e7c5b99b532df2b015651d893a6f36288a22c/clickhouse_driver/client.py#L634). Continue also providing `*args` and `**kwargs`, but only for forwards-compatibility.

---


antonpirker (approved):
Nice!

PR #4668, opened 2025-08-04
Give users classes for easier manual instrumentation.

antonpirker:
We have internal classes for setting `Span.op` and to set correct names for `Span.data/Span.attribute` items. Make them public so users can use `SpanAttr` and `SpanOp` for better manual instrumentation.


antonpirker:
closing this, will not do this.

PR #4673, opened 2025-08-05
feat(tracing): Add convenience function `update_current_span`.

antonpirker:
Manually setting data on spans for our insights modules can be made easier. This PR introduces a convenience function.

Right now users need to do some like this:

```python
import sentry_sdk

span = sentry_sdk.get_current_span()
if span is not None:
    span.description = f"Some {dynamic} name"
    span.set_attribute("key1", "value1")
    span.set_attribute("key2", "value2")
```

With this new convenience function a user can do:
```python
import sentry_sdk

sentry_sdk.update_current_span(
    name=f"Some {dynamic} name",
    attributes={
        "key1": "value1",
        "key2": "value2",
    },
)

### DIFF
--- a/tests/tracing/test_misc.py
+++ b/tests/tracing/test_misc.py
@@ -540,3 +540,48 @@ def test_span_set_data_update_data(sentry_init, capture_events):
         "thread.id": mock.ANY,
         "thread.name": mock.ANY,
     }
+
+
+def test_update_current_span(sentry_init, capture_events):
+    sentry_init(traces_sample_rate=1.0)
+
+    events = capture_events()
+
+    with sentry_sdk.start_transaction(name="test-transaction"):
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/tracing/test_misc.py [hunk 1]
```diff
@@ -540,3 +540,48 @@ def test_span_set_data_update_data(sentry_init, capture_events):
...
+                    "key2": "value2",
+                },
+            )
+
+    (event,) = events
+    span = event["spans"][0]
+
+    assert span["op"] == "updated-span-op-2"
+    assert span["description"] == "updated-span-name-3"
+    assert span["data"] == {
```
sl0thentr0py, line 581: there's a helper in conftest called ApproxDict or something, use that

antonpirker: I do not like it, I like this way more because it is better readable

PR #4681, opened 2025-08-07
fix(django): Db query recording improvements.

antonpirker:
Improve how we record db query connection parameters. This reads the database connection configuration on integration startup from the Django `settings.py` file and caches them in a thread safe manner. 

When db operations are performed we read the connection data from the cache with a fallback to an improved version to the old logic. We first try to call `db.get_connection_params()` and if this also fails access the connection directly.

refs #4604

### DIFF
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -30,7 +30,10 @@ from sentry_sdk.integrations._wsgi_common import (
     RequestExtractor,
 )
 
+# Global cache for database configurations to avoid connection pool interference
 _cached_db_configs = {}  # type: Dict[str, Dict[str, Any]]
+_cache_initialized = False  # type: bool
+_cache_lock = threading.Lock()
 
 try:
     from django import VERSION as DJANGO_VERSION
@@ -621,35 +624,47 @@ def _cache_database_configurations():
     """
     Cache database configurations from Django settings to avoid connection pool interference.
     """
-    global _cached_db_configs
+    global _cached_db_configs, _cache_initialized
 
-    try:
-        from django.conf import settings
-        from django.db import connections
+    # Fast path: if already initialized, return immediately
+    if _cache_initialized:
+        return
 
-        for alias, db_config in settings.DATABASES.items():
-            if not db_config:  # Skip empty default configs
-                continue
+    # Slow path: acquire lock and check again (double-checked locking)
+    with _cache_lock:
+        if _cache_initialized:
+            return
 
-            with capture_internal_exceptions():
-                _cached_db_configs[alias] = {
-                    "db_name": db_config.get("NAME"),
-                    "host": db_config.get("HOST", "localhost"),
-                    "port": db_config.get("PORT"),
-                    "unix_socket": db_config.get("OPTIONS", {}).get("unix_socket"),
-                    "engine": db_config.get("ENGINE"),
-                }
+        try:
+            from django.conf import settings
+            from django.db import connections
 
-                db_wrapper = connections.get(alias)
-                if db_wrapper is None:
+            for alias, db_config in settings.DATABASES.items():
+                if not db_config:  # Skip empty default configs
                     continue
 
-                if hasattr(db_wrapper, "vendor"):
-                    _cached_db_configs[alias]["vendor"] = db_wrapper.vendor
+                with capture_internal_exceptions():
+                    _cached_db_configs[alias] = {
+                        "db_name": db_config.get("NAME"),
+                        "host": db_config.get("HOST", "localhost"),
+                        "port": db_config.get("PORT"),
+                        "unix_socket": db_config.get("OPTIONS", {}).get("unix_socket"),
+                        "engine": db_config.get("ENGINE"),
+                    }
 
-    except Exception as e:
-        logger.debug("Failed to cache database configurations: %s", e)
-        _cached_db_configs = {}
+                    db_wrapper = connections.get(alias)
+                    if db_wrapper is None:
+                        continue
+
+                    if hasattr(db_wrapper, "vendor"):
+                        _cached_db_configs[alias]["vendor"] = db_wrapper.vendor
+
+            # Mark as initialized only after successful completion
+            _cache_initialized = True
+
+        except Exception as e:
+            logger.debug("Failed to cache database configurations: %s", e)
+            _cached_db_configs = {}
 
 
 def install_sql_hook():
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -779,7 +779,7 @@ def _set_db_data(span, cursor_or_db):
             span.set_data(SPANDATA.SERVER_PORT, str(cached_config["port"]))
         if cached_config.get("unix_socket"):
             span.set_data(SPANDATA.SERVER_SOCKET_ADDRESS, cached_config["unix_socket"])
-        if cached_config.get("vendor"):
+        if cached_config.get("vendor") and span._data.get(SPANDATA.DB_SYSTEM) is None:
             span.set_data(SPANDATA.DB_SYSTEM, cached_config["vendor"])
 
         return  # Success - exit early
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -1,4 +1,3 @@
-import inspect
 import sys
 import threading
 import weakref
@@ -29,6 +28,9 @@ from sentry_sdk.integrations._wsgi_common import (
     DEFAULT_HTTP_METHODS_TO_CAPTURE,
     RequestExtractor,
 )
+
+# Global cache for database configurations to avoid connection pool interference
+_cached_db_configs = {}  # type: Dict[str, Dict[str, Any]]
 
 try:
     from django import VERSION as DJANGO_VERSION
@@ -155,6 +157,9 @@ class DjangoIntegration(Integration):
     def setup_once():
         # type: () -> None
         _check_minimum_version(DjangoIntegration, DJANGO_VERSION)
+
+        # Cache database configurations to avoid connection pool interference
+        _cache_database_configurations()
 
         install_sql_hook()
         # Patch in our custom middleware.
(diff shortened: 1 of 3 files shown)


sentry_sdk/tracing_utils.py
```diff
@@ -354,7 +354,13 @@ def _format_sql(cursor, sql):
...
             real_sql = cursor.mogrify(sql)
             if isinstance(real_sql, bytes):
-                real_sql = real_sql.decode(cursor.connection.encoding)
+                # Use UTF-8 as default, with latin1 fallback for edge cases
+                bytes_sql = real_sql  # make pypy happy (type narrowing to bytes)
+                try:
+                    real_sql = bytes_sql.decode("utf-8")
+                except UnicodeDecodeError:
+                    # If UTF-8 fails, try latin1 as fallback
+                    real_sql = bytes_sql.decode("latin1", errors="replace")
```
antonpirker: this prevents accessing `cursor.connection`.

antonpirker:
I used [this example benchmarking repo](https://github.com/slafs/sentry_perf) to see how these improvements do. Here are the results:

v1 = current `master`
v2 = this branches improvements. 

v1 vs v2 (v2 averaged across 3 runs)
Scenario | RPS v1 | RPS v2 mean (sd) | Δ RPS | TPR v1 ms | TPR v2 mean (sd) ms | Δ TPR | P95 v1 ms | P95 v2 mean (sd) ms | Δ P95
---------------------------------|--------|------------------|---------|-----------|---------------------|---------|-----------|----------------------|-------
Async, no Sentry | 2474.43| 2807.11 (55.53) | +13.5% | 40.41 | 35.63 (0.71) | -11.8% | 50.0 | 45.0 (1.7) | -10.0%
Async, with Sentry | 1255.40| 1388.96 (16.45) | +10.6% | 79.66 | 72.00 (0.85) | -9.6% | 104.0 | 89.3 (1.2) | -14.3%
Sync, no Sentry | 2091.47| 2273.36 (50.57) | +8.7% | 47.81 | 44.00 (0.97) | -8.0% | 58.0 | 56.0 (2.0) | -3.4%
Sync, with Sentry | 1124.00| 1267.03 (7.41) | +12.7% | 88.97 | 78.93 (0.46) | -11.3% | 102.0 | 97.7 (1.2) | -4.2%

Notes:
TPR = ApacheBench “Time per request” (mean).
P95 = 95th percentile latency from the AB percentile section.
Sentry overhead vs no Sentry (throughput drop)
What this shows: The relative throughput loss when enabling Sentry compared to the same setup without Sentry. Computed as (RPS_with_sentry / RPS_no_sentry - 1) × 100%. More negative = larger drop; closer to 0% = better (less overhead).

Mode | v1 drop | v2 drop (3-run mean)
------|---------|----------------------
Async | -49.3% | -50.5%
Sync | -46.2% | -44.3%

Bottom line: v2 is consistently faster. Throughput +8–14%, mean latency −8–12%, tail latency improves 3–14%.

sentry_sdk/integrations/django/__init__.py
```diff
@@ -614,6 +619,54 @@ def _set_user_info(request, event):
...
+    with _cache_lock:
+        if _cache_initialized:
+            return
+
+        try:
+            from django.conf import settings
+            from django.db import connections
+
+            for alias, db_config in settings.DATABASES.items():
+                if not db_config:  # Skip empty default configs
```
sentrivana, line 641: What happens with these, caching-wise? Will they always go through the slow fallback?

sentry_sdk/integrations/django/__init__.py
```diff
@@ -698,54 +753,121 @@ def connect(self):
...
+        if cached_config.get("db_name"):
+            span.set_data(SPANDATA.DB_NAME, cached_config["db_name"])
+        if cached_config.get("host"):
+            span.set_data(SPANDATA.SERVER_ADDRESS, cached_config["host"])
+        if cached_config.get("port"):
+            span.set_data(SPANDATA.SERVER_PORT, str(cached_config["port"]))
+        if cached_config.get("unix_socket"):
+            span.set_data(SPANDATA.SERVER_SOCKET_ADDRESS, cached_config["unix_socket"])
+        if cached_config.get("vendor") and span._data.get(SPANDATA.DB_SYSTEM) is None:
+            span.set_data(SPANDATA.DB_SYSTEM, cached_config["vendor"])
```
sentrivana, line 783: Can we move these to a separate function that just picks out the fields from a dictionary? Then we can use it here with `cached_config` as well as below with `connection_params`. Otherwise it's easy to miss that you have to update three places if you ever make a change to this

sentry_sdk/integrations/django/__init__.py
```diff
@@ -698,54 +753,121 @@ def connect(self):
...
+
+        return  # Success - exit early
+
+    # Fallback to dynamic database metadata collection.
+    # This is the edge case where db configuration is not in Django's `DATABASES` setting.
+    try:
+        # Fallback 1: Try db.get_connection_params() first (NO CONNECTION ACCESS)
```
sentrivana, line 790: Shouldn't we also cache the results of Fallback 1 and 2? Otherwise we keep potentially hitting the database in these cases

sentry_sdk/integrations/django/__init__.py
```diff
@@ -698,54 +753,121 @@ def connect(self):
...
+            if port:
+                span.set_data(SPANDATA.SERVER_PORT, str(port))
+
+            unix_socket = connection_params.get("unix_socket")
+            if unix_socket:
+                span.set_data(SPANDATA.SERVER_SOCKET_ADDRESS, unix_socket)
+            return  # Success - exit early to avoid connection access
+
+        except (KeyError, ImproperlyConfigured, AttributeError):
+            # Fallback 2: Last resort - direct connection access (CONNECTION POOL RISK)
```
sentrivana, line 818: This should definitely be cached IMO

sentry_sdk/integrations/django/__init__.py
```diff
@@ -698,54 +753,121 @@ def connect(self):
...
+        if cached_config.get("host"):
+            span.set_data(SPANDATA.SERVER_ADDRESS, cached_config["host"])
+        if cached_config.get("port"):
+            span.set_data(SPANDATA.SERVER_PORT, str(cached_config["port"]))
+        if cached_config.get("unix_socket"):
+            span.set_data(SPANDATA.SERVER_SOCKET_ADDRESS, cached_config["unix_socket"])
+        if cached_config.get("vendor") and span._data.get(SPANDATA.DB_SYSTEM) is None:
+            span.set_data(SPANDATA.DB_SYSTEM, cached_config["vendor"])
+
+        return  # Success - exit early
```
sentrivana, line 785: The comment is a bit superfluous
```suggestion
        return
```

sentry_sdk/integrations/django/__init__.py
```diff
@@ -698,54 +753,121 @@ def connect(self):
...
+                span.set_data(SPANDATA.SERVER_ADDRESS, host)
+
+            port = connection_params.get("port")
+            if port:
+                span.set_data(SPANDATA.SERVER_PORT, str(port))
+
+            unix_socket = connection_params.get("unix_socket")
+            if unix_socket:
+                span.set_data(SPANDATA.SERVER_SOCKET_ADDRESS, unix_socket)
+            return  # Success - exit early to avoid connection access
```
sentrivana, line 815: This return doesn't really do anything -- fine with me to leave it though to signal that we're done with this branch/function, but in that case at least the comment should be removed since it's wrong
```suggestion
            return
```

sentry_sdk/integrations/django/__init__.py
```diff
@@ -698,54 +753,121 @@ def connect(self):
...
+
+        except (KeyError, ImproperlyConfigured, AttributeError):
```
sentrivana, line 817: I don't think we can get a `KeyError` in Fallback 1 (unless it bubbles up from Django) since we're using `.get()` everywhere, which leads me to the question whether we have some cases where Fallback 1 "fails" silently without throwing an error (and just sets everything to `None`), and we should've gone to Fallback 2, but we won't, because there were no exceptions?

sentry_sdk/integrations/django/__init__.py
```diff
@@ -30,6 +30,11 @@
     RequestExtractor,
 )
 
+# Global cache for database configurations to avoid connection pool interference
```
sentrivana, line 33: What exactly are we solving here? I understand why caching is a good idea in general but not sure what exactly is meant by connection pool interference here

sentry_sdk/integrations/django/__init__.py
```diff
@@ -614,6 +619,54 @@ def _set_user_info(request, event):
...
+                            db_wrapper.vendor
+                            if db_wrapper and hasattr(db_wrapper, "vendor")
+                            else None
+                        ),
+                    }
+
+            _cache_initialized = True
+
+        except Exception as e:
+            logger.debug("Failed to cache database configurations: %s", e)
```
sentrivana, line 666: I'd remove these `logger.debug`s, they make async code slower even if debug mode is not on :(

sentry_sdk/integrations/django/__init__.py
```diff
@@ -698,54 +753,121 @@ def connect(self):
...
+        return  # Success - exit early
+
+    # Fallback to dynamic database metadata collection.
+    # This is the edge case where db configuration is not in Django's `DATABASES` setting.
+    try:
+        # Fallback 1: Try db.get_connection_params() first (NO CONNECTION ACCESS)
+        logger.debug(
+            "Cached db connection params retrieval failed for %s. Trying db.get_connection_params().",
+            db_alias,
+        )
```
sentrivana, line 794: Same comment re: logging and async code, would remove

sentry_sdk/integrations/django/__init__.py
```diff
@@ -698,54 +753,121 @@ def connect(self):
...
-        span.set_data(SPANDATA.SERVER_PORT, str(server_port))
+            unix_socket = connection_params.get("unix_socket")
+            if unix_socket:
+                span.set_data(SPANDATA.SERVER_SOCKET_ADDRESS, unix_socket)
 
-    server_socket_address = connection_params.get("unix_socket")
-    if server_socket_address is not None:
-        span.set_data(SPANDATA.SERVER_SOCKET_ADDRESS, server_socket_address)
+    except Exception as e:
+        logger.debug("Failed to get database connection params for %s: %s", db_alias, e)
```
sentrivana, line 869: Maybe this log we could keep since it's just emitted if there was an error

tests/integrations/django/test_db_query_data.py
```diff
@@ -524,3 +533,610 @@ def test_db_span_origin_executemany(sentry_init, client, capture_events):
...
+    ]
+
+    for call in expected_calls:
+        assert call in span.set_data.call_args_list
+
+    # Verify unix_socket was not set (it's None)
+    assert mock.call("server.socket.address", None) not in span.set_data.call_args_list
+
+
+def test_set_db_data_with_cached_config_unix_socket(sentry_init):
```
sentrivana, line 578: Why does setting `unix_socket` need a separate test case? This looks like the previous test case

tests/integrations/django/test_db_query_data.py
```diff
@@ -524,3 +533,610 @@ def test_db_span_origin_executemany(sentry_init, client, capture_events):
...
+    # Verify host and port were not set (they're None)
+    assert mock.call("server.address", None) not in span.set_data.call_args_list
+    assert mock.call("server.port", None) not in span.set_data.call_args_list
+
+
+def test_set_db_data_fallback_to_connection_params(sentry_init):
+    """Test _set_db_data falls back to db.get_connection_params() when cache misses."""
+    sentry_init(integrations=[DjangoIntegration()])
+
+    _cached_db_configs.clear()
```
sentrivana, line 621: This would be good as a fixture after every test case to ensure clean state

tests/integrations/django/test_db_query_data.py
```diff
@@ -1,12 +1,15 @@
 import os
```
sentrivana, line 1: We are mocking A LOT in these tests and they're very unit-testy, any way we can have some full end-to-end tests?

antonpirker:
This PR does not really solve the root problem of the issue we are trying to fix. This PR fixes the root problem: https://github.com/getsentry/sentry-python/pull/4804

PR #4686, opened 2025-08-11
chore(Crons): Log message for debugging Crons check-ins.

antonpirker:
A debug message to see what check-ins are send including the `monitor_slug` and the check-in `status`. 

If you set `sentry_sdk.init(.., debug=True)` then there will be a debug log message:
`"[Crons] Captured check-in ({check-in-id}): {monitor-slug} -> {status}"`

### DIFF
--- a/sentry_sdk/crons/api.py
+++ b/sentry_sdk/crons/api.py
@@ -56,7 +56,7 @@ def capture_checkin(
     sentry_sdk.capture_event(check_in_event)
 
     logger.debug(
-        f"[Crons] Captured check-in({check_in_event.get('check_in_id')}): {check_in_event.get('monitor_slug')} -> {check_in_event.get('status')}"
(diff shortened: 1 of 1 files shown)


sentry_sdk/crons/api.py
```diff
@@ -54,4 +55,8 @@ def capture_checkin(
 
     sentry_sdk.capture_event(check_in_event)
 
+    logger.debug(
+        f"[Crons] Captured check-in({check_in_event.get('check_in_id')}): {check_in_event.get('monitor_slug')} -> {check_in_event.get('status')}"
```
szokeasaurusrex: I think we are missing a space character

```suggestion
        f"[Crons] Captured check-in ({check_in_event.get('check_in_id')}): {check_in_event.get('monitor_slug')} -> {check_in_event.get('status')}"
```

szokeasaurusrex:
might be worth adding a test – although idk whether we typically test such things. If we don't, I suppose this is fine

PR #4688, opened 2025-08-11
ref(gnu-integration): make path optional

MeredithAnya:
I updated the GNU Integration in https://github.com/getsentry/sentry-python/pull/4598 but I didn't make the path optional so if we didn't have the path included, then the stacktrace didn't get parsed:

```python
# got parsed
"17. DB::TCPHandler::runImpl() @ 0x00000000121bb5d8 in /usr/bin/clickhouse"

# didn't get parsed
"17. DB::TCPHandler::runImpl() @ 0x00000000121bb5d8"
```

So updated it so that regardless of whether the path is present it will get parsed


sentrivana (approved):
Thank you!

PR #4708, opened 2025-08-18
ref(wsgi): Delete custom `_werkzeug` and use new Werkzeug version

mgaligniana:
Fixes: GH-3516


antonpirker:
Hey @mgaligniana thanks for your PR!

Unfortunately we can not merge this, because it would introduce Werkzeug as a dependency for our SDK and we do not add new dependencies. We want to have a minimal amount of dependencies in the SDK. 

This is why I will close this PR. 

What you could do: You could vendor in a newer version of Werkzeug if this makes WSGI integration better or easier to maintain.

mgaligniana:
@antonpirker Perfect, no problem! Thank you!

PR #4711, opened 2025-08-18
feat(langchain): support async invoke & stream in langchain integration

shellmayr:
- Support for ainvoke and astream in langchain integration

Closes TET-1019


shellmayr:
Closing. Improvements to the integration will be done in a separate PR, but so far it seems to work as expected after all.

PR #4713, opened 2025-08-19
fix(tracing): Do not attach stacktrace to transaction

Zylphrex:
The `attach_stacktrace` option was attaching stack traces to transactions. This is an expensive operation but the results aren't used anywhere.


Zylphrex:
@Mihir-Mavalankar This is the bug we talked about

sl0thentr0py (approved):
ty 🙏

PR #4722, opened 2025-08-26
Fix clickhouse logic for generator case

sl0thentr0py:
we need to call the original function first for `db_params` to be correctly populated in the generator case

(see diff without whitespace)


sentry_sdk/integrations/clickhouse_driver.py
```diff
@@ -146,40 +146,47 @@ def _inner_send_data(
         **kwargs: Any,
     ) -> Any:
         span = getattr(self.connection, "_sentry_span", None)
+        if span is None:
```
sl0thentr0py, line 149: moved this to an early return

PR #4736, opened 2025-09-02
fix(openai): Avoid double exit causing an unraisable exception

alexander-alderman-webb:
The function `_new_embeddings_create_common()` creates a span context with the `with` keyword, whereas `_new_responses_create_common()` and `_new_chat_completion_common()` manually call `__enter__` and set `finish_span=True`. Reusing the same `_capture_exception()` for all scenarios caused a double-exit in `_wrap_async_embeddings_create()`, which calls `_new_embeddings_create_common()`.

Context:

- Unraisable exceptions were present in our OpenAI tests. See https://github.com/getsentry/sentry-python/issues/4723.
- Tests fail in the (draft) unraisable exception integration because of unraisable exceptions in the the OpenAI integrations: https://github.com/getsentry/sentry-python/pull/4733

### DIFF
--- a/sentry_sdk/integrations/openai.py
+++ b/sentry_sdk/integrations/openai.py
@@ -78,12 +78,12 @@ class OpenAIIntegration(Integration):
         return 0
 
 
-def _capture_exception(exc):
-    # type: (Any) -> None
+def _capture_exception(exc, manual_span_cleanup=True):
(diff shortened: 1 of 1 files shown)


sentrivana (approved):
Thanks for fixing -- please also close https://github.com/getsentry/sentry-python/issues/4723 once this is merged

PR #4745, opened 2025-09-03
fix: Constrain types of ai_track decorator

alexander-alderman-webb:
The following code snippet now fails mypy: https://github.com/getsentry/sentry-python/issues/4663#issuecomment-3224040493

I followed how other functions in the SDK are typed. For example, other wrappers have the signature `(F) -> F` for a type variable `F`, although here the function can be async as well.

Closes https://github.com/getsentry/sentry-python/issues/4663


sentrivana (approved):
🔥

PR #4778, opened 2025-09-08
fix(integrations): make `langchain` package optional for `LangchainIntegration`

### LINKED ISSUE #4776: `LangchainIntegration` breaks if only `langchain-core` is installed
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.35.1, 2.37.0

### Steps to Reproduce

https://github.com/getsentry/sentry-python/pull/4678 has introduced changes that made `langchain` package  required to be installed in order to use `LangchainIntegration` whereas previously only the `langchain-core` was required. The integration, however, uses `langchain` package only to wrap the `AgentExecutor` methods, which is [anyway optional](https://github.com/getsentry/sentry-python/blob/567a91a774ac251000b1bfe6c48d468964893106/sentry_sdk/integrations/langchain.py#L76).

1. create a project with `langchain-core` and `sentry-sdk` packages
2. run `sentry_sdk.integrations.langchain`
3. See `DidNotEnable("langchain not installed")` exception

Or just run this command
```sh
uv run --with "sentry-sdk" --with "langchain-core" python -c "import sentry_sdk.integrations.langchain"
```

### Expected Result

When `langchain` package is not installed it just skips wrapping `AgentExecutor` methods in [`setup_once`](https://github.com/getsentry/sentry-python/blob/567a91a774ac251000b1bfe6c48d468964893106/sentry_sdk/integrations/langchain.py#L72-L78) if `langchain` package is not present

### Actual Result

```
  File "/Users/user/.cache/uv/archive-v0/LbXO3QDNdb9rlVPdSE8LR/lib/python3.12/site-packages/sentry_sdk/integrations/langchain.py", line 32, in <module>
    from langchain.agents import AgentExecutor
ModuleNotFoundError: No module named 'langchain'

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in <module>
  File "/Users/user/.cache/uv/archive-v0/LbXO3QDNdb9rlVPdSE8LR/lib/python3.12/site-packages/sentry_sdk/integrations/langchain.py", line 44, in <module>
    raise DidNotEnable("langchain not installed")
sentry_sdk.integrations.DidNotEnable: langchain not installed
```

mshavliuk:
Fixes #4776


mshavliuk:
Fixed in https://github.com/getsentry/sentry-python/pull/4783

PR #4783, opened 2025-09-09
fix(langchain): make new langchain integration work with just langchain-core

shellmayr:
- Catch ImportError for `langchain.agents` if langchain is not present and set AgentExecutor to None so the rest of the logic still keeps working
- Add test for recording new OTEL-compliant data for `langchain-core`
- Verified AI Agents functionality (span generation, token accounting, etc) with `langchain-core` and `langchain-openai`

Closes TET-1126
Closes PY-1833
Closes https://github.com/getsentry/sentry-python/issues/4776

### DIFF
--- a/sentry_sdk/integrations/langchain.py
+++ b/sentry_sdk/integrations/langchain.py
@@ -29,7 +29,6 @@ if TYPE_CHECKING:
 
 
 try:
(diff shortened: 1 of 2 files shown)


antonpirker (approved):
Nice! thanks for quick fix!

PR #4789, opened 2025-09-11
fix(typing): Make event processor types consistent

alexander-alderman-webb:
Some of the event processors in our integrations have a return type of `Event` when they never return `None`, while others keep the more general `Optional[Event]`. Returning `Event` where possible allows us to see at a glance that the event processors set by the integration cannot drop an event.

Also make use of the `EventProcessor` in a few more places.

### DIFF
--- a/sentry_sdk/integrations/asgi.py
+++ b/sentry_sdk/integrations/asgi.py
@@ -270,7 +270,7 @@ class SentryAsgiMiddleware:
             _asgi_middleware_applied.set(False)
 
     def event_processor(self, event, hint, asgi_scope):
-        # type: (Event, Hint, Any) -> Optional[Event]
+        # type: (Event, Hint, Any) -> Event
         request_data = event.get("request", {})
         request_data.update(_get_request_data(asgi_scope))
         event["request"] = deepcopy(request_data)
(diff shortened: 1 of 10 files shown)


sentry_sdk/integrations/asgi.py
```diff
@@ -270,7 +270,7 @@ async def _sentry_wrapped_send(event):
             _asgi_middleware_applied.set(False)
 
     def event_processor(self, event, hint, asgi_scope):
-        # type: (Event, Hint, Any) -> Optional[Event]
+        # type: (Event, Hint, Any) -> Event
```
alexander-alderman-webb, line 273: Is this method considered public because there's no leading underscore?

alexander-alderman-webb:
Why does an event processor need to return `Optional[Event]` if there are no code paths where the event is dropped from the specific event processor? Currently in the SDK we have event processors typed like this:

https://github.com/getsentry/sentry-python/blob/dcefe3840181e0aa534040b3932657f0ca1c36b8/sentry_sdk/integrations/fastapi.py#L114-L117

I'm happy to close the PR if it breaks anything since it's just meant to improve readability.

Issue #4764, opened 2025-09-04
Requests to fastapi-mcp hangs if Sentry is enabled


emirsavran:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.36.0

### Steps to Reproduce

1. Clone [the example repo](https://github.com/emirsavran/fastapi-mcp-sentry-repro)
2. Run `uv sync`
3. Add DSN to `with_sentry.py`
4. Run `uv run with_sentry.py`
5. On another terminal, run MCP inspector with `npx @modelcontextprotocol/inspector`
6. Select "Streamable HTTP" as transport type
7. Enter "http://localhost:8000/mcp" as URL
8. See unsuccessful request.

### Expected Result

MCP inspector should connect and show the tools like this:
<img width="1301" height="676" alt="Image" src="https://github.com/user-attachments/assets/700b355d-71cf-4ada-a34c-4d992b452501" />

You can also find the working version in `main.py`. Just follow the repro steps, but instead of number (4), run `uv run main.py`.

### Actual Result

Requests are somehow hanging. Here are the logs:

```
DEBUG:    2025-09-05 01:00:14   [mcp.server.lowlevel.server] Initializing server 'FastAPI'
DEBUG:    2025-09-05 01:00:14   [mcp.server.lowlevel.server] Registering handler for ListToolsRequest
DEBUG:    2025-09-05 01:00:14   [mcp.server.lowlevel.server] Registering handler for CallToolRequest
INFO:     2025-09-05 01:00:14   [fastapi_mcp.server] No auth config provided, skipping auth setup
INFO:     2025-09-05 01:00:14   [fastapi_mcp.server] No auth config provided, skipping auth setup
INFO:     2025-09-05 01:00:14   [fastapi_mcp.server] MCP HTTP server listening at /mcp
INFO:     2025-09-05 01:00:14   [fastapi_mcp.server] MCP HTTP server listening at /mcp
DEBUG:    2025-09-05 01:00:14   [asyncio] Using selector: KqueueSelector
INFO:     Started server process [71787]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
DEBUG:    2025-09-05 01:00:19   [fastapi_mcp.transport.http] Starting StreamableHTTP session manager
[...]

PR #4792, opened 2025-09-12
fix(fastapi): do not consume request body

alexander-alderman-webb:
You can mount any ASGI app to a FastAPI or Starlette route. You can also parse the request body manually using the `receive` callable on FastAPI's and Starlette's `Request` objects.

In either case, reading `json()`, `form()` or `body()` in the patched route handler consumes the request body stream, leaving no stream for the application to read. If the user relies on these high-level Starlette methods then there is no issue because Starlette caches the result. Combining our FastAPI or Starlette integrations with low-level ASGI request body parsing is problematic.

Approach in this PR:
- Attach the request body in the event processor, not in the route handler itself.
- If the user calls `json()`, `form()` or `body()`, cache the result on the ASGI scope and attach the result to events as before.
- Cache the raw request body on the ASGI scope. If none of the above methods are called try to parse the request body and attach the result to events. An exception could be raised before any of the above methods are invoked.


alexander-alderman-webb:
Closing for the reasons in https://github.com/getsentry/sentry-python/issues/4764#issuecomment-3319282311

PR #4794, opened 2025-09-12
fix(fastapi): do not read request body for fastapi_mcp

alexander-alderman-webb:
Calling `request.json()` reads the request body, and nothing is left in the `receive` callable that `fastapi_mcp` passes to the `mcp` package to read the request.

The fix involves checking if we have a `fastapi_mcp` endpoint. If so, do not read the request body.


alexander-alderman-webb:
Close because the solution should also address https://github.com/getsentry/sentry-python/issues/4827

PR #4804, opened 2025-09-16
fix(Django): Avoid evaluating complex Django object in span.data/span.attributes

antonpirker:
### Description
When rendering templates of the Django Admin in Django 5.0+ the template context includes also a `QuerySet` with lots of log items.
In our `serialize()` function this was not properly serialized leading to evaluating the `QuerySet` (read: running the SQL query). 

This change updates the `serialize()` to also correctly serialize `QuerySet`s in `span.data`

#### Issues
* resolves: #4604
* resolves: PY-1781

### DIFF
--- a/tests/integrations/django/test_basic.py
+++ b/tests/integrations/django/test_basic.py
@@ -958,6 +958,7 @@ def test_render_spans(sentry_init, client, capture_events, render_span_tree):
 
 
 @pytest.mark.skipif(DJANGO_VERSION < (1, 9), reason="Requires Django >= 1.9")
+@pytest.mark.forked
 @pytest_mark_django_db_decorator()
 def test_render_spans_queryset_in_data(sentry_init, client, capture_events):
     sentry_init(


tests/integrations/django/test_basic.py
```diff
@@ -956,6 +957,41 @@ def test_render_spans(sentry_init, client, capture_events, render_span_tree):
...
+    # so we expect the total time to be way less, because the complex context is not evaluated
+    assert end - begin < 10 / 5
+
+    # Make sure complex items are not put into the span.data
+    assert transaction["spans"][-1]["data"]["context"] == {
+        "user_age": 25,
+        "complex_dict": {
+            "a": 1,
+        },
+        "complex_list": [1, 2, 3],
```
sl0thentr0py: this is actually really bad because we're giving users an **incorrect** representation of what their data was.

The ideal way is to show `complex_list` as `[1, 2, 3,  "<lambda at #foobar>"]` which does not evaluate the lambda but still provides a meaningful repr.

The serializer already does this for QuerySet through:
https://github.com/getsentry/sentry-python/blob/7a905085087d5d11285037d609f6a91cec04b138/sentry_sdk/integrations/django/__init__.py#L244-L265

I'd try to figure out similar to [my fix earlier](https://github.com/getsentry/sentry-python/pull/4621) with a concrete repro why the QuerySet's repr is still getting invoked.

sl0thentr0py: [this comment](https://github.com/getsentry/sentry-python/issues/4606#issuecomment-3108338445) explains why our repr is bypassed, maybe something similar is happening in the performance case.

antonpirker: Cool thanks!

antonpirker: Here is my repro project: https://github.com/getsentry/testing-sentry-python/tree/main/test-django-admin-query-perf

antonpirker: made some changes

sl0thentr0py (approved):

PR #4823, opened 2025-09-19
fix(langgraph): add explicit exception handling

shellmayr:
- Adds explicit exception handling to the langgraph integration
- Set the span.status field correctly in case of error


shellmayr:
Closing - this was just a test

Issue #4827, opened 2025-09-19
Sentry FastAPI middleware hangs if user middleware raises an exception


hakanutku:
### How do you use Sentry?

Sentry Saas ([sentry.io](<http://sentry.io>))

### Version

2.38.0

### Steps to Reproduce

Sentry FastAPI middleware hangs if user middleware raises an exception. If I comment out `sentry_sdk.init()` call, I get the correct 422 response.

fastapi==0.115.12<br>starlette==0.46.2

```python
import gzip
import json
import typing
import zlib

import sentry_sdk
from fastapi import FastAPI, Request
from fastapi.testclient import TestClient
from sentry_sdk.integrations.fastapi import FastApiIntegration
from sentry_sdk.integrations.starlette import StarletteIntegration
from starlette import status
from starlette.datastructures import Headers
from starlette.exceptions import HTTPException
from starlette.types import ASGIApp, Message, Receive, Scope, Send


class GZipRequestMiddleware:
    def __init__(self, app: ASGIApp):
        self.app = app

    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
        if scope["type"] == "http":
            headers = Headers(scope=scope)

            if headers.get("Content-Encoding") == "gzip":
                reader = GZipRequestReader(self.app)
                await reader(scope, receive, send)
                return

        await self.app(scope, receive, send)


class GZipRequestReader:
    def __init__(self, app: ASGIApp) -> None:
        self.app = app
        self.receive: Receive = unattached_receive
        self.scope: Scope = {}
        self.actual_body_size = 0
        self.decompressor = zlib.decompressobj(16 + zlib.MAX_WBITS)

    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
        self.scope = scope
        self.receive = receive
        await self.app(scope, self._receive_gzipped_request, send)

    async def _receive_gzipped_request(self) -> Message:
        message = await self.receive()
        body = message.get("body", b"")
        if body:
            try:
                message["body"] = self.decompressor.decompress(body)
            except zlib.error as e:
                raise HTTPException(
                    status.HTTP_422_UNPROCESSABLE_ENTITY, "Compressed request body is malformed!"
                ) from e
            else:
                self.actual_body_size += len(message["body"])

        if not message.get("more_body", False):
            if not self.decompressor.eof:
                raise HTTPException(
[...]

PR #4832, opened 2025-09-22
fix: Read request body after handler so receive stream is not deprived

alexander-alderman-webb:
### Description

Ensure endpoint handler runs before the FastAPI or Starlette integrations read the request body.

As the integrations previously called high-level Starlette `Request` methods early, the underlying ASGI receive stream was consumed before the user's handler or middleware ran. While the `Request` object caches the request body, any handler or middleware that relies on low-level stream access still failed.

---
The event handler that sets `request.data`  is now registered after the HTTP request is handled. The `request.data` attribute may no longer be attached to events emitted while the request is processed. In particular, if

- The event is not an exception; or
- The event is an exception that is caught before reaching the routing layer we patch; or
- The event is an exception that is not caught but FastAPI/Starlette request body accessors are not invoked before the exception was thrown; or
- The event is the transaction generated by the FastAPI/Starlette integration for the HTTP request, and the user's request handler never calls FastAPI/Starlette request body accessors.

Note that `pydantic` also relies on FastAPI/Starlette accessors so `request.data` may still be available even if the user only implicitly uses the high-level accessors.

#### Issues

Closes https://github.com/getsentry/sentry-python/issues/4827
Closes PY-1851
Closes PY-1831

#### Reminders

### DIFF
--- a/sentry_sdk/integrations/graphene.py
+++ b/sentry_sdk/integrations/graphene.py
@@ -106,7 +106,8 @@ def _make_event_processor(source):
         if should_send_default_pii():
             request_info = event.setdefault("request", {})
             request_info["api_target"] = "graphql"
-            request_info.setdefault("data", {})["query"] = source  # type: ignore
+            if isinstance(source, str):
+                request_info.setdefault("data", {})["query"] = source  # type: ignore
 
         elif event.get("request", {}).get("data"):
             del event["request"]["data"]


tests/integrations/starlette/test_starlette.py
```diff
@@ -942,7 +1002,11 @@ def test_active_thread_id(sentry_init, capture_envelopes, teardown_profiling, en
 
 
 def test_original_request_not_scrubbed(sentry_init, capture_events):
```
alexander-alderman-webb, line 1095: The event associated with `logging.critical` in this test no longer has the request body attached. The test only passes now because I disabled the `LoggingIntegration`.

Due to reading the request body less eagerly in the integrations, the responsible event processor is registered after the logging event is captured.

I believe you would need to buffer the request body on the ASGI stream level to avoid the breaking change. I tried that approach in https://github.com/getsentry/sentry-python/pull/4792. There is a related discussion here: https://github.com/Kludex/starlette/discussions/1729#discussioncomment-3501302.

If there is another (and easier) way to avoid breaking users please let me know!

sentry_sdk/integrations/graphene.py
```diff
@@ -99,16 +99,21 @@ async def _sentry_patched_graphql_async(schema, source, *args, **kwargs):
...
-        request_info = event.setdefault("request", {})
-        request_info["api_target"] = "graphql"
+def _make_event_processor(source):
+    # type: (Any) -> Callable[[Event, dict[str, Any]], Event]
+    def _event_processor(event, hint):
+        # type: (Event, Dict[str, Any]) -> Event
+        if should_send_default_pii():
+            request_info = event.setdefault("request", {})
+            request_info["api_target"] = "graphql"
+            request_info["data"] = {"query": source}
```
alexander-alderman-webb: May need to change because the value associated with `event["request"]["data"]` was previously subject to request body size limits.

alexander-alderman-webb:
Closing for the reasons in https://github.com/getsentry/sentry-python/issues/4764#issuecomment-3319282311

alexander-alderman-webb:
Reopening because there may be the option of changing data fields on transactions in a minor version.

Users with our FastAPI integration are facing empty receive streams when using `fastapi_mcp` or a custom middleware.

sentry_sdk/integrations/graphene.py
alexander-alderman-webb: Following up on https://github.com/getsentry/sentry-python/pull/4832#discussion_r2406658898

sentry_sdk/integrations/starlette.py
```diff
@@ -448,33 +499,50 @@ async def _sentry_async_func(*args, **kwargs):
...
+                    raise exception
+
+                info = await extractor.extract_request_info()
                 sentry_scope.add_event_processor(
-                    _make_request_event_processor(request, integration)
+                    _make_request_body_event_processor(info)
                 )
 
-                return await old_func(*args, **kwargs)
+                return response
```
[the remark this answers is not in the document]
alexander-alderman-webb: This behaviour has not changed. It's too cumbersome to extract the request body in a synchronous context.

sentry_sdk/integrations/starlette.py
```diff
@@ -448,33 +499,50 @@ async def _sentry_async_func(*args, **kwargs):
...
+                sentry_scope._name = StarletteIntegration.identifier
 
-                def _make_request_event_processor(req, integration):
-                    # type: (Any, Any) -> Callable[[Event, dict[str, Any]], Event]
+                def _make_cookies_event_processor(cookies):
+                    # type: (Optional[Dict[str, Any]]) -> Callable[[Event, Dict[str, Any]], Event]
                     def event_processor(event, hint):
                         # type: (Event, Dict[str, Any]) -> Event
+                        if cookies and should_send_default_pii():
+                            event.get("request", {})["cookies"] = deepcopy(cookies)
```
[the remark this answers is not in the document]
alexander-alderman-webb: The tests show that `event.get("request", {})` can be assigned to, so I am not sure how this is temporary.

alexander-alderman-webb:
Too many conflicts, closing for now

PR #4836, opened 2025-09-22
fix(langchain): don't record tool call output if not include_prompt / should_send_default_pii

shellmayr:
- This PR makes sure we don't record tool calls as outputs in `gen_ai.chat` spans in langchain when users deselect the option to include prompts / send pii

Closes TET-1181

### DIFF
--- a/sentry_sdk/integrations/langchain.py
+++ b/sentry_sdk/integrations/langchain.py
@@ -322,14 +322,15 @@ class SentryLangchainCallback(BaseCallbackHandler):  # type: ignore[misc]
                     pass
 
                 try:
-                    tool_calls = getattr(generation.message, "tool_calls", None)
-                    if tool_calls is not None and tool_calls != []:
-                        set_data_normalized(
-                            span,
-                            SPANDATA.GEN_AI_RESPONSE_TOOL_CALLS,
-                            tool_calls,
-                            unpack=False,
-                        )
+                    if should_send_default_pii() and self.include_prompts:
+                        tool_calls = getattr(generation.message, "tool_calls", None)
+                        if tool_calls is not None and tool_calls != []:
+                            set_data_normalized(
+                                span,
+                                SPANDATA.GEN_AI_RESPONSE_TOOL_CALLS,
+                                tool_calls,
+                                unpack=False,
+                            )
                 except AttributeError:
                     pass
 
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/langchain.py
```diff
@@ -322,14 +322,15 @@ def on_llm_end(self, response, *, run_id, **kwargs):
...
                 try:
-                    tool_calls = getattr(generation.message, "tool_calls", None)
-                    if tool_calls is not None and tool_calls != []:
-                        set_data_normalized(
-                            span,
-                            SPANDATA.GEN_AI_RESPONSE_TOOL_CALLS,
-                            tool_calls,
-                            unpack=False,
-                        )
+                    if should_send_default_pii() and self.include_prompts:
```
alexander-alderman-webb, line 325: Why does it make sense to not include tool calls when `self.include_prompts=False`?

I feel like I am missing some context.

shellmayr: `self.include_prompts` controls whether the input/output of gen_ai calls and tool calls is recorded. Both inputs and outputs of models and tools may be something our users prefer not to send for business reasons. If we record the outputs of tool calls however, we may record more data than users expect us to

alexander-alderman-webb:
My one concern is that I have not heard tool arguments referred to as "prompts". Would users be confused if setting `include_prompts=False` also hides data from the tool calls?

We could consider more granular setting, especially since I am unsure how intuitive it is for tool call input and outputs to fall under the "prompt" umbrella.

But if that's more niche than we want to cater for, this looks good to go.

shellmayr:
>My one concern is that I have not heard tool arguments referred to as "prompts". Would users be confused if setting include_prompts=False also hides data from the tool calls?

@alexander-alderman-webb My thought on this is that a tool call may include raw user input, and that raw user input is a prompt, no? And in terms of behaviour, if a user specifies that they do not want to record that type of user input in general, we should stick to it throughout the SDK. Just to be clear: we still record 1. that tool calls happen and 2. which tool is being called - just not the user input associated with it.

antonpirker (approved):
Yes, looks good. The name `include_prompts` is not the best, you are right @alexander-alderman-webb but this is the name we have in other integrations. The functionality of it evolved over time. We probably should think of changing it but as this is a library we need to be careful with this to not break users code.

Issue #2527, opened 2023-11-23
Use `ruff` for linting/formatting


antonpirker:
Would be nice to get rid of `flake8` and `black` and only use `ruff` for all our linting/formatting. To make local development a bit easier and have all the configuration in one place.

`mypy` is still needed though, because `ruff` can not replace it.

Currently this is not possible, because `ruff` does not support type annotations in comments: https://github.com/astral-sh/ruff/issues/1619

So this will be only something we from type annotations in comments to proper type annotations.

PR #4866, opened 2025-09-26
ci: Replace `black` and `flake8` with `ruff`.

antonpirker:
### Description
Replace `black` and `flake8` with `ruff`.

With this change also reformat the code base (only minor changes, because the configuration is basically the same.) The commits for the reformatting are not included in `git blame` calls (see `.git-blame-ignore-revs`).

Also updates the `pre-commit` config so we have exactly the same linting/formatting in CI and local dev.

We now have:
- Less dependencies 
- Less configuration
- Faster

#### Issues
* resolves: #2527


sentrivana (approved):
Cool!

PR #4877, opened 2025-10-01
OTLPIntegration

### LINKED ISSUE #4935: Implement OtlpIntegration for Python SDK
auto-detection of OTel setup, trace-connected, previous PoC: [https://github.com/getsentry/sentry-python/pull/4877](https://github.com/getsentry/sentry-python/pull/4877)

🛳️  in a minor release

sl0thentr0py:
* setup external propagation context to fetch `trace_id/span_id` from otel
* setup otlp exporter with endpoint and headers taken from DSN 
* setup propagator
* exporter and propagator are setup by default, can be turned off in case people want to setup manually

### Issues
* resolves #4935
* resolves PY-1894


sl0thentr0py:
sentry error + sentry log + otlp trace

2025-10-31-162117_hyprshot

sentry_sdk/integrations/otlp.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, setup_otlp_exporter=True, setup_propagator=True):
+        # type: (bool, bool) -> None
+        self.setup_otlp_exporter = setup_otlp_exporter
+        self.setup_propagator = setup_propagator
+
+    @staticmethod
+    def setup_once():
+        # type: () -> None
+        logger.debug("[OTLP] Setting up trace linking for all events")
+        register_external_propagation_context(otel_propagation_context)
```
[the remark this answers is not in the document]
sl0thentr0py: that's a different thing

sl0thentr0py:
distributed tracing also works :tada: 

image

sentry_sdk/integrations/otlp.py
```diff
@@ -0,0 +1,82 @@
...
+        logger.debug("[OTLP] No TracerProvider configured by user, creating a new one")
+        tracer_provider = TracerProvider()
+        trace.set_tracer_provider(tracer_provider)
+
+    endpoint = None
+    headers = None
+    if dsn:
+        auth = Dsn(dsn).to_auth(f"sentry.python/{VERSION}")
+        endpoint = auth.get_api_url(EndpointType.OTLP_TRACES)
+        headers = {"X-Sentry-Auth": auth.to_header()}
```
alexander-alderman-webb, line 50: Why do we set two headers in the transport and only one here? See

https://github.com/getsentry/sentry-python/blob/659bd8415cfe8f9c04099c72042964fab8abfd40/sentry_sdk/transport.py#L325-L330

sentry_sdk/integrations/otlp.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    ctx = trace.get_current_span().get_span_context()
+
+    if ctx.trace_id == trace.INVALID_TRACE_ID or ctx.span_id == trace.INVALID_SPAN_ID:
+        return None
+
+    return (trace.format_trace_id(ctx.trace_id), trace.format_span_id(ctx.span_id))
+
+
+def setup_otlp_exporter(dsn=None):
```
alexander-alderman-webb, line 36: When `dsn is None`, is there any point in setting up the provider and processor?

Not sure if I am missing something because I am not as familiar with OTel. I would have thought if `dsn is None` we could exit early, maybe with a warning message.

tests/integrations/otlp/test_otlp.py
```diff
@@ -0,0 +1,143 @@
...
+        dsn="https://mysecret@bla.ingest.sentry.io/12312012",
+        integrations=[OTLPIntegration(setup_propagator=False)],
+    )
+
+    propagator = get_global_textmap()
+    assert not isinstance(propagator, SentryPropagator)
+    assert propagator is original_propagator
+
+
+def test_otel_propagation_context(sentry_init):
```
alexander-alderman-webb, line 130: When running this test I see the message below. I'm trying to find where this is coming from.

Failed to export span batch code: 404, reason: You should be really using o*.ingest.sentry.io domains.

alexander-alderman-webb (commented):
Looks good! A few questions just so I understand what's going on better as well.

sentry_sdk/integrations/otlp.py
sl0thentr0py: good question! user agent is already set by otlp, which I think is fine to distinguish it from a normal (sentry-traces-only) user
https://github.com/open-telemetry/opentelemetry-python/blob/f3b9e6ec96b7cc3717d5c8b529da7a8923810973/exporter/opentelemetry-exporter-otlp-proto-http/src/opentelemetry/exporter/otlp/proto/http/__init__.py#L77-L80

sentry_sdk/integrations/otlp.py
sl0thentr0py: they can still manually set env variables as follows: 

```
export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT="https://o0.ingest.sentry.io/api/0/integration/otlp/v1/traces"
export OTEL_EXPORTER_OTLP_TRACES_HEADERS="x-sentry-auth=sentry sentry_key=examplePublicKey"
```

which will be picked up by otel internally when initializing the exporter. Leaving it open as an option just in case someone has a complicated ops setup.

sl0thentr0py: we'll iterate on the DX here once we get some feedback/complaints/confusion

tests/integrations/otlp/test_otlp.py
sl0thentr0py: mocked

scripts/populate_tox/tox.jinja
```diff
@@ -43,6 +43,9 @@ envlist =
     # OpenTelemetry (OTel)
     {py3.7,py3.9,py3.12,py3.13,py3.14,py3.14t}-opentelemetry
 
+    # OpenTelemetry with OTLP
+    {py3.7,py3.9,py3.12,py3.13,py3.14,py3.14t}-otlp
```
alexander-alderman-webb: Last thing that'd be good before merging.

Since `pip install "opentelemetry-distro[otlp]"` also pulls in `grpcio`, and `grpcio` does not distribute free-threading wheels, these new tests take like 17 mins to run in CI on free-threading 3.14.

https://github.com/getsentry/sentry-python/actions/runs/19139690916/job/54701021577?pr=4877

So I suggest we do not run the new test suite on `3.14t`.

Also, as an fyi: for running `populate_tox.py`, I add a `uv` command for running it with the new pre-requisites to the README in https://github.com/getsentry/sentry-python/pull/5076/files

That said, in this case you can just edit `tox.jinja` and `tox.ini` manually, since it's just removing a single version.

sentrivana (approved):

PR #4885, opened 2025-10-03
fix(ai): truncate long message histories

shellmayr:


shellmayr:
These were done individually - closing.

PR #4915, opened 2025-10-10
fix: Replace asyncio.iscoroutinefunction with compatibility shim for Python 3.16

### LINKED ISSUE #4900: Replace asyncio.iscoroutinefunction
Around two years away, but eventually we will have to address the removal of `asyncio.iscoroutinefunction`. The inspect method was added in 3.5 but we should make sure there are no differences in behaviour across the versions we support. See https://docs.python.org/3/library/inspect.html

```
DeprecationWarning: 'asyncio.iscoroutinefunction' is deprecated and slated for removal in Python 3.16; use inspect.iscoroutinefunction() instead
```

Dibbu-cell:
### Description
This PR fixes the deprecation warning for asyncio.iscoroutinefunction  that will be removed in Python 3.16. The function was deprecated in Python 3.12 in favor of   inspect.iscoroutinefunction.
#### Issues

* resolves: #4900 

#### Reminders


sentry_sdk/ai/monitoring.py
```diff
@@ -89,7 +90,7 @@ async def async_wrapped(*args, **kwargs):
                         _ai_pipeline_name.set(None)
                     return res
 
-        if inspect.iscoroutinefunction(f):
+        if iscoroutinefunction(f):
```
alexander-alderman-webb: In places we're already using `inspect.iscoroutinefunction(f)` it's best to avoid change. Can you revert the changes in this file?

tests/integrations/httpx/test_httpx.py
```diff
@@ -32,7 +33,7 @@ def before_breadcrumb(crumb, hint):
     with start_transaction():
         events = capture_events()
 
-        if asyncio.iscoroutinefunction(httpx_client.get):
+        if iscoroutinefunction(httpx_client.get):
```
alexander-alderman-webb: We should be able to call `inspect.iscoroutinefunction()` directly here, since we control what runs in the tests.

sentry_sdk/tracing_utils.py
```diff
@@ -912,7 +913,7 @@ def sync_wrapper(*args, **kwargs):
         except Exception:
             pass
 
-        if inspect.iscoroutinefunction(f):
```
alexander-alderman-webb, line 915: Same here, no need for change where we already use `inspect.iscoroutinefunction()`.

sentry_sdk/integrations/google_genai/utils.py
```diff
@@ -318,7 +319,7 @@ def wrapped_tool(tool):
     tool_name = getattr(tool, "__name__", "unknown")
     tool_doc = tool.__doc__
 
-    if inspect.iscoroutinefunction(tool):
+    if iscoroutinefunction(tool):
```
alexander-alderman-webb: The change here should also be reverted.

alexander-alderman-webb (changes requested):
Thanks for the PR @Dibbu-cell.

I have requested some initial changes.

I had a look at other libraries and there are differences at how they detect coroutines across different Python versions. I will take a another look at the changes in `sentry_sdk`.

We would be in line with `asgiref` with your changes: https://github.com/django/asgiref/blob/f587b122af17bdba5749c30b96d2237bc1c2dfdf/asgiref/sync.py#L61-L69

Starlette does a simpler version check.
https://github.com/Kludex/starlette/blob/3912d6313730cc6004dfb4436e37dbc1a81db7c8/starlette/_utils.py#L11-L15

It's worth investigating if there could be any differences with the methods for detecting full inspect and asyncio compatibility for different Python version.

Dibbu-cell:
ok, i will do that.

sentry_sdk/_compat.py
```diff
@@ -15,6 +23,38 @@
...
+            # No marker available on this Python version; return function as-is.
+            return func
+
+        try:  # pragma: no cover - defensive
+            func._is_coroutine = marker  # type: ignore[attr-defined]
+        except Exception:
+            # If assignment fails for any reason, leave func unchanged.
+            pass
+        return func
+
```
alexander-alderman-webb: Can you move this logic back to `django/asgi.py`. This would be better for two reasons:

- We don't use `markcoroutinefunction` anywhere else, so it's better to keep it close to where it's used.
- To be safe we should use the most conservative version check. See my other comment on this file.

sentry_sdk/_compat.py
```diff
@@ -15,6 +23,38 @@
 PY311 = sys.version_info[0] == 3 and sys.version_info[1] >= 11
```
alexander-alderman-webb, line 26: Switching `asyncio.iscoroutinefunction` for `inspect.iscoroutinefunction` is tricky because of historical incompatibility. Since we support Python versions 3.6 and up we have to care about this unfortunately.

We should use the most conservative cutoff to not break any users. So while CPython deprecated `asyncio.iscoroutinefunction` in version 3.12, we should probably mirror Starlette who use 3.13 because of a bug in standard libraries which have not been backported. See https://github.com/Kludex/starlette/discussions/2983

Also, since we vendor `_is_async_callable`, the change below would keep the vendor up to date. We can therefore be more confident that we won't run into edge cases, since users of `starlette` would have run into them already. 

```suggestion
PY313 = sys.version_info[0] == 3 and sys.version_info[1] >= 13
iscoroutinefunction = inspect.iscoroutinefunction if PY313 else asyncio.iscoroutinefunction
```

sentry_sdk/integrations/asgi.py
```diff
@@ -76,10 +77,10 @@ def _looks_like_asgi3(app):
     if inspect.isclass(app):
         return hasattr(app, "__await__")
     elif inspect.isfunction(app):
-        return asyncio.iscoroutinefunction(app)
+        return iscoroutinefunction(app)
```
alexander-alderman-webb: No action required, just documenting.

This logic is vendored from `uvicorn` and they recently replaced `asyncio.iscoroutinefunction` with `inspect.iscoroutinefunction`. See https://github.com/Kludex/uvicorn/pull/2659

Since we will check for Python 3.13 and up and `uvicorn` seems to be happy using `inspect` on 3.9 and up, we can rely on `uvicorn` having battle testing this change as well.

alexander-alderman-webb (changes requested):
Thanks for responding to my review @Dibbu-cell.

I've added two more changes we should implement to ensure we don't break anyone. Looks good otherwise!

Dibbu-cell:
I have done the appropriate changes.

alexander-alderman-webb:
It looks like some unrelated changes snuck in with your most recent commit @Dibbu-cell. Can you please revert everything unrelated to changing `iscoroutinefunction`? There are many ways to do this but here's a one-liner you can use on the terminal:

```bash
git checkout master -- sentry_sdk/integrations/django/asgi.py sentry_sdk/client.py tests/tracing/test_allow_n_plus_one.py
```

Also please revert `sentry_sdk/integrations/django/asgi.py` since I don't see any reason to change the detection of coroutine functions in the Django integration.

Thanks!

Dibbu-cell:
I have done the changes  and sorry for troubling you.

Dibbu-cell:
I am new in open Source .

Dibbu-cell:
check if there is problem let me know.

alexander-alderman-webb:
Hi @Dibbu-cell,

Don't worry, we're open to accept contributions! 

It looks like you've reverted too many of your changes. I can only see a change on `sentry_sdk/_compat.py` now.

Since the issue you've tackled affects many integrations and you're new to open source I suggest we narrow the scope. Changes to the integrations also have more edge cases to consider so are more complicated than I initially anticipated.

Could you **only** change `tests/integrations/httpx/test_httpx.py` by replacing `asyncio.iscoroutinefunction` with `inspect.iscoroutinefunction`, **and** leave everything else unchanged.

alexander-alderman-webb:
Closing as this has been completed with 

- https://github.com/getsentry/sentry-python/commit/5800f4320b37bcefa4994e5f8357a45aac265871
- https://github.com/getsentry/sentry-python/commit/750a5688c86e0677528d925ef43537f152a8c6da

PR #4946, opened 2025-10-16
fix(ai): introduce message truncation for openai

shellmayr:
- Splitting out introduction of the message truncation from https://github.com/getsentry/sentry-python/pull/4885
- Introduce for openai

Contributes to TET-1208

### DIFF
--- a/sentry_sdk/ai/utils.py
+++ b/sentry_sdk/ai/utils.py
@@ -110,7 +110,7 @@ def _find_truncation_index(messages, max_bytes):
     """
     running_sum = 0
     for idx in range(len(messages) - 1, -1, -1):
-        size = len(json.dumps(messages[idx], separators=(",", ":")))
+        size = len(json.dumps(messages[idx], separators=(",", ":")).encode("utf-8"))
         running_sum += size
         if running_sum > max_bytes:
             return idx + 1


sentry_sdk/ai/utils.py
```diff
@@ -1,14 +1,21 @@
...
+
     from sentry_sdk.tracing import Span
 
+from typing import TYPE_CHECKING
+
 import sentry_sdk
 from sentry_sdk.utils import logger
 
+if TYPE_CHECKING:
+    from typing import Any, Dict, List, Optional
```
alexander-alderman-webb: Can we consolidate the types here with the existing `TYPE_CHECKING` block?

sentry_sdk/ai/utils.py
```diff
@@ -95,3 +102,79 @@ def get_start_span_function():
...
+
+    serialized_json = json.dumps(truncated_messages, separators=(",", ":"))
+    current_size = len(serialized_json.encode("utf-8"))
+
+    if current_size > max_bytes and len(truncated_messages) == 1:
+        message = truncated_messages[0].copy()
+        content = message.get("content", "")
+
+        if isinstance(content, str):
+            max_content_length = max_bytes // 2
```
alexander-alderman-webb: Where does the divide by 2 come from here?

sentry_sdk/ai/utils.py
```diff
@@ -95,3 +102,79 @@ def get_start_span_function():
...
+            while True:
+                message["content"] = content[:max_content_length]
+                test_json = json.dumps([message], separators=(",", ":"))
+                if len(test_json.encode("utf-8")) <= max_bytes:
+                    break
+                max_content_length = int(max_content_length * 0.9)
+                if max_content_length < 100:
```
alexander-alderman-webb: The `0.9` and `100` also seem like magic numbers. We can keep them but imo it would be good to add a comment saying why we've set them to these values, or just adding a comment to remark that they're arbitrary.

sentry_sdk/ai/utils.py
shellmayr: Oh weird, this is an oversight - I want to do the truncation of many messages only, not the size of single messages at this point, this should not have been in here, I updated it.

tests/integrations/openai/test_openai.py
```diff
@@ -1505,3 +1506,55 @@ def test_openai_message_role_mapping(sentry_init, capture_events):
...
+        meta_path = event["_meta"]
+        if (
+            "spans" in meta_path
+            and "0" in meta_path["spans"]
+            and "data" in meta_path["spans"]["0"]
+        ):
+            span_meta = meta_path["spans"]["0"]["data"]
+            if SPANDATA.GEN_AI_REQUEST_MESSAGES in span_meta:
+                messages_meta = span_meta[SPANDATA.GEN_AI_REQUEST_MESSAGES]
+                assert "len" in messages_meta.get("", {})
```
alexander-alderman-webb: The purpose if statements are unclear to me. If you rely on this data being set we should test that it actually is set. Otherwise, we may never hit the assert because we leave the test due to not satisfying the conditions.

```suggestion
    span_meta = event["_meta"]["spans"]["0"]["data"]
    if SPANDATA.GEN_AI_REQUEST_MESSAGES in span_meta:
        messages_meta = span_meta[SPANDATA.GEN_AI_REQUEST_MESSAGES]
        assert "len" in messages_meta.get("", {})
```

sentry_sdk/scope.py
```diff
@@ -188,6 +188,7 @@ class Scope:
         "_extras",
         "_breadcrumbs",
         "_n_breadcrumbs_truncated",
+        "_gen_ai_messages_truncated",
```
alexander-alderman-webb, line 191: On a high-level, it's unclear to me why we create this dictionary on the scope.

At the place we truncate the messages, we have access to the span object. Can we just set the truncation metadata directly on the span?

alexander-alderman-webb: For reference, we discussed that adding the dictionary on the scope is preferred to not interfere with callback registered by users.

sentrivana (approved):

PR #4950, opened 2025-10-16
fix(ai): add message truncation in langchain

shellmayr:
- Add gen_ai.request.message truncation to langchain as introduced in https://github.com/getsentry/sentry-python/pull/4946

Contributes to TET-1208

### DIFF
--- a/tests/integrations/langchain/test_langchain.py
+++ b/tests/integrations/langchain/test_langchain.py
@@ -1,3 +1,4 @@
+import json
 from typing import List, Optional, Any, Iterator
 from unittest import mock
 from unittest.mock import Mock, patch
@@ -884,8 +885,6 @@ def test_langchain_message_role_mapping(sentry_init, capture_events):
 
             # Parse the message data (might be JSON string)
             if isinstance(messages_data, str):
-                import json
-
                 try:
                     messages = json.loads(messages_data)
                 except json.JSONDecodeError:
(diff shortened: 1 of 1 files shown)


tests/integrations/langchain/test_langchain.py
```diff
@@ -958,3 +958,70 @@ def test_langchain_message_role_normalization_units():
...
+    assert SPANDATA.GEN_AI_REQUEST_MESSAGES in llm_span["data"]
+
+    messages_data = llm_span["data"][SPANDATA.GEN_AI_REQUEST_MESSAGES]
+    assert isinstance(messages_data, str)
+
+    import json
+
+    parsed_messages = json.loads(messages_data)
+    assert isinstance(parsed_messages, list)
+    assert len(parsed_messages) <= len(prompts)
```
alexander-alderman-webb: This is a very trivial check. Correct me if I'm wrong but I would expect this to pass even without truncation. Can we mirror the corresponding openai test here?

shellmayr: Modified the test to assert on the messages that are supposed to remain, and the meta tag

alexander-alderman-webb (approved):

tests/integrations/langchain/test_langchain.py
```diff
@@ -958,3 +957,77 @@ def test_langchain_message_role_normalization_units():
...
+
+    messages_data = llm_span["data"][SPANDATA.GEN_AI_REQUEST_MESSAGES]
+    assert isinstance(messages_data, str)
+
+    parsed_messages = json.loads(messages_data)
+    assert isinstance(parsed_messages, list)
+    assert len(parsed_messages) == 2
+    assert "small message 4" in str(parsed_messages[0])
+    assert "small message 5" in str(parsed_messages[1])
+    assert tx["_meta"]["spans"]["0"]["data"]["gen_ai.request.messages"][""]["len"] == 5
```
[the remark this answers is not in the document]
alexander-alderman-webb: Yes but that's a broader issue about writing tests. I don't think that should block this PR.

PR #5030, opened 2025-10-24
fix(integrations): ensure that GEN_AI_AGENT_NAME is properly set for GEN_AI spans under an invoke_agent span

constantinius:
### Description

chat and execute tool spans under an invoke agent span should have the agent name property set

#### Issues
Closes https://linear.app/getsentry/issue/TET-1293/make-sure-that-agent-name-is-set-on-all-of-its-gen-ai-children

### DIFF
--- a/tests/integrations/langchain/test_langchain.py
+++ b/tests/integrations/langchain/test_langchain.py
@@ -201,20 +201,17 @@ def test_langchain_agent(
     # We can't guarantee anything about the "shape" of the langchain execution graph
     assert len(list(x for x in tx["spans"] if x["op"] == "gen_ai.chat")) > 0
 
-    assert "gen_ai.usage.input_tokens" in chat_spans[0]["data"]
-    assert "gen_ai.usage.output_tokens" in chat_spans[0]["data"]
-    assert "gen_ai.usage.total_tokens" in chat_spans[0]["data"]
+    # Token usage is only available in newer versions of langchain (v0.2+)
+    # where usage_metadata is supported on AIMessageChunk
+    if "gen_ai.usage.input_tokens" in chat_spans[0]["data"]:
+        assert chat_spans[0]["data"]["gen_ai.usage.input_tokens"] == 142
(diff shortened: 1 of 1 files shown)


sentrivana (commented):
I took the google-genai part out of this PR and merged it separately since it was straightforward: https://github.com/getsentry/sentry-python/pull/5038

Regarding the Langchain part, we talked about this and I also thought about it a bit more. Documenting my reasoning and the proposed path forward for this sort of pattern here.

The problem: We need access to some package-internal object (in this case the agent) to be able to extract something from it in different places in the integration.

In this PR, we're opting for storing it on the current scope in the `_contexts`. I'm unhappy about this since we're misusing a field whose purpose is something entirely different (populating contexts on events) to store something integration specific.

Other solutions that have been proposed:
- Store it on something coming from Langchain directly that gets passed around and is accessible everywhere we need it. This is the usual way of doing things, but we couldn't find a good candidate here.
- Store it on the parent transaction and access it from there. The problem here is there might be different agents in a transaction.
- Store it on a parent span and access it from here. But there is no good way of traversing the span tree.

**TL;DR: recommended way forward:**

My recommended solution would be to store this in a `ContextVar` on the top level of the integration. That way, we have the same execution context isolation that we have with scopes (they're also context vars), but we're not misusing `_contexts`, and we're keeping the change contained to the integration.

Alternatively, can this be filled in in relay? I.e., we set the agent name on the topmost span that makes sense, and relay copies it to all child spans where it makes sense.

constantinius:
I talked to the people at Relay and it seems to be feasible to put it in the [place where the span-buffer is evaluated](https://github.com/getsentry/sentry/blob/f8d701b742f96c91b6c9bf1c35244cd64805df25/src/sentry/spans/consumers/process_segments/enrichment.py#L69-L70) (not Relay itself).

So, I think this is a viable solution next to the `ContextVar` approach.

constantinius (commented):
Alright, this is now implemented using `ContextVars`.

PR #5051, opened 2025-10-31
Add external_propagation_context support

sl0thentr0py:
If we are on an external tracing system like otel, we allow registering a new source of `trace_id/span_id` that takes precedence over the scope's propagation_context.

* Also reworked logs and metrics to use `get_trace_context`
* Cleaned up handling of `get_trace_context` that is still messy but a bit more clearer now regarding which underlying `propagation_context` is used

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -107,6 +107,10 @@ _current_scope = ContextVar("current_scope", default=None)
 
 global_event_processors = []  # type: List[EventProcessor]
 
+# A function returning a (trace_id, span_id) tuple
+# from an external tracing source (such as otel)
+_external_propagation_context_fn = None  # type: Optional[Callable[[], Optional[Tuple[str, str]]]]
+
 
 class ScopeType(Enum):
     CURRENT = "current"
@@ -140,6 +144,25 @@ class _ScopeManager:
 def add_global_event_processor(processor):
     # type: (EventProcessor) -> None
     global_event_processors.append(processor)
+
+
+def register_external_propagation_context(fn):
+    # type: (Callable[[], Optional[Tuple[str, str]]]) -> None
+    global _external_propagation_context_fn
+    _external_propagation_context_fn = fn
+
+
+def remove_external_propagation_context():
+    # type: () -> None
+    global _external_propagation_context_fn
+    _external_propagation_context_fn = None
+
+
+def get_external_propagation_context():
+    # type: () -> Optional[Tuple[str, str]]
+    return (
+        _external_propagation_context_fn() if _external_propagation_context_fn else None
+    )
 
 
 def _attr_setter(fn):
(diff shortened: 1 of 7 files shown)


sentry_sdk/scope.py
```diff
@@ -1438,10 +1469,7 @@ def _apply_contexts_to_event(self, event, hint, options):
 
         # Add "trace" context
         if contexts.get("trace") is None:
-            if has_tracing_enabled(options) and self._span is not None:
```
sl0thentr0py, line 1441: moved these into `get_trace_context` for consistency

sentry_sdk/scope.py
```diff
@@ -562,21 +585,29 @@ def get_baggage(self, *args, **kwargs):
...
-        }  # type: Dict[str, Any]
+        # if we are tracing externally (otel), those values take precedence
+        external_propagation_context = get_external_propagation_context()
+        if external_propagation_context:
+            trace_id, span_id = external_propagation_context
+            return {"trace_id": trace_id, "span_id": span_id}
 
-        return trace_context
+        propagation_context = self.get_active_propagation_context()
+        if propagation_context is None:
```
sl0thentr0py, line 602: propagation_context is actually never None, but this implementation is whack so we need this redundant check...

sl0thentr0py: for some future cleanup, if you see `get_active_propagation_context` that's also really shady. I generally dislike that so many **instance** methods on this scope call and fetch other scopes themselves, it's entirely spaghetti but somehow works in the common code paths.

sentry_sdk/scope.py
```diff
@@ -142,6 +146,25 @@ def add_global_event_processor(processor):
...
+    # type: () -> None
+    global _external_propagation_context_fn
+    _external_propagation_context_fn = None
+
+
+def get_external_propagation_context():
+    # type: () -> Optional[Tuple[str, str]]
+    return (
+        _external_propagation_context_fn() if _external_propagation_context_fn else None
+    )
```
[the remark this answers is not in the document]
sl0thentr0py: this is an internal thing we will use properly

sentry_sdk/scope.py
```diff
@@ -562,21 +585,29 @@ def get_baggage(self, *args, **kwargs):
...
-            "trace_id": self._propagation_context.trace_id,
-            "span_id": self._propagation_context.span_id,
-            "parent_span_id": self._propagation_context.parent_span_id,
-            "dynamic_sampling_context": self.get_dynamic_sampling_context(),
-        }  # type: Dict[str, Any]
+        # if we are tracing externally (otel), those values take precedence
+        external_propagation_context = get_external_propagation_context()
+        if external_propagation_context:
+            trace_id, span_id = external_propagation_context
+            return {"trace_id": trace_id, "span_id": span_id}
```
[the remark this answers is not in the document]
sl0thentr0py: they should turn off `traces_sample_rate` while using external tracing

PR #5052, opened 2025-10-31
feat(integrations): add ability to auto-deactivate lower-level integrations based on map

shellmayr:
- Add a map that contains integrations that should be deactivated when others are active to prevent duplicate spans, etc.
- Allow positive and negative overrides by users
- Add the existing langchain deactivations for anthropic and openai
- Simplifies setup for users, because they don't have to add the lower level integrations to the disabled integrations array in the init

Closes TET-1304


sentry_sdk/integrations/__init__.py
```diff
@@ -171,6 +171,11 @@ def iter_default_integrations(with_auto_enabling_integrations):
 }
 
 
+_INTEGRATION_DEACTIVATES = {
+    "langchain": ["openai", "anthropic"],
```
shellmayr: This is not great, because we may introduce typos. Not sure if we can make this somewhat more integration without importing the integrations (which we don't want to do for obvious reasons)

sentry_sdk/integrations/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 from sentry_sdk.utils import logger
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
```
alexander-alderman-webb: please remove the import from here, as it's already below in the `if TYPE_CHECKING` block.

sentry_sdk/integrations/__init__.py
alexander-alderman-webb: nitpick: if there is no order let's make this a set

```suggestion
    "langchain": {"openai", "anthropic"},
```

tests/test_ai_integration_deactivation.py
```diff
@@ -0,0 +1,232 @@
...
+
+    has_anthropic = True
+except Exception:
+    has_anthropic = False
+
+
+pytestmark = pytest.mark.skipif(
+    not (has_langchain and has_openai and has_anthropic),
+    reason="Requires langchain, openai, and anthropic packages to be installed",
+)
```
alexander-alderman-webb, line 32: The tests are skipped in CI at the moment. They should run in CI.

alexander-alderman-webb:
Looks good overall, let's get the tests running in CI and then we're good to go.

sentry_sdk/integrations/__init__.py
shellmayr: Good idea 👍

tests/test_ai_integration_deactivation.py
alexander-alderman-webb: See https://github.com/getsentry/sentry-python/pull/5061

alexander-alderman-webb (approved):

alexander-alderman-webb:
I'll resolve the conflicts as well.

tests/test_ai_integration_deactivation.py
```diff
@@ -0,0 +1,222 @@
...
+        auto_enabling_integrations=True,
+    )
+
+    client = get_client()
+    integration_types = {
+        type(integration) for integration in client.integrations.values()
+    }
+
+    assert LangchainIntegration not in integration_types
+
```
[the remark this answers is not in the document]
alexander-alderman-webb: That's done in other tests already.

Issue #5050, opened 2025-10-31
Exceptions are formatted incorrectly (or even fail) when `message` or `detail` attribute presents


ods:
### How do you use Sentry?

Self-hosted/on-premise

### Version

2.43.0

### Steps to Reproduce

Some of our exception define properties `message` and `detail`. The formatting of error message for those exceptions is almost always incorrect (doesn't contain important information, not formatted properly) or sometimes even fails. There is a couple of such exceptions in simplified form:

```python
class GRpcError(Exception):
    def __init__(self, code: int, message: bytes) -> None:
        super().__init__(code, message)
        self.code = code
        # Original protobuf message to allow its inspection in error handlers.
        # Here we use bytes, but it can be `google.protobuf.message.Message`
        # instance as well.
        self.message = message

    def __str__(self) -> str:
        return f"code={self.code}"


class ApiError(Exception):
    def __init__(self, code: str, detail: dict[str, Any]) -> None:
        super().__init__(code, detail)
        self.code = code
        self.detail = detail

    def __str__(self) -> str:
        formatted = f"[{self.code}]"
        for key, value in self.detail.items():
            formatted += f"\n  {key}={value}"
        return formatted
```

### Expected Result

Here is how they are formatted by Python itself:
```python
grpc_exc = GRpcError(13, b"unreadable protobuf message")

print("".join(traceback.format_exception_only(grpc_exc)).strip())
# GRpcError: code=13

grpc_exc.add_note("additional context note")

print("".join(traceback.format_exception_only(grpc_exc)).strip())
# GRpcError: code=13
# additional context note

api_exc = ApiError("RATE_LIMIT_EXCEEDED", detail={"retry_after": 30})

print("".join(traceback.format_exception_only(api_exc)).strip())
# ApiError: [RATE_LIMIT_EXCEEDED]
#   retry_after=30

api_exc.add_note("additional context note")

print("".join(traceback.format_exception_only(api_exc)).strip())
# ApiError: [RATE_LIMIT_EXCEEDED]
#   retry_after=30
# additional context note
```

### Actual Result

And the formatting by sentry (`get_error_message` is used internally by `event_from_exception`, here I call it directly):

```python
grpc_exc = GRpcError(13, b"unreadable protobuf message")

print(sentry_sdk.utils.get_error_message(grpc_exc))
# No `code` field, only unreadable field in the result:
# b'unreadable protobuf message'

grpc_exc.add_note("additional context note")

[...]

PR #5117, opened 2025-11-19
feat: Remove special cases for error messages

alexander-alderman-webb:
### Description

Always stringify exception values using the `safe_str()` method.

The special-casing of the `message` and `detail` attributes resulted in missing detail. The `message` attribute is a legacy from Python 2, and the detail attribute was returned directly because an old version of Starlette did not implement `__str__` for some exceptions.

#### Issues

Closes https://github.com/getsentry/sentry-python/issues/5050

#### Reminders


sentrivana (commented):
In general looks good to me but we should make sure this is what we want because once we release it like this, it'll be tough to change.

How does this work with the old Starlette with the missing `__str__`? Or did we drop support for it already?

alexander-alderman-webb:
Good point regarding the old Starlette version.

Their `HTTPException` has a `__str__` method since version 0.29 but we support version 0.16 and up.

So our fix to add special handling of a `detail` property fixed a Starlette and FastAPI specific problem. At the same time, users that happen to have `detail` properties on their exception lost information, as before https://github.com/getsentry/sentry-python/pull/2193 `__str__` was used to generate the Sentry exception value.

To keep both worlds happy we could use `__repr__` instead of `__str__` when

- `CustomException.__str__ is Exception.__str__`; and
- `CustomException.__repr__ is not Exception.__repr__`.

Starlette users would see a more meaningful value since `__repr__` is added in Starlette 0.12.

Happy to hear other suggestions as well, especially since it'll be hard to change later.

sentrivana:
Hmm so if we detect someone explicitly redefined `__repr__` but not `__str__`, we use `__repr__`? I'm unsure how often that'd trigger in situations where `__str__` would still be more valuable.

PR #5127, opened 2025-11-20
fix(integrations): enhance input handling for embeddings in LiteLLM integration

constantinius:
#### Issues
Closes https://linear.app/getsentry/issue/TET-1461/fix-embedding-support-for-litellm

### DIFF
--- a/tests/integrations/litellm/test_litellm.py
+++ b/tests/integrations/litellm/test_litellm.py
@@ -41,8 +41,6 @@ def clear_litellm_cache():
     The LiteLLM integration uses setup_once() which only runs once per Python process.
     This fixture ensures the integration is properly re-initialized for each test.
     """
-    # Import here to avoid circular dependencies
-    from sentry_sdk.integrations import _processed_integrations, _installed_integrations
 
     # Stop all existing mocks
     mock.patch.stopall()
@@ -54,10 +52,6 @@ def clear_litellm_cache():
     ):
         litellm.in_memory_llm_clients_cache.flush_cache()
 
-    # Remove litellm from integration tracking so setup_once() can run again
-    _processed_integrations.discard("litellm")
-    _installed_integrations.discard("litellm")
-
     yield
 
     # Clean up after test as well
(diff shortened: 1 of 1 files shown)


tests/integrations/litellm/test_litellm.py
```diff
@@ -31,6 +33,46 @@ async def __call__(self, *args, **kwargs):
 LITELLM_VERSION = package_version("litellm")
 
 
+@pytest.fixture
+def clear_litellm_cache():
```
alexander-alderman-webb, line 37: Clearing the `litellm` cache seems good

Why do re-initialize the integration? I don't think the integration has state between calls, and the tests pass without the fixture.

tests/integrations/litellm/test_litellm.py
```diff
@@ -201,44 +258,145 @@ def test_streaming_chat_completion(
...
+        mock_http.return_value = ({}, mock_response)
+
+        with start_transaction(name="litellm test"):
+            response = litellm.embedding(
+                model="text-embedding-ada-002",
+                input="Hello, world!",
+                api_key="test-key",  # Provide a fake API key to avoid authentication errors
+            )
+            # Allow time for callbacks to complete (they may run in separate threads)
+            time.sleep(0.1)
```
alexander-alderman-webb, line 281: It would be best if we can avoid `time` in tests.

Is there a `litellm` option we can use to limit the number of threads? We could try to enumerate threads and just join on them.

tests/integrations/litellm/test_litellm.py
constantinius: Yep, may be a remnant of earlier tries. Will investigate.

tests/integrations/litellm/test_litellm.py
constantinius: The threads are created deep inside the `completion` function. I don't get a handle for that easily. I know this is kinda ugly, but there is no easy way to fix this without considerable effort. So I would like to keep it this way.

tests/integrations/litellm/test_litellm.py
constantinius: Yes, was not necessary. Removed

alexander-alderman-webb (approved):

PR #5143, opened 2025-11-25
fix(ai): Handle Pydantic model classes in _normalize_data

skalinchuk:
### Description

Previously, `_normalize_data()` would attempt to call `model_dump()` on Pydantic model classes (types) when they were passed as arguments, which would fail. This commonly occurs when model classes are used in schema definitions, such as when passing response format schemas to AI SDKs.

This change adds an `inspect.isclass()` check to detect when a class type is passed instead of an instance. For classes, we now return a string representation `"<ClassType: ClassName>"` instead of attempting to call `model_dump()`.

**Changes:**
- Added `inspect` import to `sentry_sdk/ai/utils.py`
- Added class type detection in `_normalize_data()` before calling `model_dump()`
- Added comprehensive test coverage for both class types and instances in `tests/utils/test_general.py`

**Behavior:**
- Pydantic model **instances**: Still normalized via `model_dump()` (existing behavior)
- Pydantic model **classes**: Now serialized as `"<ClassType: ClassName>"` (new behavior)
- Mixed data structures: Classes within dicts/lists are properly handled

### DIFF
--- a/sentry_sdk/ai/utils.py
+++ b/sentry_sdk/ai/utils.py
@@ -1,4 +1,4 @@
-import inspect 
+import inspect
 import json
 from collections import deque
 from typing import TYPE_CHECKING


sentrivana (approved):
Thank you @skalinchuk!

Issue #5060, opened 2025-11-03
Fix bad SPANSTATUS.ERROR


sl0thentr0py:
https://github.com/getsentry/sentry-python/pull/4820/files#diff-719f41f07f1275f8a1b02085180aa234b257226fc976d776ebef5d8dff9db017R768

not a valid span status in relay

PR #5146, opened 2025-11-25
Remove unsupported SPANSTATUS.(ERROR|UNSET)

sl0thentr0py:
### Description

The new `ERROR` and `UNSET` statuses introduced in #4820 are [not supported by relay](https://github.com/getsentry/relay/blob/8e6c963cdd79dc9ba2bebc21518a3553f70feeb3/relay-base-schema/src/spans.rs#L19-L94) and caused a regression.

#### Issues

* resolves: #5060
* resolves: PY-1961

### DIFF
--- a/tests/integrations/anthropic/test_anthropic.py
+++ b/tests/integrations/anthropic/test_anthropic.py
@@ -701,7 +701,7 @@ def test_exception_message_create(sentry_init, capture_events):
 
     (event, transaction) = events
     assert event["level"] == "error"
-    assert transaction["contexts"]["trace"]["status"] == "error"
+    assert transaction["contexts"]["trace"]["status"] == "internal_error"
 
 
 def test_span_status_error(sentry_init, capture_events):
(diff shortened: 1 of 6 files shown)


sentry_sdk/consts.py
```diff
@@ -864,20 +864,18 @@ class SPANSTATUS:
...
-    OK = "ok"  # HTTP 200 and OTel status code: https://opentelemetry.io/docs/concepts/signals/traces/#span-status
+    OK = "ok"
     OUT_OF_RANGE = "out_of_range"
     PERMISSION_DENIED = "permission_denied"
     RESOURCE_EXHAUSTED = "resource_exhausted"
     UNAUTHENTICATED = "unauthenticated"
     UNAVAILABLE = "unavailable"
     UNIMPLEMENTED = "unimplemented"
     UNKNOWN_ERROR = "unknown_error"
-    UNSET = "unset"  # OTel status code: https://opentelemetry.io/docs/concepts/signals/traces/#span-status
```
alexander-alderman-webb, line 880: Is removing `UNSET` intentional?

sl0thentr0py: both `UNSET` and `ERROR` are otel statuses, not relay statuses and were never supposed to be added here

Issue #5059, opened 2025-11-03
Add back `span.status`


cleptric:
Status is missing from the span in https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/tracing.py#L720-L730.
Let's add it (back). Long term, the emitted status tag on the span should not be sent any longer.

PR #5147, opened 2025-11-25
Add back span status

sl0thentr0py:
### Description
    
For some reason since Performance was implemented ages ago, we were not sending the `span.status` field at all. 
Add it back since we are moving to a Span-First world and Agent Monitoring also expects span statuses to be correct for some product features.
    
Some other notes regarding the status mess:
* the transaction status is added to the `trace_context` - this is fine
* the span status is also added as a tag, this should be removed in a major since it's redundant
* `set_http_status` further adds another tag called `http.status_code`, remove this in the major too

#### Issues

* resolves: #5059
* resolves: PY-1960

### DIFF
--- a/sentry_sdk/tracing.py
+++ b/sentry_sdk/tracing.py
@@ -666,7 +666,7 @@ class Span:
         # type: (int) -> None
         self.set_tag(
             "http.status_code", str(http_status)
-        )  # we keep this for backwards compatibility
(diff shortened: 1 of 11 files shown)


sl0thentr0py:
image

sl0thentr0py:
major cleanup tracked in https://github.com/getsentry/sentry-python/issues/5154

alexander-alderman-webb (approved):

Issue #5140, opened 2025-11-24
Test integrations with empty shadow modules


alexander-alderman-webb:
## Background

Two critical bugs came in that would have been hard to catch with human review (https://github.com/getsentry/sentry-python/issues/5134, https://github.com/getsentry/sentry-python/issues/5129). Both occurred at initialization time due to `ModuleNotFoundError` as we moved some integrations to the auto-activating list in https://github.com/getsentry/sentry-python/pull/5111.

Many imports are short and therefore easy to be shadowed by local files. For example, a user ran the `openai-agents` with a local `agents.py` file. Normally, auto-enabling integrations are not activated when the associated package is not installed, because the imports are gated. In this case, we had

```python
try:
    import agents

except ImportError:
    raise DidNotEnable("OpenAI Agents not installed")
```

Because no members of the `agents` package were imported, the import did not raise an error, in spite of `agents` referring to `agents.py` instead of `openai-agents` in the user's environment.

## Solution

Run a "smoke test" that sets the corresponding package to be empty and initializes the SDK.

PR #5150, opened 2025-11-26
test: Import integrations with empty shadow modules

alexander-alderman-webb:
Add a test parametrized on our integrations. The test detects if the integration imports modules not in the standard library, and if so, verifies that importing the integration with an empty shadow module raises a `DidNotEnable` exception.

### Motivation

Adding integrations to be auto-enabling can cause SDK crashes in two specific cases we encountered at the end of last week:
- the user has an old package version that we don’t support, resulting in an `ImportError` when we patch something that does not yet exist; or
- something like an `agents.py` in the environment shadows the import and causes an `ImportError` if the auto-activation still triggers.

All integrations with poorly gated imports are not auto-enabling, but some affected integrations are new (e.g., litellm).

#### Issues

Closes https://github.com/getsentry/sentry-python/issues/5140

#### Reminders


tests/test_shadowed_module.py
```diff
@@ -0,0 +1,87 @@
...
+            "clickhouse_driver",
+            "grpc",
+            "litellm",
+            "opentelemetry",
+            "pure_eval",
+            "ray",
+            "trytond",
+            "typer",
+        ):
+            continue
```
alexander-alderman-webb: If we merge this test, my plan is to remove integrations from the skip list one-by-one.

tests/test_shadowed_module.py
```diff
@@ -0,0 +1,87 @@
...
+            importlib.import_module(module_path)
+            continue
+        except integrations.DidNotEnable:
+            spec = importlib.util.find_spec(module_path)
+            source = pathlib.Path(spec.origin).read_text(encoding="utf-8")
+            tree = ast.parse(source, filename=spec.origin)
+            integration_dependencies = find_unrecognized_dependencies(tree)
+            for dependency in integration_dependencies:
+                sys.modules[dependency] = types.ModuleType(dependency)
+            with pytest.raises(integrations.DidNotEnable):
```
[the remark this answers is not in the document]
alexander-alderman-webb: No, integrations should raise `DidNotEnable` by importing non-trivial things from the module.

tests/test_shadowed_module.py
```diff
@@ -0,0 +1,87 @@
...
+                importlib.import_module(module_path)
```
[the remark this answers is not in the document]
alexander-alderman-webb: This is the intended test, not a logical flaw. Integrations should *not* perform a basic import check to prevent initialization crashes when integrations are made auto-enabling.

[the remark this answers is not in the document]
alexander-alderman-webb: ~~I decided to parametrize the test and fork it because messing with `sys.modules` and importing can cause side-effects and influence other tests otherwise~~.

edit: I don't like adding more forked tests. I'll move the test into it's own tox environment so there's no risk of messing with the behavior of other tests when removing the forking.

tests/test_shadowed_module.py
```diff
@@ -0,0 +1,109 @@
...
+            submodule_names
+            - {
+                "clickhouse_driver",
+                "grpc",
+                "litellm",
+                "opentelemetry",
+                "pure_eval",
+                "ray",
+                "trytond",
+                "typer",
```
sentrivana, line 39: What determines what ends up in the ignore list here?

tests/test_shadowed_module.py
```diff
@@ -0,0 +1,109 @@
...
+            except ValueError:
+                continue
+
+    return unrecognized_dependencies
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 7), reason="asyncpg imports __future__.annotations"
+)
+def test_shadowed_modules_when_importing_integrations(
```
sentrivana, line 82: I think I understand what's going on here, but it takes some effort to grasp. Maybe a good case for additional comments in the function to better explain what's happening step-by-step?

tests/test_shadowed_module.py
alexander-alderman-webb: These are packages for which the test fails. I've taken a brief look at all of them and there's no false positives.

tests/test_shadowed_module.py
alexander-alderman-webb: yes, that's a good call! Added a few comments.

PR #5153, opened 2025-11-26
feat(integration): pydantic-ai: properly report token usage and response model for invoke_agent spans

constantinius:
#### Issues

Closes https://linear.app/getsentry/issue/TET-1458/py-pydantic-ai-attributes-missing

### DIFF
--- a/sentry_sdk/integrations/pydantic_ai/spans/invoke_agent.py
+++ b/sentry_sdk/integrations/pydantic_ai/spans/invoke_agent.py
@@ -9,6 +9,7 @@ from ..utils import (
     _set_model_data,
     _should_send_prompts,
 )
+from .utils import _set_usage_data
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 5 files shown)


sentry_sdk/integrations/pydantic_ai/spans/invoke_agent.py
```diff
@@ -103,10 +104,37 @@ def invoke_agent_span(user_prompt, agent, model, model_settings, is_streaming=Fa
...
+
+    # Set token usage data if available
+    if hasattr(result, "usage") and callable(result.usage):
+        try:
+            usage = result.usage()
+            if usage:
+                _set_usage_data(span, usage)
+        except Exception:
+            # If usage() call fails, continue without setting usage data
+            pass
```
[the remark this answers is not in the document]
constantinius: That is not true, [`usage` is a function, for whatever reason.](https://ai.pydantic.dev/api/run/#pydantic_ai.run.AgentRunResult.usage)

sentry_sdk/integrations/pydantic_ai/spans/invoke_agent.py
```diff
@@ -103,10 +104,37 @@ def invoke_agent_span(user_prompt, agent, model, model_settings, is_streaming=Fa
...
+    # Set response text if prompts are enabled
+    if _should_send_prompts() and output:
         set_data_normalized(
             span, SPANDATA.GEN_AI_RESPONSE_TEXT, str(output), unpack=False
         )
+
+    # Set token usage data if available
+    if hasattr(result, "usage") and callable(result.usage):
+        try:
+            usage = result.usage()
```
sentrivana, line 125: Can calling `usage()` have any (side-)effects that we wouldn't want to trigger?

I'm thinking something in the vein of extra DB access, API request, etc. We had a similar case in Django where we were trying to get the connection params to the DB and in some users' case that actually resulted in some expensive operations being performed.

sentrivana (approved):
Thanks, looks good to me, see one comment.

Side question: would you say this is more of a feature or a bugfix? We have a fancy new changelog generation system and I want to tag this properly 😄

sentry_sdk/integrations/pydantic_ai/spans/invoke_agent.py
constantinius: No, this literally has the comment "should be a property". It is just an accessor for a nested field.

constantinius:
>Side question: would you say this is more of a feature or a bugfix? We have a fancy new changelog generation system and I want to tag this properly 😄 

It adds new fields, but those were kind of expected. So we can consider this a bugfix

PR #5158, opened 2025-11-26
Simplify continue_trace to reuse propagation_context values

sl0thentr0py:
### Description

We are already parsing both `sentry-trace` and `baggage` headers and filling in `sample_rand` while constructing the `propagation_context`. We don't need the redundant logic in `Transaction.continue_from_headers` anymore.

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -1186,27 +1186,25 @@ class Scope:
         """
         self.generate_propagation_context(environ_or_headers)
 
-        # When we generate the propagation context, the sample_rand value is set
-        # if missing or invalid (we use the original value if it's valid).
-        # We want the transaction to use the same sample_rand value. Due to duplicated
-        # propagation logic in the transaction, we pass it in to avoid recomputing it
-        # in the transaction.
-        # TYPE SAFETY: self.generate_propagation_context() ensures that self._propagation_context
-        # is not None.
-        sample_rand = typing.cast(
-            PropagationContext, self._propagation_context
-        )._sample_rand()
+        # generate_propagation_context ensures that the propagation_context is not None.
+        propagation_context = typing.cast(PropagationContext, self._propagation_context)
 
-        transaction = Transaction.continue_from_headers(
-            normalize_incoming_data(environ_or_headers),
-            _sample_rand=sample_rand,
+        optional_kwargs = {}
+        if name:
+            optional_kwargs["name"] = name
+        if source:
+            optional_kwargs["source"] = source
+
+        return Transaction(
             op=op,
             origin=origin,
-            name=name,
-            source=source,
+            baggage=propagation_context.baggage,
+            parent_sampled=propagation_context.parent_sampled,
+            trace_id=propagation_context.trace_id,
+            parent_span_id=propagation_context.parent_span_id,
+            same_process_as_parent=False,
+            **optional_kwargs,
         )
-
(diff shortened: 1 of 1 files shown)


sentry_sdk/scope.py
```diff
@@ -1186,28 +1186,23 @@ def continue_trace(
...
-
-        transaction = Transaction.continue_from_headers(
-            normalize_incoming_data(environ_or_headers),
-            _sample_rand=sample_rand,
+        # generate_propagation_context ensures that the propagation_context is not None.
+        propagation_context = typing.cast(PropagationContext, self._propagation_context)
+
+        optional_kwargs = {"source": source} if source else {}
+
+        return Transaction(
```
sl0thentr0py, line 1198: annoyingly, the type signature of `continue_trace` that we decided to expose is inconsistent with the actual `Transaction` constructor, mainly that `name` and `source` should not be `Optional` in `continue_trace` but have default values instead.

sentry_sdk/scope.py
```diff
@@ -1186,28 +1186,26 @@ def continue_trace(
...
             origin=origin,
-            name=name,
-            source=source,
+            baggage=propagation_context.baggage,
+            parent_sampled=propagation_context.parent_sampled,
+            trace_id=propagation_context.trace_id,
+            parent_span_id=propagation_context.parent_span_id,
+            same_process_as_parent=False,
+            **optional_kwargs,
         )
```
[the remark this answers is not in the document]
sl0thentr0py: This is because we froze the DSC earlier in the transaction but not in the propagation context, I will fix this separately tomorrow. Also in some SDKs, we have given up freezing completely so maybe I'll just leave it out.

sentry_sdk/scope.py
```diff
@@ -1186,28 +1186,26 @@ def continue_trace(
...
-        # in the transaction.
-        # TYPE SAFETY: self.generate_propagation_context() ensures that self._propagation_context
-        # is not None.
-        sample_rand = typing.cast(
-            PropagationContext, self._propagation_context
-        )._sample_rand()
-
-        transaction = Transaction.continue_from_headers(
-            normalize_incoming_data(environ_or_headers),
-            _sample_rand=sample_rand,
```
sentrivana, line 1202: Can we already remove the `_sample_rand` arg and associated logic from `continue_from_headers` altogether or is that still used anywhere?

sl0thentr0py: ~will remove, technically breaking but it's an underscore so fine~
nvm, let's do all this in the major, we will remove `continue_from_headers` itself.

sentry_sdk/scope.py
```diff
@@ -1186,28 +1186,26 @@ def continue_trace(
...
             op=op,
             origin=origin,
-            name=name,
-            source=source,
+            baggage=propagation_context.baggage,
+            parent_sampled=propagation_context.parent_sampled,
+            trace_id=propagation_context.trace_id,
+            parent_span_id=propagation_context.parent_span_id,
+            same_process_as_parent=False,
+            **optional_kwargs,
```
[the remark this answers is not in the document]
sl0thentr0py: we actually only call `serialize` with `include_third_party=True` in the celery case but not generally, so I will postpone this freezing / third party business for later after clarifying the requirements. As a refactor, this PR mostly does what it did before so it's fine.

PR #5169, opened 2025-11-27
fix(spotlight): align behavior with SDK spec

BYK:
Align the Spotlight implementation with the official spec at
https://develop.sentry.dev/sdk/expected-features/spotlight/

Changes:

1. Fix precedence rules:
   - When spotlight=True and SENTRY_SPOTLIGHT env var is a URL, use the
     env var URL (per spec requirement)
   - Log warning when config URL overrides env var URL
   - Log warning when spotlight=False explicitly disables despite env
     var being set

2. Route all envelopes to Spotlight:
   - Sessions, logs, and metrics now also get sent to Spotlight via the
     _capture_envelope callback
   - Move Spotlight initialization before batchers are created

3. Add exponential backoff retry logic:
   - SpotlightClient now implements proper exponential backoff when the
     Spotlight server is unreachable
   - Skips sending during backoff period to avoid blocking
   - Logs errors only once per backoff cycle

4. Update tests:
   - Fix test expectation for spotlight=True + env URL case
   - Add tests for warning scenarios
   - Add test for session envelope routing to Spotlight

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -389,68 +389,17 @@ class _Client(BaseClient):
                 if self.options["enable_backpressure_handling"]:
                     self.monitor = Monitor(self.transport)
 
-            # Setup Spotlight before creating batchers so _capture_envelope can use it
-            spotlight_config = self.options.get("spotlight")
-            spotlight_env_value = os.environ.get("SENTRY_SPOTLIGHT")
+            # Setup Spotlight before creating batchers so _capture_envelope can use it.
+            # setup_spotlight handles all config/env var resolution per the SDK spec.
+            from sentry_sdk.spotlight import setup_spotlight
 
-            # Parse env var to determine if it's a boolean or URL
-            spotlight_from_env = None  # type: Optional[bool]
-            spotlight_env_url = None  # type: Optional[str]
-            if spotlight_env_value:
-                parsed = env_to_bool(spotlight_env_value, strict=True)
-                if parsed is None:
-                    # It's a URL string
-                    spotlight_from_env = True
-                    spotlight_env_url = spotlight_env_value
-                else:
-                    spotlight_from_env = parsed
-
-            # Apply precedence rules per spec
-            if spotlight_config is False:
-                # Config explicitly disables spotlight - warn if env var was set
-                if spotlight_from_env:
-                    logger.warning(
-                        "Spotlight is disabled via spotlight=False config option, "
-                        "ignoring SENTRY_SPOTLIGHT environment variable."
-                    )
-                self.options["spotlight"] = False
-            elif spotlight_config is True:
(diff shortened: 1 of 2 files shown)


sentry_sdk/client.py
```diff
@@ -387,6 +389,69 @@ def _record_lost_event(
...
+                        spotlight_env_value,
+                    )
+                # self.options["spotlight"] already has the config URL
+            elif spotlight_config is None:
+                # No config - use env var
+                if spotlight_env_url:
+                    self.options["spotlight"] = spotlight_env_url
+                elif spotlight_from_env:
+                    self.options["spotlight"] = True
+                # else: stays None (disabled)
```
sentrivana: Can we put all of this option/env var logic into a separate function? Or better yet, put it in `setup_spotlight`?

sentry_sdk/client.py
```diff
@@ -387,6 +389,69 @@ def _record_lost_event(
...
+            if spotlight_env_value:
+                parsed = env_to_bool(spotlight_env_value, strict=True)
+                if parsed is None:
+                    # It's a URL string
+                    spotlight_from_env = True
+                    spotlight_env_url = spotlight_env_value
+                else:
+                    spotlight_from_env = parsed
+
+            # Apply precedence rules per spec
```
sentrivana: Could add the link to the spec here for future reference

sentry_sdk/spotlight.py
```diff
@@ -34,14 +35,39 @@
...
     def capture_envelope(self, envelope):
         # type: (Envelope) -> None
+
+        # Check if we're in backoff period - skip sending to avoid blocking
+        current_time = time.time()
+        if self._last_error_time > 0:
+            time_since_error = current_time - self._last_error_time
+            if time_since_error < self._retry_delay:
+                # Still in backoff period, skip this envelope
+                return
```
sentrivana: Nit: no need to save `time.time()` as `current_time` since we don't use it anywhere else
```suggestion
        if self._last_error_time > 0:
            time_since_error = time.time() - self._last_error_time
            if time_since_error < self._retry_delay:
                # Still in backoff period, skip this envelope
                return
```

sentry_sdk/spotlight.py
```diff
@@ -54,18 +80,27 @@ def capture_envelope(self, envelope):
...
+            self._retry_delay = self.INITIAL_RETRY_DELAY
+            self._last_error_time = 0.0
         except Exception as e:
-            if self.fails < 2:
-                sentry_logger.warning(str(e))
-                self.fails += 1
-            elif self.fails == 2:
-                self.fails += 1
+            current_time = time.time()
+            self._last_error_time = current_time
```
sentrivana: Nit: same as above
```suggestion
            self._last_error_time = time.time()
```

PR #5174, opened 2025-12-01
fix(openai-agents): Avoid double span exit on exception

alexander-alderman-webb:
### Description

With https://github.com/getsentry/sentry-python/pull/5165 we have a handle on the agent invocation span, as the span is now attached to the context wrapper returned from the `AgentRunner.run()` method.

Exceptions of type `AgentsException` are now caught on `AgentRunner.run()` instead of `AgentRunner._run_single_turn()`, because `openai-agents` attaches the context wrapper to these exceptions in the method.

Removing `sentry_sdk.get_current_span()` avoids double exit scenarios when the active span **is not** an agent invocation span. This is the case when the `asyncio` integration is used.

Previously, the script below results in a double span exit unhandled exception.

```python
import asyncio

from unittest import mock

import agents
from agents import Agent, Runner

import sentry_sdk
from sentry_sdk.integrations.openai_agents import OpenAIAgentsIntegration

async def main():
    with mock.patch(
        "agents.models.openai_responses.OpenAIResponsesModel.get_response"
    ) as mock_get_response:
        mock_get_response.side_effect = Exception("Model Error")

        sentry_sdk.init(
            integrations=[OpenAIAgentsIntegration(), AsyncioIntegration()],
            traces_sample_rate=1.0,
            debug=True
        )

        agent = Agent(
            name="Bug Reproduction Agent",
            instructions="You are a test agent. Always use the broken_tool.",
            model="gpt-4o-mini",
        )

        try:
            await Runner.run(
                agent, "Test input", # run_config=test_run_config
            )
        except Exception:
            pass

if __name__ == "__main__":
    asyncio.run(main())
```

#### Reminders

### DIFF
--- a/sentry_sdk/integrations/openai_agents/patches/error_tracing.py
+++ b/sentry_sdk/integrations/openai_agents/patches/error_tracing.py
@@ -1,13 +1,14 @@
 from functools import wraps
 
 import sentry_sdk
+from sentry_sdk.consts import SPANSTATUS
 from sentry_sdk.tracing_utils import set_span_errored
 from ..utils import _record_exception_on_span
 
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Any
+    from typing import Any, Callable, Optional
 
 
 def _patch_error_tracing():


sentry_sdk/integrations/openai_agents/patches/agent_run.py
```diff
@@ -73,10 +80,19 @@ async def patched_run_single_turn(cls, *args, **kwargs):
...
 
         # Call original method with all the correct parameters
-        result = await original_run_single_turn(*args, **kwargs)
+        try:
+            result = await original_run_single_turn(*args, **kwargs)
+        except Exception as exc:
+            if span is not None and span.timestamp is None:
+                _record_exception_on_span(span, exc)
+                span.__exit__(None, None, None)
+                delattr(context_wrapper, "_sentry_agent_span")
```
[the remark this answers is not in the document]
alexander-alderman-webb: In that case, `span.__exit__` fails first.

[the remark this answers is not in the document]
alexander-alderman-webb: thanks, replaced the manual stuff with `_end_invoke_agent_span()`.

sentry_sdk/integrations/openai_agents/patches/agent_run.py
```diff
@@ -65,6 +71,7 @@ async def patched_run_single_turn(cls, *args, **kwargs):
         context_wrapper = kwargs.get("context_wrapper")
         should_run_agent_start_hooks = kwargs.get("should_run_agent_start_hooks")
 
+        span = getattr(context_wrapper, "_sentry_agent_span", None)
```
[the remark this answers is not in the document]
alexander-alderman-webb: Both times `_run_single_turn` is invoked in the internals of `openai-agents`, a `context_wrapper` is passed in.

sentry_sdk/integrations/openai_agents/patches/runner.py
```diff
@@ -28,18 +34,36 @@ async def wrapper(*args, **kwargs):
...
+                        ):
+                            _record_exception_on_span(invoke_agent_span, exc)
+                            end_invoke_agent_span(context_wrapper, agent)
 
                     raise exc from None
+                except _SingleTurnException as exc:
+                    # Handled in _run_single_turn() patch.
+                    raise exc.original from None
+                except Exception as exc:
+                    # Invoke agent span is not finished in this case.
```
alexander-alderman-webb, line 59: I could add a `get_current_span()` again in this case and we would never lose spans.

You also can't guarantee that the SDK does not crash if you try to exit a span you obtain with `get_current_span()`, so I decided the just drop the span for now.

Much of `AgentRunner.run()` that is not spent in `_run_single_turn()` is before the agent invocation span is created anyway. The **only** cases in which we lose a span is that any of the post-turn steps like output guardrails raise an exception that is not an `AgentsException`.

sentry_sdk/integrations/openai_agents/patches/runner.py
```diff
@@ -28,18 +34,36 @@ async def wrapper(*args, **kwargs):
...
-                except Exception as exc:
+                    run_result = await original_func(*args, **kwargs)
+                except AgentsException as exc:
                     _capture_exception(exc)
 
-                    # It could be that there is a "invoke agent" span still open
-                    current_span = sentry_sdk.get_current_span()
-                    if current_span is not None and current_span.timestamp is None:
-                        current_span.__exit__(None, None, None)
+                    context_wrapper = getattr(exc.run_data, "context_wrapper", None)
```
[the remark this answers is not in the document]
alexander-alderman-webb: In practice it's always set as `AgentRunner.run()` has a big try-except at the top level.

sentry_sdk/integrations/openai_agents/patches/agent_run.py
```diff
@@ -65,18 +77,33 @@ async def patched_run_single_turn(cls, *args, **kwargs):
...
+            # as long as only AgentRunner.run() invokes AgentRunner._run_single_turn().
+            raise
+        except Exception as exc:
+            _capture_exception(exc)
+
+            if span is not None and span.timestamp is None:
+                _record_exception_on_span(span, exc)
+                end_invoke_agent_span(context_wrapper, agent)
+
+            raise _SingleTurnException(exc)
```
alexander-alderman-webb: I created `_SingleTurnException` to avoid calling `capture_exception()` twice.

Otherwise, exceptions raised in `_run_single_turn()` that are not of type `AgentsException` would be captured in both `run()` and `_run_single_turn()`.

sentry_sdk/integrations/openai_agents/patches/runner.py
```diff
@@ -28,18 +34,36 @@ async def wrapper(*args, **kwargs):
...
+                    # Handled in _run_single_turn() patch.
+                    raise exc.original from None
+                except Exception as exc:
+                    # Invoke agent span is not finished in this case.
+                    # This is much less likely to occur than other cases because
+                    # AgentRunner.run() is "just" a while loop around _run_single_turn.
+                    _capture_exception(exc)
+                    raise exc from None
+
+                end_invoke_agent_span(run_result.context_wrapper, agent)
```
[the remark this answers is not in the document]
alexander-alderman-webb: duplicate of https://github.com/getsentry/sentry-python/pull/5174#discussion_r2577362827

sentry_sdk/integrations/openai_agents/patches/runner.py
```diff
@@ -29,19 +35,37 @@ async def wrapper(*args, **kwargs):
...
 
                     raise exc from None
+                except _SingleTurnException as exc:
+                    # Handled in _run_single_turn() patch.
+                    raise exc.original from None
+                except Exception as exc:
+                    # Invoke agent span is not finished in this case.
+                    # This is much less likely to occur than other cases because
+                    # AgentRunner.run() is "just" a while loop around _run_single_turn.
+                    _capture_exception(exc)
```
sl0thentr0py, line 62: it's hard for me to follow the flow, but my recommendation would be this:

* we should not call `capture_exception` here at all, just have a `finally` block that takes care of the span management
* the exception should be captured only in the `runner.py` patch
* the library internally already decides what to re-raise and what to swallow and we don't need to care about those, only what the library finally emits

sl0thentr0py (changes requested):

sentry_sdk/integrations/openai_agents/patches/runner.py
sl0thentr0py: sorry I meant that we should only have this one and not the one in `agent_run`

alexander-alderman-webb: The reason there are special cases for exceptions, and for handling exceptions at different levels is that

- `openai-agents` only [attaches the context wrapper](https://github.com/openai/openai-agents-python/blob/0d2d77152a4e0f1524854938c5f3caf996421c56/src/agents/run.py#L750-L759) when `AgentsException` is raised; and
- we always have a handle on the `context_wrapper` in `agent_run.py`, and only sometimes in `runner.py`.

Regarding having a finally block, I did not go with that approach for different reasons in `agent_run.py` and `runner.py`.

In `agent_run.py`, one agent invocation can span multiple calls to the function we patch, so it's intentional to only finish the span in the except block. The broad `Exception` is caught here so we can finish the span even when an exception which does not subclass `AgentsException` is raised. In `runner.py` we do not have a handle on the span if it is not an `AgentsException`.

In `runner.py`, we do not have a finally block because you access the context wrapper from different places depending on the code path. In the non-exception case, the context wrapper is on the return value of the function, and in the exception case, the context wrapper is attached to the exception object.

sl0thentr0py: > so it's intentional to only finish the span in the except block.

in that case keeping the span management in the except block is fine. 

But I still think the `capture_exception` should only be there on the outermost layer. We should not be coupling exception handling and span management this tightly.

Regarding handling missing state on our end, we should simply write logic with presence checks so we don't rely on the actual underlying implementation of the lib so much. Currently, much of the design is very brittle to internals changing.

sl0thentr0py: the span context manager `__exit__` already responds to exceptions, but instead whoever wrote this decided to pass `(None, None, None)` here which makes everything extra complicated.

https://github.com/getsentry/sentry-python/blob/ea91e5416484cadacecdf2cfe8a9d4554d1669f2/sentry_sdk/integrations/openai_agents/spans/invoke_agent.py#L96

alexander-alderman-webb: I've moved `capture_exception` to the outer layer.

alexander-alderman-webb: Also realized we don't need this `_SingleTurnException` stuff so that's removed now.
Hopefully that makes it slightly less brittle to internals changing.

Issue #5066, opened 2025-11-04
Implement strict trace continuation


sl0thentr0py:
https://develop.sentry.dev/sdk/telemetry/traces/#stricttracecontinuation

PR #5178, opened 2025-12-01
feat: Implement strict_trace_continuation

sl0thentr0py:
### Description

Fixes a security hole where incoming traces from other orgs can cause a DOS-like attack on another org by injecting Sentry propagation headers. 

Spec: https://develop.sentry.dev/sdk/telemetry/traces/#stricttracecontinuation

#### Issues

* resolves: #5066
* resolves: PY-1963

### DIFF
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -15,7 +15,6 @@ from sentry_sdk.consts import OP, SPANDATA, SPANSTATUS, SPANTEMPLATE
 from sentry_sdk.utils import (
     capture_internal_exceptions,
     filename_for_module,
-    Dsn,
     logger,
     match_regex_list,
     qualname_from_function,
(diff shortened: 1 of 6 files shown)


sentry_sdk/tracing_utils.py
```diff
@@ -1230,6 +1237,41 @@ def _set_output_attributes(span, template, send_pii, result):
...
+
+    if (
+        client_org_id is not None
+        and baggage_org_id is not None
+        and client_org_id != baggage_org_id
+    ):
+        logger.debug(
+            f"Starting a new trace because org IDs don't match (incoming baggage org_id: {baggage_org_id}, SDK org_id: {client_org_id})"
+        )
+        return False
```
alexander-alderman-webb, line 1260: Should `_should_continue_trace` always returns true if `strict_trace_continuation=False`?

I would have expected that from the docstring for the option (although it only describes what applies when the option is `True`, so not a strict contradiction).

sl0thentr0py: if the incoming baggage does have a org_id, we always check for consistency with the `client_org_id` irrespective of `strict_trace_continuation` (which will eventually default to `True` in the next major, so this is to ease the transition).

I will improve the docstring for the other case.

alexander-alderman-webb (approved):

PR #5186, opened 2025-12-02
fix: Make PropagationContext.from_incoming_data always return a PropagationContext

sl0thentr0py:
### Description

When there is any sort of incoming data, and since `continue_trace` is always intended to be at a system boundary, we always want to force a new trace if there's no incoming propagation or a mismatched propagation headers (for `strict_trace_continuation`).

What previously happened in these cases: a single `trace_id` was kept alive in the `propagation_context` even when these were meant to be new independent traces (see screenshot that shows undesired behavior before this fix).

I think this will actually solve many complaints about long living traces that were never supposed to be such.

image

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -510,11 +510,12 @@ class Scope:
         If there is `incoming_data` overwrite existing propagation context.
         If there is no `incoming_data` create new propagation context, but do NOT overwrite if already existing.
         """
-        if incoming_data:
-            propagation_context = PropagationContext.from_incoming_data(incoming_data)
-            if propagation_context is not None:
-                self._propagation_context = propagation_context
+        if incoming_data is not None:
+            self._propagation_context = PropagationContext.from_incoming_data(
+                incoming_data
+            )
 
+        # TODO-neel this below is a BIG code smell but requires a bunch of other refactoring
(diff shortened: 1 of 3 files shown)


sentry_sdk/scope.py
```diff
@@ -511,10 +511,9 @@ def generate_propagation_context(self, incoming_data=None):
         If there is no `incoming_data` create new propagation context, but do NOT overwrite if already existing.
         """
         if incoming_data:
-            propagation_context = PropagationContext.from_incoming_data(incoming_data)
-            if propagation_context is not None:
-                self._propagation_context = propagation_context
+            self._propagation_context = PropagationContext.from_incoming_data(incoming_data)
 
+        # TODO-neel this below is a BIG code smell but requires a bunch of other refactoring
```
sl0thentr0py, line 518: @alexander-alderman-webb since you wanted me to point out more insights like these, this branch and how it interacts with the [`get_active_propagation_context` method](https://github.com/getsentry/sentry-python/blob/ad108771389395b8927775c164b15cf87e41ca63/sentry_sdk/scope.py#L690-L703) logic is still pretty badly implemented.

sl0thentr0py:
yep pretty much :/

PR #5206, opened 2025-12-09
Convert sentry_sdk type annotations into the modern format

zsol:
### Description
This PR converts all Python 2-style type annotations (in comments) into Python 3-style annotations (after `:` tokens, and inside the AST).

There's a large amount of changes in here, apologies for that. I generated these changes using:

1. A LibCST codemod to do the bulk of the work: `uv run --with libcst python -m libcst.tool codemod convert_type_comments.ConvertTypeComments --no-format -j1 sentry_sdk`
2. Some manual fixups here and there (see below)
3. Run [com2ann](https://github.com/ilevkivskyi/com2ann) to fix the rest: `uvx com2ann sentry_sdk`
4. Run [this ast-grep rule](https://ast-grep.github.io/playground.html#eyJtb2RlIjoiQ29uZmlnIiwibGFuZyI6InB5dGhvbiIsInF1ZXJ5IjoiYWxsOlxuICAtIGtpbmQiLCJyZXdyaXRlIjoibG9nZ2VyLmxvZygkTUFUQ0gpIiwic3RyaWN0bmVzcyI6InNtYXJ0Iiwic2VsZWN0b3IiOiIiLCJjb25maWciOiJsYW5ndWFnZTogcHl0aG9uXG5ydWxlOlxuICBhbGw6XG4gICAgLSBhbnk6XG4gICAgICAgIC0ga2luZDogYmluYXJ5X29wZXJhdG9yXG4gICAgICAgICAgcGF0dGVybjogJFRZUEVcbiAgICAgICAgICBpbnNpZGU6XG4gICAgICAgICAgICBraW5kOiB0eXBlXG4gICAgICAgICAgICBzdG9wQnk6IGVuZFxuICAgICAgICAtIG1hdGNoZXM6IGdlbmVyaWNcbiAgICAgICAgLSBtYXRjaGVzOiBzaW1wbGVcblxuICAgICMgSEFDSzogdHJlZS1zaXR0ZXIgaGFzIGJ1Z2d5IHN1cHBvcnQgZm9yIHR5cGUgYWxpYXMgc3RhdGVtZW50cywgYnV0IHdlIGRvbid0IGhhdmVcbiAgICAjIGFueSBvZiB0aG9zZSBpbiBhIGNvZGViYXNlIHRoYXQgd2lsbCByZXF1aXJlIHF1b3RpbmcuIExldCdzIGlnbm9yZSB0aGVzZS5cbiAgICAtIG5vdDpcbiAgICAgICAgaW5zaWRlOlxuICAgICAgICAgIGtpbmQ6IHR5cGVfYWxpYXNfc3RhdGVtZW50XG4gICAgICAgICAgc3RvcEJ5OiBlbmRcbnV0aWxzOlxuICBnZW5lcmljOlxuICAgIGFsbDpcbiAgICAgIC0gaW5zaWRlOlxuICAgICAgICAgIGtpbmQ6IHR5cGVcbiAgICAgICAgICBzdG9wQnk6IGVuZFxuICAgICAgLSBhbnk6XG4gICAgICAgICAgLSBraW5kOiBzdWJzY3JpcHRcbiAgICAgICAgICAtIGtpbmQ6IGdlbmVyaWNfdHlwZVxuICAgICAgLSBwYXR0ZXJuOiAkVFlQRVxuICBzaW1wbGU6XG4gICAgYWxsOlxuICAgICAgLSBhbnk6XG4gICAgICAgICAgLSBraW5kOiBpZGVudGlmaWVyXG4gICAgICAgICAgLSBraW5kOiBhdHRyaWJ1dGVcbiAgICAgIC0gcGF0dGVybjogJFRZUEVcbiAgICAgIC0gaW5zaWRlOlxuICAgICAgICAgIGtpbmQ6IHR5cGVcbiAgICAgICAgICBzdG9wQnk6IGVuZFxuICAgICAgLSBub3Q6XG4gICAgICAgICAgbWF0Y2hlczogYnVpbHRpblxuICBidWlsdGluOlxuICAgIGFueTpcbiAgICAgIC0gcGF0dGVybjogYm9vbFxuICAgICAgLSBwYXR0ZXJuOiBzdHJcbiAgICAgIC0gcGF0dGVybjogaW50XG4gICAgICAtIHBhdHRlcm46IE5vbmVcbiAgICAgIC0gcGF0dGVybjogZmxvYXRcbiAgICAgIC0gcGF0dGVybjogXCJUcnVlXCJcbiAgICAgIC0gcGF0dGVybjogXCJGYWxzZVwiXG4gICAgICAtIHBhdHRlcm46IG9iamVjdFxuICAgICAgLSBwYXR0ZXJuOiBieXRlc1xuICAgICAgLSBwYXR0ZXJuOiB0eXBlXG4gICAgICAtIHBhdHRlcm46IEV4Y2VwdGlvblxuICAgICAgLSBwYXR0ZXJuOiBCYXNlRXhjZXB0aW9uXG5yZXdyaXRlcnM6XG4gIC0gaWQ6IHN0cmlwLXF1b3Rlc1xuICAgIHJ1bGU6XG4gICAgICBraW5kOiBzdHJpbmdcbiAgICAgIHBhdHRlcm46ICdcIiRRVU9URURcIidcbiAgICBmaXg6IFwiJFFVT1RFRFwiXG4gIC0gaWQ6IGhhbmRsZS1tdWx0aWxpbmUgIyBhZGQgdHdvIHF1b3RlcyBoZXJlLCBzbyB0aGUgZmluYWwgdHJhbnNmb3JtIHdpbGwgYWRkIHRoZSB0aGlyZCBwYWlyXG4gICAgcnVsZTpcbiAgICAgIHJlZ2V4OiBcIlxcblwiXG4gICAgICBwYXR0ZXJuOiAkTVVMVElMSU5FXG4gICAgZml4OiAnXCJcIiRNVUxUSUxJTkVcIlwiJ1xudHJhbnNmb3JtOlxuICBTVFJJUFBFRF9UWVBFOlxuICAgIHJld3JpdGU6XG4gICAgICByZXdyaXRlcnM6IFtzdHJpcC1xdW90ZXMsIGhhbmRsZS1tdWx0aWxpbmVdXG4gICAgICBzb3VyY2U6ICRUWVBFXG5cbmZpeDogJ1wiJFNUUklQUEVEX1RZUEVcIidcbiIsInNvdXJjZSI6ImZvbzogXCJsb2xcIiA9IGJhclxuZm9vOiBsb2wgPSBiYXJcblxuZGVmIGYoZm9vOiBpbnQgfCBiYXouYmF6IHwgTm9uZSB8IExpc3RbXCJpbnRcIl0gfCBTZXRbYmF6XSkgLT4gTGlzdFtMaXN0W2ludF1dOlxuICBwYXNzXG5cbmZvbzogTGlzdFtcImJhclwiXSA9IFtdXG5cbmlmIFRZUEVfQ0hFQ0tJTkc6XG4gIHBvb2w6IFwiXCJcIlVuaW9uW1xuICAgIGZvbywgYmFyXG4gIF1cIlwiXCJcbiAgcG9vbDogVW5pb25bXG4gICAgZm9vLCBiYXJcbiAgXVxuXG50eXBlKF8pLl9fY2FsbF9fID0gZm9vXG50eXBlIChfKSA9IGZvbyJ9) to quote any non-builtin unquoted type annotations (using `ast-grep scan --rule $PATH_TO_THE_YML sentry_sdk --interactive`)
5. Reformat: `uvx ruff format sentry_sdk`

### Manual changes

- The LibCST command made a bunch of silly mistakes like turning `foo = None # type: Foo # noqa` into `foo: "Foo # noqa" = None`, which were trivial to fix
- When turning `foo = None # type: Foo` into `foo: Foo = None`, mypy starts raising an error on the second form - I suppressed these using `type: ignore[assignment]`

### Test plan

`mypy` should continue being green, as evidenced by running `uvx tox -e linters`

### DIFF
--- a/sentry_sdk/_log_batcher.py
+++ b/sentry_sdk/_log_batcher.py
@@ -115,7 +115,7 @@ class LogBatcher:
 
     @staticmethod
     def _log_to_transport_format(log: "Log") -> "Any":
-        def format_attribute(val: int | float | str | bool) -> "Any":
+        def format_attribute(val: "int | float | str | bool") -> "Any":
             if isinstance(val, bool):
                 return {"value": val, "type": "boolean"}
             if isinstance(val, int):
--- a/sentry_sdk/integrations/asyncpg.py
+++ b/sentry_sdk/integrations/asyncpg.py
@@ -91,9 +91,9 @@ SubCursor = TypeVar("SubCursor", bound=BaseCursor)
 
 @contextlib.contextmanager
 def _record(
-    cursor: "SubCursor" | None,
+    cursor: "SubCursor | None",
     query: str,
-    params_list: "tuple[Any, ...]" | None,
+    params_list: "tuple[Any, ...] | None",
     *,
     executemany: bool = False,
 ) -> "Iterator[Span]":
--- a/sentry_sdk/integrations/langgraph.py
+++ b/sentry_sdk/integrations/langgraph.py
@@ -334,7 +334,7 @@ def _set_usage_data(span: "sentry_sdk.tracing.Span", messages: "Any") -> None:
         )
 
 
-def _set_response_model_name(span: "sentry_sdk.tracing.Span", messages: Any) -> None:
+def _set_response_model_name(span: "sentry_sdk.tracing.Span", messages: "Any") -> None:
     if len(messages) == 0:
         return
 


alexander-alderman-webb:
Hi @zsol,

Thank you for the PR. Upgrading the type annotation style is a breaking change because we still support Python 3.6.

We will do this work on the major branch, and can keep you updated.

charliermarsh:
My understanding is that [PEP 526](https://peps.python.org/pep-0526/) is compatible with Python 3.6 (but not Python 3.5), so IIUC this shouldn't break any Python 3.6 users.

alexander-alderman-webb:
You're correct that using PEP 526 is possible on Python 3.6. 

The reason we didn't do the type transition yet is that we use some newer types, such as [typing.Literal](https://docs.python.org/3/library/typing.html#typing.Literal) and [typing.TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict).

We currently type check on newer Python versions, while supporting 3.6 at runtime for users. This is possible because old-style type annotations together with `typing.TYPE_CHECKING` prevent user runtimes from having to import types.

You can achieve an analogous result on Python 3.7 with new-style type annotations, but I believe you require `__future__.annotations`, which is unfortunately only available in Python 3.7. See https://peps.python.org/pep-0563/.

sl0thentr0py:
if only `Literal` and `TypedDict` are the blockers, I'm fine with removing those since I massively dislike `TypedDict` anyway

alexander-alderman-webb:
It's mainly those two. There's a few instances of other new types like `typing.NotRequired` and `typing_extensions.Unpack`.

AlexWaygood:
> You can achieve an analogous result on Python 3.7 with new-style type annotations, but I believe you require `__future__.annotations`, which is unfortunately only available in Python 3.7. See [peps.python.org/pep-0563](https://peps.python.org/pep-0563/).

You can do the same on Python 3.6 without `__future__.annotations` if you're happy to use stringized/quoted annotations. Which this PR does.

alexander-alderman-webb:
Great to know! CI seems to fail because there are some places in which the annotations are not in quotes in this PR.

What is the motivation for the PR? We need to transition regardless, but for us to prioritize bringing this over the line it would be helpful to understand if there are limitations beyond comment-based types becoming less idiomatic.
Are there new tools for which the comment-based annotations do not work?

AlexWaygood:
> Great to know! CI seems to fail because there are some places in which the annotations are not in quotes in this PR.

Ah, cool, that's a bug in the PR! Thanks for the heads-up (cc. @zsol)

> What is the motivation for the PR? We need to transition regardless, but for us to prioritize bringing this over the line it would be helpful to understand if there are limitations beyond comment-based types becoming less idiomatic. Are there new tools for which the comment-based annotations do not work?

Yes. The main motivation is that newer type checkers such as [ty](https://github.com/astral-sh/ty) do not support the legacy syntax. (Full disclosure: I'm a ty developer and Zsolt's a colleague, though he doesn't work on ty.) It would be quite a bit of work for us to add support for the legacy syntax to ty, so we're reluctant to do so given that it's only required for compatibility with Python <=3.5. Meanwhile, we're big fans of Sentry, and have other code elsewhere that uses Sentry. We'd love to type-check that code with ty, but doing so currently produces poor results, because ty can't understand any of Sentry's legacy type hints.

Note that there have also been discussions about deprecating the legacy syntax from the type system altogether, albeit those discussions have been dormant for a while now:
- https://mail.python.org/archives/list/typing-sig@python.org/thread/66JDHQ2I3U3CPUIYA43W7SPEJLLPUETG/
- https://github.com/python/mypy/issues/12947

Other tools such as [pyflakes/flake8](https://github.com/PyCQA/pyflakes/pull/684) dropped support for type comments a while back, and [Ruff](https://github.com/astral-sh/ruff/issues/1619) has never supported them.

alexander-alderman-webb:
Thank you for the detailed explanation, and thank you for using Sentry!

We didn't transition yet because of internal misunderstanding that we need `__future__.annotations`. We are happy to support your efforts so that others can use `ty` with `sentry-python`.

The steps already in the PR descriptions are great since we'll verify the manual fixes. If you have a systematic way of putting all annotations in quotes keep us updated. Otherwise, I'll pick up from here next week.

zsol:
I'm happy to get this PR green for you!

zsol:
I believe I've resolved all the unquoted type annotations and pulled in changes from master. Let's see what CI says

alexander-alderman-webb:
Great, thanks for making CI green!

I'll run some diffs and then merge if everything looks good.

alexander-alderman-webb (approved):

zsol:
Thanks for the quick turnaround @alexander-alderman-webb!

PR #5213, opened 2025-12-10
ref: Make logs, metrics go via scope

sentrivana:
### Description
Logs and metrics were going through a completely separate pipeline compared to other events.

Conceptually, they're still different from regular events since they're more lightweight and attribute-based (no data, contexts, etc., everything is an attribute) so separate handling makes sense. However, the pipeline should still conceptually resemble the one we use for other event types, for consistency.

## Current pipeline for non-log, non-metric events
- The top-level API calls `scope.capture_XXX`. This merges the active scope stack (global + isolation + current scope) and calls `client.capture_XXX` with the resulting merged scope.
- `client.capture_XXX` contains virtually all of the logic, most notably:
  - It applies the scope to the event by calling `scope.apply_to_event`, populating contexts, user data, etc.
  - It serializes the event.
  - It constructs the final envelope and sends it to the transport.

## This PR
- Instead of the logging/metrics functionality going straight to `client.capture_XXX`, we call `scope.capture_XXX`, like we do for other event types, and then call `client.capture_XXX` from there.
- Instead of inlining (and duplicating) all the attribute logic, `client.capture_XXX` now calls a new `scope.apply_to_telemetry` function internally (akin to `scope.apply_to_event`, but sets attributes instead).
- The rest of the pipeline was left as-is for now, so metrics and logs are directly put into the batcher which itself serializes them. It's questionable whether making this part of the pipeline more similar to the event one would be a good idea since in Span First it'll be beneficial to have unserialized telemetry in the buffer, as is the case now with logs and metrics.

Additionally:
- Unify attribute-related types
- Move duplicated `format_attribute` to utils

Re: naming: I'm calling the new-style, attribute-based things simply "telemetry", since not all of them are events (for example, spans v2 which are coming with span streaming).

Note: I might refactor further. I'd like to have proper classes for Logs and Metrics and give them ownership of how to serialize themselves, how to call before_send, etc., but need to see whether there's a nice way to do this without breaking backwards compat (the log/metric needs to be a dict in before_send_x).

#### Reminders


sentry_sdk/client.py
```diff
@@ -898,132 +905,41 @@ def capture_event(
 
         return return_value
 
-    def _capture_log(self, log: "Optional[Log]") -> None:
-        if not has_logs_enabled(self.options) or log is None:
+    def _capture_telemetry(
```
sentrivana, line 901: This is now essentially a dispatcher on the `ty` argument. Unfortunately we need the explicit type and can't look at the actual type of the thing coming in since it's all just dicts. I'll be exploring whether we can make this better without introducing breaking changes.

sentry_sdk/client.py
```diff
@@ -217,10 +217,10 @@ def is_active(self) -> bool:
     def capture_event(self, *args: "Any", **kwargs: "Any") -> "Optional[str]":
         return None
 
-    def _capture_log(self, log: "Log") -> None:
+    def _capture_log(self, log: "Log", scope: "Scope") -> None:
```
sentrivana, line 220: `scope` here is the merged global + iso + current scope.

sentry_sdk/client.py
```diff
@@ -898,132 +898,41 @@ def capture_event(
...
+        if telemetry is None:
             return
 
-        if self.log_batcher:
-            self.log_batcher.add(log)
-
-    def _capture_metric(self, metric: "Optional[Metric]") -> None:
-        if not has_metrics_enabled(self.options) or metric is None:
-            return
+        scope.apply_to_telemetry(telemetry)
```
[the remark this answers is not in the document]
sentrivana: https://github.com/getsentry/sentry-python/pull/5213/commits/a62d90b897a3683803750b03c396af9fe642107d

sentry_sdk/_types.py
```diff
@@ -215,13 +215,36 @@ class SDKInfo(TypedDict):
     # TODO: Make a proper type definition for this (PRs welcome!)
     Hint = Dict[str, Any]
 
+    AttributeValue = (
+        str | bool | float | int | list[str] | list[bool] | list[float] | list[int]
+    )
```
alexander-alderman-webb, line 222: I'm probably out of the loop on this. According to the develop docs we don't currently support array attributes.
Either we should update the develop docs or this type is too permissive, at least when it comes to what services after the SDK support.

https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes

sentry_sdk/client.py
```diff
@@ -898,132 +898,41 @@ def capture_event(
...
-        if before_send_log is not None:
-            log = before_send_log(log, {})
+        scope.apply_to_telemetry(telemetry)
 
-        if log is None:
-            return
+        before_send_getter = {
+            "log": lambda: get_before_send_log(self.options),
+            "metric": lambda: get_before_send_metric(self.options),
+        }.get(ty)
```
alexander-alderman-webb: Could this return a function instead of a lambda that returns a function?

sl0thentr0py: imo this causes useless allocations, just use an if else

sentry_sdk/scope.py
```diff
@@ -1521,6 +1609,21 @@ def apply_to_event(
...
         return event
 
+    @_disable_capture
+    def apply_to_telemetry(self, telemetry: "Union[Log, Metric]") -> None:
+        # Attributes-based events and telemetry go through here (logs, metrics,
+        # spansV2)
+        trace_context = self.get_trace_context()
+        trace_id = trace_context.get("trace_id")
+        if telemetry.get("trace_id") is None:
+            telemetry["trace_id"] = trace_id or "00000000-0000-0000-0000-000000000000"
```
sl0thentr0py, line 1619: actually don't know if we need this 0000 fallback

sentry_sdk/client.py
sentrivana: https://github.com/getsentry/sentry-python/pull/5213/commits/3fffa7fc502f805a526f6bcee6e3a312e5422f0e

sentry_sdk/scope.py
sentrivana: It should never happen but there needs to be a string in there in case it does since it's required. I saw that JS for instance is falling back to empty string. I think some fallback would still be good to have.

sentry_sdk/_types.py
sentrivana: It's kind of new but we do now: https://develop.sentry.dev/sdk/telemetry/attributes/#supported-types Also tried directly and I see them in Sentry. Double-checking with the ingest team on slack though to make sure it's not just something that's turned on for the sentry-sdks org.

Does it actually say arrays are not supported on the page you posted? Will update but can't find anything re: array types on the page.

sentrivana: Ok apparently they are not supported yet, will update.

sentrivana: https://github.com/getsentry/sentry-python/pull/5213/commits/7ccbd5a89a01539e1603c19f5cdb1886647c5954

PR #5215, opened 2025-12-10
ref: Cleanup outgoing propagation_context logic

sl0thentr0py:
### Description

* Add `get_baggage`, `get_traceparent` and `iter_headers` to `PropagationContext` so that the contract is similar to `Span`. 
  * This clarifies that the `PropagationContext` and `Span` are themselves responsible for constructing these values and not the `Scope`. 
  * Currently, it is all a mess on `Scope` which does not clarify state responsibilities correctly on who holds and can populate what.
* Add `Baggage.populate_from_propagation_context` as a replacement for `Baggage.from_options` (that took a `scope` argument).
* Unify all the various `propagation_context` presence checks in `Scope` methods to `get_active_propagation_context`. 
  * This `get_active_propagation_context` logic then remains the central problem to fix in the future.

### Deprecations

The following methods are thus no longer used internally and deprecated.
* `Scope.get_dynamic_sampling_context`
* `Scope.iter_headers`
* `Baggage.from_options`

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -525,13 +525,11 @@ class Scope:
         """
         Returns the Dynamic Sampling Context from the Propagation Context.
         If not existing, creates a new one.
+
+        Deprecated: Logic moved to PropagationContext, don't use directly.
         """
         if self._propagation_context is None:
             return None
-
-        baggage = self.get_baggage()
-        if baggage is not None:
-            self._propagation_context.baggage = baggage
 
         return self._propagation_context.dynamic_sampling_context
 
@@ -547,16 +545,13 @@ class Scope:
         if has_tracing_enabled(client.options) and self.span is not None:
             return self.span.to_traceparent()
 
-        # If this scope has a propagation context, return traceparent from there
-        if self._propagation_context is not None:
-            traceparent = "%s-%s" % (
-                self._propagation_context.trace_id,
-                self._propagation_context.span_id,
-            )
-            return traceparent
+        # else return traceparent from the propagation context
+        propagation_context = self.get_active_propagation_context()
+        if propagation_context is not None:
+            return propagation_context.to_traceparent()
 
-        # Fall back to isolation scope's traceparent. It always has one
-        return self.get_isolation_scope().get_traceparent()
+        # TODO-neel will never happen
+        return None
 
     def get_baggage(self, *args, **kwargs):
         # type: (Any, Any) -> Optional[Baggage]
@@ -570,12 +565,13 @@ class Scope:
         if has_tracing_enabled(client.options) and self.span is not None:
             return self.span.to_baggage()
 
-        # If this scope has a propagation context, return baggage from there
-        if self._propagation_context is not None:
-            return self._propagation_context.baggage or Baggage.from_options(self)
+        # else return baggage from the propagation context
+        propagation_context = self.get_active_propagation_context()
+        if propagation_context is not None:
+            return propagation_context.get_baggage()
 
-        # Fall back to isolation scope's baggage. It always has one
-        return self.get_isolation_scope().get_baggage()
+        # TODO-neel will never happen
+        return None
 
     def get_trace_context(self):
         # type: () -> Dict[str, Any]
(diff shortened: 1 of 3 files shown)


sentry_sdk/scope.py
```diff
@@ -547,16 +545,13 @@ def get_traceparent(self, *args, **kwargs):
...
-            )
-            return traceparent
+        # else return traceparent from the propagation context
+        propagation_context = self.get_active_propagation_context()
+        if propagation_context is not None:
+            return propagation_context.get_traceparent()
 
-        # Fall back to isolation scope's traceparent. It always has one
-        return self.get_isolation_scope().get_traceparent()
+        # TODO-neel will never happen
```
sl0thentr0py, line 553: follow up PR
https://github.com/getsentry/sentry-python/pull/5217

sentry_sdk/tracing_utils.py
```diff
@@ -506,7 +508,28 @@ def span_id(self, value):
     @property
     def dynamic_sampling_context(self):
         # type: () -> Optional[Dict[str, Any]]
-        return self.baggage.dynamic_sampling_context() if self.baggage else None
+        return self.get_baggage().dynamic_sampling_context()
+
+    def get_traceparent(self):
```
sentrivana: Nit: making this consistent with the `Span` methods
```suggestion
    def to_traceparent(self):
```

sentry_sdk/tracing_utils.py
```diff
@@ -506,7 +508,28 @@ def span_id(self, value):
...
+        # type: () -> str
+        return f"{self.trace_id}-{self.span_id}"
+
+    def get_baggage(self):
```
sentrivana, line 517: Same nit as above
```suggestion
    def to_baggage(self):
```

sentrivana (commented):
Two small comments re naming for now, will still review properly, need to understand the new flow

sentrivana (approved):
Much, much better already, thank you!

sentry_sdk/tracing_utils.py
```diff
@@ -506,7 +508,28 @@ def span_id(self, value):
...
+        # type: () -> Baggage
+        if self.baggage is None:
+            self.baggage = Baggage.populate_from_propagation_context(self)
+        return self.baggage
+
+    def iter_headers(self):
```
alexander-alderman-webb, line 523: fyi on what's coming in the future:

For MCP trace propagation we may be sending and receiving `traceparent` instead of `sentry-trace` as part of the `_meta` field in the MCP protocol. See https://github.com/getsentry/sentry-docs/pull/15752#discussion_r2606375296.

So, we should think about whether we will parametrize this function or add a new function in future. If we treat this as a generic function for HTTP and future MCP use, we probably want to rename it to something like `iter_context_items()` since the MCP protocol does not have headers.

sentry_sdk/tracing_utils.py
sl0thentr0py: we actually don't use `Span.to_baggage` anywhere and the `Span.iter_headers` uses `get_baggage` from the transaction itself so I'll leave this as `get_baggage`.

sentry_sdk/tracing_utils.py
sl0thentr0py: done

sentry_sdk/tracing_utils.py
sl0thentr0py: `iter_context_items` is too generic since these are entities that are only attached to outgoing things.
It's fine to reuse this for MCP, you can make an explicit wrapper around this function for MCP usage to signify clarity of purpose like [`trace_propagation_meta`](https://github.com/getsentry/sentry-python/blob/598e59410015ca386af27843cc35771ce1261596/sentry_sdk/scope.py#L601) if you wish.

Basically, this function will construct all relevant cross-service propagation state (like otel's `Propagator.inject`) and then you can use it elsewhere.

sentry_sdk/scope.py
```diff
@@ -616,36 +612,19 @@ def trace_propagation_meta(self, *args, **kwargs):
...
-            )
-
-        baggage = self.get_baggage()
-        if baggage is not None:
-            meta += '<meta name="%s" content="%s">' % (
-                BAGGAGE_HEADER_NAME,
-                baggage.serialize(),
-            )
+        for name, content in self.iter_trace_propagation_headers():
+            meta += f'<meta name="{name}" content="{content}">'
```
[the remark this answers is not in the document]
sl0thentr0py: that's fine, intended behavior

Issue #5232, opened 2025-12-15
Pydantic AI integration erroneously considers `ToolRetryError` to be unhandled


JakobGM:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.47.0

### Description

Pydantic AI's internal machinery raises [`pydantic_ai.exceptions.ToolRetryError`](https://github.com/pydantic/pydantic-ai/blob/1df9ca608632c51e30c2ab172fdeb591dd80c125/pydantic_ai_slim/pydantic_ai/exceptions.py#L169-L174) when the output of a model does not satisfy the specified output schema. This exception is then caught further up the call-stack by Pydantic AI, which allows the LLM to be provided with the error and try again.

Sentry SDK's `PydanticAIIntegration` erroneously considers this exception to be unhandled, I think [here](https://github.com/getsentry/sentry-python/blob/a6821989907f3c06bd8b79a8e0223a2e38ec7620/sentry_sdk/integrations/pydantic_ai/patches/tools.py#L86-L88) in `_patch_tool_execution()`. This type of exception should probably be ignored by Sentry.

If you have access to our Sentry account for debugging purposes, you can take a look at [this issue](https://crdbrd.sentry.io/issues/79973140/?project=4509354151313488) which should not have been an issue in the first place. If not, here are some screenshots:

### Screenshots
<img width="1590" height="837" alt="Image" src="https://github.com/user-attachments/assets/788797ff-ef4a-48fd-befd-90b473e15964" />
<img width="1546" height="919" alt="Image" src="https://github.com/user-attachments/assets/e9ac468a-5bdc-4778-9fd0-f1fc033b0279" />

alexander-alderman-webb:
Hi @JakobGM,

Thanks for the report and explanation. We will address this and keep you updated.

alexander-alderman-webb:
Hi @JakobGM,

We added an option to suppress these exceptions, and now report them as handled. Please initialize the integration as follows to stop seeing issues of this kind in your Sentry.

```python
sentry_init(
    ...
    integrations=[PydanticAIIntegration(handled_tool_call_exceptions=False)],
)
```

JakobGM:
Great, thanks a lot 🙇

PR #5237, opened 2025-12-15
fix(pydantic-ai): Stop capturing internal exceptions

alexander-alderman-webb:
### Description

Stop capturing broad exceptions, that bubble up to top-level run functions, at the tool call level for Pydantic AI. Only capture `ToolRetryError` at the tool call level, and mark these exceptions as handled.

#### Issues

Closes https://github.com/getsentry/sentry-python/issues/5232

#### Reminders

### DIFF
--- a/sentry_sdk/integrations/pydantic_ai/patches/tools.py
+++ b/sentry_sdk/integrations/pydantic_ai/patches/tools.py
@@ -20,7 +20,7 @@ except ImportError:
 
 try:
     from pydantic_ai._tool_manager import ToolManager  # type: ignore
-    from pydantic_ai.exceptions import ToolRetryError
+    from pydantic_ai.exceptions import ToolRetryError  # type: ignore
 except ImportError:
     raise DidNotEnable("pydantic-ai not installed")
 


sentry_sdk/integrations/pydantic_ai/patches/tools.py
```diff
@@ -73,18 +70,14 @@ async def wrapped_call_tool(
...
+                        self,
+                        call,
+                        *args,
+                        **kwargs,
+                    )
+                    update_execute_tool_span(span, result)
+                    return result
 
         # No span context - just call original
         return await original_call_tool(
```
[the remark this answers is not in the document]
alexander-alderman-webb: looks valid but not the focus of this PR

sentry_sdk/integrations/pydantic_ai/patches/tools.py
```diff
@@ -73,18 +70,14 @@ async def wrapped_call_tool(
...
-                        _capture_exception(exc)
-                        raise exc from None
+                    result = await original_call_tool(
+                        self,
+                        call,
+                        *args,
+                        **kwargs,
+                    )
+                    update_execute_tool_span(span, result)
+                    return result
```
sentrivana: I wonder if removing the exception capturing altogether is the way to go. If a user has errors happening in their tool, even if they don't necessarily bubble all the way up, they'll lose visibility into them. If e.g. a tool is consistently providing erroneous output, the dev might want to know about it and fix it, even if the whole LLM chain manages to recover.

What we could consider:
- still capturing other exceptions, just not `ToolRetryError` (that wouldn't fix the example I talk about above, but it'd at least make sure we don't stop capturing other potentially important exceptions)
- see which of the exceptions captured here will be handled by Pydantic AI internally and capture those with `handled=True`
- add an integration-level option to turn tool exceptions on/off

Since I have little experience in this type of development though, might be this is also fine as is and no one cares about recoverable errors in their tools. 😅

alexander-alderman-webb: good points, I did some more investigation.

- Other errors bubble all the way up, so if we catch them here we're catching them multiple times. At least on the code path in

```
_agent_graph.process_tool_calls()
-> _agent_graph._call_tools()
-> _agent_graph._call_tool()
-> ToolManager.handle_call()
-> ToolManager._call_tool()`
```

A few places call `process_tool_calls()`, but only `ToolRetryError` is handled, and experimentally, other errors bubble up to the top-level run methods.
- I'm open to this. Tool authors can raise `ModelRetry` to trigger a `ToolRetryError`, or `ToolRetryError` can result from a validation failure. These are documented, and the only two ways to raise `ToolRetryError` in `ToolManager._call_tool()`. See https://ai.pydantic.dev/tools-advanced/#tool-retries.
- We should add the option if we capture `ToolRetryError`.

sentrivana: Thanks for investigating!

I'm leaning towards the following:
- removing the broad `except Exception` and replacing it with `except ToolRetryError`, to prevent double capture of other errors
- gating the actual `ToolRetryError` capture behind an option (unsure about the default)
- if the option is on and we end up capturing a `ToolRetryError`, capturing it with `handled=True`

WDYT?

alexander-alderman-webb: I agree, was already working on this approach.

I set the option to `True`, since this is useful information in my view.

sentrivana:
Please also add to https://docs.sentry.io/platforms/python/integrations/pydantic-ai/#options 🙏🏻

PR #5239, opened 2025-12-16
Convert all remaining type annotations into the modern format

zsol:
### Description
Same as #5206 but for the `scripts/` and `tests/` folder (plus there was one additional place in `sentry_sdk` that snuck in).


alexander-alderman-webb:
Thanks for the PR @zsol, it's nice to have one approach to typing in the repo!

We don't actually type check `scripts/` or `tests/` at the moment, but that's a problem for the future.

PR #5251, opened 2025-12-17
fix(integrations): pydantic-ai: properly format binary input message parts to be conformant with the `gen_ai.request.messages` structure

constantinius:
#### Issues
Closes https://linear.app/getsentry/issue/TET-1634/redact-images-pydantic-ai

### DIFF
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -7,6 +7,7 @@ on:
     - reopened
     - edited
     - labeled
+    - unlabeled
 permissions:
   contents: write
   pull-requests: write
--- a/sentry_sdk/_types.py
+++ b/sentry_sdk/_types.py
@@ -217,9 +217,18 @@ if TYPE_CHECKING:
     Hint = Dict[str, Any]
 
     AttributeValue = (
-        str | bool | float | int
-        # TODO: relay support coming soon for
-        # | list[str] | list[bool] | list[float] | list[int]
+        str
+        | bool
+        | float
+        | int
+        | list[str]
+        | list[bool]
+        | list[float]
+        | list[int]
+        | tuple[str, ...]
+        | tuple[bool, ...]
+        | tuple[float, ...]
+        | tuple[int, ...]
     )
     Attributes = dict[str, AttributeValue]
 
@@ -232,11 +241,10 @@ if TYPE_CHECKING:
                 "boolean",
                 "double",
                 "integer",
-                # TODO: relay support coming soon for:
-                # "string[]",
-                # "boolean[]",
-                # "double[]",
-                # "integer[]",
+                "string[]",
+                "boolean[]",
+                "double[]",
+                "integer[]",
             ],
             "value": AttributeValue,
         },
--- a/sentry_sdk/integrations/anthropic.py
+++ b/sentry_sdk/integrations/anthropic.py
@@ -11,6 +11,7 @@ from sentry_sdk.ai.utils import (
     normalize_message_roles,
     truncate_and_annotate_messages,
     get_start_span_function,
+    transform_anthropic_content_part,
 )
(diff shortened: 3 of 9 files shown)


sentry_sdk/ai/utils.py
```diff
@@ -147,12 +147,7 @@ def redact_blob_message_parts(
     messages: "List[Dict[str, Any]]",
 ) -> "List[Dict[str, Any]]":
     """
-    Redact blob message parts from the messages by replacing blob content with "[Filtered]".
-
-    This function creates a deep copy of messages that contain blob content to avoid
-    mutating the original message dictionaries. Messages without blob content are
-    returned as-is to minimize copying overhead.
-
+    Redact blob message parts from the messages, by removing the "content" key.
```
sentrivana: This sounds a bit misleading. The `"content"` key is not removed, just its value is replaced.

sentry_sdk/integrations/pydantic_ai/spans/ai_client.py
```diff
@@ -124,8 +140,13 @@ def _set_input_messages(span: "sentry_sdk.tracing.Span", messages: "Any") -> Non
                         formatted_messages.append(message)
 
         if formatted_messages:
+            normalized_messages = normalize_message_roles(formatted_messages)
+            scope = sentry_sdk.get_current_scope()
+            messages_data = truncate_and_annotate_messages(
+                normalized_messages, span, scope
+            )
```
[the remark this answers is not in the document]
constantinius: Not sure what PII has to do with anything. Think this is a fluke

sentry_sdk/integrations/pydantic_ai/spans/ai_client.py
```diff
@@ -107,6 +114,15 @@ def _set_input_messages(span: "sentry_sdk.tracing.Span", messages: "Any") -> Non
                             for item in part.content:
                                 if isinstance(item, str):
                                     content.append({"type": "text", "text": item})
+                                elif BinaryContent and isinstance(item, BinaryContent):
+                                    content.append(
+                                        {
+                                            "type": "blob",
+                                            "modality": item.media_type.split("/")[0],
+                                            "mime_type": item.media_type,
+                                            "content": f"data:{item.media_type};base64,{base64.b64encode(item.data).decode('utf-8')}",
```
alexander-alderman-webb: I don't think we should encode `item.data` here because it is stripped away anyway by the new blob redaction. I assume it's an operation that might fail, and if we don't emit the result, there's no point risking it.

can we just remove the `"content"` entry from the dictionary here? The assignment to `BLOB_DATA_SUBSTITUTE` will happen down the line anyway.

constantinius: As you said theres two considerations: making the conversion that could fail and doing work that may be undone later.

ad 1: I think we should make it failsafe, try/catch and simply don't set the content if we fail
ad 2: In the sense of the single responsibility principle: the integrations job is to properly report the messages back and forth and IMHO should not be concerned with redaction. Because if we later decide that we do not want to redact we would just need to change one area, not look through all integrations. Sure, a little contrived example, but I would consider it cleaner. 

On the other hand I'm not the primary maintainer, so if you prefer to change it I would be okay with that. What do you say?

alexander-alderman-webb: I'm thinking more about the contract that we have with the SDK users that we do the minimum work that is reasonable to collect the telemetry we emit and nothing else.

So, while I generally agree with your point about single-responsibility, I would class decoding a potentially large object as non-trivial enough that we shouldn't do it now. But yes, we will have to keep in mind that some downstream logic is bleeding into the integration as maintainers 😄

constantinius: Done

alexander-alderman-webb (approved):

PR #5256, opened 2025-12-18
feat(api): Add `Scope.set_attribute`

sentrivana:
### Description
Added new scope APIs: `set_attribute` and `remove_attribute`
- Scopes now store their own attributes in `_attributes`
- Attributes set on the scope are applied to logs and metrics

Also:
- unify pre-serialization of attribute values a bit (`format_attribute`)

**Resources:**
- [Spec](https://develop.sentry.dev/sdk/telemetry/scopes/#setting-attributes)

#### Issues
* Closes https://github.com/getsentry/sentry-python/issues/5210
* Closes https://linear.app/getsentry/issue/PY-2009/implement-set-attributes-api

#### Reminders

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -1636,9 +1636,9 @@ class Scope:
         if telemetry.get("span_id") is None and span_id:
             telemetry["span_id"] = span_id
 
-        self._apply_global_attributes_to_telemetry(telemetry)
         self._apply_scope_attributes_to_telemetry(telemetry)
         self._apply_user_attributes_to_telemetry(telemetry)
+        self._apply_global_attributes_to_telemetry(telemetry)
 
     def update_from_scope(self, scope: "Scope") -> None:
         """Update the scope with another scope's data."""


sentry_sdk/scope.py
```diff
@@ -1487,6 +1495,13 @@ def _apply_global_attributes_to_telemetry(
         if release is not None and "sentry.release" not in attributes:
             attributes["sentry.release"] = release
 
+    def _apply_scope_attributes_to_telemetry(
+        self, telemetry: "Union[Log, Metric]"
+    ) -> None:
+        for attribute, value in self._attributes.items():
+            if attribute not in telemetry["attributes"]:
```
alexander-alderman-webb, line 1502: We should consider attributes set in `_apply_global_attributes_to_telemetry()` when it comes to precedence.

I would have expected scope attributes to override `_apply_global_attributes_to_telemetry()`, but I'm also unsure about my intuition here.

Also, probably best to stay compatible with the order enforced after the todo below the method.

> TODO: Global stuff like this should just be retrieved at init time and
    put onto the global scope's attributes and then applied to events
    from there

sentry_sdk/scope.py
```diff
@@ -1710,6 +1731,22 @@ def flags(self) -> "FlagBuffer":
...
         return self._flags
 
+    def set_attribute(self, attribute: str, value: "AttributeValue") -> None:
+        """
+        Set an attribute on the scope.
+
+        Any attributes-based telemetry (logs, metrics) captured while this scope
+        is active will inherit attributes set on the scope.
+        """
+        self._attributes[attribute] = format_attribute(value)
```
alexander-alderman-webb, line 1741: No action required.
The spec is unintuitive here, seems to suggest that attributes on a current scope only apply to a single metric, rather than all metrics sent under the scope.

> Attributes set on the current scope MUST be applied only to the current log and current metric

alexander-alderman-webb (changes requested):

sentry_sdk/scope.py
sentrivana: I think this should take care of the precedence: https://github.com/getsentry/sentry-python/pull/5256/commits/b858becba66e0a452a96262b41cc522121b1633c It's somehow a bit tricky to reason about because we're doing it the wrong way around, essentially: instead of applying the global scope attributes first, then iso, then current, and then attributes set explicitly on the log/metric, overwriting as we go, we walk the chain backwards, with a lot of "only add if the attr is not there" checks.

But: I'll be setting these attributes on the scope directly in another PR, so the whole `set_global_attributes` function will go away, and scope merging will take care of the precedence, which is cleaner.

sentry_sdk/scope.py
sentrivana: Yep this is definitely worded weird. https://github.com/getsentry/sentry-docs/pull/15866/changes

sentry_sdk/scope.py
```diff
@@ -1691,6 +1710,8 @@ def update_from_kwargs(
             self._tags.update(tags)
         if fingerprint is not None:
             self._fingerprint = fingerprint
+        if attributes is not None:
+            self._attributes.update(attributes)
```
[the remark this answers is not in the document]
sentrivana: This is not part of the public API, and also not used for setting attributes. The only way `update_from_kwargs` comes to play is when capturing non-attribute-based events.

PR #5271, opened 2025-12-31
fix(ai): Support distributed tracing for MCP tool calls

maxclaus:
### Description

MCP tool calls now properly continue distributed traces from parent services instead of creating new transactions with new trace IDs.

**Problem**: When `get_start_span_function()` executes during MCP tool handler wrapping, the HTTP transaction isn't accessible due to async context isolation. The function only checks the current scope, which often contains unrelated spans with different trace IDs.

**Solution**: Enhanced `get_start_span_function()` to check for sentry-trace headers in MCP request context. Added `_create_transaction_from_mcp_headers()` that extracts headers and uses `continue_trace()` to create transactions inheriting trace_id and parent_span_id.

**Benefits**:
- MCP tools join distributed traces correctly
- Transactions still appear in MCP Insights dashboard
- Fully backward compatible
- No new dependencies

#### Issues

N/A (External contribution)

### DIFF
--- a/sentry_sdk/ai/utils.py
+++ b/sentry_sdk/ai/utils.py
@@ -100,11 +100,85 @@ def normalize_message_roles(messages: "list[dict[str, Any]]") -> "list[dict[str,
 
 
 def get_start_span_function() -> "Callable[..., Any]":
+    """
+    Determine whether to create a span or transaction for AI operations.
+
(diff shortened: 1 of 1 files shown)


maxclaus:
I am closing this PR and will open an issue next to discuss about it first.

PR #5288, opened 2026-01-08
feat(asyncio): Add on-demand way to enable AsyncioIntegration

sentrivana:
Initializing the SDK in an async environment can be a challenge, since the requirement to initialize as early as possible might clash with the requirement to initialize when there's an event loop running, if one wants to use the AsyncioIntegration. These sort of scenarios would benefit from a dedicated way to activate the integration after `init()` has been called.

Supersedes the needlessly more general https://github.com/getsentry/sentry-python/pull/5285

### DIFF
--- a/sentry_sdk/integrations/asyncio.py
+++ b/sentry_sdk/integrations/asyncio.py
@@ -182,6 +182,9 @@ def enable_asyncio_integration(*args: "Any", **kwargs: "Any") -> None:
     # we patch the currently running event loop, and we keep the record of doing
     # that on the loop itself.
     logger.debug("Setting up integration asyncio")
+
     integration = AsyncioIntegration(*args, **kwargs)
     integration.setup_once()
-    client.integrations["asyncio"] = integration
+
+    if "asyncio" not in client.integrations:
+        client.integrations["asyncio"] = integration


sentry_sdk/integrations/__init__.py
```diff
@@ -279,6 +280,28 @@ def setup_integrations(
...
 
 
+def _enable_integration(integration: "Integration") -> "Optional[Integration]":
+    identifier = integration.identifier
+    client = sentry_sdk.get_client()
+
+    with _installer_lock:
+        if identifier in client.integrations:
+            logger.debug("Integration already enabled: %s", identifier)
+            return None
```
sentrivana: I'm explicitly not allowing overwriting the old integration with the new one so that we don't end up double patching the event loop. We could add extra logic to restore the original code and re-patch that, but it seems like that's a lot of extra work for a usecase no one will probably need.

sentry_sdk/integrations/asyncio.py
```diff
@@ -138,3 +138,39 @@ class AsyncioIntegration(Integration):
...
+    """
+    client = sentry_sdk.get_client()
+    if not client.is_active():
+        return
+
+    integration = _enable_integration(AsyncioIntegration(*args, **kwargs))
+    if integration is None:
+        return
+
+    client.integrations[integration.identifier] = integration
```
alexander-alderman-webb: it could make sense to move this assignment to `_enable_integration()`?

sentry_sdk/client.py
```diff
@@ -188,6 +188,7 @@ def __init__(self, options: "Optional[Dict[str, Any]]" = None) -> None:
         self.monitor: "Optional[Monitor]" = None
         self.log_batcher: "Optional[LogBatcher]" = None
         self.metrics_batcher: "Optional[MetricsBatcher]" = None
+        self.integrations: "dict[str, Integration]" = {}
```
alexander-alderman-webb, line 191: For my understanding, and no action required:

do you know if there is a reason we have both `Client.integrations` and the `sentry_sdk.integrations._installed_integrations` global?

sentry_sdk/integrations/asyncio.py
sentrivana: Moved the whole logic to `_enable_integration`, because why not -- we were accessing the client in both and there was no clear distinction why something is done in the API function vs in `_enable_integration`.

sentry_sdk/client.py
sentrivana: Afaik the general idea is:

- `_installed_integrations` keeps track of what integrations have had their `setup_once` already run in the current process. `setup_once` should, as the name suggests, only be run (i.e., monkeypatch things) once, so even if you, for example, set up a client with an integration and then set up a different client with the same integration, the latter client should not run `setup_once` on that integration again. (Setting up multiple clients is in general a very niche scenario though.)
- On the other hand, `client.integrations` tracks the integrations that a specific client is using. So for example, if you'd enabled integration A before in another client, its `setup_once` would have run, and it would be in `_installed_integrations`. If you then create another client that **shouldn't** have A active, A will not have its `setup_once` run again, but it will still be patched, and we'll use `client.integrations` to check whether the patch should actually be applied or if we should exit early ([this pattern](https://github.com/getsentry/sentry-python/blob/4c4f2600a849e02210ed4013623abd395c49b86c/sentry_sdk/integrations/flask.py#L97)).

So `_installed_integrations` is process-wide and means "this package has been patched", and `client.integrations` is client-specific, saying "this patch should actually be applied", to allow for these sorts of multi-client scenarios.

sentrivana: Now that I've looked into this a bit more I see that I'm not checking `_installed_integrations` correctly in this PR. Will fix

sentrivana: So AsyncioIntegration is special in that it patches the current event loop, and not anything global in asyncio, so it actually should not be affected by `_installed_integrations` 🤡 https://github.com/getsentry/sentry-python/pull/5288/changes/36ef8cb084b6108e9752c38887dc4277192cd399

alexander-alderman-webb: For the server frameworks there should be one loop per process afaict, but letting user's re-initialize would still be good if a user also initializes the SDK with `AsyncioIntegration` before the event loop was started.

So looks good to me (apart from Seer's comment that may need to be addressed)!

sentrivana: Added a bool to the patched event loop to be able to tell whether it's already been patched. That should take care of always being able to patch if not patched, and avoiding double-patching. And since this completely bypasses the `_installed_integrations` machinery (because knowing if we've patched in the current process has no value in the case of the asyncio integration), I opted for making everything specific to the `AsyncioIntegration` only instead of having a more general `_enable_integration` function.

sentry_sdk/integrations/asyncio.py
```diff
@@ -138,3 +143,48 @@ class AsyncioIntegration(Integration):
...
+
+    # This function purposefully bypasses the integration machinery in
+    # integrations/__init__.py. _installed_integrations/_processed_integrations
+    # is used to prevent double patching the same module, but in the case of
+    # the AsyncioIntegration, we don't monkeypatch the standard library directly,
+    # we patch the currently running event loop, and we keep the record of doing
+    # that on the loop itself.
+    logger.debug("Setting up integration asyncio")
+
+    integration = AsyncioIntegration(*args, **kwargs)
```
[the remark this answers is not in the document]
sentrivana: This is fine, they're there to make this future-proof in case we add args to the integration.

PR #5290, opened 2026-01-09
ci(release): Switch from action-prepare-release to Craft

BYK:
## Summary

This PR migrates from the deprecated `action-prepare-release` to the new Craft GitHub Actions.

## Changes

- Migrated `.github/workflows/release.yml` to Craft reusable workflow

## Documentation

See https://getsentry.github.io/craft/github-actions/ for more information.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     timeout-minutes: 10
 
     steps:
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6 # v6 # v2.0.1
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       - uses: actions/setup-python@v6
         with:
           python-version: 3.14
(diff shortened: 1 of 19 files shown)


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     timeout-minutes: 10
 
     steps:
-      - uses: actions/checkout@v6.0.1
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6 # v6 # v2.0.1
```
sentrivana: Is this an intentional downgrade to 2.0? Also in other files

.github/workflows/release.yml
```diff
@@ -1,38 +1,40 @@
...
-        uses: getsentry/action-prepare-release@v1
-        env:
-          GITHUB_TOKEN: ${{ steps.token.outputs.token }}
-        with:
-          version: ${{ github.event.inputs.version }}
-          force: ${{ github.event.inputs.force }}
-          merge_target: ${{ github.event.inputs.merge_target }}
+    - name: Get auth token
+      id: token
+      uses: actions/create-github-app-token@29824e69f54612133e76f7eaac726eef6c875baf # v2 # v2 # v2
```
sentrivana: Extra comments here and in other places

.github/workflows/release.yml
```diff
@@ -1,38 +1,40 @@
 name: Release
-
 on:
   workflow_dispatch:
     inputs:
       version:
-        description: Version to release
-        required: true
+        description: Version to release (or "auto")
+        required: false
```
sentrivana, line 7: If not provided, does this default to `auto`? If not, should we add logic to L38's `version: ${{ inputs.version }}` to set to `auto` if not provided?

Issue #5298, opened 2026-01-12
AttributeError: module 'channels' has no attribute '__version__'


javiertejero:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.49.0

### Steps to Reproduce

If `channels` package (e.g. a private package) does not contain `__version__` property, a simple `sentry_sdk.init()` fails with `AttributeError: module 'channels' has no attribute '__version__'`

The code in `sentry_sdk/integrations/threading.py:59` 
```
try:
    from django import VERSION as django_version  # noqa: N811
    import channels  # type: ignore[import-untyped]

    channels_version = channels.__version__  # < raises AttributeError
except ImportError:
    django_version = None
    channels_version = None
```

### Expected Result

`sentry_sdk.init()` should work regardless of the contents of `channels` module

### Actual Result

```
import sentry_sdk
sentry_sdk.init()
...
File ~/.../python3.11/site-packages/sentry_sdk/integrations/threading.py:59, in ThreadingIntegration.setup_once()
     56     from django import VERSION as django_version  # noqa: N811
     57     import channels  # type: ignore[import-untyped]
---> 59     channels_version = channels.__version__
     60 except ImportError:
     61     django_version = None

AttributeError: module 'channels' has no attribute '__version__'
```

sentrivana:
Thanks @javiertejero for bringing this to our attention, we'll have a fix for this soon.

PR #5299, opened 2026-01-12
fix(threading): Handle channels shadowing

sentrivana:
### Description
Folks might have a `channels` module that has nothing to do with the `channels` package we're expecting. This will then make the code throw an `AttributeError` which we don't handle.

#### Issues
* Closes https://github.com/getsentry/sentry-python/issues/5298

#### Reminders

### DIFF
--- a/sentry_sdk/integrations/threading.py
+++ b/sentry_sdk/integrations/threading.py
@@ -54,11 +54,14 @@ class ThreadingIntegration(Integration):
 
         try:
             from django import VERSION as django_version  # noqa: N811
+        except ImportError:
+            django_version = None
+
+        try:
             import channels  # type: ignore[import-untyped]
 
             channels_version = channels.__version__
         except (ImportError, AttributeError):
-            django_version = None
             channels_version = None
 
         is_async_emulated_with_threads = (


alexander-alderman-webb (approved):
💯

sentry_sdk/integrations/threading.py
```diff
@@ -57,7 +57,7 @@ def setup_once() -> None:
             import channels  # type: ignore[import-untyped]
 
             channels_version = channels.__version__
-        except ImportError:
+        except (ImportError, AttributeError):
```
javiertejero, line 64: this implies that if there is an AttributeError because of channels line, `django_version` will be set to `None` and that is not ideal... perhaps we should split this try/except in 2 parts, 1 for `django` and another for `channels`?

javiertejero, line 64: this implies that if there is an AttributeError because of channels line, `django_version` will be set to `None` and that is not ideal... perhaps we should split this try/except in 2 parts, 1 for `django` and another for `channels`?

sentrivana: That's ok, `django_version` and `channels_version` both are only used for warning about a very specific Django/channels/sentry-sdk incompatibility that only appears in specific versions.

But you raise a good point in that it's not clear what the variable is used for. It might happen that someone extends the logic at some point using `django_version` for some other purpose, not realizing it's not always populated correctly. So it's worth making this nicer.

sentrivana: https://github.com/getsentry/sentry-python/pull/5299/changes/d34f79657614b0162e0ec5a517c6a2586ddf6625

Issue #5335, opened 2026-01-19
Adjust AI messages trimming logic to only keep the latest message


sentrivana:
Adjust the logic in https://github.com/getsentry/sentry-python/blob/1970486d0069165fa4a4313a9e36a155a90710f1/sentry_sdk/ai/utils.py#L648 to always only include the last (newest) message rather than all that fit into the 20kB limit.

All our integrations should use the function linked above already, so changing the logic in it should be all that's necessary (and adjusting the tests).

Context (sorry, Sentry employees only): https://sentry.slack.com/archives/C06KVE7V0DA/p1768578892574999?thread_ts=1768311204.311329&cid=C06KVE7V0DA

PR #5351, opened 2026-01-20
fix(openai): normalize Pydantic objects before JSON serialization in truncate_messages_by_size

### LINKED ISSUE #5350: OpenAI Integration: `truncate_messages_by_size` fails on Pydantic SDK objects (e.g., `ResponseFunctionToolCall`)
### How do you use Sentry?

Sentry Saas ([sentry.io](<http://sentry.io>))

### Version

2.49.x (and likely all versions with the OpenAI integration)

### Steps to Reproduce

1. Enable the OpenAI integration with `sentry-sdk[openai]`
2. Use the OpenAI Responses API with tool calls
3. When building the follow-up API call after tool execution, include SDK objects (like `ResponseFunctionToolCall`) in the input list - as shown in OpenAI's documentation

```python
from openai.types.responses import ResponseFunctionToolCall

# After a tool call, building input for follow-up request
input_list = [
    {"role": "user", "content": "Please notify the team"},
    tool_call,  # ResponseFunctionToolCall - a Pydantic object
    {"type": "function_call_output", "call_id": tool_call.call_id, "output": "Success"},
]

# This triggers Sentry's instrumentation which calls json.dumps() on input_list
response = client.responses.create(model="gpt-4o", input=input_list)
```

### Expected Result

Sentry's instrumentation should handle Pydantic/SDK objects gracefully, either by:

* Calling `.model_dump()` on them (as `_normalize_data()` already does)
* Using a fallback serialization strategy

### Actual Result

```
TypeError: Object of type ResponseFunctionToolCall is not JSON serializable
```

The error originates from `sentry_sdk/ai/utils.py:truncate_messages_by_size()` which calls `json.dumps(messages)` directly without normalizing Pydantic objects first.

### Root Cause

The SDK already has `_normalize_data()` in `sentry_sdk/ai/utils.py` that correctly handles Pydantic objects:

```python
def _normalize_data(data: "Any", unpack: bool = True) -> "Any":
    # convert pydantic data (e.g. OpenAI v1+) to json compatible format
    if hasattr(data, "model_dump"):
        return _normalize_data(data.model_dump(), unpack=unpack)
    ...
```

However, `truncate_messages_by_size()` doesn't use it:

```python
def truncate_messages_by_size(messages, ...):
    serialized_json = json.dumps(messages, separators=(",", ":"))  # No normalization!
    ...
```

### Suggested Fix

Call `_normalize_data()` before `json.dumps()`:

```python
def truncate_messages_by_size(messages, ...):
    normalized = _normalize_data(messages, unpack=False)
    serialized_json = json.dumps(normalized, separators=(",", ":"))
    ...
```

Or apply the same fix in `truncate_and_annotate_messages()` before calling `truncate_messages_by_size()`.

### Additional Context

This is the same class of bug as [PY-1885](https://linear.app/getsentry/issue/PY-1885/upgrade-openai-and-use-the-new-omit-convention-instead-of-not-given) (handling `Omit`/`NotGiven` types) - the OpenAI integration assumes inputs are plain Python dicts, but OpenAI's SDK uses Pydantic objects throughout.

OpenAI's own documentation shows passing SDK objects directly in follow-up requests, so this is a common pattern that Sentry's instrumentation should handle.

mecampbellsoup:
## Summary

The OpenAI SDK v1+ returns Pydantic model instances (e.g., `ResponseFunctionToolCall`) which are not directly JSON serializable. When these objects are passed to `truncate_messages_by_size()`, the `json.dumps()` call fails with a `TypeError`:

```
TypeError: Object of type ResponseFunctionToolCall is not JSON serializable
```

The `_normalize_data()` helper already exists in this file and properly handles Pydantic objects by calling `.model_dump()`, but it wasn't being used in `truncate_messages_by_size()`.

## The Fix

This PR adds normalization at the start of `truncate_messages_by_size()` to ensure all Pydantic objects are converted to JSON-compatible dicts before serialization:

```python
# Normalize messages to ensure JSON serialization works
# (handles Pydantic objects from OpenAI SDK v1+)
messages = _normalize_data(messages, unpack=False)
```

## Reproduction

When using the OpenAI SDK's Responses API with tool calls and Sentry's OpenAI integration enabled:

```python
from openai import OpenAI
from openai.types.responses import ResponseFunctionToolCall

client = OpenAI()
response = client.responses.create(
    model="gpt-4o",
    input=[
        {"role": "user", "content": "What's the weather?"},
        {
            "role": "assistant", 
            "content": [ResponseFunctionToolCall(
                type="function_call",
                id="call_123",
                call_id="call_123",
                name="get_weather",
                arguments='{"location": "NYC"}'
            )]
        }
    ]
)
```

This crashes with the `TypeError` because Sentry's integration tries to serialize the messages for tracing.

Fixes #5350

### DIFF
--- a/sentry_sdk/ai/utils.py
+++ b/sentry_sdk/ai/utils.py
@@ -660,6 +660,10 @@ def truncate_messages_by_size(
     In the single message case, the serialized message size may exceed `max_bytes`, because
     truncation is based only on character count in that case.
     """
+    # Normalize messages to ensure JSON serialization works
+    # (handles Pydantic objects from OpenAI SDK v1+)
+    messages = _normalize_data(messages, unpack=False)
+
     serialized_json = json.dumps(messages, separators=(",", ":"))
     current_size = len(serialized_json.encode("utf-8"))
 


alexander-alderman-webb:
Hi @mecampbellsoup,

Thanks for raising the bug, we're adjusting the trimming logic for the next release, so there will not be any serialization in the integrations any longer. See https://github.com/getsentry/sentry-python/issues/5335 for details.

PR #5393, opened 2026-01-29
fix(arq): handle settings_cls passed as keyword argument

nc9:
The `_sentry_create_worker` patch assumed `settings_cls` is always passed as positional arg (`args[0]`), causing IndexError when users call `create_worker(settings_cls=WorkerSettings)`.

Now checks both `args[0]` and `kwargs.get("settings_cls")`.

Test added in `tests/integrations/arq/test_arq.py` as `init_arq_with_kwarg_settings`

### Description
`settings_cls` is now checked from kwargs if no args

```py
settings_cls = args[0] if args else kwargs.get("settings_cls")
```

#### Issues
* None

### DIFF
--- a/sentry_sdk/integrations/arq.py
+++ b/sentry_sdk/integrations/arq.py
@@ -200,13 +200,14 @@ def patch_create_worker() -> None:
                 ]
 
         if hasattr(settings_cls, "functions"):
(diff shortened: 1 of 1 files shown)


nc9:
thanks was about to apply it :)

PR #5398, opened 2026-01-30
feat(span-streaming): Add span batcher (2)

sentrivana:
- Add a specialized `SpanBatcher` that batches and sends spans by trace ID. Hook it up to the client.
- Add a skeleton of a new `StreamedSpan` class that will eventually replace the current `Span` class.

Chipping away at https://github.com/getsentry/sentry-python/pull/5317 to transform it into reviewable and mergeable PRs.


sentry_sdk/_span_batcher.py
```diff
@@ -0,0 +1,118 @@
...
+                    # TODO[span-first]
+                    # dsc = spans[0].dynamic_sampling_context()
+                    dsc = None
+
+                    envelope = Envelope(
+                        headers={
+                            "sent_at": format_timestamp(datetime.now(timezone.utc)),
+                            "trace": dsc,
+                        }
+                    )
```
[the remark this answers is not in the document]
sentrivana: All coming in future PRs

sentry_sdk/_span_batcher.py
```diff
@@ -0,0 +1,118 @@
...
+        self._running = True
+        self._lock = threading.Lock()
+
+        self._flush_event: "threading.Event" = threading.Event()
+
+        self._flusher: "Optional[threading.Thread]" = None
+        self._flusher_pid: "Optional[int]" = None
+
+    def get_size(self) -> int:
+        # caller is responsible for locking before checking this
```
alexander-alderman-webb, line 49: why can we not, or should we not, lock for the caller here?

alexander-alderman-webb (approved):
looks good! just one question from me, not blocking

sentry_sdk/_span_batcher.py
sentrivana: The `add()` function that needs to call this needs to lock a bigger chunk of code including this check and it can't really be split apart.

`get_size()` is now only used from `add()`, but if we ever need to use it from elsewhere too, I think it's likely that that place will also need to already be holding a lock if it's doing anything that depends on the current size of the buffer, so it makes sense that the responsibility to lock would be on the caller.

PR #5417, opened 2026-02-04
fix: Fix list attribute type

sentrivana:
List attributes should be sent over the wire with `type: array` instead of specialized `integer[]`, `string[]` etc. types.

We might support non-homogenous arrays in the future.


sentrivana:
The lint failure is unrelated (ruff is complaining about a file I didn't touch), will fix that [separately](https://github.com/getsentry/sentry-python/pull/5418)

PR #5449, opened 2026-02-12
fix(openai-agents): Patch models functions following library refactor

alexander-alderman-webb:
### Description

The `AgentRunner._get_model()` class method has been moved to the function `agents.run_internal.turn_preparation.get_model()`.

Patch the new function by re-using existing wrapper logic.

#### Reminders


sentry_sdk/integrations/openai_agents/patches/models.py
```diff
@@ -66,141 +66,128 @@ def _inject_trace_propagation_headers(
...
+    agent._sentry_request_model = request_model_name
+
+    # Wrap _fetch_response if it exists (for OpenAI models) to capture response model
+    if hasattr(model, "_fetch_response"):
+        original_fetch_response = model._fetch_response
+
+        @wraps(original_fetch_response)
+        async def wrapped_fetch_response(*args: "Any", **kwargs: "Any") -> "Any":
+            response = await original_fetch_response(*args, **kwargs)
+            if hasattr(response, "model") and response.model:
```
ericapisani, line 95: Nitpick: I think an alternative approach that you could take here in order to make this conditional more concise (by removing the `and` part of the conditional) is to do the following:

```suggestion
            if getattr(response, "model", None):
```

This would also make things consistent with what you have on line 119 within the `wrapped_get_response` function.

sentry_sdk/integrations/openai_agents/patches/models.py
```diff
@@ -66,141 +66,128 @@ def _inject_trace_propagation_headers(
...
+    model.get_response = wrapped_get_response
+
+    # Also wrap stream_response for streaming support
+    if hasattr(model, "stream_response"):
+        original_stream_response = model.stream_response
+
+        @wraps(original_stream_response)
+        async def wrapped_stream_response(*args: "Any", **kwargs: "Any") -> "Any":
+            # Uses explicit try/finally instead of context manager to ensure cleanup
+            # even if the consumer abandons the stream (GeneratorExit).
```
ericapisani, line 137: This comment looks like it needs to be updated or removed as there's no try/finally below and the code uses a context manager 😅

sentry_sdk/integrations/openai_agents/patches/models.py
```diff
@@ -66,141 +66,128 @@ def _inject_trace_propagation_headers(
...
+                    if hasattr(event, "response"):
+                        streaming_response = event.response
+                    yield event
+
+                # Update span with response data (usage, output, model)
+                if streaming_response:
+                    response_model = (
+                        str(streaming_response.model)
+                        if hasattr(streaming_response, "model")
+                        and streaming_response.model
```
ericapisani, line 183: Also a nitpick but similar to [the comment I left above](https://github.com/getsentry/sentry-python/pull/5449/changes#r2799480545)

sentry_sdk/integrations/openai_agents/patches/models.py
alexander-alderman-webb: We need to rethink this response model business because we probably shouldn't be setting response models on agent spans either.

This was put in before I knew as much about agents, but there are libraries that let you vary the request model during agent execution as well (in particular Claude Code wrapped by the Claude Agent SDK).

Created an issue to track https://github.com/getsentry/sentry-python/issues/5458

(and the plan would be to clean up as part of that issue).

sentry_sdk/integrations/openai_agents/patches/models.py
alexander-alderman-webb: you're completely right, and it doesn't look like we catch `GeneratorExit` anywhere for that matter 😅 .

I'll keep this commit atomic since we're moving the contents of `wrapped_get_model()` to `_get_model()` as is now, and follow up with a PR after this first train is merged 🚅

sentry_sdk/integrations/openai_agents/__init__.py
```diff
@@ -109,6 +103,14 @@ async def new_wrapped_get_all_tools(
...
 
             agents.run.get_all_tools = new_wrapped_get_all_tools
+
+            @wraps(turn_preparation.get_model)
+            def new_wrapped_get_model(
+                agent: "agents.Agent", run_config: "agents.RunConfig"
+            ) -> "agents.Model":
+                return _get_model(turn_preparation.get_model, agent, run_config)
+
+            agents.run_internal.run_loop.get_model = new_wrapped_get_model
```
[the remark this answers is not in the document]
alexander-alderman-webb: I'm patching the reference used in the library.

Issue #5428, opened 2026-02-06
Get `driver_type` from elsewhere in Celery tasks


sentrivana:
Currently, we're getting the driver type (e.g. `redis`) in a Celery task by [opening a new connection](https://github.com/getsentry/sentry-python/blob/3373da2da8886d95c3e7fd4bb712377c4420ce4b/sentry_sdk/integrations/celery/__init__.py#L394-L398). This is not ideal. Is there a way to get this information from elsewhere, without having to open a new connection? If not, can it at least be cached?

ericapisani:
Turns out that we don't need to make any changes as we're not opening a socket when we check the `driver_type` as we are.

Connections are lazily established in celery, and only open sockets when either `conn.connect()` or `conn.ensure_connection()` are called.

So when we call `task.app.connection()` like the following:

```python
with task.app.connection() as conn:
    span.set_data(
        SPANDATA.MESSAGING_SYSTEM,
        conn.transport.driver_type,
    )
```
we haven't opened a socket (checking the `conn.connected` at this point in the code confirms this). We've only retrieved a [Kombu `Connection` object](https://github.com/celery/kombu/blob/900fd2d3cbf6eb1238dffaf6d0cb8f2941687e09/kombu/connection.py#L63), where accessing `transport` instantiates a `Transport` class such as the [Redis one](https://github.com/celery/kombu/blob/900fd2d3cbf6eb1238dffaf6d0cb8f2941687e09/kombu/transport/redis.py#L1389), and the `driver_type` is an attribute on that class.

PR #5456, opened 2026-02-12
fix(celery): Get transport's `driver_type` from broker's URL

ericapisani:
### Description
Under the hood within Kombu (the messaging library used by Celery), the `driver_type` is ultimately derived from the broker's URL that's provided in the Celery configuration by the user via either the `broker` or `broker_url` arguments.

We can accomplish the same thing in our context, removing the need to open a connection to the broker to determine the `driver_type` to set in the span.

The `broker_url` is set in both of the following scenarios:
- `app = Celery("tasks", broker_url=<your-broker-url>)`
- `app = Celery("tasks", broker=<your-broker-url>)`

#### Issues
- Fixes PY-2071
- Fixes https://github.com/getsentry/sentry-python/issues/5428

### DIFF
--- a/sentry_sdk/integrations/celery/__init__.py
+++ b/sentry_sdk/integrations/celery/__init__.py
@@ -1,6 +1,7 @@
 import sys
 from collections.abc import Mapping
 from functools import wraps
+from urllib.parse import urlparse
 
 import sentry_sdk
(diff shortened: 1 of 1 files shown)


ericapisani:
Closing in favour of https://github.com/getsentry/sentry-python/pull/5461

PR #5474, opened 2026-02-17
Fix transaction naming for django-ninja endpoints with transaction_style="function_name" 

avilaton:
### Description
DISCLAIMER: We patched django using a middleware to fix this and wanted to contribute it back to sentry, delegated this to copilot and it looks ok, no intention to be ai-slop, just wanted to raise the issue here.

Django-ninja endpoints were all grouped under `ninja.operation.PathView._sync_view` when using `transaction_style="function_name"`, making it impossible to distinguish between different API endpoints in Sentry.

Django's `resolve()` returns the PathView wrapper function instead of the actual endpoint. The wrapper is shared across all endpoints on the same path, causing the grouping issue.

## Changes

- **Added `_unwrap_django_ninja_view()`** - Detects PathView wrappers by checking `__qualname__`, extracts the PathView instance from the closure, matches the HTTP method to the correct Operation, and returns the actual endpoint function
- **Modified `_set_transaction_name_and_source()`** - Calls unwrapper when `transaction_style="function_name"` before extracting the transaction name
- **Added integration tests** - Validates both `function_name` and `url` transaction styles work correctly with ninja endpoints

## Example

Before this change, all these endpoints would be grouped as `ninja.operation.PathView._sync_view`:

```python
@api.get("/users")
def list_users(request):
    ...

@api.post("/users")
def create_user(request):
    ...
```

After this change, they get distinct names: `myapp.views.list_users` and `myapp.views.create_user`.

### DIFF
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -382,6 +382,41 @@ def _patch_django_asgi_handler() -> None:
     patch_django_asgi_handler_impl(ASGIHandler)
 
 
+def _unwrap_django_ninja_view(fn: "Callable[..., Any]", request: "WSGIRequest") -> "Callable[..., Any]":
+    """
+    Unwrap django-ninja PathView to get the actual endpoint function.
+
+    Django-ninja wraps endpoint functions in PathView.get_view() which returns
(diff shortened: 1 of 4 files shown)


avilaton:
Sorry we tried this and it failed

PR #5490, opened 2026-02-20
fix(anthropic): fix token accounting

shellmayr:
- Current token accounting is off
- This PR makes sure that the logic we expect for token sums is adhered to

### DIFF
--- a/tests/integrations/anthropic/test_anthropic.py
+++ b/tests/integrations/anthropic/test_anthropic.py
@@ -2261,6 +2261,10 @@ def test_cache_tokens_nonstreaming(sentry_init, capture_events):
         )
 
     (span,) = events[0]["spans"]
+    # input_tokens normalized: 100 + 80 (cache_read) + 20 (cache_write) = 200
+    assert span["data"][SPANDATA.GEN_AI_USAGE_INPUT_TOKENS] == 200
+    assert span["data"][SPANDATA.GEN_AI_USAGE_OUTPUT_TOKENS] == 50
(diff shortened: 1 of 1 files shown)


sentry_sdk/integrations/anthropic.py
```diff
@@ -108,6 +108,11 @@ def _get_token_usage(result: "Messages") -> "tuple[int, int, int, int]":
         ):
             cache_write_input_tokens = usage.cache_creation_input_tokens
 
+    # Anthropic's input_tokens excludes cached/cache_write tokens.
+    # Normalize to total input tokens so downstream cost calculations
+    # (input_tokens - cached) don't produce negative values.
+    input_tokens += cache_read_input_tokens + cache_write_input_tokens
```
alexander-alderman-webb, line 114: you're already adding cache read and write outside afterwards, can remove this.

shellmayr: No I don't think we can remove this - there are two codepaths below, one of them for non-streaming and one for streaming. The streaming one does not use this function to retrieve the cost, so it needs to normalize separately

alexander-alderman-webb (changes requested):
I would expect existing tests to change, can you please verify why they still pass even though the behavior has changed @shellmayr 🙏

shellmayr:
@alexander-alderman-webb had a look at the tests - they didn't assert on the total tokens, etc, only the cache tokens. 

adapted them so they would have failed, as a regression test going forward

PR #5526, opened 2026-02-24
fix(integrations): asyncio: continuation trace transactions

constantinius:
### Description

Starting a task within a transaction may result in dropped spans because the task could outlive the transaction.

This PR introduces opening transactions instead with continuation headers applied, so that the trace is continued.

#### Issues

- Closes https://linear.app/getsentry/issue/TET-1993/had-to-manually-create-a-transaction-to-see-agent-spans-when-using

### DIFF
--- a/sentry_sdk/integrations/asyncio.py
+++ b/sentry_sdk/integrations/asyncio.py
@@ -57,6 +57,12 @@ def patch_asyncio() -> None:
             coro: "Coroutine[Any, Any, Any]",
             **kwargs: "Any",
         ) -> "asyncio.Future[Any]":
+            # we get the current span here, as later it may already be closed
+            in_span = sentry_sdk.get_current_span() is not None
+            headers = dict(
+                sentry_sdk.get_current_scope().iter_trace_propagation_headers()
+            )
+
             @_wrap_coroutine(coro)
             async def _task_with_sentry_span_creation() -> "Any":
                 result = None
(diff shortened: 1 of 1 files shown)


sentry_sdk/integrations/asyncio.py
```diff
@@ -57,6 +57,12 @@ def _sentry_task_factory(
             coro: "Coroutine[Any, Any, Any]",
             **kwargs: "Any",
         ) -> "asyncio.Future[Any]":
+            # we get the current span here, as later it may already be closed
+            in_span = sentry_sdk.get_current_span() is not None
```
constantinius, line 61: We have to get the info whether we are in a transaction straight away before it closes.

sentry_sdk/integrations/asyncio.py
```diff
@@ -66,16 +72,16 @@ async def _task_with_sentry_span_creation() -> "Any":
                 )
                 task_spans = integration.task_spans if integration else False
 
+                if task_spans and in_span:
```
constantinius, line 75: We should probably add a new option if we want to go that route

constantinius:
Another way would be to have a function that wraps `asyncio.create_task`. something like:
```python
def create_continuation_task(coro, *args, **kwargs):
    async def _wrapper():
        headers = dict(
            sentry_sdk.get_current_scope().iter_trace_propagation_headers()
        )
        transaction = sentry_sdk.continue_trace(
            headers, op="task", name="downstream"
        )
        with sentry_sdk.start_transaction(transaction):
            await coro
    asyncio.create_task(_wrapper(), *args, **kwargs)
```

PR #5532, opened 2026-02-25
feat(cohere): upgrade integration from ai to gen_ai

shellmayr:
This PR encompasses multiple changes to accomplish this
- Add support for the "new" v2 of cohere which was released at the end of 2024. The cohere package continues to contain the v1 and v2 APIs, therefore we still need to support both. The instrumentation is in a separate file to prevent confusion between the versions. 
- Move the auto-instrumentation for v1 from ai.* attributes to gen_ai.* attributes
- Adapt / add tests for v1 and v2
- Has been run against two new templates for cohere in the testing framework
- Semantically, Cohere is a model-provider on the level of e.g. anthropic and openai, meaning that there is no agentic flavor to this, it is "just" an API for LLM-calling for Cohere's own models

Closes TET-2023

### DIFF
--- a/sentry_sdk/ai/span_config.py
+++ b/sentry_sdk/ai/span_config.py
@@ -33,27 +33,31 @@ def set_request_span_data(span, kwargs, integration, config, span_data=None):
             set_data_normalized(span, span_attr, value)
 
     if should_send_default_pii() and integration.include_prompts:
-        extract = config.get("extract_messages")
-        if extract is not None:
-            messages = extract(kwargs)
-            if messages:
-                messages = normalize_message_roles(messages)
-                scope = sentry_sdk.get_current_scope()
-                messages = truncate_and_annotate_messages(messages, span, scope)
-                if messages is not None:
-                    target = config.get(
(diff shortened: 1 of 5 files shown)


sentry_sdk/ai/utils.py
```diff
@@ -30,7 +30,7 @@ class GEN_AI_ALLOWED_MESSAGE_ROLES:
 GEN_AI_MESSAGE_ROLE_REVERSE_MAPPING = {
     GEN_AI_ALLOWED_MESSAGE_ROLES.SYSTEM: ["system"],
     GEN_AI_ALLOWED_MESSAGE_ROLES.USER: ["user", "human"],
-    GEN_AI_ALLOWED_MESSAGE_ROLES.ASSISTANT: ["assistant", "ai"],
+    GEN_AI_ALLOWED_MESSAGE_ROLES.ASSISTANT: ["assistant", "ai", "chatbot"],
```
shellmayr, line 33: This is needed because cohere sometimes uses the role `chatbot` in their message structure

shellmayr:
Closing - splitting this up into multiple PRs for easier reviews.

PR #5553, opened 2026-02-26
ref(openai): Only handle streamed results when applicable

alexander-alderman-webb:
### Description

Remove iterator patches from shared output-handling code, and only apply them in the dedicated functions for patching streamed responses. Only keep code relevant to the Completions and Responses APIs in the respective iterator wrappers.

Add `stream=True` to tests where a streamed response is returned from `openai` functions.

#### Reminders

### DIFF
--- a/sentry_sdk/integrations/openai.py
+++ b/sentry_sdk/integrations/openai.py
@@ -462,7 +462,7 @@ def _set_embeddings_input_data(
     _commmon_set_input_data(span, kwargs)
 
 
-def _common_set_output_data(
+def _set_common_output_data(
     span: "Span",
     response: "Any",
     input: "Any",
@@ -582,7 +582,7 @@ def _set_completions_api_output_data(
     if messages is not None and isinstance(messages, str):
         messages = [messages]
 
-    _common_set_output_data(
+    _set_common_output_data(
         span,
         response,
         messages,
@@ -711,7 +711,7 @@ def _set_responses_api_output_data(
     if input is not None and isinstance(input, str):
         input = [input]
 
-    _common_set_output_data(
+    _set_common_output_data(
         span,
         response,
         input,
@@ -849,7 +849,7 @@ def _set_embeddings_output_data(
     if input is not None and isinstance(input, str):
         input = [input]
 
-    _common_set_output_data(
+    _set_common_output_data(
         span,
         response,
         input,
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,6 +126,20 @@ lldb -- ../../uwsgi/uwsgi --pythonpath "$PWD/.venv/lib/python3.14/site-packages"
   --need-app \
 ```
 
+### Troubleshooting "module not found" error in relation to the `sentry-sdk` when it's installed in editable mode
+
+If you are trying to debug a Django applicaton such as that described above, and have the Sentry SDK installed in editable mode in the Django project, you will likely encounter this error because the editable package is not being found in the uWSGI's python path. To fix this, the above command needs to be updated to include a `--pythonpath` argument that passes in where your local `sentry-python` codebase is:
+
+```bash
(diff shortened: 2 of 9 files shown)


sentry_sdk/integrations/openai.py
```diff
@@ -473,8 +473,6 @@ def _common_set_output_data(
     if hasattr(response, "model"):
         set_data_normalized(span, SPANDATA.GEN_AI_RESPONSE_MODEL, response.model)
 
-    ttft: "Optional[float]" = None
```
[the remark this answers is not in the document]
alexander-alderman-webb: will remove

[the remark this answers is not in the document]
alexander-alderman-webb: https://github.com/getsentry/sentry-python/pull/5553/changes/edf0a4f9c719d31014b15aed1b09edc749913459

sentry_sdk/integrations/openai.py
```diff
@@ -746,22 +604,109 @@ def _set_streaming_completions_api_output_data(
...
-        finish_span,
-    )
+    ttft: "Optional[float]" = None
+    data_buf: "list[list[str]]" = []  # one for each choice
+
+    old_iterator = response._iterator
+
+    def new_iterator() -> "Iterator[ChatCompletionChunk]":
+        nonlocal ttft
+        count_tokens_manually = True
```
ericapisani: This doesn't appear to be needed in this iterator function - we never set it to false, so the conditional on line 644 will always run.

sentry_sdk/integrations/openai.py
```diff
@@ -746,22 +604,109 @@ def _set_streaming_completions_api_output_data(
...
+                        all_responses,
+                        integration.count_tokens,
+                    )
+
+        if finish_span:
+            span.__exit__(None, None, None)
+
+    async def new_iterator_async() -> "AsyncIterator[ChatCompletionChunk]":
+        nonlocal ttft
+        count_tokens_manually = True
```
ericapisani: Same situation here as above - this is never set to `False` so the conditional further down on line 688 will always run.

sentry_sdk/integrations/openai.py
alexander-alderman-webb: nice! always great to remove some more dead code 👍 

addressed in https://github.com/getsentry/sentry-python/pull/5553/changes/3fb20f04e8b24025c15f704ad84481c702ce0612 + https://github.com/getsentry/sentry-python/pull/5553/changes/82843e2fd25f8555087e7420c75d3316314c03ec

sentry_sdk/integrations/openai.py
alexander-alderman-webb: https://github.com/getsentry/sentry-python/pull/5553/changes/3fb20f04e8b24025c15f704ad84481c702ce0612 + https://github.com/getsentry/sentry-python/pull/5553/changes/82843e2fd25f8555087e7420c75d3316314c03ec

goudyj:
Hello @alexander-alderman-webb, before this PR, the integration only wrapped streaming responses when the returned object actually had an `_iterator` attribute.
In this PR, the streaming logic was moved into `_set_streaming_completions_api_output_data()` and
  `_set_streaming_responses_api_output_data()`, and both now access `response._iterator` unconditionally whenever `stream=True`.

  That means third-party wrapper objects that are iterable/async-iterable but do not expose _iterator are no longer handled safely. That seems to be the case with Langfuse: its async response wrapper supports async iteration, but does not expose `_iterator`, which matches the error I’m seeing:

  *AttributeError: 'LangfuseResponseGeneratorAsync' object has no attribute '_iterator'*

  So I think this PR introduced a regression for wrapped streaming responses. The minimal fix would be to restore a guard like `hasattr(response, "_iterator")` before trying to patch the iterator, or otherwise unwrap/support wrapper types such as Langfuse’s response generator.

Issue #5107, opened 2025-11-13
UWSGI offload() skipped because of _ScopedResponse


czpython:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.13.0

### Steps to Reproduce

1. Setup a view in django to serve a file using FileResponse
2. Look at uWSGI logs to see switches and serve strategy
3. Because Sentry overrides wsgi handler, it'll use this ScopedResponse which doesn't have a fileno
4. uWSGI sees this and serves the file without offload()

### Expected Result

I expect a log line like
```
[pid: 7|app: 0|req: 1/1] 127.0.0.1 () {28 vars in 380 bytes} [Tue Nov 11 16:27:44 2025] GET /file/ => generated 605 bytes in 361 msecs via offload() (HTTP/1.1 200) 7 headers in 261 bytes (0 switches on core 0)
```

### Actual Result

```
[pid: 8|app: 0|req: 1/1] 127.0.0.1 () {28 vars in 402 bytes} [Thu Nov 13 13:13:09 2025] GET /file/ => generated 605 bytes in 1157 msecs (HTTP/1.1 200) 7 headers in 261 bytes (30 switches on core 0)
```

alexander-alderman-webb:
Thank you for the report @czpython! We'll take a look.

PR #5556, opened 2026-02-26
fix(wsgi): Do not wrap file responses when uWSGI offload-threads is enabled

ericapisani:
uWSGI uses pointer equality to detect file-like response objects eligible for its offload-threads optimization. When Sentry's WSGI middleware wraps the response in `_ScopedResponse`, the pointer changes and uWSGI's check always fails, silently disabling the offloading.

This fix skips the `_ScopedResponse` wrapping when all three conditions are true:
- uWSGI's `offload-threads` option is configured and > 0
- `wsgi.file_wrapper` is present in the WSGI environ (indicating the server would use it)
- The response object has a `fileno` attribute (indicating it is a file-like object)

When these conditions are met, the original response is returned directly, allowing uWSGI to handle offloading as intended.

Fixes PY-1977 and https://github.com/getsentry/sentry-python/issues/5107

### DIFF
--- a/sentry_sdk/integrations/wsgi.py
+++ b/sentry_sdk/integrations/wsgi.py
@@ -145,39 +145,20 @@ class SentryWsgiMiddleware:
         # To avoid this and ensure that the offloading mechanism works as expected when it's
         # enabled, we check if the response is a file-like object (determined by the presence
         # of `fileno`), if the wsgi.file_wrapper is available in the environment (as if so,
-        # it would've been used in handling the file in the response), and if uWSGI's
-        # offload-threads option is configured (since offloading only occurs when offload
-        # threads are enabled).
+        # it would've been used in handling the file in the response).
+        #
+        # Even if the offload mechanism is not enabled, there are optimizations that uWSGI does for file-like objects,
+        # so we want to make sure we don't interfere with those either.
         #
         # If all conditions are met, we return the original response object directly,
         # allowing uWSGI to handle it as intended.
         if (
-            _is_uwsgi_offload_threads_enabled()
-            and environ.get("wsgi.file_wrapper")
+            environ.get("wsgi.file_wrapper")
             and getattr(response, "fileno", None) is not None
         ):
             return response
 
         return _ScopedResponse(scope, response)
-
-
-def _is_uwsgi_offload_threads_enabled() -> bool:
-    try:
-        from uwsgi import opt  # type: ignore
-    except ImportError:
-        return False
-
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/wsgi.py
```diff
@@ -135,9 +135,51 @@ def __call__(
...
+        # threads are enabled).
+        #
+        # If all conditions are met, we return the original response object directly,
+        # allowing uWSGI to handle it as intended.
+        if (
+            _is_uwsgi_offload_threads_enabled()
+            and environ.get("wsgi.file_wrapper")
+            and getattr(response, "fileno", None) is not None
+        ):
+            return response
```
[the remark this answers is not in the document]
ericapisani: This is an intentional choice and will be documented.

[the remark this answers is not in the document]
ericapisani: Documentation added here: https://github.com/getsentry/sentry-docs/pull/16603

sentry_sdk/integrations/wsgi.py
```diff
@@ -135,9 +135,51 @@ def __call__(
...
+        # offload-threads option is configured (since offloading only occurs when offload
+        # threads are enabled).
+        #
+        # If all conditions are met, we return the original response object directly,
+        # allowing uWSGI to handle it as intended.
+        if (
+            _is_uwsgi_offload_threads_enabled()
+            and environ.get("wsgi.file_wrapper")
+            and getattr(response, "fileno", None) is not None
+        ):
```
alexander-alderman-webb, line 158: Is `uwsgi` always installed when `django` is installed?

If not we should consider reling on only the second and third condition. I seem to be able to run Django without needing the `uwsgi` Python package.

ericapisani: > Is uwsgi always installed when django is installed?

No, Django includes a WSGI dev server out of the box, but requires other packages to be installed (like `gunicorn`, `uwsgi`, `daphne`, etc.) for production servers. If I remember correctly there's a warning that's shown by Django when you run `runserver` about this.

> If not we should consider reling on only the second and third condition.

Sounds good 👍🏻  Interestingly, after making this change, it also seems to address what looks like the same issue with the switching that exists when `offload-threads` is disabled

ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/5556/commits/869866136a34dcd4eefc1e280b041bc43bf7bbce

alexander-alderman-webb (approved):
LGTM!

PR #5577, opened 2026-03-02
ci: Add git-town workflow to display PR stack visualization

ericapisani:
Add a GitHub Actions workflow that uses `git-town/action` to automatically render the stacked pull request visualization in PR descriptions.

When a PR is opened, reopened, or pushed to, the workflow injects the branch stack into the `<!-- branch-stack -->` marker in the PR template. This makes it easy to see the full stack context without leaving the PR page.

The PR template is updated to include the marker and a "Pull request stack" section where the visualization will appear — always below the user-provided description.

The workflow follows existing repo security conventions: both actions are pinned to full commit SHAs, permissions are deny-all at the workflow level with only `contents: read` and `pull-requests: write` granted at the job level, and `fetch-depth: 0` is set so git-town has the full history needed to compute the stack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,67 @@
 # Changelog
 
+## 2.54.0
+
+### New Features ✨
+
+- Add `set_attribute`, `remove_attribute` to global API by @sentrivana in [#5555](https://github.com/getsentry/sentry-python/pull/5555)
+
+### Bug Fixes 🐛
+
+#### Openai
+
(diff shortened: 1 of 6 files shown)


ericapisani:
Closing this as there's an underlying issue within git-town that prevents our use of it within this repository.

Specifically, if a user were to open a pull request from a forked repository that is from a `master` branch, it results in the inability for git-town to build an acyclic graph, resulting in the github action breaking for _all_ pull requests.

PR #5587, opened 2026-03-04
fix(utils): Avoid double serialization of strings in safe_serialize

ericapisani:
When `serialize_item()` inside `safe_serialize()` already returns a string
(for plain strings, callables, or objects with `__dict__`), the subsequent
`json.dumps()` call wraps it in extra quotes with escaped characters. For
example, a JSON string `'{"param": "value"}'` becomes
`'"{\\"param\\": \\"value\\"}"'`.

This was causing double-serialized tool arguments in pydantic-ai
integrations.

The fix skips `json.dumps` when `serialize_item` has already produced a
string, and adds tests covering plain strings, JSON strings, dicts,
callables, and objects.

Part of PY-2114

### DIFF
--- a/sentry_sdk/traces.py
+++ b/sentry_sdk/traces.py
@@ -9,15 +9,11 @@ import uuid
 from enum import Enum
 from typing import TYPE_CHECKING
 
-from sentry_sdk.consts import SPANDATA
-from sentry_sdk.utils import format_attribute
+from sentry_sdk.utils import format_attribute, logger
 
 if TYPE_CHECKING:
     from typing import Optional, Union
     from sentry_sdk._types import Attributes, AttributeValue
-
-
-FLAGS_CAPACITY = 10
 
 
 class SpanStatus(str, Enum):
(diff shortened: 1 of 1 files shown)


sentry_sdk/utils.py
```diff
@@ -2005,7 +2005,11 @@ def serialize_item(
 
     try:
         serialized = serialize_item(data)
-        return json.dumps(serialized, default=str)
+        return (
+            json.dumps(serialized, default=str)
+            if not isinstance(serialized, str)
+            else serialized
+        )
```
alexander-alderman-webb, line 2012: can you please check if we can use parameters to `json.dumps()` to avoid this complex logic and possibly error-prone logic?

something like `default=` or `cls=`?

https://docs.python.org/3/library/json.html

ericapisani: Unfortunately there's no parameter to `json.dumps` that will help us avoid the double serialization problem that we're running into here that the conditional prevents.

`default=` is for handling objects that aren't serializable (which isn't the case for us in this context), and `cls=` is for if we want to introduce a custom `JSONEncoder` to replace the default one which I don't think is necessary here.

PR #5591, opened 2026-03-05
ref: Add no-op streaming span class (4)

sentrivana:
### Description
We'll use `NoOpStreamedSpan`s for spans that have a negative sampling decision or were filtered out via `ignore_spans`.

They will get some minimal `__slots__` and `__init__` in a future PR.

#### Reminders


sentry_sdk/traces.py
```diff
@@ -165,3 +165,56 @@ def trace_id(self) -> str:
...
+    __slots__ = (
+    )
+
+    def __init__(
+        self,
+    ) -> None:
+        pass
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}(sampled={self.sampled})>"
```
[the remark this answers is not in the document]
sentrivana: Removed, will add in future PR

sentry_sdk/traces.py
```diff
@@ -165,3 +165,45 @@ def trace_id(self) -> str:
...
+    def active(self) -> bool:
+        return True
+
+    @property
+    def span_id(self) -> str:
+        return "0000000000000000"
+
+    @property
+    def trace_id(self) -> str:
+        return "00000000000000000000000000000000"
```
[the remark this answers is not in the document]
sentrivana: Very true, tests coming later.

sentry_sdk/traces.py
```diff
@@ -164,3 +164,45 @@ def trace_id(self) -> str:
             self._trace_id = uuid.uuid4().hex
 
         return self._trace_id
+
+
+class NoOpStreamedSpan(StreamedSpan):
```
alexander-alderman-webb, line 169: generally sounds good.

I don't have the picture of why we need all the same methods on this class.
In what scenario would we call `NoOpStreamedSpan.span_id()`?
Can we just return early in the case of a `NoOpStreamedSpan` if any code uses `NoOpStreamedSpan.span_id()`?

alexander-alderman-webb: and analogous questions apply, basically, why is inheritance appropriate here 😅

sentrivana: The idea is that the rest of the SDK and users can interact with the span without having to care whether it's a sampled or an unsampled span. So you can still call the same methods, access the same attributes, etc., they'll just give you dummy sentinel values back. It's nicer than the alternative, which would be to check whether we're dealing with a `StreamedSpan` or a `NoOpStreamedSpan` in all calling code.

See also https://develop.sentry.dev/sdk/telemetry/spans/span-api/#noop-span

sentrivana: Also, we're following OTel's lead here too with their `NonRecordingSpan`.

PR #5598, opened 2026-03-06
ref: Allow to start and finish `StreamedSpans` (9)

sentrivana:
### Description
This PR adds logic for starting and finishing spans, both via a context manager `with start_span(...)`, as well as directly with `span = start_span(...)` and `span.end()`.

Summary
- Added internal `_start()` and `_end()` functions that handle the starting and ending logic. 
  - `_start()` handles keeping track of what was called `context_manager_state` on old Spans, i.e., remembering which span was set on the scope before this span so that it can be restored once this span is over.
  - `_end()` handles restoring the old span, setting the end timestamp, adding final attributes on it, and queueing it in the span batcher.
- Added `__enter__` and `__exit__` methods on both `StreamedSpan` and `NoOpStreamedSpan`. These call the internal methods above, with `__exit__` additionally setting the span status to error if it encountered one.
- Added a new `span.end()` public function that allows to set an optional `end_timestamp`. If provided, it sets it and calls `_end()` internally.
- There is a `span.finish()` too, which is deprecated from the get go, just to make the migration to span first a bit easier (since the method is called `finish()` on legacy spans).

#### Reminders


sentry_sdk/traces.py
```diff
@@ -302,10 +398,27 @@ def trace_id(self) -> str:
     def sampled(self) -> "Optional[bool]":
         return True
 
+    @property
+    def start_timestamp(self) -> "Optional[datetime]":
+        return self._start_timestamp
+
+    @property
+    def timestamp(self) -> "Optional[datetime]":
+        return self._timestamp
```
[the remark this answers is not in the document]
sentrivana: Added dummy `return None` implementation to the class

sentry_sdk/_span_batcher.py
```diff
@@ -91,8 +91,12 @@ def _to_transport_format(item: "StreamedSpan") -> "Any":
             "span_id": item.span_id,
             "name": item._name,
             "status": item._status,
+            "start_timestamp": item._start_timestamp.timestamp(),
         }
 
+        if item._timestamp:
+            res["end_timestamp"] = item._timestamp.timestamp()
```
sentrivana, line 98: There is no scenario where a span that's already in the span batcher wouldn't have an end `_timestamp`, but mypy doesn't know that. 🤡

sentry_sdk/traces.py
```diff
@@ -223,11 +235,23 @@ def __init__(
         self._parent_sampled = parent_sampled
         self._baggage = baggage
 
+        self._start_timestamp = datetime.now(timezone.utc)
+        self._timestamp: "Optional[datetime]" = None
+
+        try:
+            # profiling depends on this value and requires that
+            # it is measured in nanoseconds
+            self._start_timestamp_monotonic_ns = nanosecond_time()
```
sentrivana, line 244: Anything related to `_start_timestamp_monotonic_ns` is just copy-pasted from the `Span` implementation

sentry_sdk/traces.py
```diff
@@ -244,7 +268,83 @@ def __enter__(self) -> "StreamedSpan":
...
+    def end(self, end_timestamp: "Optional[Union[float, datetime]]" = None) -> None:
+        """
+        Finish this span and queue it for sending.
+
+        :param end_timestamp: End timestamp to use instead of current time.
+        :type end_timestamp: "Optional[Union[float, datetime]]"
+        """
+        self._end(end_timestamp)
+
+    def finish(self, end_timestamp: "Optional[Union[float, datetime]]" = None) -> None:
```
sentrivana, line 289: Just here for compat reasons, to make the transition from the old API easier (very slightly).

sentry_sdk/traces.py
```diff
@@ -244,7 +268,83 @@ def __enter__(self) -> "StreamedSpan":
...
+        warnings.warn(
+            "span.finish() is deprecated. Use span.end() instead.",
+            stacklevel=2,
+            category=DeprecationWarning,
+        )
+
+        self.end(end_timestamp)
+
+    def _start(self) -> None:
+        if self._active:
```
sentrivana, line 299: `_active` basically controls whether a span will be set on the scope, so if it's `False`, we don't need to keep track of the old parent span to restore it later, because we're never replacing it.

sentry_sdk/traces.py
```diff
@@ -323,7 +441,36 @@ def __enter__(self) -> "NoOpStreamedSpan":
...
-        pass
+        self._end()
+
+    def _start(self) -> None:
+        if self._scope is None:
+            return
+
+        old_span = self._scope.span
+        self._scope.span = self
+        self._previous_span_on_scope = old_span
```
sentrivana, line 451: Unlike the main span class, no-op spans don't have an `active` flag. The user cannot control whether a no-op span will be set on the scope; the SDK does this internally, and it uses the `_scope` attribute to remember the decision. In the case of normal spans, the `_scope` attribute plays a different role: they all need a scope, regardless of whether they'll actually be set on it or not, because they will be captured using that scope in either case.

Coming back to no-op spans, the decision whether to set them on the scope or not is tightly coupled to how we want [`ignore_spans` to work](https://develop.sentry.dev/sdk/telemetry/spans/filtering/#implementation-requirements) (coming in a future PR). Basically, if a segment/root span would be ignored, all of its children should be as well. That's easiest to accomplish if we actually set the root no-op span on the scope. On the other hand, if a non-root span is ignored, it's not set on scope, so that any children spans effectively use the last not ignored span that's set on the scope as parent.

sentry_sdk/traces.py
```diff
@@ -323,7 +437,36 @@ def __enter__(self) -> "NoOpStreamedSpan":
...
+
+    def _end(self, end_timestamp: "Optional[Union[float, datetime]]" = None) -> None:
+        if self._scope is None:
+            return
+
+        with capture_internal_exceptions():
+            old_span = self._previous_span_on_scope
+            del self._previous_span_on_scope
+            self._scope.span = old_span
```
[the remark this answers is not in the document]
sentrivana: Good point, added a guard

sentry_sdk/traces.py
```diff
@@ -244,7 +268,79 @@ def __enter__(self) -> "StreamedSpan":
     def __exit__(
         self, ty: "Optional[Any]", value: "Optional[Any]", tb: "Optional[Any]"
     ) -> None:
-        pass
+        if value is not None and should_be_treated_as_error(ty, value):
+            self.status = SpanStatus.ERROR
+
+        self._end()
```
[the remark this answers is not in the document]
alexander-alderman-webb: looks like we should define what happens in this scenario? @sentrivana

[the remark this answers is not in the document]
sentrivana: This is ok, maybe even desirable.

sentry_sdk/traces.py
```diff
@@ -244,7 +268,79 @@ def __enter__(self) -> "StreamedSpan":
...
+        # Detach from scope
+        if self._active:
+            with capture_internal_exceptions():
+                old_span = self._previous_span_on_scope
+                del self._previous_span_on_scope
+                self._scope.span = old_span
+
+        # Set attributes from the segment
+        self.set_attribute("sentry.segment.id", self._segment.span_id)
+        self.set_attribute("sentry.segment.name", self._segment.name)
```
alexander-alderman-webb: why do we set the segment attributes `_end()`?

It would be easier to reason about if they're set on `__init__()` (unless there's a reason we have to set them in `_end()`)

sentry_sdk/traces.py
```diff
@@ -244,7 +268,79 @@ def __enter__(self) -> "StreamedSpan":
...
+        if self._timestamp is not None:
+            # This span is already finished, ignore.
+            return
+
+        # Detach from scope
+        if self._active:
+            with capture_internal_exceptions():
+                old_span = self._previous_span_on_scope
+                del self._previous_span_on_scope
+                self._scope.span = old_span
```
alexander-alderman-webb, line 314: In the `NoOpStreamedSpan` there's an early return which we maybe want in this function as well?


```
if not hasattr(self, "_previous_span_on_scope"):
    return
```

sentry_sdk/traces.py
sentrivana: Segment name might change during its lifetime since we often pre-set it to something generic and only update it once a better name is available. We set the attribute here to capture the segment name as late as possible, essentially, so that it's more likely that we capture the final name.

Will add this to the comment

sentry_sdk/traces.py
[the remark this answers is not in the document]
sentrivana: But it does make things kinda nondeterministic. It's cleaner if we don't allow any updates after the span has been queued. I'll update.

sentry_sdk/traces.py
sentrivana: In `NoOpStreamedSpan` it's an early return because there's nothing else that `_end()` does, but in `StreamedSpan`, the situation is a bit different as we might not have a `_previous_span_on_scope` set at all because the span was not set on the scope (happens when `active=False`), but we still want to do all the rest (set attributes, timestamp).

alexander-alderman-webb: sounds good, that all makes sense.

as long as `self._active` can't be changed between `_start()` and `_end()` there is no ungated access.

sentry_sdk/traces.py
sentrivana: https://github.com/getsentry/sentry-python/pull/5598/changes/1d203bf57af81b75e5fbde0d6482eec0fcfbfc6c

sentry_sdk/traces.py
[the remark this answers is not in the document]
sentrivana: https://github.com/getsentry/sentry-python/pull/5598/changes/1d203bf57af81b75e5fbde0d6482eec0fcfbfc6c

PR #5617, opened 2026-03-09
ref: Add sampling to span first (11)

sentrivana:
Add sampling to span first.

- In span first, we make a sampling decision BEFORE a span is actually created. Based on the sampling decision (and potentially other filtering mechanisms like `ignore_spans` -- coming soon) we either create a `StreamedSpan` or a `NoopStreamedSpan`.
- No-op spans will remember why they were unsampled and will emit a client report accordingly when finished.
- The logic that makes the sampling decision has been ported from [`_set_initial_sampling_decision`](https://github.com/getsentry/sentry-python/blob/a04ba6e1e5d67a587962f06837952bea4546193f/sentry_sdk/tracing.py#L1145-L1244).
- The effective sampling rate, which might potentially differ from the incoming baggage (because e.g. backpressure handling is active and lowers the effective sample rate), is then updated in the propagation context (ported from [here](https://github.com/getsentry/sentry-python/blob/a04ba6e1e5d67a587962f06837952bea4546193f/sentry_sdk/scope.py#L1086-L1092)).

### DIFF
--- a/sentry_sdk/traces.py
+++ b/sentry_sdk/traces.py
@@ -356,7 +356,9 @@ class StreamedSpan:
             if isinstance(end_timestamp, datetime):
                 self._timestamp = end_timestamp
             else:
-                logger.debug("Failed to set end_timestamp. Using current time instead.")
+                logger.debug(
+                    "[Tracing] Failed to set end_timestamp. Using current time instead."
+                )
 
         if self._timestamp is None:
             try:
@@ -401,7 +403,7 @@ class StreamedSpan:
 
         if status not in {e.value for e in SpanStatus}:
             logger.debug(
-                f'Unsupported span status {status}. Expected one of: "ok", "error"'
+                f'[Tracing] Unsupported span status {status}. Expected one of: "ok", "error"'
             )
             return
 
@@ -491,7 +493,7 @@ class NoOpStreamedSpan(StreamedSpan):
             client = sentry_sdk.get_client()
             if client.is_active() and client.transport:
                 logger.debug(
-                    f"Discarding span because sampled=False (reason: {self._unsampled_reason})"
+                    f"[Tracing] Discarding span because sampled=False (reason: {self._unsampled_reason})"
                 )
                 client.transport.record_lost_event(
                     reason=self._unsampled_reason,


sentry_sdk/scope.py
```diff
@@ -1227,6 +1245,15 @@ def start_streamed_span(
...
             )
 
+    def _update_sample_rate(self, sample_rate: float) -> None:
+        # If we had to adjust the sample rate when setting the sampling decision
+        # for a span, it needs to be updated in the propagation context too
+        propagation_context = self.get_active_propagation_context()
+        baggage = propagation_context.baggage
+
+        if baggage is not None:
+            baggage.sentry_items["sample_rate"] = str(sample_rate)
```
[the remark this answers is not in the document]
sentrivana: This is a really good point but this is also the one place in the SDK where we want to do this regardless of whether the baggage is frozen.

sentry_sdk/traces.py
```diff
@@ -471,16 +484,28 @@ def _start(self) -> None:
...
+        if self._finished:
             return
 
-        if not hasattr(self, "_previous_span_on_scope"):
-            return
+        if self._unsampled_reason is not None:
+            client = sentry_sdk.get_client()
+            if client.is_active() and client.transport:
+                logger.debug(
+                    f"Discarding span because sampled=False (reason: {self._unsampled_reason})"
```
ericapisani: I've noticed in other areas of the codebase (specifically [tracing.py](https://github.com/getsentry/sentry-python/blob/adcd90ca92aa173ae39aac094adf71f1a4f51b22/sentry_sdk/tracing.py#L1239-L1244)) that we prefix messages like these with `[Tracing]` - I think we should consider doing the same here to make it easier to find these messages in logs should someone need to

sentry_sdk/tracing_utils.py
```diff
@@ -1379,6 +1382,89 @@ def add_sentry_baggage_to_headers(
...
+    else:
+        if propagation_context.parent_sampled is not None:
+            sample_rate = propagation_context.parent_sampled
+        else:
+            sample_rate = client.options["traces_sample_rate"]
+
+    # Validate whether the sample_rate we got is actually valid. Since
+    # traces_sampler is user-provided, it could return anything.
+    if not is_valid_sample_rate(sample_rate, source="Tracing"):
+        logger.warning(f"[Tracing] Discarding {name} because of invalid sample rate.")
```
ericapisani, line 1433: This is a bigger refactor that would happen outside of this PR (if we do decide to do this), but I wonder if we should consider leveraging the `extra` property within the `logger` in order to both hold dynamic values like `name`, but also to include a structured enum reason `discard_reason=TraceDiscardReasons.INVALID_SAMPLE_RATE`.

If a user is trying to determine if/why spans were dropped, this would help with that effort.

sentry_sdk/traces.py
sentrivana: https://github.com/getsentry/sentry-python/pull/5617/changes/62deebf9a5c90a6832aec2bc437e6e6eb09ae896

sentry_sdk/traces.py
```diff
@@ -471,16 +486,28 @@ def _start(self) -> None:
...
+            with capture_internal_exceptions():
+                old_span = self._previous_span_on_scope
+                del self._previous_span_on_scope
+                self._scope.span = old_span
 
-        with capture_internal_exceptions():
-            old_span = self._previous_span_on_scope
-            del self._previous_span_on_scope
-            self._scope.span = old_span
+        self._finished = True
```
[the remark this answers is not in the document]
sentrivana: Coming in a future pr

sentry_sdk/scope.py
```diff
@@ -1199,6 +1200,21 @@ def start_streamed_span(
...
+            )
+
+            if sample_rate is not None:
+                self._update_sample_rate(sample_rate)
+
+            if sampled is False:
+                return NoOpStreamedSpan(
+                    scope=self,
+                    unsampled_reason=outcome,
+                )
```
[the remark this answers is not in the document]
sentrivana: This is by design

sentry_sdk/tracing_utils.py
sentrivana: I think more info in log messages is always nice, so we can def do something like this in the future.

sentry_sdk/tracing_utils.py
```diff
@@ -1379,6 +1382,89 @@ def add_sentry_baggage_to_headers(
...
+            "trace_id": propagation_context.trace_id,
+            "parent_span_id": propagation_context.parent_span_id,
+            "parent_sampled": propagation_context.parent_sampled,
+            "attributes": dict(attributes) if attributes else {},
+        }
+
+        sample_rate = client.options["traces_sampler"](sampling_context)
+    else:
+        if propagation_context.parent_sampled is not None:
+            sample_rate = propagation_context.parent_sampled
```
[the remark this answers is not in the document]
sentrivana: This is the same on master.

sentry_sdk/tracing_utils.py
ericapisani: Made PY-2133 to track this, might be a future maintenance day exploration for me 👀

PR #5628, opened 2026-03-10
ref: Add support for custom sampling context to span first (14)

sentrivana:
Custom sampling context allows folks to have arbitrary data accessible in the `traces_sampler` in order to make a sampling decision. The SDK sets custom sampling context as well in some integrations, for example, in ASGI frameworks the ASGI scope will be available.

Previously, you could provide custom sampling context as an argument to the `start_span` function. In the spirit of keeping the new `start_span` API minimal, we'll be moving `custom_sampling_context` to the propagation context and providing a dedicated API function to set it. In this PR, it's a scope method (`scope.set_custom_sampling_context()`). We can (and probably should) promote it to top-level API at some point in the future.

### DIFF
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -1421,7 +1421,7 @@ def _make_sampling_decision(
     traces_sampler_defined = callable(client.options.get("traces_sampler"))
     if traces_sampler_defined:
         sampling_context = {
-            "span": {
+            "span_context": {
                 "name": name,
                 "trace_id": propagation_context.trace_id,
                 "parent_span_id": propagation_context.parent_span_id,
--- a/tests/tracing/test_span_streaming.py
+++ b/tests/tracing/test_span_streaming.py
@@ -120,7 +120,7 @@ def test_span_sampled_when_created(sentry_init, capture_envelopes):
     # at start_span() time
 
     def traces_sampler(sampling_context):
-        assert "delayed_attribute" not in sampling_context["span"]["attributes"]
+        assert "delayed_attribute" not in sampling_context["span_context"]["attributes"]
         return 1.0
 
     sentry_init(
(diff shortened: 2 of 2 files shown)


sentry_sdk/tracing_utils.py
```diff
@@ -537,6 +540,11 @@ def update(self, other_dict: "Dict[str, Any]") -> None:
             except AttributeError:
                 pass
 
+    def _set_custom_sampling_context(
+        self, custom_sampling_context: "dict[str, Any]"
+    ) -> None:
+        self.custom_sampling_context = custom_sampling_context
```
ericapisani, line 546: Related to the [comment Cursor made](https://github.com/getsentry/sentry-python/pull/5628/changes#r2912292611) - should we consider adding validation to the value that's going be set on the `custom_sampling_context`?

tests/tracing/test_span_streaming.py
```diff
@@ -233,6 +233,62 @@ def traces_sampler(sampling_context):
...
+            "class": my_class,
+            "string": "my string",
+        }
+    )
+
+    with sentry_sdk.traces.start_span(name="span"):
+        ...
+
+
+def test_new_custom_sampling_context(sentry_init):
```
ericapisani: I think both of the test cases introduced here are good, but I think the names could be improved slightly to make it easier to understand (in terms of what broke) should they ever fail.

For this particular test, maybe something like `test_custom_sampling_context_update_to_context_value_persists` could work?

sentrivana: https://github.com/getsentry/sentry-python/pull/5628/commits/d1618c4e2ca0c10391e06bf840d1ad112f3a46d3

sentry_sdk/tracing_utils.py
sentrivana: Ideally you could set whatever you want on the custom sampling context, so I wouldn't validate. But mistakenly overriding existing keys is a concern. I'm thinking we could put the keys set by the SDK into their own subdictionary -- we're also doing this in the old code path with `transaction_context`. Just need to think of a nice new name since there are no transactions anymore in span first.

sentrivana: Maybe just [`span`](https://github.com/getsentry/sentry-python/pull/5628/changes/c64025159bf8c0b17acdd44170f7d5f1d9e6a8b5)? Or `span_context`? Or just `context`? Naming is hard.

sentrivana: Maybe `span_context` is actually the way to go -- very unique, so unlikely to collide with anything user-set, and similar to `transaction_context` so there's some continuity?

Went with that [here](https://github.com/getsentry/sentry-python/pull/5628/commits/3142247ecbadf8e40e608d809f90e3b063227ead)

ericapisani: I agree, naming is hard 😅 

I think `span_context` is a great name 👍🏻

Issue #5627, opened 2026-03-10
Handle base64 encoded images from pydantic-ai's ImageURL


ericapisani:
We handle binary content correctly via the `BinaryContent` class, but currently we have no handling in place for if a base64 encoded image comes through Pydantic AI's `imageURL` class.

PR #5629, opened 2026-03-10
feat(pydantic-ai): Support ImageUrl content type in span instrumentation

ericapisani:
Fixes PY-2129 and #5627 

Add handling for the pydantic-ai `ImageUrl` message content type in span instrumentation.

Previously, only `BinaryContent` was handled for non-text message parts. With recent pydantic-ai versions, users can pass `ImageUrl` objects as part of their prompts. Without handling this type, `ImageUrl` items would fall through to `safe_serialize`, losing structured information about the content.

### DIFF
--- a/tests/integrations/pydantic_ai/test_pydantic_ai.py
+++ b/tests/integrations/pydantic_ai/test_pydantic_ai.py
@@ -2928,15 +2928,14 @@ async def test_invoke_agent_image_url(
 
     chat_spans = [s for s in transaction["spans"] if s["op"] == "gen_ai.chat"]
     for chat_span in chat_spans:
-        if "gen_ai.request.messages" in chat_span.get("data", {}):
-            messages_data = _get_messages_from_span(chat_span["data"])
-            for msg in messages_data:
-                if "content" not in msg:
-                    continue
-                for content_item in msg["content"]:
-                    if content_item.get("type") == "image":
-                        assert content_item["content"] == expected_content
-                        found_image = True
+        messages_data = _get_messages_from_span(chat_span["data"])
+        for msg in messages_data:
+            if "content" not in msg:
+                continue
+            for content_item in msg["content"]:
+                if content_item.get("type") == "image":
+                    assert content_item["content"] == expected_content
+                    found_image = True
 
     assert found_image, "Image content item should be found in messages data"
 


sentry_sdk/integrations/pydantic_ai/consts.py
```diff
@@ -1 +1,9 @@
+import re
+
 SPAN_ORIGIN = "auto.ai.pydantic_ai"
+
+# Matches data URLs with base64-encoded content, e.g. "data:image/png;base64,iVBORw0K..."
+# Group 1: MIME type (e.g. "image/png"), Group 2: base64 data
+DATA_URL_BASE64_REGEX = re.compile(
+    r"^data:([a-zA-Z]+/[a-zA-Z]+);base64,([A-Za-z0-9+/\-_]+={0,2})$"
+)
```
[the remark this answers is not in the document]
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/5629/commits/2f01be8981424de3c5ce30b62c2f824a312947e6

sentry_sdk/integrations/pydantic_ai/consts.py
```diff
@@ -1 +1,9 @@
+import re
+
 SPAN_ORIGIN = "auto.ai.pydantic_ai"
+
+# Matches data URLs with base64-encoded content, e.g. "data:image/png;base64,iVBORw0K..."
+# Group 1: MIME type (e.g. "image/png"), Group 2: base64 data
+DATA_URL_BASE64_REGEX = re.compile(
+    r"^data:([a-zA-Z0-9][a-zA-Z0-9.+\-]*/[a-zA-Z0-9][a-zA-Z0-9.+\-]*);base64,([A-Za-z0-9+/\-_]+={0,2})$"
+)
```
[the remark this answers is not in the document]
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/5629/commits/eb4393dd25c7548569b186689fa80d26b1ad0ff8

sentry_sdk/integrations/pydantic_ai/spans/utils.py
```diff
@@ -1,15 +1,51 @@
...
+    """
+    data_url_matches = DATA_URL_BASE64_REGEX.match(item.url)
+
+    if data_url_matches:
+        mime_type = data_url_matches[1] or "image"
+        return {
+            "type": "image",
+            "mime_type": mime_type,
+            "content": BLOB_DATA_SUBSTITUTE,
+        }
```
alexander-alderman-webb: Let's remove the `"mime_type"` field altogether.
I know we have unreliable heuristics for detecting the `mime_type` from the `media_type` for the `BinaryContent` class as well. Ultimately we must prioritize reporting accurate data over sending the wrong data. This means that "less is more" in this case, especially since no one has explicitly requested the `mime_type`.

An example of where the current logic breaks down is when the first regex group matches an empty string. See https://www.rfc-editor.org/rfc/rfc2397. But we need not concern ourselves with these edge cases as we can just leave sending the `mime_type` out of scope.

> If <mediatype> is omitted, it defaults to text/plain;charset=US-ASCII.

```suggestion
    if data_url_matches:
        return {
            "type": "image",
            "content": BLOB_DATA_SUBSTITUTE,
        }
```

tests/integrations/pydantic_ai/test_pydantic_ai.py
```diff
@@ -2797,6 +2812,362 @@ async def test_set_usage_data_with_cache_tokens(sentry_init, capture_events):
...
+        image_url = ImageUrl(
+            url="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs"
+        )
+        user_part = UserPromptPart(content=["Look at this image:", image_url])
+        mock_msg = MagicMock()
+        mock_msg.parts = [user_part]
+        mock_msg.instructions = None
+
+        _set_input_messages(span, [mock_msg])
+        span.finish()
```
alexander-alderman-webb: Tests should call public `pydantic-ai` functions instead of relying on our private `_set_input_messages()` helper.

We're likely to move `_set_input_messages()` in the future while the library's public API will not change. To quote [point 11 in the CONTRIBUTING.md](https://github.com/getsentry/sentry-python/blob/master/CONTRIBUTING.md?rgh-link-date=2026-03-06T15%3A17%3A23Z)

> Don't call private SDK stuff directly, just use the patched library in a way that triggers the patch.

I'll add a code suggestion to `test_image_url_redacts_base64_data_url_via_agent_run()` to avoid calling private functions.

tests/integrations/pydantic_ai/test_pydantic_ai.py
```diff
@@ -2797,6 +2812,362 @@ async def test_set_usage_data_with_cache_tokens(sentry_init, capture_events):
...
+    sentry_init(
+        integrations=[PydanticAIIntegration()],
+        traces_sample_rate=1.0,
+        send_default_pii=True,
+    )
+
+    events = capture_events()
+    image_url = ImageUrl(
+        url="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs"
+    )
```
alexander-alderman-webb: You could re-use the test boilerplate by parameterizing this test on different url strings and removing the other tests.

```suggestion
@pytest.mark.parametrize("image_url,is_truncated", (
    ("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs", True)
    ("https://example.com/image.png", False),
    ("https://example.com/api?data=iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs", True),
   ("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciLz4=", True),
   ("data:text/plain;charset=utf-8;name=hello.txt;base64,SGVsbG8sIFdvcmxkIQ==", True),
))
@pytest.mark.asyncio
async def test_image_url_redacts_base64_data_url_via_agent_run(
    sentry_init, capture_events, image_url, is_truncated
):
    agent = Agent("test", name="test_image_url_agent")

    sentry_init(
        integrations=[PydanticAIIntegration()],
        traces_sample_rate=1.0,
        send_default_pii=True,
    )

    events = capture_events()
    image_url = ImageUrl(
        url=image_url
    )
```

alexander-alderman-webb (changes requested):
Looks good overall! Two points of feedback:
- We can remove another `mime_type` field, since there's not been a request for the SDK to send this information 😅 
- We can write the tests in a way that reduces future work for us.

Could you also PR to [the AI Agents Insight module devdocs](https://github.com/getsentry/sentry-docs/blob/master/develop-docs/sdk/telemetry/traces/modules/ai-agents.mdx) to document the regex 🙏 
Include a sentence about possible cases in image URLs and the regex that you suggest other SDKs to follow as well!

tests/integrations/pydantic_ai/test_pydantic_ai.py
ericapisani: Sounds good, happy to make the changes. 

As a side note - this came from seeing other tests (specifically the binary content-related tests) within the file following a similar pattern. I'll make a cleanup TODO to update the other tests so that everything is consistent with respect to using only public apis/functions.

ericapisani: Created https://linear.app/getsentry/issue/PY-2132/refactor-test-pydantic-ai-to-remove-usage-of-internal-functions-within

sentry_sdk/integrations/pydantic_ai/spans/utils.py
ericapisani: Addressed in [592b100](https://github.com/getsentry/sentry-python/pull/5629/commits/592b100df8095799373d871fcb524a8fd8890e8e)

sentry_sdk/integrations/pydantic_ai/spans/utils.py
```diff
@@ -1,15 +1,50 @@
...
+    """Serialize an ImageUrl content item for span data.
+
+    For data URLs containing base64-encoded images, the content is redacted.
+    For regular HTTP URLs, the URL string is preserved.
+    """
+    url = str(item.url)
+    data_url_match = DATA_URL_BASE64_REGEX.match(url)
+
+    if data_url_match:
+        return {
```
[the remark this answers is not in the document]
ericapisani: Updated the description as we're not doing this anymore

tests/integrations/pydantic_ai/test_pydantic_ai.py
```diff
@@ -2797,6 +2796,151 @@ async def test_set_usage_data_with_cache_tokens(sentry_init, capture_events):
...
+    image_url = ImageUrl(url=url, **image_url_kwargs)
+    await agent.run([image_url, "Describe this image"])
+
+    (transaction,) = events
+
+    found_image = False
+
+    chat_spans = [s for s in transaction["spans"] if s["op"] == "gen_ai.chat"]
+    for chat_span in chat_spans:
+        if "gen_ai.request.messages" in chat_span.get("data", {}):
```
alexander-alderman-webb: can this if statement be removed? 😄

ericapisani: Done! ✨ https://github.com/getsentry/sentry-python/pull/5629/commits/7b90b6603aaf8ac658d18449e4ed80f870ce09c6

sentry_sdk/integrations/pydantic_ai/spans/ai_client.py
```diff
@@ -50,6 +53,7 @@
     TextPart = None
     ThinkingPart = None
     BinaryContent = None
+    ImageUrl = None
```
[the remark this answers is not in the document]
ericapisani: `ImageUrl` has been around since at least [v0.0.28](https://github.com/pydantic/pydantic-ai/releases/tag/v0.0.28) so it'll definitely exist for all the versions that we support

PR #5631, opened 2026-03-10
fix(ai): Truncate list-based message content in AI monitoring

### LINKED ISSUE #5630: Investigate "Message should be trimmed" test failure within pydantic-ai integration
Within the "Long Input Agent Test" the `checkMessageTrimming` assertion is failing.

ericapisani:
The message truncation logic in `_truncate_single_message_content_if_present` only handled string `content`, so messages using list-based content (e.g. pydantic-ai multimodal format with text and image_url parts) were passed through without truncation. This could cause oversized payloads to be sent to Sentry.

This extends the truncation to handle list content by iterating through parts and truncating `text` entries against a shared character budget, while leaving non-text parts (like image URLs) untouched.

Fixes #5630


ericapisani:
Moving this back into draft as there are conversations that would impact whether we need these changes

alexander-alderman-webb (approved):

Issue #5624, opened 2026-03-10
Sentry removes chained exceptions during ASGI application crash


roy-work:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.35.0

### Steps to Reproduce

We have an ASGI application; we use Sentry in it, by way of calling `sentry_sdk.init`.

We recently had a incident; during that incident, the ASGI application was crashing. Our logs contained:
```
2026-03-03 16:59:59,994 - uvicorn.error - ERROR - Exception in ASGI application
Traceback (most recent call last):
  File "/…/.venv/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 366, in run_asgi
    result = await app(self.scope, self.receive, self.send)
[snipped]
  File "/…/.venv/lib/python3.10/site-packages/sentry_sdk/integrations/starlette.py", line 408, in _sentry_patched_asgi_app
    return await middleware(scope, receive, send)
  File "/…/.venv/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py", line 179, in _run_asgi3
    return await self._run_app(scope, receive, send, asgi_version=3)
  File "/…/.venv/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py", line 268, in _run_app
    raise exc from None
  File "/…/.venv/lib/python3.10/site-packages/sentry_sdk/integrations/asgi.py", line 263, in _run_app
    return await self.app(
  File "/…/.venv/lib/python3.10/site-packages/starlette/applications.py", line 113, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/…/.venv/lib/python3.10/site-packages/sentry_sdk/integrations/starlette.py", line 150, in _create_span_call
    return await old_call(app, scope, receive, send, **kwargs)
[snip more frames]
  File "/…/app/middleware/authorization_middleware.py", line 79, in __call__
    verifier.verify(token, request)
  File "/…/app/token_verifier.py", line 72, in verify
    raise HTTPException(
starlette.exceptions.HTTPException: 401: Failed to Authorize"
```

Now, that exception is a bug; I think we're supposed to return a 401 _response_, not raise.

But what hampered us during this incident is that line 72 there is:
```
        except Exception as e:
            raise HTTPException(
                status_code=status.HTTP_401_UNAUTHORIZED,
                detail="Failed to Authorize",
                headers={"WWW-Authenticate": "Bearer"},
            ) from e
```

[...]

PR #5632, opened 2026-03-10
fix: Just re-raise the original exception, as-is

roy-work:
### Description

In Python `raise … from None` is an explicit request to raise the exception but to suppress any exception chaining. (See: https://docs.python.org/3/tutorial/errors.html#exception-chaining)

This means that if, when an ASGI exception crashes, an exception causes another exception, that when the second exception (which has the first chained to it) reaches this point in the stack, these lines will re-raise it, but remove the chained exception. This can be confusing: ASGI frameworks will normally log a stack trace of the crashing application, but that stack trace will now be truncated, due to this `from None` removing the chained exception.

The chained exception (which is the exception that kicked off the entire stack unwind) is usually the more interesting one, when you're debugging.

Neither of the two changed lines seem to have a reason for `… from None`. Additionally, since they just re-raise the original exception, simply "`raise`" suffices.

#### Issues

* resolves: #5624

#### Reminders

### DIFF
--- a/sentry_sdk/integrations/asgi.py
+++ b/sentry_sdk/integrations/asgi.py
@@ -183,7 +183,7 @@ class SentryAsgiMiddleware:
 
             except Exception as exc:
                 self._capture_lifespan_exception(exc)
-                raise exc from None
+                raise
 
         _asgi_middleware_applied.set(True)
         try:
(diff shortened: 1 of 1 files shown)


alexander-alderman-webb:
We made the change on the major branch: https://github.com/getsentry/sentry-python/issues/5624#issuecomment-4090001685.

Thanks again for your report and effort 🙏 !

PR #5633, opened 2026-03-11
ref: Add span filtering to span first (15)

sentrivana:
In span first, we're introducing a new span filtering `init` option called `ignore_spans` which allows to ignore spans at start time based on name and attributes.

- Spans ignored this way emit a new `ignored` client report (set via `NoOpStreamedSpan.unsampled_reason`).
- If a segment matches one of the rules in `ignore_spans`, the whole span tree will be ignored. The implementation achieves this by allowing `NoOpStreamedSpan`s to be set on the scope in this one single case, so that their children can easily determine their segment was dropped.
- If a segment is not ignored but one of its children spans (regardless of depth) is, ONLY that one span will be ignored, not its whole subtree.

Spec: https://develop.sentry.dev/sdk/telemetry/spans/filtering/#filter-with-ignorespans

### DIFF
--- a/sentry_sdk/_types.py
+++ b/sentry_sdk/_types.py
@@ -1,4 +1,9 @@
-from re import Pattern
+try:
+    from re import Pattern
+except ImportError:
+    # 3.6
+    from typing import Pattern
+
 from typing import TYPE_CHECKING, TypeVar, Union
 
 
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -11,6 +11,12 @@ from random import Random
 from urllib.parse import quote, unquote
 import uuid
 
+try:
+    from re import Pattern
+except ImportError:
+    # 3.6
+    from typing import Pattern
+
 import sentry_sdk
 from sentry_sdk.consts import OP, SPANDATA, SPANSTATUS, SPANTEMPLATE
 from sentry_sdk.utils import (
@@ -1487,7 +1493,7 @@ def is_ignored_span(name: str, attributes: "Optional[Attributes]") -> bool:
         return False
 
     def _matches(rule: "Any", value: "Any") -> bool:
-        if isinstance(rule, re.Pattern):
+        if isinstance(rule, Pattern):
             if isinstance(value, str):
                 return bool(rule.fullmatch(value))
             else:
@@ -1496,7 +1502,7 @@ def is_ignored_span(name: str, attributes: "Optional[Attributes]") -> bool:
         return rule == value
 
     for rule in ignore_spans:
-        if isinstance(rule, (str, re.Pattern)):
+        if isinstance(rule, (str, Pattern)):
             if _matches(rule, name):
                 return True
 


sentry_sdk/tracing_utils.py
```diff
@@ -1478,6 +1479,51 @@ def _make_sampling_decision(
...
+            if "attributes" in rule:
+                for attribute, value in rule["attributes"].items():
+                    if attribute not in attributes or not _matches(
+                        value, attributes[attribute]
+                    ):
+                        attributes_match = False
+                        break
+
+            if name_matches and attributes_match:
+                return True
```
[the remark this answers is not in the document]
sentrivana: https://github.com/getsentry/sentry-python/pull/5633/commits/bf62edca90cd9e70aa1a292ada03f2952da4d4a0

sentry_sdk/tracing_utils.py
```diff
@@ -1478,6 +1479,51 @@ def _make_sampling_decision(
...
+    client = sentry_sdk.get_client()
+    ignore_spans = (client.options.get("_experiments") or {}).get("ignore_spans")
+
+    if not ignore_spans:
+        return False
+
+    def _matches(rule: "Any", value: "Any") -> bool:
+        if isinstance(rule, Pattern):
+            if isinstance(value, str):
+                return bool(rule.match(value))
```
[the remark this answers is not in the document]
sentrivana: https://github.com/getsentry/sentry-python/pull/5633/commits/0eca88b6172a930a9df5b5bc3b082eb9dd80bdca

sentry_sdk/_types.py
```diff
@@ -361,3 +361,11 @@ class SDKInfo(TypedDict):
     class TextPart(TypedDict):
         type: Literal["text"]
         content: str
+
+    IgnoreSpansName = Union[str, Pattern[str]]
+    IgnoreSpansContext = TypedDict(
+        "IgnoreSpansContext",
+        {"name": IgnoreSpansName, "attributes": Attributes},
+        total=False,
+    )
```
alexander-alderman-webb, line 376: Maybe we can have a proper `IgnoreSpansContext` class?

I think creating a dedicated `IgnoreSpansContext` would give us more control in the future, as we can add new parameters with default arguments, etc ...

sentrivana: Can you give me an example what that could look like? Would we require users to wrap their `ignore_spans` rules in the class, too?

alexander-alderman-webb: Yes, re-reading my comment that's super vague 😅 .
The idea would be to funnel everything through

```
class IgnoreSpansConfig:
	def __init__(self, rules: "list[IgnoreSpansContext]"):
		self.rules = rules + [
			IgnoreSpansContext(attributes = {"http.url": "https://api.openai.com/v1/traces"})
		]

class IgnoreSpansContext:
	def __init__(self, name: "IgnoreSpansName",  attributes: "Attributes"):
		self.name = name
		self.attributes = attributes
```

So the user would pass in

```python
ignore_spans=IgnoreSpansConfig([IgnoreSpansContext(name="/health")])
```

There would still be dedicated options for common filtering cases like the `trace_ignore_status_codes` option. The class could let us cater for cases where we want to filter out some spans by default, but give the user an option to override it by controlling `self.rules` themselves?

alexander-alderman-webb: Very rough idea. The motivation is that I found calls to `https://api.openai.com/v1/traces` clutter the trace view and it'd be nice to have the facility to globally disable them.

If there's another way to do this please let me know 🙏

sentrivana: So I think there's couple things here:
- Regarding automatic filtering, the way we're envisioning this in the span first future is that it'll be the [integrations that are responsible](https://develop.sentry.dev/sdk/telemetry/spans/filtering/) for that, and also for providing finer controls to the user. That being said, we're also kind of moving towards a future where we don't filter much by default, but provide tools (like `ignore_spans`) to do that.
- `ignore_spans` should be a user-friendly way for users to filter out spans in addition to the adjustable filtering mechanism we have in the integrations. So with that in mind, I'd optimize for good and easy UX, where the list/dict approach wins over custom classes.

So basically, if we want to filter out specific spans, I'd have a separate mechanism for the user-friendly way of doing things (`ignore_spans`) and another one for us doing it automatically (preferably via the owning integration).

Also, side note, since `ignore_spans` is evaluated at span start, it can't take into account status codes, so we'll need a separate mechanism for that.

Issue #5659, opened 2026-03-13
[Anthropic integration] Set the `gen_ai.response.id` property on the span


ericapisani:
Although an optional property, this is something that we have available to us through the [`result.id`](<http://result.id>) and [`event.message.id`](<http://event.message.id>) for non-streaming and streaming responses respectively.

PR #5662, opened 2026-03-13
feat(anthropic): Set gen_ai.response.id span attribute

ericapisani:
Set the `gen_ai.response.id` property on spans created by the Anthropic integration.

For non-streaming responses, the ID is read from `result.id` on the Message object. For streaming responses, it's captured from `event.message.id` in the `message_start` event and threaded through the iterator to be set when the stream completes.

The `_collect_ai_data` function's return tuple is extended with the new `response_id` field, and `_set_output_data` accepts an optional `response_id` parameter to set on the span.

Refs PY-2137
Closes https://github.com/getsentry/sentry-python/issues/5659

### DIFF
--- a/sentry_sdk/integrations/anthropic.py
+++ b/sentry_sdk/integrations/anthropic.py
@@ -147,7 +147,7 @@ def _collect_ai_data(
             # https://github.com/anthropics/anthropic-sdk-python/blob/9c485f6966e10ae0ea9eabb3a921d2ea8145a25b/src/anthropic/lib/streaming/_messages.py#L433-L518
             if event.type == "message_start":
                 model = event.message.model or model
-                response_id = getattr(event.message, "id", None) or response_id
+                response_id = event.message.id
 
                 incoming_usage = event.message.usage
                 usage.output_tokens = incoming_usage.output_tokens


sentry_sdk/integrations/anthropic.py
```diff
@@ -146,6 +147,7 @@ def _collect_ai_data(
             # https://github.com/anthropics/anthropic-sdk-python/blob/9c485f6966e10ae0ea9eabb3a921d2ea8145a25b/src/anthropic/lib/streaming/_messages.py#L433-L518
             if event.type == "message_start":
                 model = event.message.model or model
+                response_id = getattr(event.message, "id", None) or response_id
```
alexander-alderman-webb: I'm not sure if it's worth collecting this information? 😄 

I can't see a user coming to us wanting to know the ID of the `message_start` event.

alexander-alderman-webb: unless the ID represents the ID of the whole response

ericapisani: > unless the ID represents the ID of the whole response

That's 💯 what this is. It's clearer when it's seen in context, here's an example response that Claude has in their [API docs](https://platform.claude.com/docs/en/build-with-claude/streaming#basic-streaming-request):

```
event: message_start
data: {"type": "message_start", "message": {"id": "msg_1nZdL29xx5MUA1yADyHTEsnR8uuvGzszyY", "type": "message", "role": "assistant", "content": [], "model": "claude-opus-4-6", "stop_reason": null, "stop_sequence": null, "usage": {"input_tokens": 25, "output_tokens": 1}}}

event: content_block_start
data: {"type": "content_block_start", "index": 0, "content_block": {"type": "text", "text": ""}}

event: ping
data: {"type": "ping"}

event: content_block_delta
data: {"type": "content_block_delta", "index": 0, "delta": {"type": "text_delta", "text": "Hello"}}

event: content_block_delta
data: {"type": "content_block_delta", "index": 0, "delta": {"type": "text_delta", "text": "!"}}

event: content_block_stop
data: {"type": "content_block_stop", "index": 0}

event: message_delta
data: {"type": "message_delta", "delta": {"stop_reason": "end_turn", "stop_sequence":null}, "usage": {"output_tokens": 15}}

event: message_stop
data: {"type": "message_stop"}
```

alexander-alderman-webb: okay sounds good, each streamed response has one `message_start`, and only the `message_start` has an ID 👍 .

Let's not accumulate and just access the property directly since we know at this point that event is of type `RawMessageStartEvent`:


```suggestion
                response_id = event.message.id
```

ericapisani: Address in 9ab864e8

PR #5663, opened 2026-03-13
feat(anthropic): Set gen_ai.response.finish_reasons on span

### LINKED ISSUE #5658: [Anthropic integration] Set `gen_ai.response.finish_reason` on span when available
While an optional attribute, we have the ability to set the `gen_ai.response.finish_reason` span within the Anthropic integration, either through the `result.stop_reason` for non-streaming responses, or the `event.delta.stop_reason` for streaming messages. 

We need to ensure that we set the `gen_ai.response.finish_reason` span when available in the Anthropic integration, similar to what we do for other integrations like Google, Hugging Face, and LangChain.

ericapisani:
Set the `gen_ai.response.finish_reasons` span attribute from Anthropic API responses.

Extracts `stop_reason` from both non-streaming (`result.stop_reason`) and streaming (`message_delta` event) responses. The value is stored as an array of strings per [OpenTelemetry semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/).

Refs PY-2136
Fixes #5658

### DIFF
--- a/sentry_sdk/integrations/anthropic.py
+++ b/sentry_sdk/integrations/anthropic.py
@@ -359,7 +359,7 @@ def _set_output_data(
     Set output data for the span based on the AI response."""
     span.set_data(SPANDATA.GEN_AI_RESPONSE_MODEL, model)
     if finish_reason is not None:
-        span.set_data(SPANDATA.GEN_AI_RESPONSE_FINISH_REASONS, finish_reason)
+        span.set_data(SPANDATA.GEN_AI_RESPONSE_FINISH_REASONS, [finish_reason])
     if should_send_default_pii() and integration.include_prompts:
         output_messages: "dict[str, list[Any]]" = {
             "response": [],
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/anthropic.py
```diff
@@ -186,6 +187,10 @@ def _collect_ai_data(
                     usage.cache_read_input_tokens = cache_read_input_tokens
                 # TODO: Record event.usage.server_tool_use
 
+                stop_reason = getattr(event.delta, "stop_reason", None)
+                if stop_reason is not None:
+                    usage.finish_reason = stop_reason
```
alexander-alderman-webb, line 192: I find it counter-intuitive that we record finish reasons on the token usage tracker.
Maybe we follow the same approach as with `response_id`?

ericapisani: Good shout. Will clean this up 🧹

ericapisani:
Closing in favour of https://github.com/getsentry/sentry-python/pull/5678

PR #5678, opened 2026-03-16
feat(anthropic): Record finish reasons in AI monitoring spans

ericapisani:
Add `GEN_AI_RESPONSE_FINISH_REASONS` span data to the Anthropic integration by capturing `stop_reason` from API responses.

For non-streaming responses, the `stop_reason` is read directly from the `Message` result. For streaming responses, it's extracted from the `MessageDeltaEvent` delta and passed through the `_collect_ai_data` helper.

This brings the Anthropic integration in line with the OpenAI integration's finish reason tracking.

### DIFF
--- a/sentry_sdk/integrations/langchain.py
+++ b/sentry_sdk/integrations/langchain.py
@@ -554,7 +554,8 @@ class SentryLangchainCallback(BaseCallbackHandler):  # type: ignore[misc]
                     finish_reason = generation.generation_info.get("finish_reason")
                     if finish_reason is not None:
                         span.set_data(
-                            SPANDATA.GEN_AI_RESPONSE_FINISH_REASONS, finish_reason
+                            SPANDATA.GEN_AI_RESPONSE_FINISH_REASONS,
+                            [finish_reason],
                         )
                 except AttributeError:
                     pass
--- a/tests/integrations/langchain/test_langchain.py
+++ b/tests/integrations/langchain/test_langchain.py
@@ -297,6 +297,12 @@ def test_langchain_agent(
             f"and include_prompts={include_prompts}"
         )
 
+    # Verify finish_reasons is always an array of strings
+    assert chat_spans[0]["data"][SPANDATA.GEN_AI_RESPONSE_FINISH_REASONS] == [
+        "function_call"
+    ]
+    assert chat_spans[1]["data"][SPANDATA.GEN_AI_RESPONSE_FINISH_REASONS] == ["stop"]
+
     # Verify that available tools are always recorded regardless of PII settings
     for chat_span in chat_spans:
         span_data = chat_span.get("data", {})


tests/integrations/anthropic/test_anthropic.py
```diff
@@ -504,6 +508,7 @@ async def test_streaming_create_message_async(
     sentry_init(
         integrations=[AnthropicIntegration(include_prompts=include_prompts)],
         traces_sample_rate=1.0,
+        default_integrations=False,
```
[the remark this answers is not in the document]
ericapisani: I couldn't initially narrow down the "why" in terms of why this test was behaving this way vs. the others, but this test was consistently failing because subprocess spans were being captured due to default integrations being automatically enabled, leading to the `assert len(event["spans"]) == 1` within this test failing.

Adding this ensures that only the Anthropic spans are being captured.

sentry_sdk/integrations/anthropic.py
```diff
@@ -542,12 +553,15 @@ def _set_output_data(
...
     response_id: "str | None" = None,
+    finish_reasons: "list[str] | None" = None,
 ) -> None:
     """
     Set output data for the span based on the AI response."""
     span.set_data(SPANDATA.GEN_AI_RESPONSE_MODEL, model)
     if response_id is not None:
         span.set_data(SPANDATA.GEN_AI_RESPONSE_ID, response_id)
+    if finish_reasons is not None:
+        span.set_data(SPANDATA.GEN_AI_RESPONSE_FINISH_REASONS, finish_reasons)
```
alexander-alderman-webb: Since `anthropic` always generates a single output (i.e., no candidates) we can work with the single finish reason in `_collect_ai_data` (i.e., keep track of a string rather than a list).

If we only convert the finish reason to list form in `_set_output_data` the preceding code is easier to reason about.

```suggestion
    finish_reason: "str | None" = None,
) -> None:
"""
   Set output data for the span based on the AI response."""
span.set_data(SPANDATA.GEN_AI_RESPONSE_MODEL, model)
if response_id is not None:
span.set_data(SPANDATA.GEN_AI_RESPONSE_ID, response_id)
    if finish_reason is not None:
        span.set_data(SPANDATA.GEN_AI_RESPONSE_FINISH_REASONS, [finish_reason])
```

sentry_sdk/integrations/anthropic.py
```diff
@@ -221,18 +223,18 @@ def _collect_ai_data(
...
 
-                return (
-                    model,
-                    usage,
-                    content_blocks,
-                    response_id,
-                )
+                stop_reason = getattr(event.delta, "stop_reason", None)
+                if stop_reason is not None:
+                    finish_reasons = [stop_reason]
```
alexander-alderman-webb: non-blocking nitpick: We don't need the `getattr()`, since `event.type == "message_delta"` forces `event` to be an instance of `MessageDeltaEvent`.

```suggestion
                if event.delta.stop_reason is not None:
                    finish_reason = event.delta.stop_reason
```

sentry_sdk/integrations/anthropic.py
ericapisani: Addressed https://github.com/getsentry/sentry-python/pull/5678/commits/e146224f368123b39e766a865ec082fd1d9665ec

sentry_sdk/integrations/anthropic.py
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/5678/commits/e146224f368123b39e766a865ec082fd1d9665ec

PR #5682, opened 2026-03-17
feat(integrations): instrument pyreqwest tracing

servusdei2018:
### Description

Implements request tracing for the [`pyreqwest`](https://github.com/MarkusSintonen/pyreqwest) HTTP library. Includes tests and CI updates.

#### Issues

Resolves MarkusSintonen/pyreqwest#30.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,66 @@
 # Changelog
 
+## 2.57.0
+
+### New Features ✨
+
+#### Langchain
+
+- Set `gen_ai.operation.name` and `gen_ai.pipeline.name` on LLM spans by @ericapisani in [#5849](https://github.com/getsentry/sentry-python/pull/5849)
+- Broaden AI provider detection beyond OpenAI and Anthropic by @ericapisani in [#5707](https://github.com/getsentry/sentry-python/pull/5707)
+- Update LLM span operation to `gen_ai.generate_text` by @ericapisani in [#5796](https://github.com/getsentry/sentry-python/pull/5796)
+
+#### Other
+
+- Add experimental async transport by @BYK in [#5646](https://github.com/getsentry/sentry-python/pull/5646)
+
+  See https://github.com/getsentry/sentry-python/discussions/5919 for details.
+
+### Bug Fixes 🐛
+
+#### Openai
+
+- Only wrap types with `_iterator` for streamed responses by @alexander-alderman-webb in [#5917](https://github.com/getsentry/sentry-python/pull/5917)
+- Always set `gen_ai.response.streaming` for Responses by @alexander-alderman-webb in [#5697](https://github.com/getsentry/sentry-python/pull/5697)
+- Simplify Responses input handling by @alexander-alderman-webb in [#5695](https://github.com/getsentry/sentry-python/pull/5695)
+- Use `max_output_tokens` for Responses API by @alexander-alderman-webb in [#5693](https://github.com/getsentry/sentry-python/pull/5693)
+- Always set `gen_ai.response.streaming` for Completions by @alexander-alderman-webb in [#5692](https://github.com/getsentry/sentry-python/pull/5692)
+- Simplify Completions input handling by @alexander-alderman-webb in [#5690](https://github.com/getsentry/sentry-python/pull/5690)
+- Simplify embeddings input handling by @alexander-alderman-webb in [#5688](https://github.com/getsentry/sentry-python/pull/5688)
+
+#### Other
+
+- (google-genai) Guard response extraction by @alexander-alderman-webb in [#5869](https://github.com/getsentry/sentry-python/pull/5869)
+- Add cycle detection to exceptions_from_error by @ericapisani in [#5880](https://github.com/getsentry/sentry-python/pull/5880)
+
+### Internal Changes 🔧
+
+#### Ai
+
+- Remove unused GEN_AI_PIPELINE operation constant by @ericapisani in [#5886](https://github.com/getsentry/sentry-python/pull/5886)
+- Rename generate_text to text_completion by @ericapisani in [#5885](https://github.com/getsentry/sentry-python/pull/5885)
+
+#### Langchain
+
+- Add text completion test by @alexander-alderman-webb in [#5740](https://github.com/getsentry/sentry-python/pull/5740)
+- Add tool execution test by @alexander-alderman-webb in [#5739](https://github.com/getsentry/sentry-python/pull/5739)
+- Add basic agent test with Responses call by @alexander-alderman-webb in [#5726](https://github.com/getsentry/sentry-python/pull/5726)
+- Replace mocks with `httpx` types by @alexander-alderman-webb in [#5724](https://github.com/getsentry/sentry-python/pull/5724)
+- Consolidate span origin assertion by @alexander-alderman-webb in [#5723](https://github.com/getsentry/sentry-python/pull/5723)
+- Consolidate available tools assertion by @alexander-alderman-webb in [#5721](https://github.com/getsentry/sentry-python/pull/5721)
+
+#### Openai
+
+- Replace mocks with httpx types for streaming Responses by @alexander-alderman-webb in [#5882](https://github.com/getsentry/sentry-python/pull/5882)
+- Replace mocks with httpx types for streaming Completions by @alexander-alderman-webb in [#5879](https://github.com/getsentry/sentry-python/pull/5879)
+- Move input handling code into API-specific functions by @alexander-alderman-webb in [#5687](https://github.com/getsentry/sentry-python/pull/5687)
+
+#### Other
+
+- (asyncpg) Normalize query whitespace in integration by @ericapisani in [#5855](https://github.com/getsentry/sentry-python/pull/5855)
+- Exclude compromised litellm versions by @alexander-alderman-webb in [#5876](https://github.com/getsentry/sentry-python/pull/5876)
+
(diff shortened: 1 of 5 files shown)


sentrivana:
Hey @servusdei2018, thanks for the PR, we'll review next week!

sentry_sdk/integrations/pyreqwest.py
```diff
@@ -0,0 +1,186 @@
...
+                    self._sentry_instrumented = True
+                except (TypeError, AttributeError):
+                    # In case the instance itself is immutable or doesn't allow extra attributes
+                    pass
+        return original_method(self, *args, **kwargs)
+
+    setattr(cls, method_name, sentry_patched_method)
+
+
+async def sentry_async_middleware(request: "Any", next_handler: "Any") -> "Any":
```
MarkusSintonen: You could use the types and type stubs to add types eg here ^ see eg https://github.com/MarkusSintonen/pyreqwest/blob/main/python/pyreqwest/middleware/types.py

sentry_sdk/integrations/pyreqwest.py
```diff
@@ -0,0 +1,186 @@
...
+
+def sentry_sync_middleware(request: "Any", next_handler: "Any") -> "Any":
+    if sentry_sdk.get_client().get_integration(PyreqwestIntegration) is None:
+        return next_handler.run(request)
+
+    parsed_url = None
+    with capture_internal_exceptions():
+        parsed_url = parse_url(str(request.url), sanitize=False)
+
+    with start_span(
```
MarkusSintonen: There is lot of same logic in both sentry_async_middleware and sentry_sync_middleware. You could extract the common stuff into a shared contextmanager for span handling

sentry_sdk/integrations/pyreqwest.py
```diff
@@ -0,0 +1,186 @@
...
+
+def _patch_pyreqwest() -> None:
+    # Patch Client Builders
+    try:
+        from pyreqwest.client import ClientBuilder, SyncClientBuilder  # type: ignore[import-not-found]
+
+        _patch_builder_method(ClientBuilder, "build", sentry_async_middleware)
+        _patch_builder_method(SyncClientBuilder, "build", sentry_sync_middleware)
+    except ImportError:
+        pass
```
sentrivana: Can we move the imports and the `ImportError` guard to the imports at the top of the file, and raise `DidNotEnable` if anything is missing? (Unless any of the imports is optional, in which case we shouldn't raise.)

Basically you'd have:

```python
# sdk imports, etc.

try:
    from pyreqwest.client import ...
    from pyreqwest.request import ...
except ImportError:
    raise DidNotEnable("pyreqwest not installed or incompatible version installed")
```

Then we can also remove the `importlib.util.find_spec`.

sentry_sdk/integrations/pyreqwest.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def _patch_builder_method(cls: type, method_name: str, middleware: "Any") -> None:
+    if not hasattr(cls, method_name):
+        return
+
+    original_method = getattr(cls, method_name)
+
+    def sentry_patched_method(self: "Any", *args: "Any", **kwargs: "Any") -> "Any":
```
sentrivana, line 59:
```suggestion

    @functools.wraps(original_method)
    def sentry_patched_method(self: "Any", *args: "Any", **kwargs: "Any") -> "Any":
```

You'll need to import `functools`, too.

sentry_sdk/integrations/pyreqwest.py
```diff
@@ -0,0 +1,186 @@
...
+    with capture_internal_exceptions():
+        parsed_url = parse_url(str(request.url), sanitize=False)
+
+    with start_span(
+        op=OP.HTTP_CLIENT,
+        name="%s %s"
+        % (
+            request.method,
+            parsed_url.url if parsed_url else SENSITIVE_DATA_SUBSTITUTE,
+        ),
```
sentrivana:
```suggestion
        name=f"{request.method} {parsed_url.url if parsed_url else SENSITIVE_DATA_SUBSTITUTE}",
```

Luckily we can use f-strings now that we've dropped 2.7 support :)

sentry_sdk/integrations/pyreqwest.py
```diff
@@ -0,0 +1,186 @@
...
+        _patch_builder_method(RequestBuilder, "build", sentry_async_middleware)
+        _patch_builder_method(RequestBuilder, "build_streamed", sentry_async_middleware)
+        _patch_builder_method(SyncRequestBuilder, "build", sentry_sync_middleware)
+        _patch_builder_method(
+            SyncRequestBuilder, "build_streamed", sentry_sync_middleware
+        )
+        _patch_builder_method(OneOffRequestBuilder, "send", sentry_async_middleware)
+        _patch_builder_method(SyncOneOffRequestBuilder, "send", sentry_sync_middleware)
+    except ImportError:
+        pass
```
sentrivana: Same comment as https://github.com/getsentry/sentry-python/pull/5682/changes#r2980238526

tests/integrations/pyreqwest/test_pyreqwest.py
```diff
@@ -0,0 +1,154 @@
...
+    yield port
+    server.shutdown()
+
+
+@pytest.fixture(autouse=True)
+def clear_captured_requests():
+    PyreqwestMockHandler.captured_requests.clear()
+
+
+@pytest.mark.skipif(pyreqwest is None, reason="pyreqwest not installed")
```
sentrivana: These `skipif`s can be removed -- I added an `__init__.py` alongside the test file that will make pytest skip the test suite if pyreqwest is not installed, so these checks are now unnecessary.

sentrivana:
I only now noticed CI is borked. Might have to do with the merge conflicts. Will take a look soon.

sentry_sdk/integrations/pyreqwest.py
```diff
@@ -0,0 +1,186 @@
...
+    def sentry_patched_method(self: "Any", *args: "Any", **kwargs: "Any") -> "Any":
+        if not getattr(self, "_sentry_instrumented", False):
+            integration = sentry_sdk.get_client().get_integration(PyreqwestIntegration)
+            if integration is not None:
+                self.with_middleware(middleware)
+                try:
+                    self._sentry_instrumented = True
+                except (TypeError, AttributeError):
+                    # In case the instance itself is immutable or doesn't allow extra attributes
+                    pass
```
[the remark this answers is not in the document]
sentrivana: I believe this might actually be happening @servusdei2018 -- I do appear to be getting multiple spans for the same request:

Screenshot 2026-03-25 at 10 06 34

This is my test app:

```python
import asyncio
from pyreqwest.client import SyncClientBuilder, ClientBuilder

import sentry_sdk
from sentry_sdk.integrations.pyreqwest import PyreqwestIntegration

sentry_sdk.init(
    traces_sample_rate=1.0,
    integrations=[
        PyreqwestIntegration(),
    ]
)


def sync_example():
    with SyncClientBuilder().build() as client:
        resp = client.get("https://httpbin.org/get").build().send()


async def async_example():
    async with ClientBuilder().build() as client:
        resp = await client.get("https://httpbin.org/get").build().send()
        resp = await client.post("https://httpbin.org/post").body_json({"hello": "world"}).build().send()


if __name__ == "__main__":
    with sentry_sdk.start_transaction(name="pyreqwest sync"):
        sync_example()

    with sentry_sdk.start_transaction(name="pyreqwest async"):
        asyncio.run(async_example())
```

(The sync example has the same issue.)

sentry_sdk/integrations/__init__.py
```diff
@@ -94,6 +94,7 @@ def iter_default_integrations(
     "sentry_sdk.integrations.openai.OpenAIIntegration",
     "sentry_sdk.integrations.openai_agents.OpenAIAgentsIntegration",
     "sentry_sdk.integrations.pydantic_ai.PydanticAIIntegration",
+    "sentry_sdk.integrations.pyreqwest.PyreqwestIntegration",
```
sentrivana: One more thing -- please remove from auto-enabling integrations -- we usually want to give a new integration a bit of time to iron out issues before we enable it for everyone by default.

servusdei2018:
@sentrivana @MarkusSintonen thank you for the reviews!

I added those changes, and refactored the middleware so the duplicate spans are no longer present (also strengthened the test cases to catch a regression there).

sentry_sdk/integrations/pyreqwest.py
```diff
@@ -0,0 +1,136 @@
...
+
+    original_method = getattr(cls, method_name)
+
+    def sentry_patched_method(self: "Any", *args: "Any", **kwargs: "Any") -> "Any":
+        if not getattr(self, "_sentry_instrumented", False):
+            integration = sentry_sdk.get_client().get_integration(PyreqwestIntegration)
+            if integration is not None:
+                self.with_middleware(middleware)
+                try:
+                    self._sentry_instrumented = True
```
MarkusSintonen, line 65: Did we have some test to check if this works and doesn't always give  the exception? (as these are native classes)

sentrivana:
@servusdei2018 Thanks for making the changes, looks good to me! Verified that the duplicate spans are also gone.

I'll add some more tests to your branch to make the test suite on par with the other HTTP client integrations and I'll fix the unrelated CI failures, and then we should be good to go. Nothing needed from your side.

PR #5687, opened 2026-03-18
ref(openai): Move input handling code into API-specific functions

alexander-alderman-webb:
### Description

Inline code from `_commmon_set_input_data()` to remove dead code in follow up PRs.

#### Reminders


sentry_sdk/integrations/openai.py
```diff
@@ -249,18 +234,38 @@ def _commmon_set_input_data(
...
+            SPANDATA.GEN_AI_REQUEST_FREQUENCY_PENALTY, frequency_penalty
+        )
+
+    temperature = kwargs.get("temperature")
+    if temperature is not None and _is_given(temperature):
+        set_data_normalized(SPANDATA.GEN_AI_REQUEST_TEMPERATURE, temperature)
+
+    top_p = kwargs.get("top_p")
+    if top_p is not None and _is_given(top_p):
+        set_data_normalized(SPANDATA.GEN_AI_REQUEST_TOP_P, top_p)
```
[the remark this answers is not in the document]
alexander-alderman-webb: my bad fixed

sentry_sdk/integrations/openai.py
```diff
@@ -219,48 +219,54 @@ def _calculate_token_usage(
...
+    messages: "Optional[Union[str, ResponseInputParam]]" = kwargs.get("input")
+
     tools = kwargs.get("tools")
     if tools is not None and _is_given(tools) and len(tools) > 0:
         set_data_normalized(
             span, SPANDATA.GEN_AI_REQUEST_AVAILABLE_TOOLS, safe_serialize(tools)
         )
 
+    model = kwargs.get("model")
+    if model is not None and _is_given(model):
```
ericapisani, line 237: How is the `model is not None` part of this conditional distinct from the checks within the `_is_given` method? Reading this and the contents of `_is_given`, I would've expected that the `None` check would also be part of `_is_given`

alexander-alderman-webb: `_is_given()` checks for the `NotGiven` and `Omit` sentinel values (which are distinct from the user explicitly providing a `None` value).

The `Omit` sentinel value instructs to omit the parameter from the model request, the `NotGiven` sentinel is the default value, and `None` is a value that is explicitly provided by the user and can be included in the model request. In that sense `None` is given 😅 

This function seems to be adapted from `openai-python`:

https://github.com/openai/openai-python/blob/acd0c54d8a68efeedde0e5b4e6c310eef1ce7867/src/openai/_utils/_utils.py#L133

ericapisani: Gotcha. I think what's a little confusing about this is that there's ambiguity in this conditional in that `None` could mean "not provided" (the default) or, as you've pointed out, "explicitly provided by the user but omitted from the span to keep envelope size down."

If "given" always means "explicitly provided by the user" regardless of value, I'm fine leaving this as-is (no `None` check inside `is_given`).

That said, since the `not None and is_given` pattern appears throughout the SDK, should we extract it into a helper? (This is a non-blocking nit)

alexander-alderman-webb: The default values are the `omit` and `not_given` singletons (of type `Omit` and `NotGiven`, respectively). For `openai` functions we patch, `None` is always explicitly provided by the user.

https://github.com/openai/openai-python/blob/acd0c54d8a68efeedde0e5b4e6c310eef1ce7867/src/openai/resources/responses/responses.py#L859

The difficulty with extracting this into a shared function is that `NotGiven` and `Omit` are defined separately by each library. For example, `openai` and `anthropic` use the same `Omit` and `NotGiven` classes, but they declare them separately 😞 .

ericapisani: Ah, ok, thanks for all that context, learned a lot today! 😄

PR #5691, opened 2026-03-18
fix(pydantic_ai): Fix AttributeError when patching ToolManager methods

### LINKED ISSUE #5686: regression in 2.55.0: pydantic-ai integration fails at startup (AttributeError: ToolManager has no attribute '_call_tool')
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.55.0

### Steps to Reproduce

1. Install sentry-sdk==2.55.0 and pydantic-ai-slim==1.69.0
2. Initialize Sentry with PydanticAIIntegration
3. Start the application

### Expected Result

Sentry initializes successfully with no errors.

### Actual Result

```
File "/app/.venv/lib/python3.13/site-packages/sentry_sdk/integrations/pydantic_ai/patches/tools.py", line 41, in _patch_tool_execution                                                                                                                                                                    
original_call_tool = ToolManager._call_tool                                                                                                                                                                                                                                                               
AttributeError: type object 'ToolManager' has no attribute '_call_tool'
```

This is a regression introduced in 2.55.0. The same setup works fine with 2.54.0:

| sentry-sdk | pydantic-ai-slim | Result   |
|------------|------------------|----------|
| 2.54.0     | 1.69.0           | ✅ Works  |
| 2.55.0     | 1.69.0           | ❌ Broken |

BillionClaw:
## Summary

Fixes #5686

The pydantic-ai integration fails at startup in sentry-python 2.55.0 when using pydantic-ai-slim 1.69.0+, raising:

## Root Cause

The `_patch_tool_execution()` function used `hasattr()` to check for method existence, but `hasattr()` can return `True` even when accessing the attribute raises `AttributeError` (e.g., for properties or descriptors).

This caused the code to enter the wrong branch and fail when trying to access `ToolManager._call_tool`.

## Fix

Instead of using `hasattr()`, we now try to access the method directly in a try-except block. This safely handles cases where:
- The method doesn't exist (`AttributeError`)
- The method exists but can't be accessed (e.g., properties that raise)

## Testing

- Verified the fix with pydantic-ai-slim 1.70.0
- All 98 existing tests pass
- The integration now initializes without errors

### DIFF
--- a/sentry_sdk/integrations/pydantic_ai/patches/tools.py
+++ b/sentry_sdk/integrations/pydantic_ai/patches/tools.py
@@ -28,12 +28,22 @@ except ImportError:
 
 
 def _patch_tool_execution() -> None:
-    if hasattr(ToolManager, "execute_tool_call"):
+    # Try to access the method directly, as hasattr() can return True
+    # but accessing the attribute may still raise AttributeError (e.g., for properties)
+    try:
+        ToolManager.execute_tool_call
         _patch_execute_tool_call()
+        return
+    except AttributeError:
+        pass
 
-    elif hasattr(ToolManager, "_call_tool"):
+    try:
+        ToolManager._call_tool
         # older versions
         _patch_call_tool()
+        return
+    except AttributeError:
+        pass
 
 
 def _patch_execute_tool_call() -> None:


sentrivana:
> The _patch_tool_execution() function used hasattr() to check for method existence, but hasattr() can return True even when accessing the attribute raises AttributeError (e.g., for properties or descriptors).

This might be true in general but not in this case:

```
$ ipython
Python 3.12.8 (main, Dec 17 2024, 12:40:19) [Clang 19.1.5 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.11.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: IPython supports combining unicode identifiers, eg F\vec<tab> will become F⃗, useful for physics equations. Play with \dot \ddot and others.

In [1]: from pydantic_ai._tool_manager import ToolManager

In [2]: hasattr(ToolManager, "_call_tool")
Out[2]: False
```

Also, if you look at the [code](https://github.com/getsentry/sentry-python/blob/c4df76bbbb9e73ee7b6add348fd4736945f46106/sentry_sdk/integrations/pydantic_ai/patches/tools.py#L30-L36):

```python
def _patch_tool_execution() -> None:
    if hasattr(ToolManager, "execute_tool_call"):
        _patch_execute_tool_call()

    elif hasattr(ToolManager, "_call_tool"):
        # older versions
        _patch_call_tool()
```

You'll notice we don't even attempt to patch `_call_tool` if `ToolManager` has an `execute_tool_call` method, which it [does](https://github.com/pydantic/pydantic-ai/blob/529c049534d7e7e9fa9715bfbc61c14c549c093e/pydantic_ai_slim/pydantic_ai/_tool_manager.py#L295) in 1.69.0. So that branch does not even get executed.

I could not reproduce [the original issue](https://github.com/getsentry/sentry-python/issues/5686) and the traceback does not appear to be from SDK v2.55.0, so unless proven otherwise I don't believe there is a regression in the first place.

PR #5692, opened 2026-03-18
fix(openai): Always set `gen_ai.response.streaming` for Completions

alexander-alderman-webb:
#### Reminders


sentry_sdk/integrations/openai.py
```diff
@@ -614,12 +610,14 @@ def _new_chat_completion_common(f: "Any", *args: "Any", **kwargs: "Any") -> "Any
     span.__enter__()
 
     span.set_data(SPANDATA.GEN_AI_SYSTEM, "openai")
+    is_streaming_response = kwargs.get("stream", False)
```
ericapisani: I don't think this will evaluate to `True` at any point, will want to double-check this. 

If this is the `stream` property coming in from the `Completions.create` method, then the only valid accepted values are `False` or the `Omit` sentinel value.

Normally, sentinel values evaluate to something truthy, but in this case, [`Omit` would evaluate to `False`](https://github.com/openai/openai-python/blob/acd0c54d8a68efeedde0e5b4e6c310eef1ce7867/src/openai/_types.py#L180).

ericapisani: Chatted offline, the `create` method is overloaded such that this can also be `True`

PR #5705, opened 2026-03-19
feat(langchain): Change LLM span operation to generate_text

### LINKED ISSUE #5679: Update the langchain/langgraph `on_llm_start` hook to use a different `op`
At the moment, the `op`is set to `gen_ai.pipeline` which is a legacy value that we have deprecated (and only appears in the Python SDK).

We need to update this to something more appropriate, either a text completion or chat completion. Which thing we update to will be determined by the outcome of [this RFC](<https://github.com/getsentry/rfcs/pull/153>).

ericapisani:
Update the LangChain integration to use `gen_ai.generate_text` as the span operation for LLM calls instead of `gen_ai.pipeline`. This aligns with the more specific semantics of what's happening when an LLM is invoked directly, as opposed to a broader pipeline execution.

Changes:
- Span operation changed from `OP.GEN_AI_PIPELINE` to `OP.GEN_AI_GENERATE_TEXT`
- Span name updated to `generate_text {model}` to include model info
- `GEN_AI_OPERATION_NAME` span data set to `"generate_text"`
- `GEN_AI_PIPELINE_NAME` span data set when a pipeline name is provided in kwargs

Fixes #5679 and PY-2143

### DIFF
--- a/sentry_sdk/integrations/langchain.py
+++ b/sentry_sdk/integrations/langchain.py
@@ -365,7 +365,6 @@ class SentryLangchainCallback(BaseCallbackHandler):  # type: ignore[misc]
                 or ""
             )
 
-            pipeline_name = kwargs.get("name")
             watched_span = self._create_span(
                 run_id,
                 parent_run_id,
@@ -377,10 +376,7 @@ class SentryLangchainCallback(BaseCallbackHandler):  # type: ignore[misc]
 
             span.set_data(SPANDATA.GEN_AI_OPERATION_NAME, "generate_text")
 
-            agent_name = _get_current_agent()
-            if agent_name:
-                span.set_data(SPANDATA.GEN_AI_AGENT_NAME, agent_name)
-
+            pipeline_name = kwargs.get("name")
             if pipeline_name:
                 span.set_data(SPANDATA.GEN_AI_PIPELINE_NAME, pipeline_name)
 
--- a/tests/integrations/langchain/test_langchain.py
+++ b/tests/integrations/langchain/test_langchain.py
@@ -27,8 +27,6 @@ from sentry_sdk.integrations.langchain import (
     SentryLangchainCallback,
     _transform_langchain_content_block,
     _transform_langchain_message_content,
-    _push_agent,
-    _pop_agent,
 )
 
 try:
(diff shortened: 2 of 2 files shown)


sentry_sdk/integrations/langchain.py
```diff
@@ -366,15 +365,25 @@ def on_llm_start(
...
+                name=f"generate_text {model}".strip(),
                 origin=LangchainIntegration.origin,
             )
             span = watched_span.span
 
+            span.set_data(SPANDATA.GEN_AI_OPERATION_NAME, "generate_text")
+
+            agent_name = _get_current_agent()
+            if agent_name:
+                span.set_data(SPANDATA.GEN_AI_AGENT_NAME, agent_name)
```
ericapisani: `on_llm_start` can be triggered by an agent, so this adds extra information that we were previously missing (mirrors what we do for chat spans)

sentry_sdk/integrations/langchain.py
```diff
@@ -366,15 +365,25 @@ def on_llm_start(
...
+
+            if pipeline_name:
+                span.set_data(SPANDATA.GEN_AI_PIPELINE_NAME, pipeline_name)
```
alexander-alderman-webb: nitpick, not blocking. I am requesting changes to address the point raised on Slack before merging.

Can you extract the `pipeline_name` right before it's used instead?

This would reduce the amount of code that you need to read to understand where the `pipeline_name` comes from. Small reduction in cognitive overhead when reading 😅 .

The check can also be `if agent_name is not None:` instead of `if agent_name:` 🙏 

```suggestion
watched_span = self._create_span(
run_id,
parent_run_id,
                op=OP.GEN_AI_GENERATE_TEXT,
                name=f"generate_text {model}".strip(),
origin=LangchainIntegration.origin,
)
span = watched_span.span

            span.set_data(SPANDATA.GEN_AI_OPERATION_NAME, "generate_text")

            agent_name = _get_current_agent()
            if agent_name:
                span.set_data(SPANDATA.GEN_AI_AGENT_NAME, agent_name)

			pipeline_name = kwargs.get("name")
            if pipeline_name is not None:
                span.set_data(SPANDATA.GEN_AI_PIPELINE_NAME, pipeline_name)
```

sentry_sdk/integrations/langchain.py
```diff
@@ -369,12 +368,18 @@ def on_llm_start(
...
+                name=f"generate_text {model}".strip(),
                 origin=LangchainIntegration.origin,
             )
             span = watched_span.span
 
+            span.set_data(SPANDATA.GEN_AI_OPERATION_NAME, "generate_text")
+
+            pipeline_name = kwargs.get("name")
+            if pipeline_name:
+                span.set_data(SPANDATA.GEN_AI_PIPELINE_NAME, pipeline_name)
```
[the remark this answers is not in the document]
ericapisani: It was discussed offline that this piece of data was going to be removed from all langchain functions. Right now, because of the scope of this PR, this will only be removed here, but a follow-up set of changes will be done to remove it in the other spots as well.

ericapisani:
Closing this as this was broken up into separate PRs to make it easier to merge the changes once the refactoring of the tests are in

PR #5715, opened 2026-03-20
ref(anthropic): Separate sync and async .create() patches

alexander-alderman-webb:
### Description

Remove generator pattern to allow capturing exception info on AI Client Spans in upcoming PRs.

#### Reminders


sentry_sdk/integrations/anthropic.py
```diff
@@ -638,6 +647,81 @@ def _sentry_patched_create_common(f: "Any", *args: "Any", **kwargs: "Any") -> "A
...
+            ) = _get_token_usage(result)
+
+            content_blocks = []
+            for content_block in result.content:
+                if hasattr(content_block, "to_dict"):
+                    content_blocks.append(content_block.to_dict())
+                elif hasattr(content_block, "model_dump"):
+                    content_blocks.append(content_block.model_dump())
+                elif hasattr(content_block, "text"):
+                    content_blocks.append({"type": "text", "text": content_block.text})
```
ericapisani, line 666: From what I've seen of the responses in the Claude API, `result.content` will always be a dict with the `type` property within it being what determines what type of content block it is. If that holds, I'm not sure we would end up in the 2nd and 3rd `elif` blocks.

Are there scenarios where `result.content` is something else?

sentry_sdk/integrations/anthropic.py
```diff
@@ -638,6 +647,81 @@ def _sentry_patched_create_common(f: "Any", *args: "Any", **kwargs: "Any") -> "A
...
+            span.__exit__(None, None, None)
+
+    return result
+
+
+async def _sentry_patched_create_async(
+    f: "Any", *args: "Any", **kwargs: "Any"
+) -> "Any":
+    """
+    Creates and manages an AI Client Span for both non-streaming and streaming calls.
```
ericapisani: This comment mirrors what we have for the `sync` version, so I'm not sure if this provides much benefit

sentry_sdk/integrations/anthropic.py
```diff
@@ -638,6 +647,81 @@ def _sentry_patched_create_common(f: "Any", *args: "Any", **kwargs: "Any") -> "A
...
+    """
+    integration = kwargs.pop("integration")
+    if integration is None:
+        return await f(*args, **kwargs)
+
+    if "messages" not in kwargs:
+        return await f(*args, **kwargs)
```
ericapisani, line 699: This can be reduced to one conditional instead of two:

```suggestion
    if integration is None or "messages" not in kwargs:
        return await f(*args, **kwargs)
```

alexander-alderman-webb: Created an issue for removing the messages condition all together since I don't know why we have the condition: https://github.com/getsentry/sentry-python/issues/5938

sentry_sdk/integrations/anthropic.py
alexander-alderman-webb: Opened an issue: https://github.com/getsentry/sentry-python/issues/5939

I am hesitant to touch this now because that's the existing logic in `_sentry_patched_create_common`.
But 100% given some time to write more exhaustive tests and read more of the `anthropic` library we should do this!

sentry_sdk/integrations/anthropic.py
alexander-alderman-webb: I would like to convey that the AI Client Span is opened **and** closed **only** in this function when the asynchronous `anthropic` API is invoked (or equivalently open and closed in the sync variant).

What would you suggest instead of the duplicate comments?

alexander-alderman-webb: I just realized that the span will only be closed in this function with the follow-up PR.
I'll update the docstring to only say that the span is opened here

https://github.com/getsentry/sentry-python/pull/5716

alexander-alderman-webb: https://github.com/getsentry/sentry-python/pull/5715/changes/e6eba3e38573c01c7f34983c50cca347d6d99a5d

sentry_sdk/integrations/anthropic.py
```diff
@@ -638,6 +647,81 @@ def _sentry_patched_create_common(f: "Any", *args: "Any", **kwargs: "Any") -> "A
 
         return result
 
+    with capture_internal_exceptions():
```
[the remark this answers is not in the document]
alexander-alderman-webb: with `capture_internal_exceptions()` the exception is logged and the control flow continues on the happy path. why would the span not be closed in this case?

sentry_sdk/integrations/anthropic.py
```diff
@@ -624,7 +627,13 @@ def _sentry_patched_create_common(f: "Any", *args: "Any", **kwargs: "Any") -> "A
...
+    try:
+        result = f(*args, **kwargs)
+    except Exception as exc:
+        exc_info = sys.exc_info()
+        with capture_internal_exceptions():
+            _capture_exception(exc)
+        reraise(*exc_info)
 
     if isinstance(result, Stream):
         result._span = span
```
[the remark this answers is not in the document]
alexander-alderman-webb: We only catch `Exception` before the change as well.

sentry_sdk/integrations/anthropic.py
```diff
@@ -689,41 +773,20 @@ def _sentry_patched_create_common(f: "Any", *args: "Any", **kwargs: "Any") -> "A
...
-                    _capture_exception(exc)
-                reraise(*exc_info)
-
-            return gen.send(result)
-        except StopIteration as e:
-            return e.value
-
     @wraps(f)
-    def _sentry_patched_create_sync(*args: "Any", **kwargs: "Any") -> "Any":
+    def _sentry_wrapped_create_sync(*args: "Any", **kwargs: "Any") -> "Any":
```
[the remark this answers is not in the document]
alexander-alderman-webb: that's okay, sync and async paths are distinct patches.

PR #5783, opened 2026-03-23
fix(scrubber): Remove ip_address instead of replacing with [Filtered]

### LINKED ISSUE #5701: Processing error due to invalid IP but an IP is shown in the event information
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. Configure the Python SDK in a Django project with eventScrubber:
```
entry_sdk.init(
    dsn="...", 
    send_default_pii=True,
    event_scrubber=EventScrubber()
)
```
2. Add user IP via set_user, it can also be added by the SDK via `send_default_pii=True,`
3. Disable the option "Prevent storing IP address" in the project settings
4. Sent a test error event

### Expected Result

The UI shows the user IP as `[Filtered]`.

### Actual Result

The UI shows an error due to invalid IP but the event has a valid IP address:
Image

The error is showing because the IP received by the server is invalid(from the event's JSON):
```
ip_address": {
"": {
"rem": [
[
"!config",
"s"
]
],
"err": [
[
"invalid_data",
{
"reason": "expected an ip address"
}
]
],
"val": "[Filtered]"
}
}
```

However, the server also infers the IP address for the event based on the HTTP request sending the event:
```
"user": {
"ip_address": "127.0.0.1",
"sentry_user": "ip:127.0.0.1"
},
```

This is what it seems to happen:
1. An IP address is added via set_user or default_PII true
2. The SDK's event scrubber removes it and replaces with [Filtered]
3. The server sees the ip address value as Filtered and mark it as an invalid field in the event
4. The server infers the IP address from the HTTP request and add it to the user info
5. The UI shows an error (step 3), but also an IP address (Step4).

### Product Area

Ingestion and Filtering

### Link

https://dev-curumas.sentry.io/issues/7343757920/events/8184ab9875ca46aeafac1a9a53e16f0d/

### DSN

_No response_

### Version

_No response_

joaquinhuigomez:
Fixes #5701

## Problem

The `EventScrubber` replaces `ip_address` with `[Filtered]`, but `[Filtered]` is not a valid IP address. The Sentry server rejects it as a protocol violation, showing a processing error in the UI even though the server infers the real IP from the HTTP request.

## Solution

Add a `remove_list` parameter to `EventScrubber` that controls which denied keys are **deleted** from the dict instead of being replaced with `[Filtered]`.

- Defaults to `DEFAULT_REMOVE_LIST = ["ip_address"]`, so the fix works out of the box.
- Users can override via `EventScrubber(remove_list=[...])`.
- Setting `remove_list=[]` restores the previous replace-with-`[Filtered]` behavior for all fields.

This follows the approach suggested by @rodolfoBee in the issue discussion: adding an option that removes the IP address from the event instead of replacing it.

## Changes

- `sentry_sdk/scrubber.py`: New `DEFAULT_REMOVE_LIST` constant and `remove_list` parameter on `EventScrubber.__init__`. Updated `scrub_dict` to delete keys in the remove list instead of substituting them.
- `tests/test_scrubber.py`: Updated existing test expectations (ip_address now absent instead of `[Filtered]`), added four new tests covering the user dict scenario, PII-enabled passthrough, custom remove list, and empty remove list fallback.

### DIFF
--- a/sentry_sdk/scrubber.py
+++ b/sentry_sdk/scrubber.py
@@ -57,6 +57,13 @@ DEFAULT_PII_DENYLIST = [
     "remote_addr",
 ]
 
+# Fields that must be removed entirely rather than replaced with [Filtered],
+# because their value must conform to a specific format (e.g. a valid IP address)
+# and [Filtered] would be a protocol violation.
+DEFAULT_REMOVE_LIST = [
+    "ip_address",
+]
+
 
 class EventScrubber:
     def __init__(
@@ -65,6 +72,7 @@ class EventScrubber:
         recursive: bool = False,
         send_default_pii: bool = False,
         pii_denylist: "Optional[List[str]]" = None,
+        remove_list: "Optional[List[str]]" = None,
     ) -> None:
         """
         A scrubber that goes through the event payload and removes sensitive data configured through denylists.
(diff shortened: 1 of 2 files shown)


stephanie-anderson:

PR #5849, opened 2026-03-24
feat(langchain): Set gen_ai.operation.name and gen_ai.pipeline.name on LLM spans

ericapisani:
Set `gen_ai.operation.name` and `gen_ai.pipeline.name` span attributes on spans created by `on_llm_start` in the LangChain integration.

`on_llm_start` handles non-chat LLM calls (e.g. completion models). It was already setting span attributes for model, system, and token usage, but was missing the operation name and pipeline name attributes that other handlers set. This adds:

- `gen_ai.operation.name = "generate_text"` to align with the span op (`gen_ai.generate_text`)
- `gen_ai.pipeline.name` from the `name` kwarg passed by LangChain, when present

Tests are updated in `test_langchain_message_truncation` and `test_langchain_response_model_extraction` to assert these attributes are set correctly.

Part of SDK-669, is the second part of changes that are being pulled out of #5705

### DIFF
--- a/tests/integrations/langchain/test_langchain.py
+++ b/tests/integrations/langchain/test_langchain.py
@@ -256,6 +256,7 @@ def test_langchain_create_agent(
     assert len(chat_spans) == 1
     assert chat_spans[0]["origin"] == "auto.ai.langchain"
 
+    assert chat_spans[0]["data"]["gen_ai.system"] == "openai-chat"
     assert chat_spans[0]["data"]["gen_ai.usage.input_tokens"] == 10
     assert chat_spans[0]["data"]["gen_ai.usage.output_tokens"] == 20
     assert chat_spans[0]["data"]["gen_ai.usage.total_tokens"] == 30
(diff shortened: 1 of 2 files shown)


tests/integrations/langchain/test_langchain.py
```diff
@@ -966,6 +1040,7 @@ def test_langchain_message_truncation(sentry_init, capture_events):
             serialized=serialized,
             prompts=prompts,
             run_id=run_id,
+            name="my_pipeline",
```
alexander-alderman-webb, line 1272: following test changes, can you pass `config={"run_name": "my_pipeline"}` to `llm.invoke()` in `test_langchain_text_completion()`?

That way we ensure that we pick up the `run_name` as the pipeline name, so the contract with the user is tested (whereas by calling the hook manually we guarantee that whatever ends up in `name` is set as the pipeline name, which is unclear 😅 )

Issue #5850, opened 2026-03-24
Multiline string support in `before_send_transaction`


ericapisani:
We had a report come in that a user is attempting to use a multi-line string within `before_send_transaction` that's copied to clipboard via the "Traces" page in Sentry UI, but events that contain that string aren't being successfully matched.

Investigate the root cause of why this isn't currently working and see how we can add support for this

PR #5855, opened 2026-03-24
ref(asyncpg): Normalize query whitespace in integration

ericapisani:
Normalize multiline SQL query whitespace in the asyncpg integration so that span descriptions contain single-line queries with collapsed whitespace.

asyncpg passes raw multiline SQL strings as span descriptions. This makes it difficult for users to match queries in `before_send_transaction` callbacks — they'd need to account for newlines and varying indentation instead of writing simple substring checks like `"SELECT id, name FROM users" in desc`.

Fixes PY-2255 and #5850


ericapisani:
bugbot review

Issue #5812, opened 2026-03-24
Bug: Streaming responses don't capture `response_id` or `model_version`


ericapisani:
## Problem

In `streaming.py`, the `accumulate_streaming_response()` function initializes `response_id` and `model` to `None` (lines 53-54) but **never populates them** from the streaming chunks. The accumulation loop (lines 56-83) never reads `chunk.response_id` or `chunk.model_version`.

This means streaming responses will always have `null` for `gen_ai.response.id` and `gen_ai.response.model`.

The non-streaming path correctly captures both in `utils.py:886-890`:

```python
if getattr(response, "response_id", None):
    span.set_data(SPANDATA.GEN_AI_RESPONSE_ID, response.response_id)
if getattr(response, "model_version", None):
    span.set_data(SPANDATA.GEN_AI_RESPONSE_MODEL, response.model_version)
```

## Files

* `sentry_sdk/integrations/google_genai/streaming.py` — `accumulate_streaming_response()`

## Acceptance Criteria

* Streaming chunks are checked for `response_id` and `model_version`
* `gen_ai.response.id` and `gen_ai.response.model` are populated on streaming spans
* Test added to verify streaming responses capture these fields

NIK-TIGER-BILL:
I tracked down the root cause of this issue.

In `accumulate_streaming_response()`, `response_id` and `model` are initialised to `None` but the accumulation loop never reads `chunk.response_id` or `chunk.model_version` from the chunks. The non-streaming path in `utils.py` already reads these fields correctly via `getattr(response, 'response_id', None)` etc.

**Proposed fix:** Inside the for-loop, read `response_id = getattr(chunk, 'response_id', None)` and `model = getattr(chunk, 'model_version', None)` (keeping the first non-None value). This is a minimal, low-risk change.

Would this approach be acceptable? I can open a PR with a test that asserts the fields are populated on streaming spans.

Bahtya:
I'd like to work on this. The fix is straightforward — extract `response_id` and `model_version` from streaming chunks in the accumulation loop, using `getattr` to safely handle missing attributes. I've identified the exact location in `sentry_sdk/integrations/google_genai/streaming.py`. Happy to submit a PR if this approach looks correct.

zkasuran:
I hit this on 2.67.0 and verified it end to end. Would you like me to pick it up? A patch is
ready if so.

Confirmation that both attributes are lost on every streaming call, not just some. I ran the
[...]

PR #5858, opened 2026-03-25
fix(google_genai): populate response_id and model_version in streaming accumulator

NIK-TIGER-BILL:
### Description

`accumulate_streaming_response()` in `streaming.py` initialises `response_id` and `model` to `None` (lines 53-54) but **never populates them** from the streaming chunks. As a result, streaming spans always have `null` for `gen_ai.response.id` and `gen_ai.response.model`, while the non-streaming path in `utils.py:888-892` correctly captures both fields.

**Fix:** after processing tool-calls for each chunk, extract `response_id` and `model_version` from the first chunk that carries them — the same pattern the non-streaming path already uses.

**Changed files:**
- `sentry_sdk/integrations/google_genai/streaming.py` — extract `response_id` / `model_version` in accumulation loop
- `tests/integrations/google_genai/test_google_genai.py` — add assertions to the existing streaming test that verify both fields are now recorded on the span

#### Issues
* resolves: #5812

#### Reminders
- [x] Tests updated (`test_streaming_generate_content` now asserts `GEN_AI_RESPONSE_ID` and `GEN_AI_RESPONSE_MODEL`)
- [x] No new lint issues
- [x] PR title uses conventional commit style

### DIFF
--- a/sentry_sdk/integrations/google_genai/streaming.py
+++ b/sentry_sdk/integrations/google_genai/streaming.py
@@ -72,6 +72,12 @@ def accumulate_streaming_response(
         if extracted_tool_calls:
             tool_calls.extend(extracted_tool_calls)
 
+        # Capture response_id and model_version from the first chunk that has them
+        if response_id is None:
+            response_id = getattr(chunk, "response_id", None)
+        if model is None:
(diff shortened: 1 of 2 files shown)


stephanie-anderson:

PR #5859, opened 2026-03-25
fix(anthropic): stop setting transaction status on AI span errors

NIK-TIGER-BILL:
### Description

AI integrations should not interfere with HTTP transactions.

The previous `_capture_exception()` called `set_span_errored()`, which **not only** marked the current AI span as `INTERNAL_ERROR` but **also** propagated that status to the **containing HTTP transaction** via:

```python
span.containing_transaction.set_status(SPANSTATUS.INTERNAL_ERROR)
```

This is incorrect: the HTTP layer owns the transaction status, and an Anthropic API error is an application-level event that should only affect the AI-specific span.

**Fix:** Replace `set_span_errored()` with a direct `span.set_status()` call that does **not** touch the containing transaction.

**Affected files:**
- `sentry_sdk/integrations/anthropic.py` — removed `set_span_errored()`, now only marks the current span
- `tests/integrations/anthropic/test_anthropic.py` — updated 4 error tests to assert the transaction trace status is NOT set to `internal_error` by the integration

#### Issues
* resolves: #5790

#### Reminders
- [x] Tests updated (4 assertions changed from `== internal_error` to `!= internal_error`)
- [x] No new lint issues
- [x] PR title uses conventional commit style

### DIFF
--- a/sentry_sdk/integrations/anthropic.py
+++ b/sentry_sdk/integrations/anthropic.py
@@ -17,7 +17,7 @@ from sentry_sdk.ai.utils import (
 from sentry_sdk.consts import OP, SPANDATA, SPANSTATUS
 from sentry_sdk.integrations import _check_minimum_version, DidNotEnable, Integration
 from sentry_sdk.scope import should_send_default_pii
-from sentry_sdk.tracing_utils import set_span_errored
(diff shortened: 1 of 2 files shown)


stephanie-anderson:

PR #5872, opened 2026-03-25
fix: handle None content.parts in google_genai

### LINKED ISSUE #5854: TypeError: 'NoneType' object is not iterable in google_genai integration
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.50.0

### Steps to Reproduce

1. Make a Gemini call where a candidate's content.parts is explicitly `None` (attribute present but set to None)
2. Sentry's tracing fires and calls `_extract_response_text`
3. hasattr check passes (attribute exists), then iterating over `None` crashes

### Expected Result

No exception is raised. Sentry skips recording the response text span attribute when content.parts is None.

### Actual Result

`TypeError: 'NoneType' object is not iterable`
Traceback points to `_extract_response_text` in `sentry_sdk/integrations/google_genai/utils.py`.

thakoreh:
Fixes #5854

Gemini sometimes returns candidates where content.parts exists but
is None. hasattr lets this through and we crash on iteration.

Switched to a truthiness check and added a regression test.

### DIFF
--- a/sentry_sdk/integrations/google_genai/utils.py
+++ b/sentry_sdk/integrations/google_genai/utils.py
@@ -693,7 +693,7 @@ def _extract_response_text(
 
     texts = []
     for candidate in response.candidates:
-        if not hasattr(candidate, "content") or not hasattr(candidate.content, "parts"):
+        if not hasattr(candidate, "content") or not candidate.content.parts:
             continue
 
         for part in candidate.content.parts:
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb:
Resolved by https://github.com/getsentry/sentry-python/commit/bcfb7886e9d63a9ff7bcb4b90831608ac5a6ecec

PR #5880, opened 2026-03-26
fix: Add cycle detection to exceptions_from_error

ericapisani:
## Summary
- Add cycle detection to `exceptions_from_error()` to prevent infinite recursion when exception chains contain cycles
- This fixes a production issue where FastAPI + multiple `BaseHTTPMiddleware` instances cause Starlette's `collapse_excgroups()` to create cyclic `__context__` references (ExceptionGroup → inner exception → `__context__` → ExceptionGroup)
- Without this fix, the SDK silently drops the event due to a `RecursionError`

## Test plan
- [x] `test_exceptiongroup_starlette_collapse` — realistic reproduction of the Starlette `collapse_excgroups()` pattern with exact expected exception values
- [x] `test_cyclic_exception_group_cause` — verifies cycle detection works for `__cause__` chains
- [x] `test_deeply_nested_cyclic_exception_group` — verifies cycle detection across multiple nested ExceptionGroups

Fixes https://github.com/getsentry/sentry-python/issues/5025 and PY-1949

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -828,6 +828,22 @@ def exceptions_from_error(
 
     See the Exception Interface documentation for more details:
     https://develop.sentry.dev/sdk/event-payloads/exception/
+
+    Args:
+        exception_id (int):
+
+            Sequential counter for assigning ``mechanism.exception_id``
+            to each processed exception. Is NOT the result of calling `id()` on the exception itself.
+
+        parent_id (int):
+
+            The ``mechanism.exception_id`` of the parent exception.
+
+            Written into ``mechanism.parent_id`` in the event payload so Sentry can
+            reconstruct the exception tree.
+
+            Not to be confused with ``seen_exception_ids``, which tracks Python ``id()``
+            values for cycle detection.
     """
 
     if seen_exception_ids is None:


ericapisani:
bugbot run

sentry_sdk/utils.py
```diff
@@ -828,6 +830,19 @@ def exceptions_from_error(
...
+        seen_exception_ids = set()
+
+    if seen_exceptions is None:
+        seen_exceptions = []
+
+    if exc_value is not None and id(exc_value) in seen_exception_ids:
+        return (exception_id, [])
+
+    if exc_value is not None:
+        seen_exceptions.append(exc_value)
```
sl0thentr0py, line 859: we don't seem to use `seen_exceptions` directly, do you know why it's there (also in the other code from before)?

my hunch would be it's only to keep the memory reference alive otherwise GC could reuse the id.

sentry_sdk/utils.py
```diff
@@ -828,6 +830,19 @@ def exceptions_from_error(
     https://develop.sentry.dev/sdk/event-payloads/exception/
     """
 
+    if seen_exception_ids is None:
+        seen_exception_ids = set()
+
+    if seen_exceptions is None:
+        seen_exceptions = []
+
+    if exc_value is not None and id(exc_value) in seen_exception_ids:
```
sl0thentr0py, line 855: is `exception_id` here the same as `id(exc_value)`? If so, please use the one already present.

sl0thentr0py: nvm that's a counter, but then the naming is kinda confusing, could you add a small docstring regarding `exception_id` and `parent_id` that they are counters and not the actual object ids of the exception.

sentry_sdk/utils.py
ericapisani: > my hunch would be it's only to keep the memory reference alive otherwise GC could reuse the id.

You've got it - it ensures that exceptions don't get garbage collected mid-traversal of the exception chain

sentry_sdk/utils.py
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/5880/commits/d516f1a8f40ab0c24c1f62e0a70bb19369398710

PR #5899, opened 2026-03-27
fix(tests): Fix flaky continuous profiler test under gevent

stephanie-anderson:
Fix flaky `test_continuous_profiler_manual_start_and_stop_sampled` test
that intermittently fails under the gevent scheduler on slow CI runners.

The test asserts that `profile_chunk` envelope items exist immediately
after a transaction completes, but the profiler thread may not have
flushed its buffer yet — especially under gevent's cooperative scheduling
on resource-constrained CI. The fix adds polling with a timeout to
`assert_single_transaction_with_profile_chunks` so it waits up to 2
seconds for chunks to arrive before asserting, rather than assuming
they're already present.

Example failure: https://github.com/getsentry/sentry-python/actions/runs/23643367140/job/68869340875


alexander-alderman-webb:
Closing in favor of https://github.com/getsentry/sentry-python/pull/6174 (simpler to just assert after we know flushing is done)

PR #5920, opened 2026-03-30
feat(asgi): Make integration fully span first compatible

sentrivana:
### Description
In span first, there are no event processors. Therefore, we need to be able to set the data we were setting in event processors differently.

As we migrate our integrations one by one, this will be an exercise in whether it's possible to achieve this without some sort of callback/lifecycle hooks. So far, in ASGI, it seems we can get by by simply using `scope.set_attribute()` for setting request related data, and updating the segment name/source just before the span ends.

Adding this enables us to actually test the new functionality.

To sum up, this PR:
- makes the ASGI integration fully span-first-compatible: data set on spans currently should appear as attributes in span first
- introduces a test helper that makes it easier to unpack envelope items (useful especially when a test checks both old- and new-style telemetry)
- fully tests the ASGI span streaming implementation

Like in other span first PRs, there is quite a bit of intentional code duplication so that it's easier to remove the legacy implementation in the next major.

#### Issues
* Closes https://github.com/getsentry/sentry-python/issues/5722
* Closes https://linear.app/getsentry/issue/PY-2152/event-processor-replacement-for-streamed-spans

#### Reminders

### DIFF
--- a/sentry_sdk/_span_batcher.py
+++ b/sentry_sdk/_span_batcher.py
@@ -91,13 +91,21 @@ class SpanBatcher(Batcher["StreamedSpan"]):
     @staticmethod
     def _estimate_size(item: "StreamedSpan") -> int:
         # Rough estimate of serialized span size that's quick to compute.
-        # 210 is the rough size of the payload without attributes, and we
-        # estimate additional 70 bytes on top of that per attribute.
-        return 210 + 70 * len(item._attributes)
+        # 210 is the rough size of the payload without attributes, and then we
+        # estimate the attributes separately.
+        estimate = 210
+        for value in item._attributes.values():
+            estimate += 50
+
+            if isinstance(value, str):
+                estimate += len(value)
+            else:
+                estimate += len(str(value))
+
+        return estimate
 
     @staticmethod
     def _to_transport_format(item: "StreamedSpan") -> "Any":
-        # TODO[span-first]
         res: "dict[str, Any]" = {
             "trace_id": item.trace_id,
             "span_id": item.span_id,
@@ -126,7 +134,7 @@ class SpanBatcher(Batcher["StreamedSpan"]):
                 return
 
             envelopes = []
-            for trace_id, spans in self._span_buffer.items():
+            for spans in self._span_buffer.values():
                 if spans:
                     dsc = spans[0]._dynamic_sampling_context()
 
--- a/sentry_sdk/integrations/huggingface_hub.py
+++ b/sentry_sdk/integrations/huggingface_hub.py
@@ -1,6 +1,7 @@
-import sys
 import inspect
+import sys
 from functools import wraps
+from typing import TYPE_CHECKING
 
 import sentry_sdk
 from sentry_sdk.ai.monitoring import record_token_usage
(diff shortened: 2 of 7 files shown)


sentry_sdk/integrations/_asgi_common.py
```diff
@@ -105,3 +106,31 @@ def _get_request_data(asgi_scope: "Any") -> "Dict[str, Any]":
         request_data["env"] = {"REMOTE_ADDR": _get_ip(asgi_scope)}
 
     return request_data
+
+
+def _get_request_attributes(asgi_scope: "Any") -> "dict[str, Any]":
```
sentrivana, line 110: This is an attributes based copy of `_get_request_data` just above

sentry_sdk/integrations/asgi.py
```diff
@@ -424,3 +443,40 @@ def _get_transaction_name_and_source(
             return name, source
 
         return name, source
+
+    def _get_segment_name_and_source(
```
sentrivana, line 463: This is a copy of `_get_transaction_name_and_source` above, just adapted for segments

sentry_sdk/integrations/_asgi_common.py
```diff
@@ -105,3 +105,33 @@ def _get_request_data(asgi_scope: "Any") -> "Dict[str, Any]":
...
+    """
+    Return attributes related to the HTTP request from the ASGI scope.
+    """
+    attributes: "dict[str, Any]" = {}
+
+    ty = asgi_scope["type"]
+    if ty in ("http", "websocket"):
```
alexander-alderman-webb: Why do we only set these attributes if it's HTTP or WebSocket?
Maybe I misunderstand the transaction logic but I thought we set the attributes regardless, and the connection type only determines if we continue the trace.

sentry_sdk/integrations/asgi.py
```diff
@@ -301,6 +307,9 @@ async def _run_app(
                             else nullcontext()
                         )
 
+                    for attribute, value in _get_request_attributes(scope).items():
+                        sentry_scope.set_attribute(attribute, value)
```
alexander-alderman-webb: should this run only if span-streaming? If it runs for both paths we could remove the event processor?

sentrivana: We need to keep the event processor since it still enriches other kinds of events (like errors) -- those also won't get any attributes from the scope since they don't support attributes.

But I just talked about this with folks in the span first sync and we don't actually want to set the attrs on the scope since most (all?) of these attrs should just be set on the segment, and this way they'd be set on all spans (and metrics, and logs). So I'll fix that.

alexander-alderman-webb: Cool, I forgot for a moment that the `scope.set_attribute()` doesn't affect old spans.
If we don't want the attributes on logs or metrics that sounds good.

sentrivana: Here's the change: https://github.com/getsentry/sentry-python/pull/5920/changes/2202e0cfadd13f3f6de5976148496ee60aa9bc89

PR #6074, opened 2026-04-14
feat(celery): Support span streaming

### LINKED ISSUE #6009: Migrate celery to span first

sentrivana:
### Description
Support streaming mode in the Celery integration.

This is the first integration that contains explicit guards against creating certain spans as segments, mirroring how `start_span` works currently -- it won't create a span if there's no transaction. In Celery, there were multiple spots that, if we just replaced all legacy `start_span`s with the new `traces.start_span`, would've created new segment spans, leading to a change in behavior.

- [x] replace APIs
- [x] map data to attributes from conventions
- [x] run tests in streaming mode

Best viewed with `Hide whitespace` 😎  

#### Issues
* Closes https://github.com/getsentry/sentry-python/issues/6009
* Closes PY-2311


sentry_sdk/integrations/celery/__init__.py
```diff
@@ -487,31 +553,52 @@ def sentry_publish(self: "Producer", *args: "Any", **kwargs: "Any") -> "Any":
             # method will still work.
             kwargs_headers = {}
 
-        task_name = kwargs_headers.get("task")
+        task_name = kwargs_headers.get("task") or "<unknown Celery task>"
```
sentrivana, line 554: This is used for naming the span and spans v2 must always have a name, `name=None` is not supported.

sentry_sdk/_span_batcher.py
```diff
@@ -109,7 +109,7 @@ def _to_transport_format(item: "StreamedSpan") -> "Any":
         res: "dict[str, Any]" = {
             "trace_id": item.trace_id,
             "span_id": item.span_id,
-            "name": item._name,
+            "name": item._name if item._name is not None else "<unknown>",
```
sentrivana, line 112: Guarding against sending null span names also here, directly in the batcher, since it'd be a protocol violation

With the fix for `task_name` above, there are no known cases where this would happen, but better to be safe

sentry_sdk/traces.py
```diff
@@ -258,7 +258,9 @@ def __init__(
     ):
         self._name: str = name
         self._active: bool = active
-        self._attributes: "Attributes" = {}
+        self._attributes: "Attributes" = {
+            "sentry.origin": "manual",
```
sentrivana, line 262: This was an oversight in the core logic -- we were never setting `sentry.origin` for custom instrumentation

sentry_sdk/scope.py
```diff
@@ -893,6 +893,19 @@ def span(self, span: "Optional[Union[Span, StreamedSpan]]") -> None:
...
+        # is used for populating events and linking them to segments
+        if (
+            isinstance(span, StreamedSpan)
+            and not isinstance(span, NoOpStreamedSpan)
+            and span._is_segment()
+        ):
+            self._transaction = span.name
+            if span._attributes.get("sentry.span.source"):
+                self._transaction_info["source"] = str(
+                    span._attributes["sentry.span.source"]
```
sentrivana, line 906: If you're wondering what the `str()` is doing here, the answer is mypy 😎 😭

sentry_sdk/_span_batcher.py
alexander-alderman-webb: We currently fallback to `<unlabeled_transaction>` for transactions, maybe we want a fallback value that's more similar?

https://github.com/getsentry/sentry-python/blob/5d26c2f60e289c89ab260ee25607a3bea5e174c9/sentry_sdk/tracing.py#L1007

```suggestion
            "name": item._name if item._name is not None else "<unlabeled span>",
```

alexander-alderman-webb (approved):
Nice! Just one optional suggestion.

PR #6084, opened 2026-04-16
feat(httpx): Migrate to span first

### LINKED ISSUE #6028: Migrate httpx to span first

ericapisani:
Migrates the httpx integration to the span-first (streaming span) architecture.

When `_experiments={"trace_lifecycle": "stream"}` is enabled, the integration
now uses `StreamedSpan` via `sentry_sdk.traces.start_span` instead of the legacy
`Span`-based (transactions) path. The legacy path remains unchanged for backwards compatibility.

Supporting changes:
- `add_http_request_source_for_streamed_span()` added to `tracing_utils` — a
  `StreamedSpan`-compatible variant that must be called inside the span context
  manager so source data is attached before the span is flushed
- `add_source()` updated to dispatch between `set_data`/`set_attribute` depending
  on span type
- `start_timestamp_monotonic_ns` property added to `StreamedSpan` and
  `NoOpStreamedSpan` to support the request source duration threshold check

Part of the broader span-first integration migration.

Fixes PY-2330
Fixes #6028

### DIFF
--- a/sentry_sdk/integrations/httpx.py
+++ b/sentry_sdk/integrations/httpx.py
@@ -95,12 +95,13 @@ def _install_httpx_client() -> None:
                         else:
                             request.headers[key] = value
 
-                rv = real_send(self, request, **kwargs)
+                try:
+                    rv = real_send(self, request, **kwargs)
 
-                streamed_span.status = "error" if rv.status_code >= 400 else "ok"
-                attributes["http.response.status_code"] = rv.status_code
-
-                streamed_span.set_attributes(attributes)
+                    streamed_span.status = "error" if rv.status_code >= 400 else "ok"
+                    attributes["http.response.status_code"] = rv.status_code
+                finally:
+                    streamed_span.set_attributes(attributes)
 
                 # Needs to happen within the context manager as we want to attach the
                 # final data before the span finishes and is sent for ingesting.
--- a/sentry_sdk/integrations/httpx.py
+++ b/sentry_sdk/integrations/httpx.py
@@ -1,13 +1,10 @@
 import sentry_sdk
-from sentry_sdk import start_span as legacy_start_span
 from sentry_sdk.consts import OP, SPANDATA
 from sentry_sdk.integrations import Integration, DidNotEnable
-from sentry_sdk.traces import start_span
 from sentry_sdk.tracing import BAGGAGE_HEADER_NAME
 from sentry_sdk.tracing_utils import (
-    add_http_request_source_for_streamed_span,
-    should_propagate_trace,
     add_http_request_source,
+    should_propagate_trace,
     add_sentry_baggage_to_headers,
     has_span_streaming_enabled,
 )
@@ -61,7 +58,7 @@ def _install_httpx_client() -> None:
             parsed_url = parse_url(str(request.url), sanitize=False)
 
         if is_span_streaming_enabled:
-            with start_span(
+            with sentry_sdk.traces.start_span(
                 name="%s %s"
                 % (
                     request.method,
@@ -70,15 +67,15 @@ def _install_httpx_client() -> None:
                 attributes={
                     "sentry.op": OP.HTTP_CLIENT,
                     "sentry.origin": HttpxIntegration.origin,
-                    SPANDATA.HTTP_METHOD: request.method,
+                    "http.request.method": request.method,
                 },
-            ) as segment:
+            ) as streamed_span:
                 attributes: "Attributes" = {}
 
                 if parsed_url is not None:
-                    attributes["url"] = parsed_url.url
-                    attributes[SPANDATA.HTTP_QUERY] = parsed_url.query
-                    attributes[SPANDATA.HTTP_FRAGMENT] = parsed_url.fragment
+                    attributes["url.full"] = parsed_url.url
+                    attributes["url.query"] = parsed_url.query
+                    attributes["url.fragment"] = parsed_url.fragment
 
                 if should_propagate_trace(client, str(request.url)):
                     for (
@@ -98,17 +95,17 @@ def _install_httpx_client() -> None:
 
                 rv = real_send(self, request, **kwargs)
 
-                segment.status = "error" if rv.status_code >= 400 else "ok"
-                attributes[SPANDATA.HTTP_STATUS_CODE] = rv.status_code
+                streamed_span.status = "error" if rv.status_code >= 400 else "ok"
+                attributes["http.response.status_code"] = rv.status_code
 
-                segment.set_attributes(attributes)
+                streamed_span.set_attributes(attributes)
 
                 # Needs to happen within the context manager as we want to attach the
                 # final data before the span finishes and is sent for ingesting.
                 with capture_internal_exceptions():
-                    add_http_request_source_for_streamed_span(segment)
+                    add_http_request_source(streamed_span)
         else:
-            with legacy_start_span(
+            with sentry_sdk.start_span(
                 op=OP.HTTP_CLIENT,
                 name="%s %s"
                 % (
@@ -168,7 +165,7 @@ def _install_httpx_async_client() -> None:
             parsed_url = parse_url(str(request.url), sanitize=False)
 
         if is_span_streaming_enabled:
-            with start_span(
+            with sentry_sdk.traces.start_span(
                 name="%s %s"
                 % (
                     request.method,
@@ -177,15 +174,17 @@ def _install_httpx_async_client() -> None:
                 attributes={
                     "sentry.op": OP.HTTP_CLIENT,
                     "sentry.origin": HttpxIntegration.origin,
-                    SPANDATA.HTTP_METHOD: request.method,
+                    "http.request.method": request.method,
                 },
-            ) as segment:
+            ) as streamed_span:
                 attributes: "Attributes" = {}
 
                 if parsed_url is not None:
-                    attributes["url"] = parsed_url.url
-                    attributes[SPANDATA.HTTP_QUERY] = parsed_url.query
-                    attributes[SPANDATA.HTTP_FRAGMENT] = parsed_url.fragment
+                    attributes["url.full"] = parsed_url.url
+                    if parsed_url.query:
+                        attributes["url.query"] = parsed_url.query
+                    if parsed_url.fragment:
+                        attributes["url.fragment"] = parsed_url.fragment
 
                 if should_propagate_trace(client, str(request.url)):
                     for (
@@ -205,17 +204,17 @@ def _install_httpx_async_client() -> None:
 
                 rv = await real_send(self, request, **kwargs)
 
-                segment.status = "error" if rv.status_code >= 400 else "ok"
-                attributes[SPANDATA.HTTP_STATUS_CODE] = rv.status_code
+                streamed_span.status = "error" if rv.status_code >= 400 else "ok"
+                attributes["http.response.status_code"] = rv.status_code
 
-                segment.set_attributes(attributes)
+                streamed_span.set_attributes(attributes)
 
                 # Needs to happen within the context manager as we want to attach the
                 # final data before the span finishes and is sent for ingesting.
                 with capture_internal_exceptions():
-                    add_http_request_source_for_streamed_span(segment)
+                    add_http_request_source(streamed_span)
         else:
-            with legacy_start_span(
+            with sentry_sdk.start_span(
                 op=OP.HTTP_CLIENT,
                 name="%s %s"
                 % (
--- a/sentry_sdk/integrations/httpx.py
+++ b/sentry_sdk/integrations/httpx.py
@@ -74,8 +74,10 @@ def _install_httpx_client() -> None:
 
                 if parsed_url is not None:
                     attributes["url.full"] = parsed_url.url
-                    attributes["url.query"] = parsed_url.query
-                    attributes["url.fragment"] = parsed_url.fragment
+                    if parsed_url.query:
+                        attributes["url.query"] = parsed_url.query
+                    if parsed_url.fragment:
+                        attributes["url.fragment"] = parsed_url.fragment
 
                 if should_propagate_trace(client, str(request.url)):
                     for (
--- a/sentry_sdk/integrations/httpx.py
+++ b/sentry_sdk/integrations/httpx.py
@@ -205,12 +205,13 @@ def _install_httpx_async_client() -> None:
                         else:
                             request.headers[key] = value
 
-                rv = await real_send(self, request, **kwargs)
+                try:
+                    rv = await real_send(self, request, **kwargs)
 
-                streamed_span.status = "error" if rv.status_code >= 400 else "ok"
-                attributes["http.response.status_code"] = rv.status_code
-
-                streamed_span.set_attributes(attributes)
+                    streamed_span.status = "error" if rv.status_code >= 400 else "ok"
+                    attributes["http.response.status_code"] = rv.status_code
+                finally:
+                    streamed_span.set_attributes(attributes)
 
                 # Needs to happen within the context manager as we want to attach the
                 # final data before the span finishes and is sent for ingesting.
--- a/sentry_sdk/integrations/httpx.py
+++ b/sentry_sdk/integrations/httpx.py
@@ -1,12 +1,15 @@
 import sentry_sdk
-from sentry_sdk import start_span
+from sentry_sdk import start_span as legacy_start_span
 from sentry_sdk.consts import OP, SPANDATA
 from sentry_sdk.integrations import Integration, DidNotEnable
+from sentry_sdk.traces import start_span
 from sentry_sdk.tracing import BAGGAGE_HEADER_NAME
 from sentry_sdk.tracing_utils import (
+    add_http_request_source_for_streamed_span,
     should_propagate_trace,
     add_http_request_source,
     add_sentry_baggage_to_headers,
+    has_span_streaming_enabled,
 )
 from sentry_sdk.utils import (
     SENSITIVE_DATA_SUBSTITUTE,
@@ -20,6 +23,7 @@ from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from typing import Any
+    from sentry_sdk._types import Attributes
 
 
 try:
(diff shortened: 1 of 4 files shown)


ericapisani:
bugbot run

sentry_sdk/traces.py
```diff
@@ -467,6 +467,10 @@ def sampled(self) -> "Optional[bool]":
     def start_timestamp(self) -> "Optional[datetime]":
         return self._start_timestamp
 
+    @property
+    def start_timestamp_monotonic_ns(self) -> "Optional[int]":
+        return self._start_timestamp_monotonic_ns
```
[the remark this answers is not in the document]
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6084/commits/6815bc621e36ceba1e468df0d395dc9444921067

sentry_sdk/integrations/httpx.py
```diff
@@ -49,48 +53,99 @@ def _install_httpx_client() -> None:
...
+                rv = real_send(self, request, **kwargs)
 
-            span.set_http_status(rv.status_code)
-            span.set_data("reason", rv.reason_phrase)
+                segment.status = "error" if rv.status_code >= 400 else "ok"
+                attributes[SPANDATA.HTTP_STATUS_CODE] = rv.status_code
 
-        with capture_internal_exceptions():
-            add_http_request_source(span)
+                segment.set_attributes(attributes)
```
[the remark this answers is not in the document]
ericapisani: We're dropping this in span-first, this is intentional

sentry_sdk/traces.py
```diff
@@ -467,6 +464,10 @@ def sampled(self) -> "Optional[bool]":
...
```
sentrivana, line 469: Let's not expose new properties on the span -- the idea is to keep the API surface minimal. We can still use the attribute directly in our code if we need to, but I think we don't (see other comments).

sentry_sdk/traces.py
```diff
@@ -278,12 +278,9 @@ def __init__(
...
 
-        try:
-            # profiling depends on this value and requires that
-            # it is measured in nanoseconds
-            self._start_timestamp_monotonic_ns = nanosecond_time()
-        except AttributeError:
-            pass
+        # profiling depends on this value and requires that
+        # it is measured in nanoseconds
+        self._start_timestamp_monotonic_ns = nanosecond_time()
```
sentrivana, line 283: For context, this code including the `try..except` was ported as-is from the existing tracing implementation. I also don't know why we're guarding against `AttributeError`s here (and later in `_end`), but it seems deliberate (this was before LLMs 😆). I wouldn't change it, or at least not in this PR unless we're 100% sure the exception handling is not necessary.

sentry_sdk/integrations/httpx.py
```diff
@@ -103,50 +158,100 @@ def _install_httpx_async_client() -> None:
...
+            with start_span(
+                name="%s %s"
+                % (
+                    request.method,
+                    parsed_url.url if parsed_url else SENSITIVE_DATA_SUBSTITUTE,
+                ),
+                attributes={
+                    "sentry.op": OP.HTTP_CLIENT,
+                    "sentry.origin": HttpxIntegration.origin,
+                    SPANDATA.HTTP_METHOD: request.method,
```
sentrivana: `SPANDATA.HTTP_METHOD` is `http.method`, which is [deprecated](https://getsentry.github.io/sentry-conventions/attributes/http/#http-method) and superseded by `http.request.method`.

Unfortunately a lot of the `SPANDATA` consts are outdated and need to be double checked and potentially replaced in streaming mode 

```suggestion
                    "http.request.method": request.method,
```

sentry_sdk/integrations/httpx.py
```diff
@@ -103,50 +158,100 @@ def _install_httpx_async_client() -> None:
...
+                },
+            ) as segment:
+                attributes: "Attributes" = {}
+
+                if parsed_url is not None:
+                    attributes["url"] = parsed_url.url
```
sentrivana: `url` is [deprecated](https://getsentry.github.io/sentry-conventions/attributes/general/#url), this should be `url.full`

sentry_sdk/integrations/httpx.py
```diff
@@ -103,50 +158,100 @@ def _install_httpx_async_client() -> None:
...
+                    attributes[SPANDATA.HTTP_QUERY] = parsed_url.query
```
sentrivana: There's `http.query` and there's also `url.query`, one contains the starting question mark while the other [doesn't](https://getsentry.github.io/sentry-conventions/attributes/url/#url-query). I haven't checked which applies to `parsed_url.query`, but I believe it's the latter.

sentry_sdk/integrations/httpx.py
```diff
@@ -103,50 +158,100 @@ def _install_httpx_async_client() -> None:
...
+                    attributes[SPANDATA.HTTP_FRAGMENT] = parsed_url.fragment
```
sentrivana: Same case as https://github.com/getsentry/sentry-python/pull/6084/changes#r3098336387 -- there's `http.fragment` and `url.fragment`, one [contains](https://getsentry.github.io/sentry-conventions/attributes/http/#http-fragment) the leading `#` while the other one doesn't, we might want `url.fragment` here.

sentry_sdk/tracing_utils.py
```diff
@@ -273,14 +275,20 @@ def add_source(
         except Exception:
             lineno = None
         if lineno is not None:
-            span.set_data(SPANDATA.CODE_LINENO, frame.f_lineno)
+            if isinstance(span, LegacySpan):
+                span.set_data(SPANDATA.CODE_LINENO, lineno)
+            else:
+                span.set_attribute(SPANDATA.CODE_LINENO, lineno)
```
sentrivana: [Deprecated](https://getsentry.github.io/sentry-conventions/attributes/code/#code-lineno) in favor of `code.line.number`
```suggestion
                span.set_attribute("code.line.number", lineno)
```

sentry_sdk/tracing_utils.py
```diff
@@ -273,14 +275,20 @@ def add_source(
...
         try:
             namespace = frame.f_globals.get("__name__")
         except Exception:
             namespace = None
         if namespace is not None:
-            span.set_data(SPANDATA.CODE_NAMESPACE, namespace)
+            if isinstance(span, LegacySpan):
+                span.set_data(SPANDATA.CODE_NAMESPACE, namespace)
+            else:
+                span.set_attribute(SPANDATA.CODE_NAMESPACE, namespace)
```
sentrivana: https://getsentry.github.io/sentry-conventions/attributes/code/#code-namespace

```suggestion
                span.set_attribute("code.function.name", namespace)
```

sentry_sdk/integrations/httpx.py
```diff
@@ -49,48 +53,99 @@ def _install_httpx_client() -> None:
 
     @ensure_integration_enabled(HttpxIntegration, real_send)
     def send(self: "Client", request: "Request", **kwargs: "Any") -> "Response":
```
sentrivana, line 52: Please see my comments on the async send since this is essentially the same changeset

sentry_sdk/integrations/httpx.py
```diff
@@ -103,50 +158,100 @@ def _install_httpx_async_client() -> None:
...
+                            request.headers[key] = value
 
-            span.set_http_status(rv.status_code)
-            span.set_data("reason", rv.reason_phrase)
+                rv = await real_send(self, request, **kwargs)
 
-        with capture_internal_exceptions():
-            add_http_request_source(span)
+                segment.status = "error" if rv.status_code >= 400 else "ok"
+                attributes[SPANDATA.HTTP_STATUS_CODE] = rv.status_code
```
sentrivana: https://getsentry.github.io/sentry-conventions/attributes/http/#http-status_code
```suggestion
                attributes["http.response.status_code"] = rv.status_code
```

sentry_sdk/tracing_utils.py
```diff
@@ -290,15 +298,23 @@ def add_source(
                 in_app_path = filepath.replace(project_root, "").lstrip(os.sep)
             else:
                 in_app_path = filepath
-            span.set_data(SPANDATA.CODE_FILEPATH, in_app_path)
+
+            if isinstance(span, LegacySpan):
+                span.set_data(SPANDATA.CODE_FILEPATH, in_app_path)
+            else:
+                if in_app_path is not None:
+                    span.set_attribute(SPANDATA.CODE_FILEPATH, in_app_path)
```
sentrivana: https://getsentry.github.io/sentry-conventions/attributes/code/#code-filepath
```suggestion
                    span.set_attribute("code.file.path", in_app_path)
```

sentry_sdk/tracing_utils.py
```diff
@@ -290,15 +298,23 @@ def add_source(
...
             code_function = frame.f_code.co_name
         except Exception:
             code_function = None
 
         if code_function is not None:
-            span.set_data(SPANDATA.CODE_FUNCTION, frame.f_code.co_name)
+            if isinstance(span, LegacySpan):
+                span.set_data(SPANDATA.CODE_FUNCTION, frame.f_code.co_name)
+            else:
+                span.set_attribute(SPANDATA.CODE_FUNCTION, frame.f_code.co_name)
```
sentrivana: https://getsentry.github.io/sentry-conventions/attributes/code/#code-function-name

```suggestion
                span.set_attribute("code.function.name", frame.f_code.co_name)
```

sentry_sdk/integrations/httpx.py
```diff
@@ -103,50 +158,100 @@ def _install_httpx_async_client() -> None:
...
+                % (
+                    request.method,
+                    parsed_url.url if parsed_url else SENSITIVE_DATA_SUBSTITUTE,
+                ),
+                attributes={
+                    "sentry.op": OP.HTTP_CLIENT,
+                    "sentry.origin": HttpxIntegration.origin,
+                    SPANDATA.HTTP_METHOD: request.method,
+                },
+            ) as segment:
```
sentrivana: Let's call this `span` instead of `segment` -- HTTP client spans are not segments so it might lead to confusion

sentry_sdk/tracing_utils.py
```diff
@@ -331,6 +347,47 @@ def add_query_source(span: "sentry_sdk.tracing.Span") -> None:
...
+
+    should_add_request_source = client.options.get("enable_http_request_source", True)
+    if not should_add_request_source:
+        return
+
+    if span.start_timestamp_monotonic_ns is not None:
+        elapsed = nanosecond_time() - span.start_timestamp_monotonic_ns
+        end_timestamp = span.start_timestamp + timedelta(microseconds=elapsed / 1000)
+    else:
+        end_timestamp = datetime.now(timezone.utc)
```
sentrivana: Why work with the monotonic start timestamp here at all? The original `add_http_request_source` just uses `start_timestamp`, so I believe that should be good enough?
```suggestion
    end_timestamp = datetime.now(timezone.utc)
```

sentrivana: If we make the change I proposed above, then as far as I can see the only difference to the original `add_http_request_source` is just the source of `end_timestamp` and then I believe we don't need a separate function for the streaming case at all. We just need something like this in the original function:

```python
if span.timestamp is not None:
    end_timestamp = span.timestamp
else:
    end_timestamp = datetime.now(timezone.utc)
```

sentry_sdk/integrations/httpx.py
```diff
@@ -49,48 +53,99 @@ def _install_httpx_client() -> None:
...
-        with capture_internal_exceptions():
-            add_http_request_source(span)
+                segment.set_attributes(attributes)
+
+                # Needs to happen within the context manager as we want to attach the
+                # final data before the span finishes and is sent for ingesting.
+                with capture_internal_exceptions():
+                    add_http_request_source_for_streamed_span(segment)
+        else:
+            with legacy_start_span(
```
sentrivana: Can we use the full name for the `start_span` functions instead of an alias? The earlier ports do that so it'd be consistent, and it might make deleting code easier later. So I'd go with `sentry_sdk.start_span` for the legacy and `sentry_sdk.traces.start_span` for the new way.

sentrivana (changes requested):
TY! Left some comments, the gist:
- Any attributes we set on the span in streaming mode need to be in [Sentry Conventions](https://getsentry.github.io/sentry-conventions/). I've tried to point out some that need changing but please go through the rest too and double check 🙏🏻 
- I'd remove all of the `start_timestamp_monotonic_ms` changes. I don't really see the benefit over just using the `start_timestamp` and removing it makes the new and old `add_http_request_source` similar enough that we can unify them.

sentry_sdk/traces.py
ericapisani: > I also don't know why we're guarding against AttributeErrors here (and later in _end), but it seems deliberate (this was before LLMs 😆).

From my commit splunking, it looks like this try-except was [introduced a while ago](https://github.com/getsentry/sentry-python/commit/d31e805f1) in the existing tracing implementation when Python 2 was still supported, and the `perf_counter` didn't exist on the time module just yet as it was first made available in [Python 3.3.](https://docs.python.org/3/library/time.html#time.perf_counter). 

Since we've dropped support for Python 2 within the SDK, this is safe to remove 🔥

sentry_sdk/integrations/httpx.py
ericapisani: You're correct - just did a test in the Python interpreter and `parsed_url.query` won't contain a starting question mark here

sentry_sdk/tracing_utils.py
ericapisani: > Why work with the monotonic start timestamp here at all? The original add_http_request_source just uses start_timestamp, so I believe that should be good enough?

I opted to use it in this way because  the `span.timestamp` is set within `_end` using the `start_timestamp_monotonic_ns` (snippet below from within `StreamedSpan` but here's the [link](https://github.com/getsentry/sentry-python/blob/6ce096bc83fba9defff06b5b2a0c236073de26c9/sentry_sdk/traces.py#L387-L394) on main as well):

```python
if self._timestamp is None:
            if self._start_timestamp_monotonic_ns is not None:
                elapsed = nanosecond_time() - self._start_timestamp_monotonic_ns
                self._timestamp = self._start_timestamp + timedelta(
                    microseconds=elapsed / 1000
                )
            else:
                self._timestamp = datetime.now(timezone.utc)
```

and I was trying to emulate the behaviour as closely as possible (which is also why I exposed it as a property on the class).

If we want to specifically avoid exposing `start_timestamp_monotonic_ns` as a property, we could also explore adding a method that does this calculation for us to retain the behaviour.

~~I'm also realizing that, since the changes that I introduced unconditionally set the `self._start_timestamp_monotonic_ns` within the `__init__`, this can never be `None` and so the `else` falling back to a datetime stamp is dead code and needs to be removed.~~
-- edit: this is wrong, I forgot about the `NoOpStreamedSpan` class which would leave this uninitialized

Also - I think you're right that this can be consolidated into the single function (`add_http_request_source`), so will look to make those changes regardless.

sentry_sdk/tracing_utils.py
ericapisani: If a namespace and code_function are both available, the code_function, once updated to also set `code.function.name`, will overwrite the namespace value if this change is made.

To avoid overloading the meaning of this attribute, I'm going to remove the setting of this property (namespace) to `code.function.name` in favour of solely relying on `frame.f_code.co_name` being set as the attribute name

sentry_sdk/integrations/httpx.py
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6084/commits/25a953e9399d56118480620a5ab5218fe5bf1907

sentry_sdk/integrations/httpx.py
ericapisani: Renamed to `streamed_span` in https://github.com/getsentry/sentry-python/pull/6084/commits/25a953e9399d56118480620a5ab5218fe5bf1907 because `mypy` doesn't like it when `span` is used for both the streamed and non-streamed context managers (it scopes them at the function-level and will raise errors about the `StreamedSpan` not having `set_data`)

sentry_sdk/tracing_utils.py
ericapisani: I've consolidated into a single function in https://github.com/getsentry/sentry-python/pull/6084/commits/25a953e9399d56118480620a5ab5218fe5bf1907 , but have left the `start_timestamp_monotonic_ns` changes in there for now

sentry_sdk/tracing_utils.py
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6084/commits/25a953e9399d56118480620a5ab5218fe5bf1907

sentry_sdk/tracing_utils.py
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6084/commits/25a953e9399d56118480620a5ab5218fe5bf1907

sentry_sdk/integrations/httpx.py
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6084/commits/25a953e9399d56118480620a5ab5218fe5bf1907

sentry_sdk/tracing_utils.py
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6084/commits/25a953e9399d56118480620a5ab5218fe5bf1907

sentry_sdk/integrations/httpx.py
ericapisani: I've asked about `http` vs `url` in the project channel just in case others have thoughts, but have optimistically picked `url` in https://github.com/getsentry/sentry-python/pull/6084/commits/25a953e9399d56118480620a5ab5218fe5bf1907

sentry_sdk/integrations/httpx.py
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6084/commits/25a953e9399d56118480620a5ab5218fe5bf1907

sentry_sdk/integrations/httpx.py
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6084/commits/25a953e9399d56118480620a5ab5218fe5bf1907

sentry_sdk/tracing_utils.py
```diff
@@ -273,13 +275,16 @@ def add_source(
...
+                span.set_data(SPANDATA.CODE_LINENO, lineno)
+            else:
+                span.set_attribute("code.line.number", lineno)
 
         try:
             namespace = frame.f_globals.get("__name__")
         except Exception:
             namespace = None
-        if namespace is not None:
+        if namespace is not None and isinstance(span, LegacySpan):
```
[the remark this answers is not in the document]
ericapisani: This is an intentional change

sentry_sdk/traces.py
sentrivana: Ok, sounds good then 🤞🏻

sentry_sdk/tracing_utils.py
sentrivana: > I opted to use it in this way because the span.timestamp is set within _end using the `start_timestamp_monotonic_ns`

I see. I still think it's sufficient to use just `start_timestamp` in this case. We just need to know if the operation the span is capturing is longer than a threshold in milliseconds (since the option to configure this is in ms). Both `start_timestamp` and `datetime.now(timezone.utc)` have more than millisecond precision, so they should be sufficient for a simple heuristic like this. And using them simplifies the code a great deal.

ericapisani: Sounds good, I'll make the change 😄

sentry_sdk/traces.py
ericapisani: Removed in https://github.com/getsentry/sentry-python/pull/6084/commits/19eeac58f7867075af5047d16af958baddc0c131

sentry_sdk/tracing_utils.py
ericapisani: Removed in https://github.com/getsentry/sentry-python/pull/6084/commits/19eeac58f7867075af5047d16af958baddc0c131

sentrivana (approved):
Ty for making the changes 🙏🏻 

Cursor left one more comment: https://github.com/getsentry/sentry-python/pull/6084#discussion_r3110944445 Might make sense to harden the code path if an exception gets thrown from `send` with a `finally:` or something similar.

Looks great otherwise! Approving to not block this unnecessarily.

sentry_sdk/integrations/httpx.py
```diff
@@ -49,48 +50,101 @@ def _install_httpx_client() -> None:
...
+                attributes["http.response.status_code"] = rv.status_code
 
-        with capture_internal_exceptions():
-            add_http_request_source(span)
+                streamed_span.set_attributes(attributes)
+
+                # Needs to happen within the context manager as we want to attach the
+                # final data before the span finishes and is sent for ingesting.
+                with capture_internal_exceptions():
+                    add_http_request_source(streamed_span)
```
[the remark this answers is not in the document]
ericapisani: Fixed in [a632a81](https://github.com/getsentry/sentry-python/pull/6084/commits/a632a818a3d2e8c26eee980d0c886ee56513ebd6)

PR #6102, opened 2026-04-20
fix(openai-agents): Resolve agent from `bindings` for openai-agents >= 0.14

### LINKED ISSUE #6101: Update patches in openai-agents due to introduction of AgentBindings
Tox [test suite is failing](<https://github.com/getsentry/sentry-python/pull/6100>) on the latest versions of OpenAI-agent libraries due to replacement of the `agent` argument with `bindings` (see [the latest](<https://github.com/openai/openai-agents-python/blob/da82b2cd663ee263b206ce65179c26b598d61f73/src/agents/run_internal/run_loop.py#L1686>) vs. [v.0.13.6](<https://github.com/openai/openai-agents-python/blob/bfc8c6355ff0cc2b7f43980ed1b459f411dc42e5/src/agents/run_internal/run_loop.py#L1500>)).

ericapisani:
Update the openai-agents integration's run-loop patches to resolve the current agent from the new `AgentBindings` dataclass introduced in openai-agents 0.14, falling back to the pre-0.14 `agent` kwarg.

Prior to 0.14, `run_single_turn`, `run_single_turn_streamed`, `execute_handoffs`, and `execute_final_output` all received the current agent as `agent=...`. In 0.14 the library refactored these to pass `bindings: AgentBindings` (carrying `public_agent` and `execution_agent`) to the `run_single_turn*` functions and renamed the `agent` kwarg to `public_agent` on the `execute_*` functions. Our patches still called `kwargs.get("agent")`, which returned `None` — so `_maybe_start_agent_span` short-circuited and no `invoke_agent` span was ever created. Every `test_agent_invocation_span*` parametrization failed on the `openai_agents-latest` suite with `not enough values to unpack (expected 2, got 1)`.

`public_agent` is used rather than `execution_agent` because it's the user-facing identity the library itself uses for lifecycle hooks (`on_agent_start`, `on_llm_start`), `streamed_result.current_agent`, and `RunItem.agent` fields. The two only diverge in sandbox execution (`agents/sandbox/runtime.py`), where `execution_agent` is a cloned sandbox-prepared agent — using that would produce unstable span names that don't match what users see from the library's own callbacks.

Verified locally against `py3.14-openai_agents-latest` (bindings path) and `py3.14-openai_agents-v0.8.4` (legacy kwarg fallback path).

Refs PY-2382
Fixes #6101

### DIFF
--- a/sentry_sdk/integrations/openai_agents/patches/agent_run.py
+++ b/sentry_sdk/integrations/openai_agents/patches/agent_run.py
@@ -89,7 +89,13 @@ async def _run_single_turn(
     - creates agent invocation spans if there is no already active agent invocation span.
     - ends the agent invocation span if and only if an exception is raised in `_run_single_turn()`.
     """
-    agent = kwargs.get("agent")
+    # openai-agents >= 0.14 passes `bindings: AgentBindings` instead of `agent`.
+    bindings = kwargs.get("bindings")
+    agent = (
+        getattr(bindings, "public_agent", None)
+        if bindings is not None
+        else kwargs.get("agent")
+    )
     context_wrapper = kwargs.get("context_wrapper")
     should_run_agent_start_hooks = kwargs.get("should_run_agent_start_hooks", False)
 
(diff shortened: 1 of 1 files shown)


sentrivana (approved):
🚀

PR #6104, opened 2026-04-21
ref(transport): Remove redundant checks for dsn

### LINKED ISSUE #6089: ref(transport): Simplify redundant DSN check in Transport.__init__
In Transport.**init** ([transport.py#L109](<http://transport.py#L109>)), the DSN check contains a redundant is not None guard:                                     <br>   <br>if options and options\["dsn"\] is not None and options\["dsn"\]:                                                                        <br>                                                                                         <br>The is not None check is unnecessary — the subsequent truthiness check already evaluates None as falsy, making it redundant.         <br>                                                                                         <br>This is also inconsistent with the existing pattern later in the same file:                                                          <br>                                                                                         <br>[transport.py](<http://transport.py>) L1165                                                                                                                 <br>if options\["dsn"\]:

Suggested Change

* if options and options\["dsn"\] is not None and options\["dsn"\]:
* if options and options\["dsn"\]:

psh9508:
#### Related Issue
* Fixes #6089

#### What changed
Removed the redundant `is not None` check from the DSN conditional statement in `Transport.init` (`transport.py`).

**Code Comparison:**
```python
# Before
if options and options["dsn"] is not None and options["dsn"]:

# After
if options and options["dsn"]:
```

####  Why
In Python, the truthiness check (`options["dsn"]`) inherently evaluates `None` as falsy. Therefore, the explicit `is not None` guard is redundant. Removing it simplifies the logic while maintaining the same behavior.

####  Notes
* **Why not use `options.get('dsn')`?** While `.get()` is often used for safe dictionary access, a default value for `"dsn"` is already guaranteed in this context, so there is no risk of a `KeyError`. 
* **Consistency:** I intentionally kept the direct dictionary access (`options["dsn"]`) to remain consistent with the existing coding pattern used later in the same file (e.g., line 1165: `if options["dsn"]:`).


sentrivana (approved):

psh9508:
@sentrivana I've merged the latest master to keep the branch up to date, but the Warden workflow is failing with an authentication error ([@octokit/auth-app] appId option is required). It seems like a secret-access limitation for PRs from forks. Could you please take a look and help re-trigger or bypass the workflow? Thank you!

Issue #5386, opened 2026-01-29
Migrate all integrations to span first


sentrivana:

alexander-alderman-webb:
Would be great to wait on AI integrations until we have more clarity on what happens with sending `gen_ai.*` spans as V2 envelope items in https://github.com/getsentry/sentry-python/pull/6079. Otherwise there will be a lot of conflicts in the test files.

And for now there's enough other integrations to port 😄.
Let me know if we need to prioritize porting AI integrations though. 

cc @sentrivana @ericapisani

sentrivana:
Makes sense to me. I'll adjust the prio of the individual issues.

PR #6123, opened 2026-04-21
feat(starlette): Support span streaming

ericapisani:
Bring the Starlette integration to span-streaming parity with the FastAPI change that landed in #6118.

Under the `trace_lifecycle: "stream"` experiment the scope's `transaction` is `None` and the scope instead holds a `StreamedSpan`, so the existing middleware instrumentation (which called `start_span(op=..., origin=...) + set_tag`) and the profiler's `current_scope.transaction.update_active_thread()` hook in `patch_request_response` didn't function under streaming. This PR routes both through the streaming APIs when streaming is enabled while leaving the legacy transaction-based path untouched:

- `_enable_span_for_middleware` now uses a `_start_middleware_span` helper that calls `sentry_sdk.traces.start_span` with `sentry.op`, `sentry.origin`, and `starlette.middleware_name` attributes under streaming, and the existing `sentry_sdk.start_span(op=..., origin=...) + set_tag` otherwise.
- The profiler hook detects `StreamedSpan` (excluding `NoOpStreamedSpan`) and calls `_segment._update_active_thread()`; otherwise it takes the legacy `current_scope.transaction.update_active_thread()` branch.

Tests are parametrized across streaming and static modes for `test_middleware_spans`, `test_middleware_spans_disabled`, `test_middleware_callback_spans`, and `test_span_origin`. A new `test_active_thread_id_span_streaming` covers the segment's `thread.id` attribute under streaming. `auto_enabling_integrations=False` is set in streaming tests where auto-instrumented spans would otherwise leak into the captured span stream.

Refs PY-2362

### DIFF
--- a/sentry_sdk/integrations/starlette.py
+++ b/sentry_sdk/integrations/starlette.py
@@ -238,7 +238,7 @@ def _serialize_body_data(data: "Any") -> str:
     # data may be a JSON-serializable value, an AnnotatedValue, or a dict with AnnotatedValue values
     def _default(value: "Any") -> "Any":
         if isinstance(value, AnnotatedValue):
-            return {"value": value.value, "metadata": value.metadata}
+            return value.value
         return str(value)
 
     return json.dumps(data, default=_default)
--- a/sentry_sdk/integrations/starlette.py
+++ b/sentry_sdk/integrations/starlette.py
@@ -521,10 +521,12 @@ def patch_request_response() -> None:
                         and not isinstance(current_span, NoOpStreamedSpan)
                     ):
                         data = info["data"]
-                        current_span._segment.set_attribute(
-                            "http.request.body.data",
-                            _serialize_body_data(data),
-                        )
+
+                        with capture_internal_exceptions():
+                            current_span._segment.set_attribute(
+                                "http.request.body.data",
+                                _serialize_body_data(data),
+                            )
 
                 return await old_func(*args, **kwargs)
 
--- a/sentry_sdk/integrations/starlette.py
+++ b/sentry_sdk/integrations/starlette.py
@@ -20,10 +20,12 @@ from sentry_sdk.integrations._wsgi_common import (
 )
 from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 from sentry_sdk.scope import should_send_default_pii
+from sentry_sdk.traces import NoOpStreamedSpan, StreamedSpan
 from sentry_sdk.tracing import (
     SOURCE_FOR_STYLE,
     TransactionSource,
 )
+from sentry_sdk.tracing_utils import has_span_streaming_enabled
 from sentry_sdk.utils import (
     AnnotatedValue,
     capture_internal_exceptions,
@@ -147,7 +149,8 @@ def _enable_span_for_middleware(middleware_class: "Any") -> type:
         send: "Callable[[Dict[str, Any]], Awaitable[None]]",
         **kwargs: "Any",
     ) -> None:
-        integration = sentry_sdk.get_client().get_integration(StarletteIntegration)
+        client = sentry_sdk.get_client()
+        integration = client.get_integration(StarletteIntegration)
         if integration is None:
             return await old_call(app, scope, receive, send, **kwargs)
 
@@ -164,22 +167,38 @@ def _enable_span_for_middleware(middleware_class: "Any") -> type:
             return await old_call(app, scope, receive, send, **kwargs)
 
         middleware_name = app.__class__.__name__
+        is_span_streaming_enabled = has_span_streaming_enabled(client.options)
 
-        with sentry_sdk.start_span(
-            op=OP.MIDDLEWARE_STARLETTE,
-            name=middleware_name,
-            origin=StarletteIntegration.origin,
+        def _start_middleware_span(op: str, name: str) -> "Any":
+            if is_span_streaming_enabled:
+                return sentry_sdk.traces.start_span(
+                    name=name,
+                    attributes={
+                        "sentry.op": op,
+                        "sentry.origin": StarletteIntegration.origin,
+                        "starlette.middleware_name": middleware_name,
+                    },
+                )
+            return sentry_sdk.start_span(
+                op=op,
+                name=name,
+                origin=StarletteIntegration.origin,
+            )
+
+        with _start_middleware_span(
+            op=OP.MIDDLEWARE_STARLETTE, name=middleware_name
         ) as middleware_span:
-            middleware_span.set_tag("starlette.middleware_name", middleware_name)
+            if not is_span_streaming_enabled:
+                middleware_span.set_tag("starlette.middleware_name", middleware_name)
 
             # Creating spans for the "receive" callback
             async def _sentry_receive(*args: "Any", **kwargs: "Any") -> "Any":
-                with sentry_sdk.start_span(
+                with _start_middleware_span(
                     op=OP.MIDDLEWARE_STARLETTE_RECEIVE,
                     name=getattr(receive, "__qualname__", str(receive)),
-                    origin=StarletteIntegration.origin,
                 ) as span:
-                    span.set_tag("starlette.middleware_name", middleware_name)
+                    if not is_span_streaming_enabled:
+                        span.set_tag("starlette.middleware_name", middleware_name)
                     return await receive(*args, **kwargs)
 
             receive_name = getattr(receive, "__name__", str(receive))
@@ -188,12 +207,12 @@ def _enable_span_for_middleware(middleware_class: "Any") -> type:
 
             # Creating spans for the "send" callback
             async def _sentry_send(*args: "Any", **kwargs: "Any") -> "Any":
-                with sentry_sdk.start_span(
+                with _start_middleware_span(
                     op=OP.MIDDLEWARE_STARLETTE_SEND,
                     name=getattr(send, "__qualname__", str(send)),
-                    origin=StarletteIntegration.origin,
                 ) as span:
-                    span.set_tag("starlette.middleware_name", middleware_name)
+                    if not is_span_streaming_enabled:
+                        span.set_tag("starlette.middleware_name", middleware_name)
                     return await send(*args, **kwargs)
 
             send_name = getattr(send, "__name__", str(send))
@@ -496,7 +515,13 @@ def patch_request_response() -> None:
                     return old_func(*args, **kwargs)
 
                 current_scope = sentry_sdk.get_current_scope()
-                if current_scope.transaction is not None:
+                current_span = current_scope.span
+
+                if isinstance(current_span, StreamedSpan) and not isinstance(
+                    current_span, NoOpStreamedSpan
+                ):
+                    current_span._segment._update_active_thread()
+                elif current_scope.transaction is not None:
                     current_scope.transaction.update_active_thread()
 
                 sentry_scope = sentry_sdk.get_isolation_scope()
--- a/sentry_sdk/integrations/starlette.py
+++ b/sentry_sdk/integrations/starlette.py
@@ -1,5 +1,6 @@
 import asyncio
 import functools
+import json
 import warnings
 from collections.abc import Set
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/starlette.py
```diff
@@ -164,22 +167,38 @@ async def _create_span_call(
...
+        is_span_streaming_enabled = has_span_streaming_enabled(client.options)
+
+        def _start_middleware_span(op: str, name: str) -> "Any":
+            if is_span_streaming_enabled:
+                return sentry_sdk.traces.start_span(
+                    name=name,
+                    attributes={
+                        "sentry.op": op,
+                        "sentry.origin": StarletteIntegration.origin,
+                        "starlette.middleware_name": middleware_name,
```
sentrivana, line 179: `starlette.middleware_name` is not in Sentry semantic conventions -- anything we set as an attribute on a streamed span needs to be there. So we either need to add it or not set it at all.

In this case, if I read the code right, it seems to be the same as the span `name`? In that case I'd omit the attribute. If we want to keep it, I'd propose a new attribute name in Sentry conventions that's not tied to Starlette specifically, so that we can use it across different frameworks and languages.

sentry_sdk/integrations/starlette.py
sentrivana, line 501: We're adding an event processor here. In legacy mode, it would enrich any transaction event with additional data. Since streamed spans are not events, it'll not apply, so anytime there's an event processor in an integration, we need to come up with an equivalent way to apply the data to streaming segments as well.

The gist is basically:
- Leave the event processor in place since it should continue working for other event types, like errors, as well, in the future.
- The span streaming version of this will basically be us setting equivalent attributes (which need to be in conventions) on the segment span.
- Event processors are normally run at the very end of the transaction lifecycle. Depending on what data they set, the span streaming equivalent might also need to be run towards the end of the segment's lifetime -- it depends on whether any of the data we set can be expected to be populated only later (like, for instance, the response status code).

Have a look at ASGI for prior art.

Ideally, this would've popped up in the tests. Looks like we're not really testing that the event processor applies the correct data, since the tests are passing even without it. That's not great. But we can also address that at a later point since it's out of scope of this PR.

sentrivana, line 560: We'll [also](https://github.com/getsentry/sentry-python/pull/6123/changes#r3122209592) need to port this event processor

sentrivana (changes requested):
Nice, left a few comments -- we'll need to migrate the event processors as well.

sentry_sdk/integrations/starlette.py
ericapisani: > starlette.middleware_name is not in Sentry semantic conventions -- anything we set as an attribute on a streamed span needs to be there. So we either need to add it or not set it at all.

I had assumed the following from the span-first migration docs applied since `starlette.middleware_name` is being set as a tag in the legacy approach:

> In span streaming mode, spans have no contexts, data, or tags. Everything is a span attribute. It's therefore necessary to migrate all existing span.set_data(), span.set_context(), and span.set_tag() to span.set_attribute()

Re:

> In this case, if I read the code right, it seems to be the same as the span name?

Not in this case - `middleware_name` is set to the value of `app.__class__.__name__` (just above the `is_span_streaming_enabled` check) whereas the `name` set on the span differs depending on if this is invoked in the patched `receive` or `send` callbacks a little further below ([receive](https://github.com/getsentry/sentry-python/pull/6123/changes#diff-6f6cf0fb18998238da86355c38e101792b5fe0582ea0c09c615a68eb82fbea57R198) and [send](https://github.com/getsentry/sentry-python/pull/6123/changes#diff-6f6cf0fb18998238da86355c38e101792b5fe0582ea0c09c615a68eb82fbea57R212) links) and ultimately makes the `name` a different value from `middleware_name`

sentrivana: > In span streaming mode, spans have no contexts, data, or tags. Everything is a span attribute. It's therefore necessary to migrate all existing span.set_data(), span.set_context(), and span.set_tag() to span.set_attribute()

This is true, but it's from the migration guide, which is a user facing artifact -- it doesn't mention sentry conventions since that only applies to attributes set by the SDK.

This https://github.com/getsentry/sentry-python/issues/5386 has more detailed info on the SDK-set attrs:

> When translating old contexts, data, tags etc. to span attributes, follow Sentry conventions. If anything is unclear or there's an attribute missing, raise on Slack in #proj-span-first-sentry as product folks also need to be aligned on attribute changes.

---

> Not in this case - middleware_name is set to the value of app.__class__.__name__ (just above the is_span_streaming_enabled check) whereas the name set on the span differs depending on if this is invoked in the patched receive or send callbacks a little further below (receive and send links) and ultimately makes the name a different value from middleware_name

Gotcha. In that case adding a convention for middleware names would be the way to go.

ericapisani: > When translating old contexts, data, tags etc. to span attributes, follow Sentry conventions. If anything is unclear or there's an attribute missing, raise on Slack in #proj-span-first-sentry as product folks also need to be aligned on attribute changes.

Ah, that's my oversight, I'll be more careful checking that list instead going forward!

And otherwise, sounds good, will add the missing things to the conventions 🚀

sentry_sdk/integrations/starlette.py
ericapisani: Because Starlette leverages Sentry's ASGI  middleware, the cookies that we're enriching the event data with in the event processor will be added on automatically to the span when span-first is enabled by logic within the ASGI integration (so no changes are needed here)

sentry_sdk/integrations/starlette.py
ericapisani: Have PRs open for adding the [starlette middleware](https://github.com/getsentry/sentry-conventions/pull/337) and [`http.request.body.data`](https://github.com/getsentry/sentry-conventions/pull/336) to the conventions

sentry_sdk/integrations/starlette.py
ericapisani: There's some information that's already being added to the spans via the ASGI middleware and its span streaming support (such as cookies via headers). The request body data is now being added as of commit https://github.com/getsentry/sentry-python/pull/6123/commits/32286b8c4789648664115064f6961465fe065501

Re: the test coverage - I've created an issue to address that separately. https://github.com/getsentry/sentry-python/issues/6130

sentry_sdk/integrations/starlette.py
```diff
@@ -164,22 +168,38 @@ async def _create_span_call(
...
+        is_span_streaming_enabled = has_span_streaming_enabled(client.options)
+
+        def _start_middleware_span(op: str, name: str) -> "Any":
+            if is_span_streaming_enabled:
+                return sentry_sdk.traces.start_span(
+                    name=name,
+                    attributes={
+                        "sentry.op": op,
+                        "sentry.origin": StarletteIntegration.origin,
+                        "middleware.name": middleware_name,
```
ericapisani, line 180: The associated [convention PR](https://github.com/getsentry/sentry-conventions/pull/337) changed this from `starlette.middleware.name` to `middleware.name`. I left the legacy stream alone because I didn't want to break existing functionality for users.

sentry_sdk/integrations/starlette.py
```diff
@@ -481,6 +510,22 @@ def event_processor(
...
+                        info
+                        and "data" in info
+                        and isinstance(current_span, StreamedSpan)
+                        and not isinstance(current_span, NoOpStreamedSpan)
+                    ):
+                        data = info["data"]
+                        current_span._segment.set_attribute(
+                            "http.request.body.data",
+                            _serialize_body_data(data),
+                        )
```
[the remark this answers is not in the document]
ericapisani: Going forward, the sanitization of data will need to be handled by the user via the `before_send_span` hook. 

As a result, we are not scrubbing the data the same way that we do for the current spans, which is handled via the `EventScrubber`.

sentrivana: We should probably take `max_request_body_size` into account here

sentry_sdk/integrations/starlette.py
```diff
@@ -214,6 +234,16 @@ async def _sentry_send(*args: "Any", **kwargs: "Any") -> "Any":
...
 
 
+def _serialize_body_data(data: "Any") -> str:
+    # data may be a JSON-serializable value, an AnnotatedValue, or a dict with AnnotatedValue values
+    def _default(value: "Any") -> "Any":
+        if isinstance(value, AnnotatedValue):
+            return {"value": value.value, "metadata": value.metadata}
+        return str(value)
+
+    return json.dumps(data, default=_default)
```
sentrivana, line 244: Can the data here have `AnnotatedValue`s in it? It seems to come straight from Starlette (i.e., I don't believe it's gone through any SDK sanitization at this point)?

I hope we can just get rid of the whole `AnnotatedValue` business in span first.

tests/integrations/starlette/test_starlette.py
```diff
@@ -676,55 +683,107 @@ def test_middleware_spans(sentry_init, capture_events):
...
+        middleware_spans = sorted(
+            [
+                item.payload
+                for item in items
+                if item.payload.get("attributes", {})
+                .get("sentry.op", "")
+                .startswith("middleware.starlette")
+            ],
+            key=lambda s: s["start_timestamp"],
+        )
```
sentrivana: I've noticed the clanker loves to do this sort of span filtering with matching on the `op` -- it also did it on my PRs initially. This removes some implicit assumptions we're testing though, like there not being any extra unexpected spans. I think we should still retain the implicit assertions on the exact spans that are supposed to be emitted and their order (they'll be queued up by end time in both the legacy and span streaming branch).

So here e.g. I'd do something like:
```python
# No filtering (assign each span to where we expect it to be) and no sorting 
segment = items.pop(-1).payload
middleware_spans = [item.payload for item in items]
... # assertions...
```

So basically, I'd advocate for following the logic in the old branch as close as possible, without any extra filtering or sorting of spans -- not just here but also in other tests (the clanker seems to get it right when you tell it).

sentry_sdk/integrations/starlette.py
ericapisani: It's already being taken into account through the [`request_body_within_bounds`](https://github.com/getsentry/sentry-python/blob/12efdd58490aaa08b3f172b4198e88db6f0f7b0a/sentry_sdk/integrations/_wsgi_common.py#L62) method that's called within the [`extract_request_info`](https://github.com/getsentry/sentry-python/blob/12efdd58490aaa08b3f172b4198e88db6f0f7b0a/sentry_sdk/integrations/starlette.py#L635) method higher up in this file.

If the body is too large, the request's `data` will be replaced with the returned value from the `AnnotatedValue.removed_because_over_size_limit` method

sentry_sdk/integrations/starlette.py
ericapisani: > Can the data here have AnnotatedValues in it? 

It can when a raw value is present, or if the request body exceeds the size limit set by the `max_request_body_size` property.

Both of these occur within the `extract_request_info` method I linked to in the [comment above](https://github.com/getsentry/sentry-python/pull/6123#discussion_r3146805519).

> I don't believe it's gone through any SDK sanitization at this point?

From [this thread](https://sentry.slack.com/archives/C09RR0KD2N7/p1776951331206129?thread_ts=1776951227.440659&cid=C09RR0KD2N7), sanitization, at least for the request body, will be the user's responsibility using the `before_send_span` once span-first launches.

tests/integrations/starlette/test_starlette.py
ericapisani: Have updated all the middleware tests where this pattern appears in [this commit](https://github.com/getsentry/sentry-python/pull/6123/commits/29de94942dec392d8011e92ad9c0050b79b862f5)

sentry_sdk/integrations/starlette.py
```diff
@@ -214,6 +234,16 @@ async def _sentry_send(*args: "Any", **kwargs: "Any") -> "Any":
     return middleware_class
 
 
+def _serialize_body_data(data: "Any") -> str:
+    # data may be a JSON-serializable value, an AnnotatedValue, or a dict with AnnotatedValue values
+    def _default(value: "Any") -> "Any":
+        if isinstance(value, AnnotatedValue):
+            return {"value": value.value, "metadata": value.metadata}
```
sentrivana:
```suggestion
            return value.value
```

sentry_sdk/integrations/starlette.py
sentrivana: Gotcha. In that case looks good -- but in case of `AnnotatedValue`s, I'd just include the sanitized `value`, and not the metadata part (see https://github.com/getsentry/sentry-python/pull/6123/changes#r3151943009) as that's internal event protocol and it won't be of any value to the user.

Stamping this already to unblock.

sentrivana (approved):
Stamping to unblock, but please have a look at one last thing: https://github.com/getsentry/sentry-python/pull/6123/changes#r3151965692 🙏🏻

PR #6124, opened 2026-04-21
feat(huggingface_hub): Migrate to span first

### LINKED ISSUE #6030: Migrate huggingface_hub to span first

ericapisani:
Migrates the huggingface_hub integration to the span-first (streaming span) architecture.

When `_experiments={"trace_lifecycle": "stream"}` is enabled, the integration now uses `StreamedSpan` via `sentry_sdk.traces.start_span` instead of the legacy `Span`-based (transactions) path. The legacy path remains unchanged for backwards compatibility.

This is the first AI integration to emit `StreamedSpan`s, so the shared AI monitoring utilities needed to accept both span types. `record_token_usage` (`sentry_sdk/ai/monitoring.py`) and `set_data_normalized` (`sentry_sdk/ai/utils.py`) now accept `Union[Span, StreamedSpan]` and route attribute writes through a new private `_set_span_data_attribute` helper in `sentry_sdk.ai.utils` — calling `set_attribute` on `StreamedSpan` and `set_data` on `Span`. Existing AI integrations (anthropic, openai, cohere, langchain, langgraph, litellm, google_genai, openai_agents, pydantic_ai) are unaffected at runtime: they continue to pass `Span` instances, which route to `set_data` via the isinstance check.

Test coverage mirrors the existing non-streaming suite under streaming mode (text_generation / chat_completion, sync and streaming, with and without tools, plus an error-path test) parametrized over `send_default_pii` and `include_prompts`.

Part of the broader span-first integration migration.

Fixes PY-2332
Fixes #6030

### DIFF
--- a/sentry_sdk/ai/monitoring.py
+++ b/sentry_sdk/ai/monitoring.py
@@ -2,10 +2,12 @@ import inspect
 import sys
 from functools import wraps
 
+from sentry_sdk.ai.utils import _set_span_data_attribute
 from sentry_sdk.consts import SPANDATA
 import sentry_sdk.utils
 from sentry_sdk import start_span
 from sentry_sdk.tracing import Span
+from sentry_sdk.traces import StreamedSpan
 from sentry_sdk.utils import ContextVar, reraise, capture_internal_exceptions
 
 from typing import TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


alexander-alderman-webb (approved):
Looks good to me.

sentrivana (approved):
Looking great! I didn't double check if all the attributes are in the conventions but I assume you've checked -- if not please do before merging 🙏🏻

ericapisani:
@sentrivana Yup, we're all good on the attributes being in the conventions - nothing new was added in this one! 🚀

PR #6136, opened 2026-04-24
feat(fastapi): Capture request body on segment span under streaming

### LINKED ISSUE #6020: Migrate fastapi to span first

ericapisani:
Capture the request body as `http.request.body.data` on the streaming segment span for FastAPI requests, matching the behavior already in place for Starlette.

FastAPI runs its own request-body extraction path, so it was bypassing the data-capture logic that lives in the Starlette integration. This extracts that logic into a shared helper in the Starlette integration and invokes it from FastAPI's request handler once the body has been read.

Stacked on top of #6123.

Refs PY-2322
Fixes #6020


ericapisani:
bugbot run

sentry_sdk/integrations/starlette.py
```diff
@@ -244,6 +244,22 @@ def _default(value: "Any") -> "Any":
...
+    if (
+        info
+        and "data" in info
+        and isinstance(current_span, StreamedSpan)
+        and not isinstance(current_span, NoOpStreamedSpan)
+    ):
+        current_span._segment.set_attribute(
+            "http.request.body.data",
+            _serialize_request_body_data(info["data"]),
+        )
```
sentrivana: Same concern as on the Starlette PR: We should respect `max_request_body_size` before setting the attribute

ericapisani: It's being taken into account through Starlette's [`extract_request_info`](https://github.com/getsentry/sentry-python/pull/6123#discussion_r3145939091) so we're all good here 😄

sentrivana (approved):
We should def make sure to call out the change in automatic sanitization and let folks know they'll need to sanitize themselves. Added it to https://linear.app/getsentry/project/span-first-sdk-python-727da28dd037/overview -- we can keep a running list there so that we don't forget about these.

PR #6167, opened 2026-04-29
test(fastmcp): Span streaming tests

alexander-alderman-webb:
### Description

The following tests are **not** parametrized since they have no assertions on spans:

- `test_fastmcp_tool_with_none_return()`
- `test_fastmcp_tool_with_no_arguments()`
- `test_standalone_fastmcp_specific_features()`
- `test_mcp_fastmcp_specific_features()`
- `test_fastmcp_prompt_async()`

--- 

## Adapting Tests

`sed` commands used for converting transaction context managers:

- sed -i '' 's/start_transaction(name="standalone fastmcp tx")/sentry_sdk.traces.start_span(name="custom parent")/g'
- sed -i '' 's/start_transaction(name="fastmcp tx")/sentry_sdk.traces.start_span(name="custom parent")/g'

`sed` commands used for converting event capture:

- sed -i '' 's/capture_events,/capture_items,/g'
- sed -i '' 's/events = capture_events()/items = capture_items("event", "transaction", "span")/g'
- sed -i '' 's/tx\["spans"\]/spans/g'
- sed -i '' 's/assert len(tx\["spans"\]) == 1/assert len(spans) == 2/g'
- sed -i '' 's/span = tx\["spans"\]\[0\]/span = spans[0]/g'
- sed -i '' 's/span = tx\["spans"\]\[0\]/span = next(item.payload for item in items if item.type == "span")/g'
- sed -i '' '/(tx,) = events/d'

`sed` commands used for converting `op`:

- sed -i '' 's/\["op"\]/\["attributes"\]\["sentry.op"\]/g'
- sed -i '' 's/span\["op"\]/span\["attributes"\]\["sentry.op"\]/g'

`sed` commands used for converting origin:

- sed -i '' 's/\["origin"\]/\["attributes"\]\["sentry.origin"\]/g'

`sed` commands used for converting `description`:

- sed -i '' 's/description/name/g'

`sed` commands used for converting `data` to `attributes`:

- sed -i '' 's/\["data"\]/\["attributes"\]/g'

`sed` commands to remove transaction assertions:

- sed -i '' '/assert tx\["type"\] == "transaction"/d'

#### Reminders

### DIFF
--- a/tests/integrations/fastmcp/test_fastmcp.py
+++ b/tests/integrations/fastmcp/test_fastmcp.py
@@ -963,6 +963,79 @@ async def test_fastmcp_prompt_sync(
                 assert SPANDATA.MCP_PROMPT_RESULT_MESSAGE_CONTENT not in span["data"]
 
 
+@pytest.mark.parametrize("FastMCP", fastmcp_implementations, ids=fastmcp_ids)
+@pytest.mark.asyncio
+async def test_fastmcp_prompt_async(
+    sentry_init,
+    capture_events,
+    FastMCP,
+    json_rpc,
+    select_transactions_with_mcp_spans,
+):
+    """Test that FastMCP async prompt handlers create proper spans"""
+    sentry_init(
+        integrations=[MCPIntegration()],
+        traces_sample_rate=1.0,
+    )
+    events = capture_events()
+
+    mcp = FastMCP("Test Server")
+
+    session_manager = StreamableHTTPSessionManager(
+        app=mcp._mcp_server,
+        json_response=True,
+    )
+
+    app = Starlette(
+        routes=[
+            Mount("/mcp", app=session_manager.handle_request),
+        ],
+        lifespan=lambda app: session_manager.run(),
+    )
+
+    # Try to register an async prompt handler
+    if hasattr(mcp, "prompt"):
+
+        @mcp.prompt()
+        async def async_prompt(topic: str):
+            """Get async prompt for a topic"""
+            message1 = {
+                "role": "user",
+                "content": {"type": "text", "text": f"What is {topic}?"},
+            }
+
+            message2 = {
+                "role": "assistant",
+                "content": {
+                    "type": "text",
+                    "text": "Let me explain that",
+                },
+            }
+
+            if FASTMCP_VERSION is not None and FASTMCP_VERSION.startswith("3"):
+                message1 = Message(message1)
+                message2 = Message(message2)
+
+            return [message1, message2]
+
+        _, result = json_rpc(
+            app,
+            method="prompts/get",
+            params={
+                "name": "async_prompt",
+                "arguments": {"topic": "MCP"},
+            },
+            request_id="req-async-prompt",
+        )
+
+        assert len(result.json()["result"]["messages"]) == 2
+
+        transactions = select_transactions_with_mcp_spans(
+            events, method_name="prompts/get"
+        )
+        assert len(transactions) == 1
+
+
 # =============================================================================
 # Resource Handler Tests (if supported)
 # =============================================================================
--- a/tests/integrations/fastmcp/test_fastmcp.py
+++ b/tests/integrations/fastmcp/test_fastmcp.py
@@ -9,7 +9,7 @@ both FastMCP implementations:
 Tests focus on verifying Sentry integration behavior:
 - Integration doesn't break FastMCP functionality
 - Span creation when tools/prompts/resources are called through MCP protocol
-- Span data accuracy (operation, description, origin, etc.)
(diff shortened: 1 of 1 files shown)


tests/integrations/fastmcp/test_fastmcp.py
```diff
@@ -632,112 +888,79 @@ def code_help_prompt(language: str):
...
 
-        transactions = select_transactions_with_mcp_spans(
-            events, method_name="prompts/get"
-        )
-        assert len(transactions) == 1
+            # Check PII-sensitive data
+            if send_default_pii and include_prompts:
+                assert SPANDATA.MCP_PROMPT_RESULT_MESSAGE_CONTENT in span["data"]
+            else:
+                assert SPANDATA.MCP_PROMPT_RESULT_MESSAGE_CONTENT not in span["data"]
```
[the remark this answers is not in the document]
alexander-alderman-webb: oops, will bring back

tests/integrations/fastmcp/test_fastmcp.py
```diff
@@ -292,42 +301,83 @@ def add_numbers(a: int, b: int) -> dict:
...
+
+        sentry_sdk.flush()
+        spans = [item.payload for item in items if item.type == "span"]
+        assert len(spans) == 2
+
+        # Verify span structure
+        span = spans[0]
+        assert span["attributes"]["sentry.op"] == OP.MCP_SERVER
+        assert span["attributes"]["sentry.origin"] == "auto.ai.mcp"
+        assert span["name"] == "tools/call add_numbers"
```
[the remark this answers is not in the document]
alexander-alderman-webb: out of scope.

tests/integrations/fastmcp/test_fastmcp.py
```diff
@@ -426,46 +534,94 @@ def failing_tool(value: int) -> int:
...
+            )
+
+        # If no exception raised, check if result indicates error
+        assert result.message.root.result["isError"] is True
+
+        sentry_sdk.flush()
+        # Check span was created
+        spans = [item.payload for item in items if item.type == "span"]
+        tool_spans = [
+            s for s in spans if s["attributes"].get("sentry.op") == OP.MCP_SERVER
```
ericapisani, line 557: I had followed a somewhat similar pattern in another PR and had received [feedback](https://github.com/getsentry/sentry-python/pull/6123#discussion_r3146010068) that we're trying to avoid this approach of selectively getting spans like this (filtering and matching on the `op`) because it removes some of the implicit assumptions that we're making about the amount of spans that we're emitting and their order.

Maybe something to sync about when we're all online so that we're all using the same approach going forward 😄

alexander-alderman-webb: In this case, the logic in the `else:` branch for the transaction-based variant has the same filter.

In general I'm trying to stay as close as possible to this philosophy (from the review comment you linked).
> I'd advocate for following the logic in the old branch as close as possible, without any extra filtering or sorting of spans

If there are other improvements to the tests we can still discuss and make them in a follow-up but I'd consider them out of scope for porting the tests to span-first.

What do you think?

ericapisani: > If there are other improvements to the tests we can still discuss and make them in a follow-up but I'd consider them out of scope for porting the tests to span-first.

Yeah that's fair to leave these improvements out of scope for this PR, I wouldn't want this to be overloaded with a number of different changes. 

My earlier comment was more to note the inconsistency we have if testing the implicit assumptions is important within this project, and wanting to ensure tests are ported over to the correct approach as part of the migration process if we can fit it in since the old version of the tests will be removed soon and we likely won't come back to this afterwards.

ericapisani (approved):

tests/integrations/fastmcp/test_fastmcp.py
alexander-alderman-webb: Gotcha. I think there are different levels of implicit assumptions:

- Lines similar to `(span, ) = spans` have an implicit assumption that there is only one span, as otherwise the test fails with a `ValueError`.
- Lines like `(client_span, tool_span, ) = spans` also have an ordering assumption. There is an implicit assumption that exactly 2 spans have been captured, and that the `client_span` is captured before the `tool_span`.

As for preferring one approach over another, I think it depends if you only test positive features (the span that I intend to create is there), or also the absence of other spans. I'm not aware of either approach being prescribed over the other.

I also don't think we provide a guarantee on the ordering of spans, so lines like `(client_span, tool_span, ) = spans` probably just exist because it's the shortest syntax to extract the spans 🤷 .

PR #6168, opened 2026-04-29
ref(batcher): Only flush the bucket that triggered the flush event

### LINKED ISSUE #5585: Only flush trace buckets that triggered a flush event
Flushing in the span batcher is currently global: if a flush event is emitted (e.g. because a bucket became full), we flush all buckets. Ideally, all the limits (size, weight, time) should be per-bucket, and we should only flush that specific bucket instead of the whole buffer.

sentrivana:
### Description
The span batcher uses multiple buffers (buckets) to store spans, keyed by `trace_id`. If a bucket fills up, we flush the whole buffer (all the buckets) instead of just the one bucket that actually should be flushed. This leads to too-frequent, busy flushes and is also against the [spec](https://develop.sentry.dev/sdk/telemetry/spans/span-buffer/#span-buffer-requirements).

With this change, we keep a set of buckets to flush. We still support full flushes for cases when `sentry_sdk.flush()` is called manually, or when the SDK is shutting down. Time-based flushes are also still global, i.e., not per bucket.

#### Issues
* closes https://github.com/getsentry/sentry-python/issues/5585
* closes https://linear.app/getsentry/issue/PY-2123/only-flush-trace-buckets-that-triggered-a-flush-event

#### Reminders


tests/tracing/test_span_streaming.py
```diff
@@ -1504,7 +1504,7 @@ def test_profile_stops_when_segment_ends(
     assert get_profiler_id() is None, "profiler should have stopped"
 
 
-def test_transport_format(sentry_init, capture_envelopes):
```
sentrivana, line 1508: This one was split in two (`test_default_attributes` and `test_transport_format`) and the latter was moved to `test_span_buffer`

sentry_sdk/_span_batcher.py
```diff
@@ -79,15 +101,23 @@ def add(self, span: "StreamedSpan") -> None:
...
         finally:
             self._active.flag = False
 
+    def kill(self) -> None:
+        if self._flusher is None:
+            return
+
+        self._running = False
+        self._flush_queue.put(None)
+        self._flusher = None
```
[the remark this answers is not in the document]
sentrivana: https://github.com/getsentry/sentry-python/pull/6168/commits/c3ee7530c6d3423792d1e71847bfe6dc3b85b540

sentry_sdk/_span_batcher.py
```diff
@@ -79,15 +101,24 @@ def add(self, span: "StreamedSpan") -> None:
...
             self._active.flag = False
 
+    def kill(self) -> None:
+        if self._flusher is None:
+            return
+
+        self._flush()
+        self._running = False
+        self._flush_queue.put(None)
+        self._flusher = None
```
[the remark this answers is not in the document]
sentrivana: I can't reliably determine whether this is even a real concern. The `active` flag was added to prevent capturing a recursive log while another log was being added. The equivalent here would be a span being created while another span was being added. I don't see how that would happen. Nevertheless, I changed this to use the `flush()` method (rather than `_flush()`) which sets the flag in https://github.com/getsentry/sentry-python/pull/6168/commits/41a98c025c43c2d268afc7656368e6af90b206e0

tests/tracing/test_span_buffer.py
```diff
@@ -0,0 +1,435 @@
...
+        pass
+
+    time.sleep(0.1)
+
+    assert len(envelopes) == 2
+
+    assert len(envelopes[0].items[0].payload.json["items"]) == 1
+    assert envelopes[0].items[0].payload.json["items"][0]["name"] == "span 1"
+
+    assert len(envelopes[1].items[0].payload.json["items"]) == 1
```
alexander-alderman-webb: Would creating one span be sufficient for the weight-based flushing test?

alexander-alderman-webb (approved):
LGTM!
It's probably not great having so many sleeps in the test, but the alternative is relying on private details (which wouldn't be good either).

tests/tracing/test_span_buffer.py
sentrivana: Definitely, not sure why I did it like this anymore: https://github.com/getsentry/sentry-python/pull/6168/commits/e88a50da1f4ec520262e4e9e3f376aff97e7a16a

sentrivana:
> It's probably not great having so many sleeps in the test, but the alternative is relying on private details (which wouldn't be good either).

Yeah definitely. Tried to keep them as low as possible at least.

PR #6177, opened 2026-04-30
fix: Introduce `_get_current_streamed_span()` to keep types backwards compatible

### LINKED ISSUE #6166: span.set_data does not exist on StreamedSpan
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.58.0

### Steps to Reproduce

With https://github.com/getsentry/sentry-python/commit/0c4a75db83c9b34c51079106f2c512f55f4dbe51, `span.get_current_span()` can now return a `StreamedSpan`. 
This break the recommended way in the documentation https://docs.sentry.io/platforms/python/tracing/span-lifecycle/#improving-span-data to set data on a span, as `set_data` does not exist on a `StreamedSpan` anymore.
As a workaround, I can add a check `if isinstance(span, Span):` before, but that does not seem ideal.  
 

### Expected Result

```python
sentry_sdk.get_current_span().set_data('foo', 'bar') # Type checker happy 
```

### Actual Result

```python
sentry_sdk.get_current_span().set_data('foo', 'bar') # Type checker not happy
```

alexander-alderman-webb:
…atible

### Description

Create streaming-lifecycle variants of the `tracing_utils.get_current_span()` function and the `Scope.span` property. The existing functions are narrowed to be backwards compatible for users of the transaction-based span lifecycle, allowing users to upgrade their SDK version without their type checkers raising errors.

Note that this breaks all current uses of these functions when the streaming lifecycle option is enabled (which is experimental). All call sites internal to the SDK are updated.

The `Scope.span` property returns `None` if a `StreamedSpan` is held, and `Scope.streamed_span` returns `None` if a `Span` is held on the scope.

#### Issues

Closes https://github.com/getsentry/sentry-python/issues/6166

#### Reminders

### DIFF
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -543,12 +543,7 @@ def update_current_span(
             attributes={"user_id": 123, "batch_size": 50}
         )
     """
-    current_span = get_current_span()
-
-    if current_span is None:
-        return
-
-    if isinstance(current_span, StreamedSpan):
+    if isinstance(get_current_streamed_span(), StreamedSpan):
         warnings.warn(
             "The `update_current_span` API isn't available in streaming mode. "
             "Retrieve the current span with get_current_span() and use its API "
@@ -556,6 +551,11 @@ def update_current_span(
             DeprecationWarning,
             stacklevel=2,
         )
+        return
+
+    current_span = get_current_span()
+
+    if current_span is None:
         return
 
     if op is not None:
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -6,7 +6,7 @@ from sentry_sdk import tracing_utils, Client
 from sentry_sdk._init_implementation import init
 from sentry_sdk.consts import INSTRUMENTER
 from sentry_sdk.scope import Scope, _ScopeManager, new_scope, isolation_scope
-from sentry_sdk.traces import StreamedSpan
+from sentry_sdk.traces import StreamedSpan, _get_current_streamed_span
 from sentry_sdk.tracing import NoOpSpan, Transaction, trace
 from sentry_sdk.crons import monitor
 
@@ -67,7 +67,6 @@ __all__ = [
     "get_isolation_scope",
     "get_current_scope",
     "get_current_span",
-    "get_current_streamed_span",
     "get_traceparent",
     "is_initialized",
     "isolation_scope",
@@ -430,15 +429,6 @@ def get_current_span(
     return tracing_utils.get_current_span(scope)
 
 
-def get_current_streamed_span(
-    scope: "Optional[Scope]" = None,
-) -> "Optional[StreamedSpan]":
-    """
-    Returns the currently active streamed span if there is one running, otherwise `None`
-    """
-    return tracing_utils.get_current_streamed_span(scope)
-
-
 def get_traceparent() -> "Optional[str]":
     """
     Returns the traceparent either from the active span or from the scope.
@@ -543,7 +533,7 @@ def update_current_span(
             attributes={"user_id": 123, "batch_size": 50}
         )
     """
-    if isinstance(get_current_streamed_span(), StreamedSpan):
+    if isinstance(_get_current_streamed_span(), StreamedSpan):
         warnings.warn(
             "The `update_current_span` API isn't available in streaming mode. "
             "Retrieve the current span with get_current_span() and use its API "
--- a/sentry_sdk/api.py
+++ b/sentry_sdk/api.py
@@ -67,6 +67,7 @@ __all__ = [
     "get_isolation_scope",
     "get_current_scope",
     "get_current_span",
+    "get_current_streamed_span",
     "get_traceparent",
     "is_initialized",
     "isolation_scope",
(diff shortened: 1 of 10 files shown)


sentry_sdk/tracing_utils.py
```diff
@@ -1203,6 +1203,17 @@ def get_current_span(
     return current_span
 
 
+def get_current_streamed_span(
```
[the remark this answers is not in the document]
alexander-alderman-webb: damn we shouldn't have this function anyway. I'll update

sentry_sdk/scope.py
```diff
@@ -580,12 +580,18 @@ def get_traceparent(self, *args: "Any", **kwargs: "Any") -> "Optional[str]":
         """
         client = self.get_client()
 
+        if not has_tracing_enabled(client.options):
+            return self.get_active_propagation_context().to_traceparent()
```
sentrivana, line 584: Do we need this early return? Isn't this covered by the fallback on line 598?

```python
        # else return traceparent from the propagation context
        return self.get_active_propagation_context().to_traceparent()
```

sentry_sdk/scope.py
```diff
@@ -598,12 +604,18 @@ def get_baggage(self, *args: "Any", **kwargs: "Any") -> "Optional[Baggage]":
         """
         client = self.get_client()
 
+        if not has_tracing_enabled(client.options):
+            return self.get_active_propagation_context().get_baggage()
```
sentrivana, line 608: Same q: https://github.com/getsentry/sentry-python/pull/6177/changes#r3167511101

sentry_sdk/api.py
```diff
@@ -422,13 +423,22 @@ def set_measurement(name: str, value: float, unit: "MeasurementUnit" = "") -> No
...
 
+def get_current_streamed_span(
+    scope: "Optional[Scope]" = None,
+) -> "Optional[StreamedSpan]":
+    """
+    Returns the currently active streamed span if there is one running, otherwise `None`
+    """
+    return tracing_utils.get_current_streamed_span(scope)
+
+
```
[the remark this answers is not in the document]
alexander-alderman-webb: https://github.com/getsentry/sentry-python/pull/6177/commits/a8adaeebe70a16dae670d4977896cd93cb2fff82

sentry_sdk/api.py
```diff
@@ -422,13 +423,22 @@ def set_measurement(name: str, value: float, unit: "MeasurementUnit" = "") -> No
...
     return tracing_utils.get_current_span(scope)
 
 
+def get_current_streamed_span(
+    scope: "Optional[Scope]" = None,
+) -> "Optional[StreamedSpan]":
+    """
+    Returns the currently active streamed span if there is one running, otherwise `None`
+    """
+    return tracing_utils.get_current_streamed_span(scope)
```
sentrivana: I'd not expose this as a public API function, at least while the functionality is experimental. Maybe we can move this to `traces.py`, or at least prefix it with an underscore.

Would also be good to add more context to the docstring so that users know when this can be used (i.e., only if `trace_lifecycle="stream"`).

sentrivana (commented):
TY for untangling the two 🙏🏻 

Left two small questions and a comment.

sentry_sdk/tracing_utils.py
```diff
@@ -1203,6 +1203,17 @@ def get_current_span(
...
+    scope: "Optional[sentry_sdk.Scope]" = None,
+) -> "Optional[StreamedSpan]":
+    """
+    Returns the currently active span if there is one running, otherwise `None`
+    """
+    scope = scope or sentry_sdk.get_current_scope()
+    current_span = scope.streamed_span
+    return current_span
+
+
```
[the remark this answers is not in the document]
alexander-alderman-webb: duplicate of https://github.com/getsentry/sentry-python/pull/6177#discussion_r3167478322

sentry_sdk/scope.py
```diff
@@ -893,6 +907,15 @@ def span(self, span: "Optional[Union[Span, StreamedSpan]]") -> None:
...
                     self._transaction_info["source"] = transaction.source
 
+    @property
+    def streamed_span(self) -> "Optional[StreamedSpan]":
+        """Get/set current tracing span."""
+        return self._span if isinstance(self._span, StreamedSpan) else None
+
+    @streamed_span.setter
+    def streamed_span(self, span: "Optional[StreamedSpan]") -> None:
+        self._span = span
```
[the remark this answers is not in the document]
alexander-alderman-webb: yes but only one mode is intended to be used

sentry_sdk/scope.py
alexander-alderman-webb: We would have to check `has_tracing_enabled(client.options)` in both the `if` and `elif` conditions to rely only on the fallback, so I tried to simplify by adding an early return. Previously there was no `elif` branch.

I don't mind either way. What do you think?

sentry_sdk/api.py
alexander-alderman-webb: Yes that's a great point, it would be best to not expose too much. I've limited the API surface to `_get_current_streamed_span` in `traces.py` and expanded the docstring

sentry_sdk/scope.py
sentrivana: Gotcha. No then it's okay this way.

sentrivana (approved):
🙏🏻

Issue #6175, opened 2026-04-29
Introduce a substitute value for AnnotatedValue.removed_because_over_size_limit and AnnotatedValue.removed_because_raw_data


ericapisani:
We generally try to avoid setting none or empty strings on span attributes because it doesn't provide any meaningful information to users and increases the size of the envelope.

In scenarios where, within an integration, we are setting the request body data on a span attribute, and that request data is removed/annotated for some reason, we should set a value to indicate the reason why that data was removed.

PR #6178, opened 2026-04-30
feat(wsgi,asgi): Introduce substitute values for filtered fields in span-streaming mode

ericapisani:
In the legacy (event-based) pipeline, filtered/removed values in request data are represented as `AnnotatedValue` objects — wrappers that carry metadata (`rem` remarks) consumed by Relay during ingestion. In span-streaming mode, request attributes are plain strings sent directly on spans, so `AnnotatedValue` is not appropriate there.

Previously, `_filter_headers` had a `use_annotated_value: bool` parameter that callers set to `False` when building span attributes. This was fragile — callers had to know the right value, and the logic was split between call sites instead of living in one place.

This PR:
- Adds two new substitute string constants (`OVER_SIZE_LIMIT_SUBSTITUTE`, `UNPARSABLE_RAW_DATA_SUBSTITUTE`) for use in the span-streaming path.
- Adds two new `AnnotatedValue` factory methods (`substituted_because_raw_data`, `substituted_because_over_size_limit`) that use remark type `"s"` (substituted) rather than `"x"` (removed), for future use as the SDK fully transitions to span-first.

Refs GH-6175
Fixes PY-2396


ericapisani:
bugbot run

ericapisani:
@sentry review

sentry_sdk/integrations/_wsgi_common.py
```diff
@@ -213,16 +214,18 @@ def _is_json_content_type(ct: "Optional[str]") -> bool:
...
-    else:
-        substitute = SENSITIVE_DATA_SUBSTITUTE
+    client_options = sentry_sdk.get_client().options
+    is_span_streaming_enabled = has_span_streaming_enabled(client_options)
+
+    substitute: "Union[AnnotatedValue, str]" = (
+        SENSITIVE_DATA_SUBSTITUTE
+        if is_span_streaming_enabled
+        else AnnotatedValue.removed_because_over_size_limit()
+    )
```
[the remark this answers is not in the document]
ericapisani: In span-first related code that would call this method, it was calling `filter_headers` with `use_annotated_value=False`, resulting in the string value (`SENSITIVE_DATA_SUBSTITUTE`) to be returned, and is set on the `http.request.header.<key>` span attribute (which in this case expects a string).

The legacy approach expects the annotated value, which is why it's gated this way.

edit: I realize I missed something, this is actually a valid issue

sentry_sdk/_types.py
```diff
@@ -13,6 +13,10 @@
 
 SENSITIVE_DATA_SUBSTITUTE = "[Filtered]"
 BLOB_DATA_SUBSTITUTE = "[Blob substitute]"
+OVER_SIZE_LIMIT_SUBSTITUTE = (
+    "[Value removed due to size of field exceeding configured maximum size.]"
+)
+UNPARSABLE_RAW_DATA_SUBSTITUTE = "[Value removed due to being unparsable.]"
```
sentrivana: These are maybe a tad too long, how about:

```suggestion
OVER_SIZE_LIMIT_SUBSTITUTE = "[Too large]"
UNPARSABLE_RAW_DATA_SUBSTITUTE = "[Unparsable]"
```

ericapisani: How does splitting it down the middle sound?

```
OVER_SIZE_LIMIT_SUBSTITUTE = "[Value removed: Exceeds maximum size]"
UNPARSABLE_RAW_DATA_SUBSTITUTE = "[Value removed: Unparsable]"
```

sentry_sdk/_types.py
```diff
@@ -6,13 +6,14 @@
...
 
-
 # Re-exported for compat, since code out there in the wild might use this variable.
 MYPY = TYPE_CHECKING
 
 
 SENSITIVE_DATA_SUBSTITUTE = "[Filtered]"
 BLOB_DATA_SUBSTITUTE = "[Blob substitute]"
+OVER_SIZE_LIMIT_SUBSTITUTE = "[Value removed: Exceeds maximum size]"
+UNPARSABLE_RAW_DATA_SUBSTITUTE = "[Value removed: Unparsable]"
```
sentrivana: I'd remove the `Value removed:`, think it's self explanatory that it's a substitute value

sentrivana (approved):
Left one more comment re: the substitute value, otherwise lgtm!

PR #6179, opened 2026-04-30
feat(aiohttp): Add span streaming support

### LINKED ISSUE #5996: Migrate aiohttp to span first

ericapisani:
Migrate the aiohttp integration to support the new span-first streaming model alongside the existing transaction-based approach.

Fixes PY-2298
Fixes #5996


ericapisani:
bugbot run

ericapisani:
@sentry review

sentry_sdk/integrations/aiohttp.py
```diff
@@ -121,38 +137,130 @@ async def sentry_app_handle(
...
+                        )
+
+                        header_dict: "dict[str, Any]" = {}
+                        for header, header_value in _filter_headers(
+                            headers, use_annotated_value=False
+                        ).items():
+                            header_dict[f"http.request.header.{header.lower()}"] = (
+                                # header_value will always be a string because we set `use_annotated_value` to false above
+                                header_value
+                            )
```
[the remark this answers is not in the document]
ericapisani: We do this on event payloads (legacy) and within the ASGI and WSGI integrations (streamed spans/span-first), this follows the new approach

sentry_sdk/integrations/aiohttp.py
```diff
@@ -121,38 +137,130 @@ async def sentry_app_handle(
...
+                        )
+
+                        client_address = (
+                            {
+                                "client.address": request.remote,
+                                "user.ip_address": request.remote,
+                            }
+                            if should_send_default_pii() and request.remote
+                            else {}
+                        )
```
ericapisani: This pattern was followed so that if there are no values in the query string or in the client address, we are not adding empty strings to the envelope.

sentry_sdk/integrations/aiohttp.py
```diff
@@ -121,38 +137,130 @@ async def sentry_app_handle(
...
+
+                                if (
+                                    e.status_code
+                                    in integration._failed_request_status_codes
+                                ):
+                                    _capture_exception()
+                                raise
+                            except (asyncio.CancelledError, ConnectionResetError):
+                                if isinstance(span, StreamedSpan):
+                                    span.status = SpanStatus.ERROR.value
```
ericapisani, line 233: Note that despite a `NoOpStreamedSpan` doing nothing when `status` is set on it, this is fine because ultimately these types of spans aren't sampled or are being ignored.

sentry_sdk/integrations/aiohttp.py
```diff
@@ -121,38 +137,130 @@ async def sentry_app_handle(
...
+                                with capture_internal_exceptions():
+                                    raw_data = get_aiohttp_request_data(request)
+                                    body_data = (
+                                        raw_data.value
+                                        if isinstance(raw_data, AnnotatedValue)
+                                        else raw_data
+                                    )
+                                    if body_data is not None:
+                                        span._segment.set_attribute(
+                                            "http.request.body.data", body_data
```
ericapisani, line 257: As this is currently written, there is a potential for this to be an empty string instead of a value - specifically in the case of when an `AnnotatedValue` was returned in `get_aiohttp_request_data`. 

There's a [pull request](https://github.com/getsentry/sentry-python/pull/6178) that's open in order to ensure that this gets a value set on it in those situations, which will also have the benefit of making it clearer to users on why a value was replaced (i.e. too large, raw data that was present)

sentry_sdk/integrations/aiohttp.py
```diff
@@ -232,19 +362,38 @@ async def on_request_start(
...
+            legacy_span = sentry_sdk.start_span(
+                op=OP.HTTP_CLIENT,
+                name=span_name,
+                origin=AioHttpIntegration.origin,
+            )
+            legacy_span.set_data(SPANDATA.HTTP_METHOD, method)
+            if parsed_url is not None:
+                legacy_span.set_data("url", parsed_url.url)
+                legacy_span.set_data(SPANDATA.HTTP_QUERY, parsed_url.query)
+                legacy_span.set_data(SPANDATA.HTTP_FRAGMENT, parsed_url.fragment)
```
ericapisani, line 390: These were not updated from `http` to `url` in order to preserve existing behaviour

sentry_sdk/_types.py
ericapisani, line 1: The changes here were added in from https://github.com/getsentry/sentry-python/pull/6178 so they could be used right away

sentry_sdk/integrations/aiohttp.py
```diff
@@ -194,10 +314,21 @@ async def sentry_urldispatcher_resolve(
                 pass
 
             if name is not None:
-                sentry_sdk.get_current_scope().set_transaction_name(
-                    name,
-                    source=SOURCE_FOR_STYLE[integration.transaction_style],
-                )
+                current_scope = sentry_sdk.get_current_scope()
+                current_span = current_scope.span
```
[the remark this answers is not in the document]
ericapisani: This is essentially what `get_current_span` does under the hood:

```
def get_current_span(
    scope: "Optional[sentry_sdk.Scope]" = None,
) -> "Optional[Union[Span, StreamedSpan]]":
    """
    Returns the currently active span if there is one running, otherwise `None`
    """
    scope = scope or sentry_sdk.get_current_scope()
    current_span = scope.span
    return current_span
```

I wonder if the bot is tripping over the return types or something? Switching to `get_current_span` is more concise, so I'll switch to that and see if that resolves this

sentry_sdk/integrations/aiohttp.py
```diff
@@ -121,38 +135,134 @@ async def sentry_app_handle(
...
+                                    # There are a number of sub-exceptions that should have an "ok" status,
+                                    # but will actually have a status of error once the `raise` happens below
+                                    # and we pass through the `should_be_treated_as_error` method invoked
+                                    # within a span's `__exit__` method.
+                                    #
+                                    # As a result, made this behaviour explicit
+                                    # so people don't think that the "ok" status persists in that scenario.
+                                    #
+                                    # Although not obvious, this overwriting behaviour occurs in the legacy
+                                    # approach as well, so this matches what we do today.
```
sentrivana: > There are a number of sub-exceptions that should have an "ok" status,
> but will actually have a status of error once the `raise` happens below
> and we pass through the `should_be_treated_as_error` method invoked
> within a span's `__exit__` method.

Would it be a lot of effort to fix this? I imagine we'd need to add some status code parsing to `should_be_treated_as_error`? Because the current situation def sounds like a bug (so we could also fix it for the legacy code path)

sentry_sdk/integrations/aiohttp.py
sentrivana: Yup this would be my preferred way to work with `StreamedSpan`s and `NoOpStreamedSpan`s -- just using the method and not caring what the actual class is, since the class itself will take care of it.

(At least whenever possible -- which is not always, especially if something else depends on the result of calling the method, or if something external needs to happen only if a span is sampled (like the span needing to be set on the scope).)

sentry_sdk/integrations/aiohttp.py
ericapisani: This is trickier than it seems, at the heart of the issue is how to deal with raised exceptions that are used for certain framework behaviours that aren't necessarily "errors". In this case, these `HTTPException`s are used to create the final HTTP response, whether they're 200s or 500s.

We could call `__end__` on the `StreamedSpan` manually to avoid this prior to the `raise`, but that would mean risking loss of data/spans if things happen afterwards.

We could also add status code parsing, as you had suggested, within `should_be_treated_as_error`, but I worry that there could be unintentional side effects. 

For instance, could the SDK could be running in a non-web framework application such that a custom status code meant to represent an error is a success HTTP status code (200, etc.) or an exit code that's not an error but is one in HTTP?

sentrivana: > For instance, could the SDK could be running in a non-web framework application such that a custom status code meant to represent an error is a success HTTP status code (200, etc.) or an exit code that's not an error but is one in HTTP?

I would check whether we're dealing with an instance of aiohttp's `HTTPException`, specifically, and only apply the status code logic then. That way it's only scoped to the one case we're aware of that needs this special handling.

sentrivana (approved):
Approving to not block, two things outstanding from my pov:
- Please consider https://github.com/getsentry/sentry-python/pull/6179/changes#r3186741314 -- not necessary as it's out of scope of the PR, but would be great to fix this if we already have eyes on it
- The `AnnotatedValue` changes from the other PR probably need merging first if I understand correctly (this PR has an old snapshot of the `AnnotatedValue` changes)

sentry_sdk/utils.py
```diff
@@ -27,6 +27,11 @@
     # Python 3.10 and below
     BaseExceptionGroup = None  # type: ignore
 
+try:
+    from aiohttp.web_exceptions import HTTPException as AIOHttpHttpException
+except ImportError:
+    AIOHttpHttpException = None
```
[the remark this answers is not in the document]
ericapisani: There's not really a good way around this because of how `aiohttp` works under the hood. We're considering this an acceptable tradeoff in order to ensure the SDK works for users of this integration.

sentry_sdk/utils.py
```diff
@@ -1986,6 +1991,12 @@ def should_be_treated_as_error(ty: "Any", value: "Any") -> bool:
         # https://docs.python.org/3/library/exceptions.html#SystemExit
         return False
 
+    # In the aiohttp integration, all of their HTTP responses are Exceptions.
+    # Because they have to be raised and handled by the framework, we need this check so
+    # that we don't accidentally overwrite a status of "ok" with "error" here.
+    if AIOHttpHttpException and isinstance(value, AIOHttpHttpException):
+        return False
```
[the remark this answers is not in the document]
ericapisani: See https://github.com/getsentry/sentry-python/pull/6179#discussion_r3188483245

sentry_sdk/utils.py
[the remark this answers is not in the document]
ericapisani: See https://github.com/getsentry/sentry-python/pull/6179#discussion_r3188483245

ericapisani:
@sentrivana Addressed your last comments in [d7b51c2](https://github.com/getsentry/sentry-python/pull/6179/commits/d7b51c23261e64530edb48a29ddf19e5ab1ec59a) 🚀

PR #6180, opened 2026-05-01
ref: Introduce inline type check for whether a span is StreamedSpan

### LINKED ISSUE #6182: Create helper method to replace frequent "is streamed span but not a noop streamed span" check
Quality of life refactor. We often have to check that a streamed span is also not a noop streamed span, and having a helper function do this would save a lot of code review comments from the bots that we missed the `not isinstance(span, NoOpStreamedSpan)` check.

ericapisani:
The `isinstance(span, StreamedSpan) and not isinstance(span, NoOpStreamedSpan)` pattern was repeated in multiple places. Omitting the `NoOpStreamedSpan` guard is a frequent code review catch — it silently treats unsampled spans as sampled.

Introduce `_is_streamed_span()` in `traces.py` with a `TypeGuard[StreamedSpan]` return type for proper type narrowing, and replace all combined-check instances in `scope.py`, `starlette.py`, and `fastapi.py`.

Locations where `_span` can also be a regular `Span` (not just `StreamedSpan`) were intentionally left unchanged, as the semantics differ there.

Fixes PY-2397
Fixes #6182


ericapisani:
bugbot run

ericapisani:
@sentry review

alexander-alderman-webb (commented):
It looks like in blocks gated by `not isinstance(current_span, NoOpStreamedSpan)` we rely on private attributes and functions of `StreamedSpan` such as `_segment` and `_is_segment()`.

If we're relying on private details that we expect to only exist on `StreamedSpan` (and not subclasses) it would be more appropriate to replace logic like

```python
isinstance(span, StreamedSpan) and not isinstance(span, NoOpStreamedSpan)
```

with 

```python
type(span) is StreamedSpan
```

sentrivana (commented):
Wondering if it'd be better to replace 

```python
isinstance(span, StreamedSpan) and not isinstance(span, NoOpStreamedSpan)
```

with

```python
isinstance(span, StreamedSpan) and span.sampled
```

But I assume that might cause mypy to complain about nonexistent methods on `NoOpStreamedSpan`s since it won't be able to realize the condition effectively means it can't be a `NoOpStreamedSpan`?

Edit: Only saw Alex's [comment](https://github.com/getsentry/sentry-python/pull/6180#pullrequestreview-4218429328) above now. That also sounds good.

ericapisani:
Re: using the `type` approach rather than `isinstance` (i.e.: `type(span) is StreamedSpan`)

I think it depends on how strict we want to be with ensuring that the only spans that pass this are `StreamedSpan` classes specifically.

I wrote this check with `isinstance` in case there are additional children introduced in the future of `StreamedSpan` that we would want to handle similar to their parent, and that we ultimately didn't want to filter out with this guard - using `type` would unfortunately cause that to happen.

Re: using the `sampled` property:

> But I assume that might cause mypy to complain about nonexistent methods on NoOpStreamedSpans since it won't be able to realize the condition effectively means it can't be a NoOpStreamedSpan?

We would achieve the same outcome using the `sampled` property as is written here because it exists on both the `StreamedSpan` and the `NoOpStreamedSpan`, but I think it doesn't make the conditional as explicitly clear that it's more about removing any instances of `NoOpStreamedSpan` due to the reason that you mentioned in that they lack the methods that `StreamedSpan` has.

alexander-alderman-webb:
> I think it depends on how strict we want to be with ensuring that the only spans that pass this are StreamedSpan classes specifically.

The spans touched by the filtering logic are created by the SDK, so we know that they are of type `StreamedSpan` and not a subclass other than `NoOpStreamedSpan`. If we introduce other subclasses in the future, I think we have full flexibility to change 😄.

ericapisani:
@alexander-alderman-webb @sentrivana Made the change in [203c816](https://github.com/getsentry/sentry-python/pull/6180/commits/203c816c66856a89d483fb77de71a1b55e2d7e85)

sentry_sdk/integrations/fastapi.py
```diff
@@ -95,9 +95,7 @@ def _sentry_call(*args: "Any", **kwargs: "Any") -> "Any":
                 if has_span_streaming_enabled(client.options):
                     current_span = current_scope.streamed_span
 
-                    if isinstance(current_span, StreamedSpan) and not isinstance(
-                        current_span, NoOpStreamedSpan
-                    ):
+                    if _is_streamed_span(current_span):
```
alexander-alderman-webb: Let's inline the method now since it's only one condition. Removes one indirection.

```suggestion
                    if type(span) is StreamedSpan:
```

sentry_sdk/integrations/starlette.py
```diff
@@ -255,12 +255,7 @@ def _set_request_body_data_on_streaming_segment(
     info: "Optional[Dict[str, Any]]",
 ) -> None:
     current_span = _get_current_streamed_span()
-    if (
-        info
-        and "data" in info
-        and isinstance(current_span, StreamedSpan)
-        and not isinstance(current_span, NoOpStreamedSpan)
-    ):
+    if info and "data" in info and _is_streamed_span(current_span):
```
alexander-alderman-webb:
```suggestion
    if info and "data" in info and type(current_span) is StreamedSpan:
```

sentry_sdk/integrations/starlette.py
```diff
@@ -557,9 +552,7 @@ def _sentry_sync_func(*args: "Any", **kwargs: "Any") -> "Any":
                 if span_streaming:
                     current_span = current_scope.streamed_span
 
-                    if isinstance(current_span, StreamedSpan) and not isinstance(
-                        current_span, NoOpStreamedSpan
-                    ):
+                    if _is_streamed_span(current_span):
```
alexander-alderman-webb:
```suggestion
                    if type(current_span) is StreamedSpan:
```

sentry_sdk/scope.py
```diff
@@ -585,11 +588,7 @@ def get_traceparent(self, *args: "Any", **kwargs: "Any") -> "Optional[str]":
 
         span_streaming = has_span_streaming_enabled(client.options)
         # If we have an active span, return traceparent from there
-        if (
-            span_streaming
-            and self.streamed_span is not None
-            and not isinstance(self.streamed_span, NoOpStreamedSpan)
-        ):
+        if span_streaming and _is_streamed_span(self.streamed_span):
```
alexander-alderman-webb:
```suggestion
        if type(self.streamed_span) is StreamedSpan:
```

sentry_sdk/scope.py
```diff
@@ -609,11 +608,7 @@ def get_baggage(self, *args: "Any", **kwargs: "Any") -> "Optional[Baggage]":
 
         span_streaming = has_span_streaming_enabled(client.options)
         # If we have an active span, return baggage from there
-        if (
-            span_streaming
-            and self.streamed_span is not None
-            and not isinstance(self.streamed_span, NoOpStreamedSpan)
-        ):
+        if span_streaming and _is_streamed_span(self.streamed_span):
```
alexander-alderman-webb:
```suggestion
        if type(self.streamed_span) is StreamedSpan:
```

sentry_sdk/scope.py
```diff
@@ -918,11 +913,7 @@ def streamed_span(self, span: "Optional[StreamedSpan]") -> None:
 
         # Also set _transaction and _transaction_info in streaming mode as this
         # is used for populating events and linking them to segments
-        if (
-            isinstance(span, StreamedSpan)
-            and not isinstance(span, NoOpStreamedSpan)
-            and span._is_segment()
-        ):
+        if _is_streamed_span(span) and span._is_segment():
```
alexander-alderman-webb:
```suggestion
       if type(self.streamed_span) is StreamedSpan:
```

alexander-alderman-webb (changes requested):

alexander-alderman-webb (approved):
Code LGTM! Change the title when you merge 🙏

PR #6181, opened 2026-05-01
feat(transport): Add EnvelopePrinterTransport for debug logging

### LINKED ISSUE #6183: Add transport for pretty-printing envelope contents when debugging span-first logic
Quality of life feature. 

It's currently annoying to pull and log the contents of an envelope, so having a transport that handles this for us that's easily toggled would make life easier for developers working in the span-first logic.

ericapisani:
Adds `EnvelopePrinterTransport`, a decorator transport that wraps the real transport and pretty-prints each envelope's headers and item payloads to the SDK debug logger before forwarding the envelope.

Enabled via `SENTRY_PRINT_ENVELOPES=1` (also accepts `true`/`yes`). When unset or falsy, no wrapping occurs and there's no runtime cost. Useful for local debugging without having to run a local Sentry instance or intercept network traffic.

Fixes PY-2398
Fixes #6183


ericapisani:
bugbot run

ericapisani:
@sentry review

ericapisani:
bugbot run

ericapisani:
@sentry review

sentry_sdk/transport.py
```diff
@@ -1081,6 +1083,74 @@ def _make_pool(
...
+    """Wraps another transport, printing envelope contents to the SDK debug logger before sending."""
+
+    def __init__(self, transport: "Transport") -> None:
+        Transport.__init__(self, options=transport.options)
+        self._inner = transport
+        self.parsed_dsn = transport.parsed_dsn
+
+    @property  # type: ignore[misc]
+    def __class__(self) -> type:
+        return self._inner.__class__
```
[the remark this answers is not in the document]
ericapisani: Here's an example of this approach working in the Python repl:

```
class Foo():
...     def __init__(self):
...         pass
...
>>> class Wrap(Foo):
...     def __init__(self, wrapped):
...         self._inner = wrapped
...     @property
...     def __class__(self):
...         return self._inner.__class__
...
>>> class Bar(Foo):
...     def __init__(self):
...         pass
...
>>> b = Bar()
>>> w = Wrap(b)
>>> isinstance(w, Bar)
True
```
and a script using the `EnvelopePrinterTransport` and `AsyncHttpTransport` classes directly
[demo_envelope_printer.py](https://github.com/user-attachments/files/27286537/demo_envelope_printer.py)

alexander-alderman-webb (changes requested):
The snippet below should probably work without crashing, since that's the canonical way to provide an alternative transport.

Maybe we create a transport which only prints to the terminal without wrapping another transport?

```python
import sentry_sdk
from sentry_sdk.transport import _EnvelopePrinterTransport
sentry_sdk.init(
    dsn="...",
    transport=_EnvelopePrinterTransport,
)
```

ericapisani:
@alexander-alderman-webb Given this transport is for debugging purposes on our end, I opted for the wrapping approach so that we don't need to change how we've configured the SDK, which was intended to include specific transports we may be leveraging for our use case.

Is there something in particular that you're concerned about with this approach?

alexander-alderman-webb:
My main concern is that we need to set `debug=True` to see the envelopes with this approach (otherwise the logs are filtered out).
A smaller concern is that we will have to update the transport when we add new methods to the interface defined by the `Transport` base class. Ideally the new transport would be no maintenance effort at all 😄.

I agree that wrapping would be the nicest approach but I don't think our current abstractions were designed for this kind of wrapping 😞.

sentry_sdk/transport.py
```diff
@@ -1081,6 +1083,74 @@ def _make_pool(
...
+
+    def capture_envelope(self, envelope: "Envelope") -> None:
+        try:
+            logger.debug("--- Sentry Envelope ---")
```
alexander-alderman-webb: As discussed update to info so they're printed without `debug=True`

```suggestion
            logger.info("--- Sentry Envelope ---")
```

tests/test_transport.py
```diff
@@ -1041,3 +1045,329 @@ async def send_message(i):
     # New request should be dropped due to rate limiting
     assert len(capturing_server.captured) == 0
     await client.close_async(timeout=2.0)
+
+
+# --- _EnvelopePrinterTransport tests ---
+
+
+class FakeTransport(Transport):
```
alexander-alderman-webb: Let's remove tests to reduce maintenance burden (and there's no cost of this breaking).

sentry_sdk/transport.py
```diff
@@ -1081,6 +1083,74 @@ def _make_pool(
             return httpcore.ConnectionPool(**opts)
 
 
+class _EnvelopePrinterTransport(Transport):
+    """Wraps another transport, printing envelope contents to the SDK debug logger before sending."""
+
+    def __init__(self, transport: "Transport") -> None:
+        Transport.__init__(self, options=transport.options)
+        self._inner = transport
+        self.parsed_dsn = transport.parsed_dsn
```
alexander-alderman-webb, line 1093:
```suggestion
    def __init__(self, transport: "Transport") -> None:
        Transport.__init__(self, options=transport.options)
        self._inner = transport
        self.parsed_dsn = transport.parsed_dsn

        self.envelope_logger = logging.getLogger("sentry_sdk.envelopes")
        self.envelope_logger.setLevel(logging.DEBUG)
```

tests/test_transport.py
```diff
@@ -30,23 +31,22 @@
 import sentry_sdk
 from sentry_sdk import (
     Client,
+    Hub,
     add_breadcrumb,
     capture_message,
-    isolation_scope,
     get_isolation_scope,
-    Hub,
+    isolation_scope,
```
alexander-alderman-webb, line 38: Probably best to remove the changes in `test_transport.py` since we're not adding any tests. Although it's only import ordering so this is a nitpick.

sentry_sdk/transport.py
```diff
@@ -1147,8 +1227,10 @@ def make_transport(options: "Dict[str, Any]") -> "Optional[Transport]":
             "You tried to use AsyncHttpTransport but don't have httpcore[asyncio] installed. Falling back to sync transport."
         )
 
+    transport = None  # type: Optional[Transport]
```
alexander-alderman-webb: Since we're not using the old style annotations anymore (and they don't work well with modern typing tooling).

```suggestion
    transport: "Optional[Transport]" = None
```

alexander-alderman-webb (approved):
Looks good! I ran a simple example and can confirm that envelopes appear in the terminal.

Thanks for making the changes!

sentry_sdk/transport.py
ericapisani: Addressed in [aeb6830](https://github.com/getsentry/sentry-python/pull/6181/commits/aeb68304a9965e66535575c24c941879fdf9f86b)

PR #6199, opened 2026-05-05
feat(integrations): pass along the conversation ID for openai `responses` calls

constantinius:
Closes https://linear.app/getsentry/issue/TET-2287/see-if-we-can-auto-extract-conversationid-from-openai-python

### DIFF
--- a/tests/integrations/openai/test_openai.py
+++ b/tests/integrations/openai/test_openai.py
@@ -2843,11 +2843,6 @@ def test_ai_client_span_responses_api(
         pytest.param(None, None, id="none"),
         pytest.param("conv_abc123", "conv_abc123", id="string"),
         pytest.param({"id": "conv_abc123"}, "conv_abc123", id="dict"),
-        pytest.param(
-            mock.Mock(spec=["id"], id="conv_abc123"),
-            "conv_abc123",
-            id="object_with_id_attr",
-        ),
     ],
 )
 @pytest.mark.skipif(SKIP_RESPONSES_TESTS, reason="Responses API not available")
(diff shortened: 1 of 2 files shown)


tests/integrations/openai/test_openai.py
```diff
@@ -2836,6 +2836,49 @@ def test_ai_client_span_responses_api(
...
+    [
+        pytest.param(omit, None, id="omit"),
+        pytest.param(None, None, id="none"),
+        pytest.param("conv_abc123", "conv_abc123", id="string"),
+        pytest.param({"id": "conv_abc123"}, "conv_abc123", id="dict"),
+        pytest.param(
+            mock.Mock(spec=["id"], id="conv_abc123"),
+            "conv_abc123",
+            id="object_with_id_attr",
+        ),
```
alexander-alderman-webb: When is an object with the `id` property accepted?

I can only find strings and dictionaries in the accepted type: https://github.com/openai/openai-python/blob/acd0c54d8a68efeedde0e5b4e6c310eef1ce7867/src/openai/types/responses/response_create_params.py#L295

```python
from openai import OpenAI

client = OpenAI(api_key="test")

class MyConversation:
    def __init__(self, id):
        self.id = id

client.responses.create(
    model="gpt-4o",
    input="hello",
    conversation=MyConversation("conv_abc123"),
)
```

constantinius: Yes, you're correct, I was confused about the `ResponseConversationParam` but that's actually a TypedDict

constantinius: Resolved with [9f88ee7](https://github.com/getsentry/sentry-python/pull/6199/commits/9f88ee7150cb576e38119b2ab3b1a92806dc77b9)

alexander-alderman-webb (approved):

PR #6205, opened 2026-05-05
feat(integrations): add support for the `litellm` `responses`/`aresponses` APIs

constantinius:
### Description

Adds support for `responses` and `aresponses`, and their differences in output tracking. Also checking the conversation ID if it is passed in the extra_args.

Contributes to https://linear.app/getsentry/issue/TET-2287/see-if-we-can-auto-extract-conversationid-from-openai-python

### DIFF
--- a/sentry_sdk/integrations/litellm.py
+++ b/sentry_sdk/integrations/litellm.py
@@ -40,6 +40,23 @@ def _get_metadata_dict(kwargs: "Dict[str, Any]") -> "Dict[str, Any]":
     return metadata
 
 
+def _read_usage_field(usage: "Any", *names: str) -> "Optional[int]":
+    """Read the first non-None field from a usage container.
+
+    The usage object can be either a typed Pydantic model (attribute access) or
+    a plain dict (litellm hands us a dict for the assembled async-streaming
+    response), so we try both shapes.
+    """
+    for name in names:
+        if isinstance(usage, dict):
+            value = usage.get(name)
+        else:
+            value = getattr(usage, name, None)
+        if value is not None:
+            return value
+    return None
+
+
 def _convert_message_parts(messages: "List[Dict[str, Any]]") -> "List[Dict[str, Any]]":
     """
     Convert the message parts from OpenAI format to the `gen_ai.request.messages` format
@@ -313,16 +330,18 @@ def _success_callback(
         #   Responses API (non-streaming): input_tokens  / output_tokens
         #   Responses API (streaming):     prompt_tokens / completion_tokens
         #     (litellm normalizes to chat-completion names when assembling the
-        #      streaming response).
+        #      streaming response). For the async-streaming variant, the
+        #      assembled `usage` is a plain dict, not a Pydantic model — hence
+        #      `_read_usage_field` supports both shapes.
         if hasattr(response, "usage"):
             usage = response.usage
             record_token_usage(
                 span,
-                input_tokens=getattr(usage, "prompt_tokens", None)
-                or getattr(usage, "input_tokens", None),
-                output_tokens=getattr(usage, "completion_tokens", None)
-                or getattr(usage, "output_tokens", None),
-                total_tokens=getattr(usage, "total_tokens", None),
+                input_tokens=_read_usage_field(usage, "prompt_tokens", "input_tokens"),
+                output_tokens=_read_usage_field(
+                    usage, "completion_tokens", "output_tokens"
+                ),
+                total_tokens=_read_usage_field(usage, "total_tokens"),
             )
 
     finally:
--- a/sentry_sdk/integrations/litellm.py
+++ b/sentry_sdk/integrations/litellm.py
@@ -6,6 +6,7 @@ from sentry_sdk import consts
 from sentry_sdk.ai.monitoring import record_token_usage
 from sentry_sdk.ai.utils import (
     get_start_span_function,
+    normalize_message_roles,
     set_data_normalized,
     truncate_and_annotate_messages,
     transform_openai_content_part,
@@ -17,7 +18,7 @@ from sentry_sdk.scope import should_send_default_pii
 from sentry_sdk.utils import event_from_exception
 
 if TYPE_CHECKING:
-    from typing import Any, Dict, List
+    from typing import Any, Dict, List, Optional
     from datetime import datetime
 
 try:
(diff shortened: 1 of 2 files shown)


tests/integrations/litellm/test_litellm.py
```diff
@@ -2135,3 +2135,261 @@ def test_convert_message_parts_image_url_missing_url():
...
+    )
+    events = capture_events()
+
+    kwargs = _build_responses_kwargs(extra_body_conversation=conversation)
+
+    with start_transaction(name="litellm test"):
+        _input_callback(kwargs)
+        _success_callback(
+            kwargs, MockResponsesResponse(), datetime.now(), datetime.now()
+        )
```
alexander-alderman-webb: With SDK tests we aim to verify that we generate some telemetry based on the user's interaction with the library. We want to assert the presence of telemetry if the patched library is used as the user would use the library.

Currently, we assert that telemetry is generated if `_input_callback` and `_success_callback` are each invoked exactly once.

This is _not_ always the case, and the assumption has resulted in unhandled SDK exceptions that were fixed in the commit below:

https://github.com/getsentry/sentry-python/commit/96ebbf67d91146ebb30be6e0898a637ea6c2c697

tests/integrations/litellm/test_litellm.py
```diff
@@ -2135,3 +2135,261 @@ def test_convert_message_parts_image_url_missing_url():
...
+    def __init__(
+        self,
+        model="gpt-4.1-nano",
+        output=None,
+        usage=None,
+    ):
+        self.id = "resp-test"
+        self.model = model
+        self.output = output or [MockResponsesOutputMessage("the model response")]
+        self.usage = usage or MockResponsesUsage()
```
alexander-alderman-webb: Related to https://github.com/getsentry/sentry-python/pull/6205/changes#r3201008608, we should aim to avoid custom types in our test suites.

As soon as we introduce custom types our tests are not coupled to the concrete types used in a library, and the tests no longer verify the SDK contract (namely, that telemetry is generated when a library is used like a user would interact with the library).

We can't hit real LLM APIs in the tests but we can do the next best thing: couple the sample response to the types in the library and patch at the lowest possible level.

This is done most of the tests in this test file, and there are helpers in the repo to accomplish writing effective tests (such as [get_model_response()](https://github.com/getsentry/sentry-python/blob/master/tests/conftest.py#L1111)).

sentry_sdk/integrations/litellm.py
```diff
@@ -213,15 +292,56 @@ def _success_callback(
...
                 span,
-                input_tokens=getattr(usage, "prompt_tokens", None),
-                output_tokens=getattr(usage, "completion_tokens", None),
-                total_tokens=getattr(usage, "total_tokens", None),
+                input_tokens=_read_usage_field(usage, "prompt_tokens", "input_tokens"),
+                output_tokens=_read_usage_field(
+                    usage, "completion_tokens", "output_tokens"
+                ),
+                total_tokens=_read_usage_field(usage, "total_tokens"),
             )
```
alexander-alderman-webb: We already probe above to determine which API is used.

As a result, reading `prompt_tokens` or `input_tokens` is dead code conditioned on knowing which API you are handling (adding cognitive overhead when reading).

sentry_sdk/integrations/litellm.py
```diff
@@ -213,15 +292,56 @@ def _success_callback(
                     set_data_normalized(
                         span, SPANDATA.GEN_AI_RESPONSE_TEXT, response_messages
                     )
+            elif hasattr(response, "output"):
```
alexander-alderman-webb: You are adding code here which runs for all possible types of object that have an `output` field.

As a result the branch can easily be accidentally triggered as `litellm` evolves. There are multiple approaches to narrow down if you have a response in the Chat Completion API schema or a response in the Responses API schema. For example, you can check

```python
isinstance(response, (ResponsesAPIResponse, BaseResponsesAPIStreamingIterator))
```

based on the signature of the library function

https://github.com/BerriAI/litellm/blob/a67b7a7e87f11bed01f9e073125a7f8f180105a2/litellm/responses/main.py#L449.

sentry_sdk/integrations/litellm.py
```diff
@@ -129,11 +149,50 @@ def _input_callback(kwargs: "Dict[str, Any]") -> None:
...
+            responses_input = kwargs.get("input")
+            if responses_input:
+                if isinstance(responses_input, str):
+                    input_messages = [responses_input]
+                else:
+                    input_messages = list(responses_input)
+                normalized = normalize_message_roles(input_messages)  # type: ignore[arg-type]
+                messages_data = truncate_and_annotate_messages(normalized, span, scope)
+                if messages_data is not None:
+                    set_data_normalized(
```
alexander-alderman-webb: Based on the marshaling above you know that `messages_data` is a list. You should just use `span.set_data()` when you know the type of an attribute (again, removing cognitive overhead by avoiding dead code).

sentry_sdk/integrations/litellm.py
```diff
@@ -39,6 +40,23 @@ def _get_metadata_dict(kwargs: "Dict[str, Any]") -> "Dict[str, Any]":
     return metadata
 
 
+def _read_usage_field(usage: "Any", *names: str) -> "Optional[int]":
+    """Read the first non-None field from a usage container.
+
+    The usage object can be either a typed Pydantic model (attribute access) or
+    a plain dict (litellm hands us a dict for the assembled async-streaming
+    response), so we try both shapes.
```
alexander-alderman-webb: Why don't we just read from the dictionary int he asynchronous streaming scenario and otherwise access the attribute on the Pydantic model 😄 ?

These responses have types, so an `isinstance` check can tell you which branch you are in.

In the end we're developing against a library with a finite number of return types, and we should just check which case we are handling instead of probing around. Probing around is less robust, since new return types accidentally trigger `hasattr()` checks.

sentry_sdk/integrations/litellm.py
```diff
@@ -213,15 +292,56 @@ def _success_callback(
...
+                    if output_type == "function_call":
+                        if hasattr(output, "model_dump"):
+                            tool_calls.append(output.model_dump())
+                        elif hasattr(output, "dict"):
+                            tool_calls.append(output.dict())
+                    elif output_type == "message":
+                        for content_item in getattr(output, "content", []) or []:
+                            text = getattr(content_item, "text", None)
+                            if text is not None:
+                                output_text.append(text)
```
alexander-alderman-webb: This has reached **a lot** of indentation for Python code. Usually you can keep code readable by adding early returns or breaking up into functions where appropriate.

alexander-alderman-webb (changes requested):

sentry_sdk/integrations/litellm.py
constantinius: I don't understand, we should stringify lists, right? That is what `set_data_normalized` does. Otherwise I'd have to pull in `json` for that reason alone and stringify myself. So I would actually opt to keep it here.

sentry_sdk/integrations/litellm.py
constantinius: I'm fine with using the `isinstance` check in the future. I just see that the `hasattr` checks are way more prevalent throughout the SDK, but maybe that is something that will be changed.

alexander-alderman-webb:
Closing because our contribution guidelines require an issue in `sentry-python` before a PR is opened:

https://github.com/getsentry/sentry-python/blob/master/CONTRIBUTING.md

PR #6206, opened 2026-05-05
feat(tornado): Support span streaming

### LINKED ISSUE #6067: Migrate tornado to span first

sl0thentr0py:
#### Issues

* resolves: #6067
* resolves: PY-2369

### DIFF
--- a/sentry_sdk/integrations/tornado.py
+++ b/sentry_sdk/integrations/tornado.py
@@ -10,15 +10,17 @@ from sentry_sdk.integrations._wsgi_common import (
     RequestExtractor,
     _filter_headers,
     _is_json_content_type,
+    request_body_within_bounds,
 )
 from sentry_sdk.integrations.logging import ignore_logger
 from sentry_sdk.scope import should_send_default_pii
-from sentry_sdk.traces import SegmentSource, StreamedSpan
+from sentry_sdk.traces import NoOpStreamedSpan, SegmentSource, StreamedSpan
 from sentry_sdk.tracing import TransactionSource
 from sentry_sdk.tracing_utils import has_span_streaming_enabled
 from sentry_sdk.utils import (
     CONTEXTVARS_ERROR_MESSAGE,
     HAS_REAL_CONTEXTVARS,
+    AnnotatedValue,
     capture_internal_exceptions,
     ensure_integration_enabled,
     event_from_exception,
@@ -155,14 +157,16 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
             )
 
         with span_ctx as span:
-            if isinstance(span, StreamedSpan):
+            if isinstance(span, StreamedSpan) and not isinstance(
+                span, NoOpStreamedSpan
+            ):
                 with capture_internal_exceptions():
                     for attr, value in _get_request_attributes(self.request).items():
                         span.set_attribute(attr, value)
 
                     method = getattr(self, self.request.method.lower(), None)
                     if method is not None:
-                        span_name = transaction_from_function(method) or ""
+                        span_name = transaction_from_function(method)
                         if span_name:
                             span.name = span_name
                             span.set_attribute(
@@ -173,7 +177,9 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
             try:
                 yield
             finally:
-                if isinstance(span, StreamedSpan):
+                if isinstance(span, StreamedSpan) and not isinstance(
+                    span, NoOpStreamedSpan
+                ):
                     with capture_internal_exceptions():
                         status_int = self.get_status()
                         span.set_attribute(SPANDATA.HTTP_STATUS_CODE, status_int)
@@ -188,7 +194,7 @@ def _get_request_attributes(request: "Any") -> "Dict[str, Any]":
 
     headers = _filter_headers(dict(request.headers), use_annotated_value=False)
     for header, value in headers.items():
-        attributes[f"http.request.header.{header.lower()}"] = value
+        attributes[f"{SPANDATA.HTTP_REQUEST_HEADER}.{header.lower()}"] = value
 
     if request.query:
         attributes[SPANDATA.HTTP_QUERY] = request.query
@@ -200,13 +206,32 @@ def _get_request_attributes(request: "Any") -> "Dict[str, Any]":
     )
 
     if request.protocol:
-        attributes["network.protocol.name"] = request.protocol
+        attributes[SPANDATA.NETWORK_PROTOCOL_NAME] = request.protocol
 
     if should_send_default_pii() and request.remote_ip:
-        attributes["client.address"] = request.remote_ip
-        attributes["user.ip_address"] = request.remote_ip
+        attributes[SPANDATA.CLIENT_ADDRESS] = request.remote_ip
+        attributes[SPANDATA.USER_IP_ADDRESS] = request.remote_ip
+
+    with capture_internal_exceptions():
+        raw_data = _get_tornado_request_data(request)
+        body_data = raw_data.value if isinstance(raw_data, AnnotatedValue) else raw_data
+        if body_data is not None:
+            attributes[SPANDATA.HTTP_REQUEST_BODY_DATA] = body_data
 
     return attributes
+
+
+def _get_tornado_request_data(
+    request: "Any",
+) -> "Union[Optional[str], AnnotatedValue]":
+    body = request.body
+    if not body:
+        return None
+
+    if not request_body_within_bounds(sentry_sdk.get_client(), len(body)):
+        return AnnotatedValue.substituted_because_over_size_limit()
+
+    return body.decode("utf-8", "replace")
 
 
 @ensure_integration_enabled(TornadoIntegration)
--- a/sentry_sdk/integrations/tornado.py
+++ b/sentry_sdk/integrations/tornado.py
@@ -4,7 +4,7 @@ from inspect import iscoroutinefunction
 
 import sentry_sdk
 from sentry_sdk.api import continue_trace
-from sentry_sdk.consts import OP
+from sentry_sdk.consts import OP, SPANDATA
 from sentry_sdk.integrations import DidNotEnable, Integration, _check_minimum_version
 from sentry_sdk.integrations._wsgi_common import (
     RequestExtractor,
@@ -13,7 +13,9 @@ from sentry_sdk.integrations._wsgi_common import (
 )
 from sentry_sdk.integrations.logging import ignore_logger
 from sentry_sdk.scope import should_send_default_pii
+from sentry_sdk.traces import SegmentSource, StreamedSpan
 from sentry_sdk.tracing import TransactionSource
+from sentry_sdk.tracing_utils import has_span_streaming_enabled
 from sentry_sdk.utils import (
     CONTEXTVARS_ERROR_MESSAGE,
     HAS_REAL_CONTEXTVARS,
@@ -33,9 +35,10 @@ except ImportError:
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Any, Callable, Dict, Generator, Optional
+    from typing import Any, Callable, ContextManager, Dict, Generator, Optional, Union
 
     from sentry_sdk._types import Event, EventProcessor
+    from sentry_sdk.tracing import Span
 
 
 class TornadoIntegration(Integration):
@@ -97,6 +100,9 @@ class TornadoIntegration(Integration):
         RequestHandler.log_exception = sentry_log_exception
 
 
+_DEFAULT_TRANSACTION_NAME = "generic Tornado request"
+
+
 @contextlib.contextmanager
 def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]":
     integration = sentry_sdk.get_client().get_integration(TornadoIntegration)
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/tornado.py
```diff
@@ -118,22 +129,90 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
+            if isinstance(span, StreamedSpan):
+                with capture_internal_exceptions():
+                    for attr, value in _get_request_attributes(self.request).items():
+                        span.set_attribute(attr, value)
+
+                    method = getattr(self, self.request.method.lower(), None)
+                    if method is not None:
+                        tx_name = transaction_from_function(method) or ""
+                        if tx_name:
+                            span.name = tx_name
```
alexander-alderman-webb: Fallback is dead code and I'm not sure if we want transaction concepts to bleed into the streaming path when it comes to variable naming.

```suggestion
                        span_name = transaction_from_function(method)
                        if span_name:
                            span.name = span_name
```

alexander-alderman-webb (approved):
Comment not blocking, looks good to me!

tests/integrations/tornado/test_tornado.py
```diff
@@ -104,86 +104,152 @@ def test_basic(tornado_testcase, sentry_init, capture_events):
...
+        # client tx + server segment span
+        assert len(spans) >= 2
+        server_segment = next(
+            s for s in spans if s["attributes"].get("sentry.op") == "http.server"
+        )
+        client_segment = next(
+            s
+            for s in spans
+            if s["attributes"].get("sentry.op") != "http.server" and s.get("is_segment")
         )
```
sentrivana: Can we unfurl the spans directly without this filtering? In the legacy path, we implicitly assert on the number and order of spans by assigning them directly; would be good to preserve this.

The clanker loves to do this sort of filtering instead for some reason, but it does get it if you tell it to unfurl the items directly.

tests/integrations/tornado/test_tornado.py
```diff
@@ -438,15 +504,35 @@ def test_error_has_existing_trace_context_performance_disabled(
...
-    (_, event) = events
+    sentry_sdk.flush()
 
-    assert event["contexts"]["trace"]["origin"] == "auto.http.tornado"
+    if span_streaming:
+        spans = [i.payload for i in items]
+        segment = next(
+            s for s in spans if s["attributes"].get("sentry.op") == "http.server"
+        )
+        assert segment["attributes"]["sentry.origin"] == "auto.http.tornado"
```
sentrivana: Same comment re: unfurling here

sentry_sdk/integrations/tornado.py
```diff
@@ -118,22 +129,90 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
+                yield
+            finally:
+                if isinstance(span, StreamedSpan):
+                    with capture_internal_exceptions():
+                        status_int = self.get_status()
+                        span.set_attribute(SPANDATA.HTTP_STATUS_CODE, status_int)
+                        span.status = "error" if status_int >= 400 else "ok"
+
+
+def _get_request_attributes(request: "Any") -> "Dict[str, Any]":
```
sentrivana, line 187: Are we still capturing the request body in the streaming path? See https://getsentry.github.io/sentry-conventions/attributes/http/#http-request-body-data

For prior art IIRC @ericapisani did this for another integration, not sure which exactly

sentry_sdk/integrations/tornado.py
```diff
@@ -110,6 +119,8 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
         return
 
     weak_handler = weakref.ref(self)
+    client = sentry_sdk.get_client()
+    span_streaming = has_span_streaming_enabled(client.options)
```
ericapisani: Because this is communicating a state rather than a "thing", would suggest a name like `is_span_streaming_enabled`

sentry_sdk/integrations/tornado.py
```diff
@@ -118,22 +129,90 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
+                        span.set_attribute(attr, value)
+
+                    method = getattr(self, self.request.method.lower(), None)
+                    if method is not None:
+                        tx_name = transaction_from_function(method) or ""
+                        if tx_name:
+                            span.name = tx_name
+                            span.set_attribute(
+                                "sentry.span.source",
+                                SegmentSource.COMPONENT.value,
```
ericapisani: 🧹 : both approaches work, but I noticed that this spot uses `.value` and line [143](https://github.com/getsentry/sentry-python/pull/6206/changes#diff-0d5064b36068f4bf2c2a4548c9977e3bed85aed9ead93ae926d29a06318237d2R143) doesn't. As a clean up, would be good to make this consistent.

sentry_sdk/integrations/tornado.py
ericapisani: [starlette](https://github.com/getsentry/sentry-python/blob/bc380bfcf0c7df3ead8c8f5e942937c4f974d91b/sentry_sdk/integrations/starlette.py#L261-L264) and [aiohttp](https://github.com/getsentry/sentry-python/blob/bc380bfcf0c7df3ead8c8f5e942937c4f974d91b/sentry_sdk/integrations/aiohttp.py#L249-L258) were the couple that I worked on that had this

sentry_sdk/integrations/tornado.py
```diff
@@ -118,22 +129,90 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
+            scope.set_custom_sampling_context({"tornado_request": self.request})
+
+            span_ctx = sentry_sdk.traces.start_span(
+                name=_DEFAULT_TRANSACTION_NAME,
+                attributes={
+                    "sentry.op": OP.HTTP_SERVER,
+                    "sentry.origin": TornadoIntegration.origin,
+                    "sentry.span.source": SegmentSource.ROUTE,
+                },
+            )
```
sentrivana, line 141:
```suggestion
            span_ctx = sentry_sdk.traces.start_span(
                name=_DEFAULT_TRANSACTION_NAME,
                attributes={
                    "sentry.op": OP.HTTP_SERVER,
                    "sentry.origin": TornadoIntegration.origin,
                    "sentry.span.source": SegmentSource.ROUTE,
                },
                parent_span=None,
            )
```

to make sure this will be a segment

sentry_sdk/integrations/tornado.py
```diff
@@ -118,22 +129,91 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
-
-        with sentry_sdk.start_transaction(
-            transaction, custom_sampling_context={"tornado_request": self.request}
-        ):
-            yield
+        span_ctx: "ContextManager[Union[Span, StreamedSpan, None]]"
+
+        if is_span_streaming_enabled:
+            sentry_sdk.traces.continue_trace(dict(headers))
+            scope.set_custom_sampling_context({"tornado_request": self.request})
```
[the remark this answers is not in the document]
sl0thentr0py: the api handles this itself

sentry_sdk/integrations/tornado.py
sl0thentr0py: added req body

sentry_sdk/integrations/tornado.py
```diff
@@ -114,22 +124,112 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
+
+    headers = _filter_headers(dict(request.headers), use_annotated_value=False)
+    for header, value in headers.items():
+        attributes[f"{SPANDATA.HTTP_REQUEST_HEADER}.{header.lower()}"] = value
+
+    if request.query:
+        attributes[SPANDATA.HTTP_QUERY] = request.query
+
+    attributes[SPANDATA.URL_FULL] = "%s://%s%s" % (
+        request.protocol,
```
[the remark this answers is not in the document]
sl0thentr0py: changed to URL_QUERY and changed URL_FULL to internal tornado helper

sentry_sdk/integrations/tornado.py
```diff
@@ -114,22 +124,110 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
+    request: "Any",
+) -> "Union[Optional[str], AnnotatedValue]":
+    body = request.body
+    if not body:
+        return None
+
+    if not request_body_within_bounds(sentry_sdk.get_client(), len(body)):
+        return AnnotatedValue.substituted_because_over_size_limit()
+
+    return body.decode("utf-8", "replace")
```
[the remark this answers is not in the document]
sl0thentr0py: older path also doesn't respect pii, we will handle this separately later

sentry_sdk/integrations/tornado.py
```diff
@@ -114,22 +124,110 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
-        ):
-            yield
+        span_ctx: "ContextManager[Union[Span, StreamedSpan, None]]"
+
+        if is_span_streaming_enabled:
+            sentry_sdk.traces.continue_trace(dict(headers))
+            scope.set_custom_sampling_context({"tornado_request": self.request})
+
+            span_ctx = sentry_sdk.traces.start_span(
+                name=_DEFAULT_TRANSACTION_NAME,
```
ericapisani: `_DEFAULT_TRANSACTION_NAME` is a bit strange to read in the context of streamed spans where there is no longer a "transaction".

I'm ok with either leaving this as-is until we remove the transaction v1 spans and updating this then, or doing a rename that indicates that this name is the default for a "root" span.

sentry_sdk/integrations/tornado.py
```diff
@@ -114,22 +124,110 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
+                custom_sampling_context={"tornado_request": self.request},
+            )
+
+        with span_ctx as span:
+            try:
+                yield
+            finally:
+                if isinstance(span, StreamedSpan) and not isinstance(
+                    span, NoOpStreamedSpan
+                ):
```
ericapisani: This can be made a bit more concise with `type(span) is StreamedSpan`:


```suggestion
                if type(span) is StreamedSpan:
```

sentry_sdk/integrations/tornado.py
```diff
@@ -114,22 +124,110 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
+
+                    with capture_internal_exceptions():
+                        method = getattr(self, self.request.method.lower(), None)
+                        if method is not None:
+                            span_name = transaction_from_function(method)
+                            if span_name:
+                                span.name = span_name
+                                span.set_attribute(
+                                    "sentry.span.source",
+                                    SegmentSource.COMPONENT,
```
ericapisani, line 178: Is there a reason why setting this attribute is being gated by whether or not a `span_name `exists? This looks like something that we could set regardless.

sentry_sdk/integrations/tornado.py
```diff
@@ -114,22 +124,110 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
+
+
+def _get_tornado_request_data(
+    request: "Any",
+) -> "Union[Optional[str], AnnotatedValue]":
+    body = request.body
+    if not body:
+        return None
+
+    if not request_body_within_bounds(sentry_sdk.get_client(), len(body)):
```
ericapisani, line 225: Is the `body` at this point in time the "realized" result? For example, the result of `request.json()`?

I'm also wondering if using the `content-length` header - does tornado set this automatically for incoming requests?

tests/integrations/tornado/test_tornado.py
```diff
@@ -103,86 +103,147 @@ def test_basic(tornado_testcase, sentry_init, capture_events):
...
     "handler,code",
     [
         (CrashingHandler, 500),
         (HelloHandler, 200),
     ],
 )
-def test_transactions(tornado_testcase, sentry_init, capture_events, handler, code):
-    sentry_init(integrations=[TornadoIntegration()], traces_sample_rate=1.0)
-    events = capture_events()
+def test_transactions(
```
ericapisani, line 114: Since this is now testing both segments and transactions, this test name should be updated

sentry_sdk/integrations/tornado.py
sentrivana: `sentry.span.source` captures the source of the span name, so it only makes sense to set it if the name is actually set too

sentrivana (approved):
LGTM

sentry_sdk/integrations/tornado.py
sl0thentr0py: it's just whatever the raw body is, since this is a span attribute, the parsing can be left to relay imo

as to content-length, the previous extractor doesn't use content-length so i assume not
https://github.com/getsentry/sentry-python/blob/ee5bfee76935ffd651b3f248fbbb031aa9091250/sentry_sdk/integrations/tornado.py#L294-L298

sentry_sdk/integrations/tornado.py
```diff
@@ -114,22 +124,110 @@ def _handle_request_impl(self: "RequestHandler") -> "Generator[None, None, None]
...
+        return AnnotatedValue.substituted_because_over_size_limit()
+
```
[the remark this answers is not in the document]
sl0thentr0py: same as before, will be handled later

sentry_sdk/integrations/tornado.py
sl0thentr0py: resolved in slack

tests/integrations/tornado/test_tornado.py
sl0thentr0py: what should I update it to instead? I took wsgi tests as a reference and they're still called transactions

PR #6216, opened 2026-05-06
fix(openai): Guard against `choices=None`

cla7aye15I4nd:
## Summary

Some providers (e.g. OpenRouter) can return `choices=None` on upstream error responses. The existing `hasattr(response, "choices")` check passes because the attribute exists — it just doesn't guard against `None`.

This causes two distinct bugs:

### 1. Non-streaming: crash in caller

```
  File "sentry_sdk/integrations/openai.py", in _calculate_token_usage
    for choice in response.choices:
```

This propagates out of Sentry's patched `create()` wrapper before the calling code ever receives the response, making it impossible to handle on the caller side without catching a bare `TypeError`.

### 2. Streaming: silent data loss in span

Inside `_wrap_synchronous_completions_chunk_iterator` and `_wrap_asynchronous_completions_chunk_iterator`, the iteration is wrapped in `capture_internal_exceptions()`. When `choices=None`, the `TypeError` is silently suppressed — but this causes a premature exit from the `with` block, so `streaming_message_total_token_usage = x.usage` is never reached. Token usage and response text are then missing from the Sentry span.

## Fix

Add an explicit `is not None` guard alongside the existing `hasattr` check at all four affected locations (non-streaming, sync streaming, async streaming).

## Reproduction

Use the OpenAI SDK client pointed at OpenRouter (`base_url="https://openrouter.ai/api/v1"`). When an upstream provider fails, OpenRouter returns a response with `choices=null`, which triggers both bugs.

### DIFF
--- a/sentry_sdk/_compat.py
+++ b/sentry_sdk/_compat.py
@@ -54,7 +54,7 @@ def check_uwsgi_thread_support() -> bool:
             except Exception:
                 pass
 
-        return value and str(value).lower() not in FALSE_VALUES
+        return value and str(value).lower() not in FALSE_VALUES  # type: ignore[return-value]
 
     # When `threads` is passed in as a uwsgi option,
     # `enable-threads` is implied on.
--- a/sentry_sdk/_span_batcher.py
+++ b/sentry_sdk/_span_batcher.py
@@ -168,8 +168,8 @@ class SpanBatcher(Batcher["StreamedSpan"]):
             "start_timestamp": item._start_timestamp.timestamp(),
         }
 
-        if item._timestamp:
-            res["end_timestamp"] = item._timestamp.timestamp()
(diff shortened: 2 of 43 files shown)


alexander-alderman-webb:
Hi @cla7aye15I4nd,

Thanks a lot for the PR! We seem to have pulled in a new mypy major so merging is blocked by our required checks (very unfortunate timing).
I'll give you a few days to update your PR (merge in master or rebase), and otherwise I'll take over.
The type failures are all fixed in https://github.com/getsentry/sentry-python/commit/bd567bd14c9333f3426dc34d7137586ab6043fcd.

cla7aye15I4nd:
Thanks you ! I have fixed.

alexander-alderman-webb (approved):

PR #6239, opened 2026-05-08
feat(span-first): Support `before_send_span`

### LINKED ISSUE #5388: Add `before_send_span`
[Scrubbing data](https://develop.sentry.dev/sdk/telemetry/spans/scrubbing-data/)

Note: this cannot drop spans

sentrivana:
### Description
Add support for `before_send_span` in span streaming mode.

`before_send_span` is different from `before_send_metric` and `before_send_log` in that:
- it doesn't allow users to drop a span (i.e., return `None`)
- it only allows to modify specific parts of the span

To that end, we're now serializing the span earlier, and exposing the serialized dictionary in the `before_send` callback. This is consistent with metrics and logs. It also means we're now queuing dictionaries instead of `StreamedSpan` instances in the span batcher, which should also decrease our memory footprint.

This aligns our implementation with JS.

See https://develop.sentry.dev/sdk/telemetry/spans/scrubbing-data/ for spec.

#### Issues

* resolves: https://github.com/getsentry/sentry-python/issues/5388
* resolves: https://linear.app/getsentry/issue/PY-2057/add-before-send-span


sentry_sdk/client.py
```diff
@@ -938,34 +939,72 @@ def _capture_telemetry(
...
             before_send = get_before_send_metric(self.options)
+            serialized = telemetry
+
+        elif ty == "span":
+            before_send = get_before_send_span(self.options)
+            serialized = telemetry._to_json()  # type: ignore[union-attr]
 
         if before_send is not None:
-            telemetry = before_send(telemetry, {})  # type: ignore
+            serialized = before_send(serialized, {})  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
sentrivana: This is how all of the new telemetry `before_send_X` callbacks work. We can revisit if it's how it should be, but for now it's consistent.

sentry_sdk/client.py
```diff
@@ -938,34 +939,72 @@ def _capture_telemetry(
...
+
+            if ty in ("log", "metric"):
+                # Logs and metrics can be dropped in their respective
+                # before_send, so if we get None, don't queue them for sending.
+                if serialized is None:
+                    return
+
+            elif ty == "span" and isinstance(telemetry, StreamedSpan):
```
sentrivana, line 1207: The `isinstance(telemetry, StreamedSpan)` part of the condition is only there because mypy was complaining :( It can't deal with dispatchers/generics very well in general.

tests/integrations/sqlalchemy/test_sqlalchemy.py
```diff
@@ -1080,19 +1080,19 @@ class Person(Base):
 
             class fake_record_sql_queries:  # noqa: N801
                 def __init__(self, *args, **kwargs):
-                    with record_sql_queries_supporting_streaming(
+                    self._ctx_mgr = record_sql_queries_supporting_streaming(
```
sentrivana, line 1082: Had to change this test because it was failing -- it was closing the span too early.

sentry_sdk/_span_batcher.py
```diff
@@ -158,26 +159,15 @@ def _estimate_size(item: "StreamedSpan") -> int:
...
+    def _to_transport_format(item: "SpanJSON") -> "Any":
+        res = {k: v for k, v in item.items() if k not in ("_segment_span",)}
+
+        if item.get("attributes"):
             res["attributes"] = {
-                k: serialize_attribute(v) for (k, v) in item._attributes.items()
+                k: serialize_attribute(v) for (k, v) in item["attributes"].items()
             }
+        else:
+            del res["attributes"]
```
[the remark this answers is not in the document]
sentrivana: The `attributes` key is always there, it's explicitly added in the pre-serialization step in the Client

sentry_sdk/utils.py
```diff
@@ -2111,6 +2112,15 @@ def get_before_send_metric(
     )
 
 
+def get_before_send_span(
+    options: "Optional[dict[str, Any]]",
+) -> "Optional[Callable[[StreamedSpan, Hint], Optional[StreamedSpan]]]":
+    if options is None:
+        return None
+
+    return options["_experiments"].get("before_send_span")
```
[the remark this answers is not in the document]
sentrivana: https://github.com/getsentry/sentry-python/pull/6239/commits/177cc5b2395703da28cb2998146ce0eaa55907f3

alexander-alderman-webb (approved):
Looks good!

sentry_sdk/_span_batcher.py
```diff
@@ -142,12 +143,12 @@ def add(self, span: "StreamedSpan") -> None:
...
 
     @staticmethod
-    def _estimate_size(item: "StreamedSpan") -> int:
+    def _estimate_size(item: "SpanJSON") -> int:
         # Rough estimate of serialized span size that's quick to compute.
         # 210 is the rough size of the payload without attributes, and then we
         # estimate the attributes separately.
         estimate = 210
-        for value in item._attributes.values():
+        for value in (item.get("attributes") or {}).values():
```
ericapisani, line 152: This can be made slightly more concise by doing the following:
```suggestion
        for value in (item.get("attributes", {})).values():
```

sentry_sdk/client.py
```diff
@@ -1175,34 +1176,72 @@ def _capture_telemetry(
...
+                # Spans can't be dropped in before_send_span by design. They can
+                # be altered though (e.g. to sanitize). Only allow changes to
+                # name and attributes.
+                if isinstance(serialized, dict) and serialized and "name" in serialized:
```
ericapisani: The `and serialized` part is unnecessary because if the `isinstance(serialized, dict)` is truthy, then you can perform the `"name" in serialized` check without worrying about an exception being raised.

```suggestion
                if isinstance(serialized, dict) and "name" in serialized:
```

sentry_sdk/client.py
```diff
@@ -1175,34 +1176,72 @@ def _capture_telemetry(
...
+                    telemetry.name = serialized["name"]  # type: ignore[typeddict-item]
+                    telemetry._attributes = {}
+                    for k, v in (serialized.get("attributes") or {}).items():
```
ericapisani, line 1214: Similar to my comment above, you can remove the `or` by adding the `{}` as the fallback on the `get`:


```suggestion
                    for k, v in (serialized.get("attributes", {})).items():
```

sentry_sdk/client.py
```diff
@@ -1175,34 +1176,72 @@ def _capture_telemetry(
...
         batcher = None
         if ty == "log":
             batcher = self.log_batcher
+
         elif ty == "metric":
             batcher = self.metrics_batcher
+
         elif ty == "span":
+            # We need a reference to the segment span in the batcher to populate
+            # the DSC
```
ericapisani: I'd avoid using acronyms for comments that are explaining behaviour or are intended to provide context - it makes it difficult for those not familiar with Sentry's features/terminology to follow along


```suggestion
            # the dynamic sampling context
```

ericapisani (approved):
LGTM to me overall, just saw a few polishing opportunities. Approving so as not to block :shipit:

sentry_sdk/_span_batcher.py
sentrivana: If `item["attributes"]` exists and is `None`, then `item.get("attributes", {})` would return `None` and the subsequent `.values()` call would fail.

FWIW I don't think `attributes` _can_ be `None` at this point, but wrote it that way just to be extra safe.

sentry_sdk/client.py
sentrivana: Probably a leftover of my attempts to convince mypy to not complain. Fixed https://github.com/getsentry/sentry-python/pull/6239/commits/be1f8d2e2994354d39c5cdbb67814f784eb9ddaa

sentry_sdk/client.py
sentrivana: Same concern as above -- there's a case where the two are not equivalent and I'd opt for the more paranoid version

sentry_sdk/client.py
sentrivana: https://github.com/getsentry/sentry-python/pull/6239/commits/5fa6c96d549aad5df8baa560a2c8301f80da5130

sentry_sdk/client.py
```diff
@@ -1169,34 +1170,72 @@ def _capture_telemetry(
...
+            elif ty == "span" and isinstance(telemetry, StreamedSpan):
+                # Spans can't be dropped in before_send_span by design. They can
+                # be altered though (e.g. to sanitize). Only allow changes to
+                # name and attributes.
+                if isinstance(serialized, dict) and "name" in serialized:
+                    telemetry.name = serialized["name"]  # type: ignore[typeddict-item]
+                    telemetry._attributes = {}
+                    for k, v in (serialized.get("attributes") or {}).items():
+                        telemetry.set_attribute(k, v)
+
```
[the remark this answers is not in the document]
sentrivana: The callback returns the span as it should be sent over the wire. So if someone removes the `attributes` key (why would they even do that?), we'll also remove the attributes from the result.

PR #6241, opened 2026-05-08
Add option to drop scrubbed user IP addresses

### LINKED ISSUE #5701: Processing error due to invalid IP but an IP is shown in the event information

juliosuas:
## Summary

Fixes #5701 by adding an EventScrubber option to remove `user.ip_address` instead of replacing it with `[Filtered]`.

This avoids emitting `[Filtered]` in a field Relay expects to be a valid IP address, while preserving the existing default behavior for backward compatibility. By default, `user.ip_address` is still scrubbed through the normal scrubber path and keeps the `_meta` annotation.

## Testing

- `python -m pytest tests/test_scrubber.py tests/new_scopes_compat/test_new_scopes_compat_event.py -q`
- `ruff check sentry_sdk/scrubber.py tests/test_scrubber.py`
- `ruff format --check sentry_sdk/scrubber.py tests/test_scrubber.py`
- `git diff --check`

### DIFF
--- a/sentry_sdk/scrubber.py
+++ b/sentry_sdk/scrubber.py
@@ -66,7 +66,6 @@ class EventScrubber:
         recursive: bool = False,
         send_default_pii: bool = False,
         pii_denylist: "Optional[List[str]]" = None,
-        remove_user_ip_address: bool = False,
     ) -> None:
         """
         A scrubber that goes through the event payload and removes sensitive data configured through denylists.
@@ -75,7 +74,6 @@ class EventScrubber:
         :param recursive: Whether to scrub the event payload recursively, default False.
         :param send_default_pii: Whether pii is sending is on, pii fields are not scrubbed.
         :param pii_denylist: The denylist to use for scrubbing when pii is not sent, defaults to DEFAULT_PII_DENYLIST.
-        :param remove_user_ip_address: Whether to remove ``user.ip_address`` instead of replacing it with ``[Filtered]``.
         """
         self.denylist = DEFAULT_DENYLIST.copy() if denylist is None else denylist
 
@@ -87,7 +85,6 @@ class EventScrubber:
 
         self.denylist = [x.lower() for x in self.denylist]
         self.recursive = recursive
-        self.remove_user_ip_address = remove_user_ip_address
 
     def scrub_list(self, lst: object) -> None:
         """
@@ -141,11 +138,7 @@ class EventScrubber:
         with capture_internal_exceptions():
             if "user" in event:
                 user = event["user"]
-                if (
-                    "ip_address" in self.denylist
-                    and self.remove_user_ip_address
-                    and isinstance(user, dict)
-                ):
+                if "ip_address" in self.denylist and isinstance(user, dict):
                     user.pop("ip_address", None)
                 self.scrub_dict(user)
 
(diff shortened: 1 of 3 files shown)


sentry_sdk/scrubber.py
```diff
@@ -137,7 +137,10 @@ def scrub_extra(self, event: "Event") -> None:
     def scrub_user(self, event: "Event") -> None:
         with capture_internal_exceptions():
             if "user" in event:
-                self.scrub_dict(event["user"])
+                user = event["user"]
+                if isinstance(user, dict) and "ip_address" in self.denylist:
+                    user.pop("ip_address", None)
```
ericapisani: This is the right idea, but let's make one further optimization:

```suggestion
                if "ip_address" in self.denylist and isinstance(user, dict):
                    user.pop("ip_address", None)
```

By moving the `"ip_address" in self.denylist` earlier in the conditional and evaluating it first, it means that we can ignore the second condition if the first isn't satisfied.

ericapisani (changes requested):
Thanks for opening a PR! These changes are generally looking good, just one small adjustment before we're good to ship :shipit:

juliosuas:
Thanks — applied the suggested condition order and reran the targeted scrubber tests locally:\n\n```bash\n.venv/bin/python -m pytest tests/test_scrubber.py -q\n# 11 passed\n```

sentry_sdk/scrubber.py
```diff
@@ -66,6 +66,7 @@ def __init__(
         recursive: bool = False,
         send_default_pii: bool = False,
         pii_denylist: "Optional[List[str]]" = None,
+        remove_user_ip_address: bool = False,
```
alexander-alderman-webb: Let's please no add API surface to address the bug.
Can you just always drop the `user.ip_address` field when it is scrubbed 🙏 ? 

```suggestion
```

alexander-alderman-webb (changes requested):

ericapisani (approved):

alexander-alderman-webb (approved):

PR #6242, opened 2026-05-08
feat(asyncpg): Add query source to execute, executemany, and cursor methods

ericapisani:
## Summary
- Add `add_query_source()` calls to `_wrap_connection_method` and `_wrap_cursor_method`, which cover `execute()`, `executemany()`, `prepare()`, `_bind_exec()`, and `_exec`. Previously only `fetch()` had query source tracking via the patched `_wrap_execute` method.
- Handle `StreamedSpan` vs regular span cases separately (StreamedSpan needs the call inside the context manager).


tests/integrations/asyncpg/test_asyncpg.py
```diff
@@ -740,40 +743,71 @@ async def test_query_source_with_module_in_search_path(sentry_init, capture_even
...
 
-    events = capture_events()
-
     from asyncpg_helpers.helpers import execute_query_in_connection
 
-    with start_transaction(name="test_transaction", sampled=True):
-        conn: Connection = await connect(PG_CONNECTION_URI)
+    if span_streaming:
+        items = capture_items("span")
+        with sentry_sdk.traces.start_span(name="test_transaction"):
```
sentrivana: Can we call this `test_span` or `test_segment` in the span first code path? Since there are no transactions in span first

Also if we mention transactions in other span first tests we should change that 🙏🏻

sentrivana (approved):
One comment re: naming in tests https://github.com/getsentry/sentry-python/pull/6242/changes#r3225427754

Otherwise lgtm!

alexander-alderman-webb (changes requested):
Why do we add query source to `_wrap_cursor_creation()`?

My understanding is that the `add_query_source()` function is intended to add the query source to functions that execute SQL queries, which are only the following:

- `asyncpg.Connection.execute`
- `asyncpg.Connection._execute`
- `asyncpg.Connection._executemany`

The parameter description says the following:

https://github.com/getsentry/sentry-python/blob/17cc8c7b2c31c2df130418bb49137814d4d35f7b/sentry_sdk/consts.py#L1559

Can we ensure that _only_ spans for these functions have query source added?

ericapisani:
@alexander-alderman-webb 

> Why do we add query source to _wrap_cursor_creation()?

When this PR was initially created, I hadn't yet noticed the underlying issue with how we've patched cursors, and that this only captured the creation of the `CursorFactory` (which included the SQL query first passed in). Spans wouldn't be created as the cursor iterated through a result set.

At this point, due to the introduction of #6252 , this PR is somewhat out of date in that cursor creation is no longer patched - the underlying `BaseCursor` methods are.

This means that, with those changes:
- spans are created when `_bind_exec` is called, and we want to record the query source because this results in the first set of rows being retrieved
- spans are created when `_exec` is called, as this method processes the next fetch

and we'll want to add query sources to those.

tests/integrations/asyncpg/test_asyncpg.py
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6242/commits/ede05373324068409848204bee918eb0adeebed4

sentry_sdk/integrations/asyncpg.py
```diff
@@ -53,8 +53,10 @@ def setup_once() -> None:
         )
         asyncpg.Connection.prepare = _wrap_connection_method(asyncpg.Connection.prepare)
```
alexander-alderman-webb, line 54: Should source context be attached to spans for `asyncpg.Connection.prepare`?

If it doesn't execute an SQL query we should probably not attach query source for the same reason as in https://github.com/getsentry/sentry-python/pull/6242#pullrequestreview-4271462167

ericapisani: > Should source context be attached to spans for asyncpg.Connection.prepare?

I think in this case we should because while it's not executing a SQL query such that rows are fetched, this is a public method that a user can invoke within their application that has performance implications. 

From the [docs](https://magicstack.github.io/asyncpg/current/api/index.html#prepared-statements) (I've highlighted the specific part of interest with italics):

> When a query is prepared by a call to Connection.prepare(), _the server parses, analyzes and compiles the query_ allowing to reuse that work once there is a need to run the same query again.

alexander-alderman-webb: Since this is sent to the database, I think it's valid to consider this to be covered by `enable_db_query_source` 👍

alexander-alderman-webb (approved):

PR #6252, opened 2026-05-11
feat(asyncpg): Add cursor span support via BaseCursor method patching

### LINKED ISSUE #6240: Investigate missing span coverage on queries run by an asyncpg `Cursor`
Cursors use the `protocol`rather than `connection` to execute their queries. This means our current, Connection-focused patches are not capturing those queries.

The methods of interest are:

* [BaseCursor.\_exec](<https://github.com/MagicStack/asyncpg/blob/db8ecc2a38e16fb0c090aef6f5506547c2831c24/asyncpg/cursor.py#L149-L159>)
* [BaseCursor.\_bind_exec](<https://github.com/MagicStack/asyncpg/blob/db8ecc2a38e16fb0c090aef6f5506547c2831c24/asyncpg/cursor.py#L118-L131>)
* [BaseCursor.\_bind](<https://github.com/MagicStack/asyncpg/blob/db8ecc2a38e16fb0c090aef6f5506547c2831c24/asyncpg/cursor.py#L133-L147>)

ericapisani:
Patch `BaseCursor._bind_exec`, `_bind`, and `_exec` instead of wrapping `Connection.cursor` creation. The old approach only captured cursor creation, not actual query execution. The new approach creates spans for each cursor operation (iteration via `_bind_exec`, explicit cursor usage via `_bind`/`_exec`) and supports both static and streaming span lifecycles.

Builds on #6215 which added streaming span support to the asyncpg integration.

Removes the patch on `asyncpg.Connection.cursor` because the patches on `BaseCursor` introduced in this pull request will create the correct span when the cursor is used.

Fixes PY-2408
Fixes #6240


ericapisani:
bugbot run

ericapisani:
@sentry review

sentry_sdk/integrations/asyncpg.py
```diff
@@ -52,8 +51,12 @@ def setup_once() -> None:
         asyncpg.Connection._executemany = _wrap_connection_method(
             asyncpg.Connection._executemany, executemany=True
         )
-        asyncpg.Connection.cursor = _wrap_cursor_creation(asyncpg.Connection.cursor)
```
ericapisani, line 55: This is safe to remove because this patches the function that led to the creation/return of a `CursorFactory`. When that cursor is `await`ed, the `CursorFactory` calls `_bind` through the underlying cursor's `init` [method](https://github.com/MagicStack/asyncpg/blob/db8ecc2a38e16fb0c090aef6f5506547c2831c24/asyncpg/cursor.py#L75), essentially capturing the creation and invocation of the cursor (as opposed to just capturing the creation of the factory which is what was happening previously)

tests/integrations/asyncpg/test_asyncpg.py
```diff
@@ -1102,3 +1114,188 @@ def before_send_transaction(event, hint):
...
+        assert bind_exec_span["data"]["db.system"] == "postgresql"
+        assert bind_exec_span["data"]["db.driver.name"] == "asyncpg"
+        assert bind_exec_span["data"]["server.address"] == PG_HOST
+        assert bind_exec_span["data"]["server.port"] == PG_PORT
+        assert bind_exec_span["data"]["db.name"] == PG_NAME
+        assert bind_exec_span["data"]["db.user"] == PG_USER
+
+
+@pytest.mark.asyncio
+async def test_cursor__bind_and__exec_methods_create_spans(
```
ericapisani, line 1240: I did my best to create some distinction between this and `test_cursor__bind_exec_creates_spans` but in case this naming causes confusion:

This test tests the `_bind` and `_exec` patches, whereas the other test above checks the `_bind_exec` patch.

tests/integrations/asyncpg/test_asyncpg.py
```diff
@@ -1102,3 +1114,188 @@ def before_send_transaction(event, hint):
...
+                ("Alice", "pw", datetime.date(1990, 12, 25)),
+            ],
+        )
+
+        async with conn.transaction():
+            # This exercises the `_bind` patch and the `cursor` patch
+            cur = await conn.cursor(
+                "SELECT * FROM users WHERE dob > $1", datetime.date(1970, 1, 1)
+            )
+            # These exercises the `_exec` patch
```
sentrivana:
```suggestion
            # These exercise the `_exec` patch
```

sentrivana (approved):
LGTM

sk-:
Hi @ericapisani, this change started triggering lots of detected N+1 queries when using server side cursors.

In our case we are using sqlalchemy with asyncpg, and all the places where we do `await conn.stream(query)` are being flagged as N+1 queries.

What is the proper workaround? Should we just disable the AsyncPGIntegration if using sqlalchemy?

ericapisani:
Hi @sk- ,

Could you open an issue for what you're describing?

It'd also be helpful to know how these n+1 queries are being detected - are you seeing these through [Sentry's detection](https://docs.sentry.io/product/issues/issue-details/performance-issues/n-one-queries/) or is this being determined through some other means?

If you're able to, it'd also be great to have a repro.

Thanks!

PR #6273, opened 2026-05-15
feat(dramatiq): Support span streaming

### LINKED ISSUE #6016: Migrate dramatiq to span first

sl0thentr0py:
## Issues

* resolves: #6016 
* resolves: PY-2318

### DIFF
--- a/sentry_sdk/integrations/dramatiq.py
+++ b/sentry_sdk/integrations/dramatiq.py
@@ -6,11 +6,13 @@ from sentry_sdk.api import continue_trace, get_baggage, get_traceparent
 from sentry_sdk.consts import OP, SPANSTATUS
 from sentry_sdk.integrations import DidNotEnable, Integration
 from sentry_sdk.integrations._wsgi_common import request_body_within_bounds
+from sentry_sdk.traces import SegmentSource
 from sentry_sdk.tracing import (
     BAGGAGE_HEADER_NAME,
     SENTRY_TRACE_HEADER_NAME,
     TransactionSource,
 )
+from sentry_sdk.tracing_utils import has_span_streaming_enabled
 from sentry_sdk.utils import (
     AnnotatedValue,
     capture_internal_exceptions,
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/dramatiq.py
```diff
@@ -129,21 +132,35 @@ def before_process_message(self, broker: "Broker", message: "Message[R]") -> Non
...
-        transaction.__enter__()
+        if has_span_streaming_enabled(client.options):
+            sentry_sdk.traces.continue_trace(sentry_headers)
+            span = sentry_sdk.traces.start_span(
+                name=message.actor_name,
+                attributes={
+                    "sentry.op": OP.QUEUE_TASK_DRAMATIQ,
+                    "sentry.origin": DramatiqIntegration.origin,
+                    "sentry.span.source": SegmentSource.TASK.value,
+                },
```
sentrivana, line 143:
```suggestion
                },
                parent_span=None,
```

sentry_sdk/integrations/dramatiq.py
```diff
@@ -129,21 +132,35 @@ def before_process_message(self, broker: "Broker", message: "Message[R]") -> Non
...
+            )
+            message._sentry_span_ctx = span
+            span.__enter__()
```
sentrivana:
```suggestion
```

`__enter__` in span first doesn't do anything, the span is set on the scope at `start_span` time (unless explicitly called with `active=False`)

sentry_sdk/integrations/dramatiq.py
```diff
@@ -172,7 +189,7 @@ def after_process_message(
         )
         if not is_event_capture_required:
             # normal transaction finish
-            transaction.__exit__(None, None, None)
+            span_ctx.__exit__(None, None, None)
```
sentrivana, line 192: The canonical way to do this in span first would just be `span.end()`, but I get that this way it's compatible with both legacy and span streaming mode

sentrivana (commented):
Couple things, please see comments

sentry_sdk/integrations/dramatiq.py
sl0thentr0py: right, i keep forgetting

sentrivana (approved):

PR #6297, opened 2026-05-19
feat(aiohttp): Remove request body capture from span streaming

### LINKED ISSUE #6292: Remove request body capture in aiohttp
In span first, we're capturing request bodies on the `http.request.body.data` attribute. Remove this functionality and simplify the code.<br><br>Note: this is only about removing the functionality for v2 spans. Old code paths that capture the request body for events should stay as is.

ericapisani:
Removes request body capture from aiohttp span streaming to reduce complexity.

The aiohttp integration previously captured the request body and set it as the
`http.request.body.data` span attribute during span streaming. This change
removes that functionality to simplify the integration, along with the
associated tests that verified the behavior.

Closes #6292
Fixes PY-2422

### DIFF
--- a/sentry_sdk/integrations/aiohttp.py
+++ b/sentry_sdk/integrations/aiohttp.py
@@ -201,60 +201,41 @@ class AioHttpIntegration(Integration):
 
                     with span_ctx as span:
                         try:
-                            try:
-                                response = await old_handle(self, request)
-                            except HTTPException as e:
-                                if isinstance(span, StreamedSpan) and not isinstance(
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/aiohttp.py
ericapisani, line 1: This is an indentation change after the removal of the `finally` block that was setting the `http.request.body.data` attribute - it might be easier to review this outside the diff view

alexander-alderman-webb (approved):
Thanks!

PR #6304, opened 2026-05-20
fix(serializer): Don't call `__iter__` on arbitrary sequences

### LINKED ISSUE #5586: Deadlock during GC with Sentry + Opentelemetry + Tornado + aiohttp
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.53.0

### Steps to Reproduce

Detailed minimal repro project here:

https://github.com/brettdh/sentry-aiohttp-deadlock-repro/

Note: this is a very specific example of a more general issue: if an error occurs in a context where a lock is held, Sentry's inspection and serialization of the stack variables can cause a deadlock when inspecting a variable causes another attempt to acquire a lock that's already held.

In this specific case, these details fill in the general pattern above:
1. the context is an aiohttp client session's `__del__` via garbage collection
   - This error also occurs with other aiohttp `__del__` methods 
   - but it could have been a different chain of events that led to the inspection of an object on the stack that implicitly tries to grab a lock that's already held by the same thread
2. the lock is held in the Opentelemetry SDK, on an internal variable used when starting or ending a span
   - because the span starts in Sentry's Tornado instrumentation, which itself relies on the Opentelemetry SDK 
   - but it could have been anything in any code, first- or third-party
3. the lock is being held by the thread that is attempting to re-acquire it
   - so in this case, making it a re-entrant lock avoids the deadlock
   - but in another case, it could be held by another thread, and acquiring it during garbage collection (while other threads are stopped) could still lead to a deadlock
   - for example, the garbage-collection could have been triggered on the main thread while a worker thread was holding the lock, and the deadlock would still occur, since the worker thread cannot wake up and release the lock
   - in fact, we think we have seen this happen in production as well (but not 100% sure)
     - edit: it seems like this might not be the case after all, looking at the [PEP 703](https://peps.python.org/pep-0703/#stop-the-world) description of stop-the-world garbage collection in Python 3.12 and 3.13

and these workarounds are available:
1. Fix all the "Unclosed connector" issues in our application that cause the logging during `__del__`
   - but this isn't always straightforward, and it may regress
2. Ignore the `aiohttp` logger in Sentry config (or disable all logging integration)
   - but then we don't get notified in Sentry about the above warnings or others
   - and actually, this was a red herring and doesn't fix the problem
   - The `__del__` method calls the loop's exception handler, which is what logs at `error` level and gets captured by Sentry
   - we'd have to ignore the `asyncio` logger instead - but we couldn't know this without digging in to the code
   - what other risky code paths do we not know about yet?
3. Disable stack variable capture altogether
   - obvious downsides; local variable capture is extremely valuable for debugging
4. Something else that I haven't thought of yet

As the repro project shows, this particular issue can be resolved by changing the `Lock` to a `RLock` in the otel SDK. That may be a good change to make, or it may have unacceptable side effects; I'm not sure yet. The reason I am reporting this as a Sentry issue is that enabling Sentry in our application has had the surprising effect of exposing us to the risk of deadlock (which we have seen happen reliably in production), and we've had to disable Sentry in our Python contexts for the time being.

I'd like to start a conversation about what changes in the SDK might be possible to reduce or eliminate this risk and similar risks caused by a large operation such as stack introspection in various contexts where locks could be held.

### Expected Result

No deadlock triggered by Sentry, regardless of what code/stack Sentry is attempting to inspect

### Actual Result

Deadlock triggered by Sentry (does not occur when Sentry is disabled)

sentrivana:
### Description
The serializer will attempt to serialize whatever objects are thrown into it. This includes arbitrary local variables in folks' programs that might have all sorts of side-effects.

Notably, if an object is a [`Sequence`](https://docs.python.org/3/glossary.html#term-sequence) (it has `__iter__`, `__len__`, and `__getitem__`), we will try to `__iter__` over it. Which is potentially an unwise thing to do, as we know nothing about the object's custom `__iter__` implementation. It can, for example, [acquire a non-reentrant lock](https://github.com/getsentry/sentry-python/issues/5586). In some scenarios, this can lead to deadlocks caused by the SDK.

In this PR, the serializer is changed so that we don't try to `__iter__` over all `Sequence`s, just built-in stdlib ones. Of course, nothing is preventing folks from defining side-effecty lists, tuples and sets, but it's a bit less likely to happen than on a custom class. Custom sequences will now be simply `__repr__`'d.

#### Issues
- Closes https://github.com/getsentry/sentry-python/issues/5586
- Closes https://linear.app/getsentry/issue/PY-2124/deadlock-during-gc-with-sentry-opentelemetry-tornado-aiohttp

#### Reminders

### DIFF
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -1,6 +1,6 @@
 import math
 import sys
-from collections.abc import Mapping, Sequence, Set
+from collections.abc import Mapping
 from datetime import datetime
 from typing import TYPE_CHECKING
 
@@ -53,7 +53,7 @@ def add_global_repr_processor(processor: "ReprProcessor") -> None:
     global_repr_processors.append(processor)
 
 
-sequence_types: "List[type]" = [Sequence, Set]
+sequence_types: "List[type]" = [tuple, list, set]
 
 
 def add_repr_sequence_type(ty: type) -> None:
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -53,7 +53,7 @@ def add_global_repr_processor(processor: "ReprProcessor") -> None:
     global_repr_processors.append(processor)
 
 
-sequence_types: "List[type]" = [tuple, list, set]
+sequence_types: "List[type]" = [tuple, list, set, frozenset]
 
 
 def add_repr_sequence_type(ty: type) -> None:
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -53,7 +53,7 @@ def add_global_repr_processor(processor: "ReprProcessor") -> None:
     global_repr_processors.append(processor)
 
 
-sequence_types: "List[type]" = [tuple, list, set, frozenset]
+sequence_types: "List[type]" = [tuple, list, set, frozenset, bytes]
 
 
 def add_repr_sequence_type(ty: type) -> None:
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -1,5 +1,6 @@
 import math
 import sys
+from array import array
 from collections.abc import Mapping
 from datetime import datetime
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb (approved):
Possibly leads to some data loss but it's worth it to not interfere with users apps IMO.

ericapisani (approved):

PR #6305, opened 2026-05-20
ref(wsgi): Remove unused segment name setting

sentrivana:
### Description
In span first, there is logic in the WSGI integration that sets the segment name and the source attribute to a generic default at the end of a segment span's lifecycle, unless the name/source have been set explicitly.

This logic is unnecessary. We create the segment with the default name and source, so there's no reason to rename again at the end. If the segment was renamed during its lifetime, that's the name we want to keep; the default is just a fallback.

#### Reminders


alexander-alderman-webb (approved):
Perfect makes the Django migration simpler 🙏 
I'll wait to merge [mine](https://github.com/getsentry/sentry-python/pull/6248) after this is in

PR #6308, opened 2026-05-20
feat(strawberry): Support span streaming

### LINKED ISSUE #6063: Migrate strawberry to span first

ericapisani:
Add span streaming support to the Strawberry GraphQL integration,
mirroring the pattern used in the starlite and aiohttp
integrations.

Fixes PY-2366
Fixes #6063

### DIFF
--- a/sentry_sdk/integrations/strawberry.py
+++ b/sentry_sdk/integrations/strawberry.py
@@ -8,7 +8,8 @@ from sentry_sdk.consts import OP
 from sentry_sdk.integrations import DidNotEnable, Integration, _check_minimum_version
 from sentry_sdk.integrations.logging import ignore_logger
 from sentry_sdk.scope import should_send_default_pii
-from sentry_sdk.tracing import TransactionSource
+from sentry_sdk.tracing import Span, TransactionSource
+from sentry_sdk.tracing_utils import StreamedSpan, has_span_streaming_enabled
 from sentry_sdk.utils import (
     capture_internal_exceptions,
     ensure_integration_enabled,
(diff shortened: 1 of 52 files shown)


sentry_sdk/integrations/strawberry.py
```diff
@@ -180,53 +181,116 @@ def on_operation(self) -> "Generator[None, None, None]":
...
+            validation_span = sentry_sdk.start_span(
+                op=OP.GRAPHQL_VALIDATE,
+                name="validation",
+                origin=StrawberryIntegration.origin,
+            )
 
         yield
 
-        validation_span.finish()
+        if is_span_streaming_enabled and type(validation_span) is StreamedSpan:
```
ericapisani: The extra `type` check here (and in other areas) is because the type checker would raise errors.

sentry_sdk/integrations/strawberry.py
```diff
@@ -168,65 +169,128 @@ def on_operation(self) -> "Generator[None, None, None]":
...
+                origin=StrawberryIntegration.origin,
+            )
 
         yield
 
-        self.validation_span.finish()
+        if is_span_streaming_enabled and type(validation_span) is StreamedSpan:
+            validation_span.end()
+        else:
+            validation_span.finish()
```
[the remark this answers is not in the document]
ericapisani: `on_validate` already had [this pattern](https://github.com/getsentry/sentry-python/blob/c44340a69f6cab619dd8bf0c226566dc051fe871/sentry_sdk/integrations/strawberry.py#L209-L218) as did [`on_parse`](https://github.com/getsentry/sentry-python/blob/c44340a69f6cab619dd8bf0c226566dc051fe871/sentry_sdk/integrations/strawberry.py#L221-L229)

sentry_sdk/integrations/strawberry.py
```diff
@@ -8,7 +8,8 @@
 from sentry_sdk.integrations import DidNotEnable, Integration, _check_minimum_version
 from sentry_sdk.integrations.logging import ignore_logger
 from sentry_sdk.scope import should_send_default_pii
-from sentry_sdk.tracing import TransactionSource
+from sentry_sdk.tracing import Span, TransactionSource
+from sentry_sdk.tracing_utils import StreamedSpan, has_span_streaming_enabled
```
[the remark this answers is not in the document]
ericapisani: Have been removed.

ericapisani:
bugbot run
@sentry review

sentry_sdk/integrations/strawberry.py
```diff
@@ -180,53 +181,116 @@ def on_operation(self) -> "Generator[None, None, None]":
...
+        client = sentry_sdk.get_client()
+        is_span_streaming_enabled = has_span_streaming_enabled(client.options)
+        if is_span_streaming_enabled:
+            additional_attributes: "dict[str, Any]" = {}
+
+            if should_send_default_pii():
+                additional_attributes["graphql.document"] = execution_context.query
+
+            if operation_name:
+                additional_attributes["graphql.operation.name"] = operation_name
```
ericapisani, line 197: The operation_name may be `None` which is why we need this conditional here.

sentry_sdk/integrations/strawberry.py
```diff
@@ -180,53 +181,116 @@ def on_operation(self) -> "Generator[None, None, None]":
...
+            graphql_span.set_data("graphql.resource_name", self._resource_name)
 
         yield
 
-        transaction = graphql_span.containing_transaction
-        if transaction and self.execution_context.operation_name:
-            transaction.name = self.execution_context.operation_name
-            transaction.source = TransactionSource.COMPONENT
-            transaction.op = op
+        if type(graphql_span) is StreamedSpan:
```
ericapisani, line 227: I would have preferred to have `has_span_streamed_enabled` check here, but the type checker becomes unhappy if I do that.

sentry_sdk/integrations/strawberry.py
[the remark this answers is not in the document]
ericapisani: Follow up - have made an issue to investigate this further - https://github.com/getsentry/sentry-python/issues/6309

sentry_sdk/integrations/strawberry.py
```diff
@@ -180,53 +181,116 @@ def on_operation(self) -> "Generator[None, None, None]":
...
+                origin=StrawberryIntegration.origin,
+            )
 
         yield
 
-        parsing_span.finish()
+        if is_span_streaming_enabled and type(parsing_span) is StreamedSpan:
+            parsing_span.end()
+        else:
+            parsing_span.finish()
```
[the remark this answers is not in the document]
ericapisani: This pattern existed prior to these changes.  Have made a separate issue to investigate this further - https://github.com/getsentry/sentry-python/issues/6309

sentry_sdk/integrations/strawberry.py
```diff
@@ -290,6 +366,18 @@ def resolve(
...
+        if is_span_streaming_enabled:
+            with sentry_sdk.traces.start_span(
+                name=f"resolving {field_path}",
+                attributes={
+                    "sentry.origin": StrawberryIntegration.origin,
+                    "sentry.op": OP.GRAPHQL_RESOLVE,
+                },
+            ):
+                return _next(root, info, *args, **kwargs)
+
```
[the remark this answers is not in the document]
ericapisani: Those properties are not in our conventions and are being removed, so this code is correct

sentry_sdk/integrations/strawberry.py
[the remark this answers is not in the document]
ericapisani: Addressed in [ec14f15](https://github.com/getsentry/sentry-python/pull/6308/commits/ec14f158ba71045684c00a4babfd4f165f547f6e)

sentry_sdk/integrations/strawberry.py
```diff
@@ -180,53 +181,116 @@ def on_operation(self) -> "Generator[None, None, None]":
         )
 
         scope = sentry_sdk.get_isolation_scope()
-        event_processor = _make_request_event_processor(self.execution_context)
+        execution_context = self.execution_context
```
sentrivana: Why this? IMO it's better if it's clear throughout the function that `execution_context` is not an ephemeral local var, but rather something living on the Extension and used by it, so that we don't e.g. arbitrarily change it.

Also, if we can get rid of extra vars that's always a good thing.

sentry_sdk/integrations/strawberry.py
```diff
@@ -180,53 +181,116 @@ def on_operation(self) -> "Generator[None, None, None]":
...
+                },
+            )
+        else:
+            graphql_span = sentry_sdk.start_span(
+                op=op,
+                name=description,
+                origin=StrawberryIntegration.origin,
+            )
+
         graphql_span.__enter__()
```
sentrivana: `graphql_span.__enter__()` is no-op for streamed spans -- unlike legacy spans, they're started and put on the scope (made active) automatically when you call `start_span`, even without the context manager.

It doesn't hurt to have it here, but I'd suggest we move the line to the legacy span `else` branch above, so that when we get rid of the legacy logic, we also remove this no-op.

sentry_sdk/integrations/strawberry.py
```diff
@@ -180,53 +181,116 @@ def on_operation(self) -> "Generator[None, None, None]":
...
 
-        transaction = graphql_span.containing_transaction
-        if transaction and self.execution_context.operation_name:
-            transaction.name = self.execution_context.operation_name
-            transaction.source = TransactionSource.COMPONENT
-            transaction.op = op
+        if type(graphql_span) is StreamedSpan:
+            if execution_context.operation_name:
+                segment = graphql_span._segment
+                segment.set_attribute("sentry.span.source", TransactionSource.COMPONENT)
```
sentrivana:
```suggestion
                segment.set_attribute("sentry.span.source", SegmentSource.COMPONENT)
```

`SegmentSource` is the streaming equivalent of `TransactionSource`. It can be imported from `traces.py`

sentry_sdk/integrations/strawberry.py
```diff
@@ -180,53 +181,116 @@ def on_operation(self) -> "Generator[None, None, None]":
...
+            validation_span = sentry_sdk.start_span(
+                op=OP.GRAPHQL_VALIDATE,
+                name="validation",
+                origin=StrawberryIntegration.origin,
+            )
 
         yield
 
-        validation_span.finish()
+        if is_span_streaming_enabled and isinstance(validation_span, StreamedSpan):
```
sentrivana: Does this need both parts of the condition? Would just checking for `StreamedSpan` be enough?

```suggestion
        if isinstance(validation_span, StreamedSpan):
```

sentry_sdk/integrations/strawberry.py
```diff
@@ -180,53 +181,116 @@ def on_operation(self) -> "Generator[None, None, None]":
...
+            parsing_span = sentry_sdk.start_span(
+                op=OP.GRAPHQL_PARSE,
+                name="parsing",
+                origin=StrawberryIntegration.origin,
+            )
 
         yield
 
-        parsing_span.finish()
+        if is_span_streaming_enabled and isinstance(parsing_span, StreamedSpan):
```
sentrivana: Similar comment as [above](https://github.com/getsentry/sentry-python/pull/6308/changes#r3279101667):

```suggestion
        if isinstance(parsing_span, StreamedSpan):
```

sentrivana (approved):
Looking great, couple minor things, please see comments. Approving in advance

sentry_sdk/integrations/strawberry.py
ericapisani: When I wrote this, I was initially thinking that doing this made it slightly easier to read but looking at this function again, it doesn't really do much, and if I was going to do this, I should've done it at the start of the function 😅 

I think you're right though regarding your point about trying to indicate to future devs working in this space to not arbitrarily modify the `execution_context`, am ok with reverting this 👍🏻

sentry_sdk/integrations/strawberry.py
ericapisani: You're right, no, the first part of this condition isn't necessary, will get this cleaned up 🧹

PR #6319, opened 2026-05-21
feat(sanic): Support span streaming

### LINKED ISSUE #6055: Migrate sanic to span first

sl0thentr0py:
The `_unsampled_statuses` option is not respected for span streaming since we currently don't allow tail-based sampling in the new system.

#### Issues

* resolves: #6055
* resolves: PY-2357

### DIFF
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -230,7 +230,7 @@ async def _context_exit(
                     span.set_attribute(SPANDATA.HTTP_STATUS_CODE, response_status)
                     span.status = "error" if response_status >= 400 else "ok"
 
-                span._end()
+                span.end()
             else:
                 span.set_http_status(response_status)
                 span.sampled &= (
--- a/sentry_sdk/integrations/sanic.py
+++ b/sentry_sdk/integrations/sanic.py
@@ -6,11 +6,14 @@ from urllib.parse import urlsplit
 
 import sentry_sdk
 from sentry_sdk import continue_trace
-from sentry_sdk.consts import OP
+from sentry_sdk.consts import OP, SPANDATA
 from sentry_sdk.integrations import DidNotEnable, Integration, _check_minimum_version
 from sentry_sdk.integrations._wsgi_common import RequestExtractor, _filter_headers
 from sentry_sdk.integrations.logging import ignore_logger
+from sentry_sdk.scope import should_send_default_pii
+from sentry_sdk.traces import SegmentSource, StreamedSpan
 from sentry_sdk.tracing import TransactionSource
+from sentry_sdk.tracing_utils import has_span_streaming_enabled
 from sentry_sdk.utils import (
     CONTEXTVARS_ERROR_MESSAGE,
     HAS_REAL_CONTEXTVARS,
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/sanic.py
```diff
@@ -194,16 +217,31 @@ async def _context_exit(
...
+                    span.set_attribute(SPANDATA.HTTP_STATUS_CODE, response_status)
+                    span.status = "error" if response_status >= 400 else "ok"
+                # If the status is in unsampled_statuses, skip ending the span so
+                # it is never queued for sending. The orphaned scope is discarded
+                # when the isolation scope exits below.
+                if should_sample:
+                    span.__exit__(None, None, None)
+            else:
+                span.set_http_status(response_status)
+                span.sampled &= should_sample
```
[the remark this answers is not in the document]
sl0thentr0py: added `_end_cleanup` method

sentry_sdk/integrations/sanic.py
```diff
@@ -194,16 +217,34 @@ async def _context_exit(
...
+                    span.set_attribute(SPANDATA.HTTP_STATUS_CODE, response_status)
+                    span.status = "error" if response_status >= 400 else "ok"
+
+                # If the status is in unsampled_statuses, skip ending the span so
+                # it is never queued for sending but just cleanup the state.
+                # The orphaned scope is discarded when the isolation scope exits below.
+                if should_sample:
+                    span._end()
+                else:
+                    span._end_cleanup()
```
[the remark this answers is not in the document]
sl0thentr0py: added to noop

sentry_sdk/integrations/sanic.py
```diff
@@ -198,12 +221,24 @@ async def _context_exit(
...
+            span = request.ctx._sentry_root_span
+            if isinstance(span, StreamedSpan):
+                with capture_internal_exceptions():
+                    for attr, value in _get_request_attributes(request).items():
+                        span.set_attribute(attr, value)
+                if response_status is not None:
+                    span.set_attribute(SPANDATA.HTTP_STATUS_CODE, response_status)
+                    span.status = "error" if response_status >= 400 else "ok"
+
+                span._end()
```
ericapisani: We have a public `end` method that we can use here instead of the private `_end` (and the public method calls `_end` under the hood giving us the same outcome)

sentry_sdk/integrations/sanic.py
```diff
@@ -315,6 +350,36 @@ def _capture_exception(exception: "Union[ExcInfo, BaseException]") -> None:
...
+    headers = _filter_headers(dict(request.headers), use_annotated_value=False)
+    for header, value in headers.items():
+        attributes[f"{SPANDATA.HTTP_REQUEST_HEADER}.{header.lower()}"] = value
+
+    urlparts = urlsplit(request.url)
+
+    if urlparts.query:
+        attributes[SPANDATA.HTTP_QUERY] = urlparts.query
+
+    attributes[SPANDATA.URL_FULL] = request.url
```
ericapisani, line 371: [This appears to now be treated as PII](https://getsentry.github.io/sentry-conventions/attributes/url/#url-full), so would need to also be gated behind `should_send_default_pii()`

tests/integrations/sanic/test_sanic.py
```diff
@@ -408,57 +421,124 @@ def test_transactions(
...
-        test_config.expected_transaction_name is None
-    )
+    sentry_sdk.flush()
+
+    if span_streaming:
+        segments = [
+            i.payload
+            for i in items
+            if i.type == "span"
+            and i.payload["attributes"].get("sentry.origin") == "auto.http.sanic"
```
ericapisani: Because this is specifically looking for segments, this filter may be too broad since any child span originating from the integration would be returned. To be safe, it might be worth adding a check on the `_is_segment` property and confirming that it's `True`

ericapisani (commented):
Some small things but otherwise LGTM. Approving so as not to block it from merging

sentry_sdk/integrations/sanic.py
alexander-alderman-webb: I think the PII designation in conventions is enforced in Relay and not the SDK

sentry_sdk/integrations/sanic.py
sentrivana:
```suggestion
                span.end()
```

sentry_sdk/integrations/sanic.py
sentrivana, line 251: There's a `SegmentSource` too for the streaming path. Would be good to use that if span streaming is enabled

sentrivana, line 328: Here too we should do `SegmentSource` if streaming

sentrivana, line 332: And here

sentrivana (approved):
Looks good, couple small things `_end` -> `end` and `TransactionSource` -> `SegmentSource`

tests/integrations/sanic/test_sanic.py
sentrivana: If you do `capture_items("span")`, there won't be anything but spans in the list, so no need to check explicitly
```suggestion
            if i.payload["attributes"].get("sentry.origin") == "auto.http.sanic"
```

tests/integrations/sanic/test_sanic.py
```diff
@@ -408,57 +421,124 @@ def test_transactions(
...
-    (_, event) = events
+    sentry_sdk.flush()
 
-    assert event["contexts"]["trace"]["origin"] == "auto.http.sanic"
+    if span_streaming:
+        (segment,) = [
+            i.payload
+            for i in items
+            if i.type == "span"
+            and i.payload["attributes"].get("sentry.origin") == "auto.http.sanic"
```
sentrivana:
```suggestion
            if i.payload["attributes"].get("sentry.origin") == "auto.http.sanic"
```

sentry_sdk/integrations/sanic.py
sl0thentr0py: well the method itself is called `set_transaction_name`. If we want to clean this up properly, we should make a segment based method, just switching the source is half-assed

sentry_sdk/integrations/sanic.py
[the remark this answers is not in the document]
sl0thentr0py: intentional, see PR description

sentry_sdk/integrations/sanic.py
sl0thentr0py: I'll merge this for now, let's clean this up separately all together

PR #6385, opened 2026-05-21
test(pyramid): Add span streaming support to integration tests

### LINKED ISSUE #6048: Migrate pyramid to span first

ericapisani:
This integration gets span streaming automatically via the SentryWsgiMiddleware and the branching logic in helper methods like `set_transaction_name`.

Added test coverage to confirm span streaming works as expected.

Fixes PY-2350
Fixes #6048

### DIFF
--- a/sentry_sdk/integrations/pyramid.py
+++ b/sentry_sdk/integrations/pyramid.py
@@ -8,7 +8,9 @@ from sentry_sdk.integrations import DidNotEnable, Integration
 from sentry_sdk.integrations._wsgi_common import RequestExtractor
 from sentry_sdk.integrations.wsgi import SentryWsgiMiddleware
 from sentry_sdk.scope import should_send_default_pii
-from sentry_sdk.tracing import SOURCE_FOR_STYLE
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb (approved):

sentrivana (commented):
Can't comment on the diff since there is none for the main integration file, so just a separate comment:

- `_set_transaction_name_and_source` technically works as is, but would be nice to use `SegmentSource` instead of `TransactionSource` in it when span streaming. There's a streaming `SOURCE_FOR_STYLE` equivalent in `traces.py`.
- The integration defines an `event_processor`; is the data that would be set there set on the span elsewhere (e.g. WSGI), or is there anything we're missing?

ericapisani:
@sentrivana 

> _set_transaction_name_and_source technically works as is, but would be nice to use SegmentSource instead of TransactionSource in it when span streaming. There's a streaming SOURCE_FOR_STYLE equivalent in traces.py.

👍🏻 

> The integration defines an event_processor; is the data that would be set there set on the span elsewhere (e.g. WSGI), or is there anything we're missing?

The event processor adds the parsed request body to the event, so we're technically missing data, but given we had discussed not adding the `http.request.body.data` span attribute for now on streamed spans, I made no changes there.

Issue #6402, opened 2026-05-26
Django cache.add operations do not emit cache spans


immanuwell:
### Problem

`DjangoIntegration(cache_spans=True)` traces `cache.set()` and `cache.get()`, but `cache.add()` emits no cache span.

`cache.add()` is a normal Django cache write API, so this shows up in practice for apps that use add-on-miss / write-if-absent flows.

### Repro

```py
from django.core.cache import cache
import sentry_sdk
from sentry_sdk.integrations.django import DjangoIntegration

sentry_sdk.init(
    integrations=[DjangoIntegration(cache_spans=True)],
    traces_sample_rate=1.0,
)

with sentry_sdk.start_transaction(name="t", op="test"):
    cache.add("k", "value")
```

Actual: no `cache.put` span for `cache.add()`.

Expected: `cache.add()` should be traced like the other cache write methods.

### Notes

I verified locally that `cache.set()` emits a span while `cache.add()` does not, and a targeted fix is just to instrument `add` as a cache write.

PR #6403, opened 2026-05-26
fix(django): trace cache.add operations

immanuwell:
Refs #6402

`cache.add()` was a blind spot. `cache.set()` emitted `cache.put`, `cache.add()` emitted nothing. kinda odd for a normal Django cache write path.

Repro
```py
from django.core.cache import cache
import sentry_sdk

sentry_sdk.init(
    integrations=[DjangoIntegration(cache_spans=True)],
    traces_sample_rate=1.0,
)

with sentry_sdk.start_transaction(name="t", op="test"):
    cache.add("k", "value")
```

Before this patch: no cache span.
After this patch: one `cache.put` span for `cache.add()`.

What changed
- instrument `cache.add()`
- treat it as `cache.put`
- add a regression test for static + stream mode

Tested
- `tox -e py3.12-django-v4.2.30 -- -k test_cache_spans_add`

I also checked a wider Django cache test slice, but this env doesnt have the repo's local Postgres test service, so the db-marked Django tests bail early.

### DIFF
--- a/tests/integrations/django/test_cache_module.py
+++ b/tests/integrations/django/test_cache_module.py
@@ -4,6 +4,7 @@ import uuid
 
 import pytest
 from django import VERSION as DJANGO_VERSION
+from django.core.cache import cache
 from werkzeug.test import Client
 
 try:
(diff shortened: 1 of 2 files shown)


tests/integrations/django/test_cache_module.py
```diff
@@ -542,6 +543,56 @@ def test_cache_spans_templatetag(
...
+            )
+        ],
+        traces_sample_rate=1.0,
+        _experiments={"trace_lifecycle": "stream" if span_streaming else "static"},
+    )
+
+    if span_streaming:
+        items = capture_items("span")
+
+        with sentry_sdk.start_transaction(name="cache-add", op="test"):
```
ericapisani, line 570: In the context of span streaming, we no longer have the concept of a transaction, we instead have a segment.

Because of that, we shouldn't be invoking this method, but if we need to create a span to attach the generated cache span to, use `with sentry_sdk.traces.start_span(...)` instead.

The assertions from lines 575 to 579 will likely need to be updated as a result, as there will be another span caught by the  `capture_items` line above, so the index value used will change.

ericapisani (changes requested):
Thanks for opening this PR! These changes are overall looking good, just need to update something in the span streaming test before we look to get it across the finish line.

tests/integrations/django/test_cache_module.py
alexander-alderman-webb: Yes, just adapt the test to follow the existing pattern in the same file @immanuwell

alexander-alderman-webb:
Please merge master into your branch as well to resolve unrelated CI failures.

PR #6404, opened 2026-05-26
feat(openai-agents): Support span streaming

### LINKED ISSUE #6041: Migrate openai_agents to span first

alexander-alderman-webb:
> **Note: Depends on https://github.com/getsentry/sentry-python/pull/6424**

### Description

Use `sentry_sdk.traces.start_span,` replace `Span.set_data()` with `StreamedSpan.set_attribute()` and `Span.set_status(SPANSTATUS.INTERNAL_ERROR)` with `StreamedSpan.status = SpanStatus.ERROR` when in span streaming mode.

Parametrize tests on the trace lifecycle option.

#### Issues

Closes https://github.com/getsentry/sentry-python/issues/6041

#### Reminders


sentry_sdk/integrations/openai_agents/patches/runner.py
```diff
@@ -43,9 +44,15 @@ async def wrapper(*args: "Any", **kwargs: "Any") -> "Any":
...
-                    )
+
+                    if isinstance(workflow_span, StreamedSpan):
+                        workflow_span.set_attribute(
+                            SPANDATA.GEN_AI_CONVERSATION_ID, conversation_id
+                        )
+                    else:
+                        workflow_span.set_data(
+                            SPANDATA.GEN_AI_CONVERSATION_ID, conversation_id
+                        )
```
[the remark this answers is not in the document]
alexander-alderman-webb: https://github.com/getsentry/sentry-python/pull/6404/commits/53a84ae504adfec8ee56c27bcf40ebfbc24a3443

sentry_sdk/integrations/openai_agents/utils.py
```diff
@@ -173,9 +191,9 @@ def _set_input_data(
     client = sentry_sdk.get_client()
     scope = sentry_sdk.get_current_scope()
     messages_data = (
-        normalized_messages
-        if client.options.get("stream_gen_ai_spans", False)
-        else truncate_and_annotate_messages(normalized_messages, span, scope)
+        truncate_and_annotate_messages(normalized_messages, span, scope)
+        if should_truncate_gen_ai_input(client.options)
+        else normalized_messages
```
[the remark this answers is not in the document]
alexander-alderman-webb: https://github.com/getsentry/sentry-python/pull/6404/commits/dcdeb8dd77e2aa4ded7afe8d5aa034fbeae2dc1d

sentry_sdk/integrations/openai_agents/patches/agent_run.py
```diff
@@ -108,7 +114,13 @@ async def _run_single_turn(
         context_wrapper, agent, should_run_agent_start_hooks, kwargs
     )
 
-    if span is None or span.timestamp is not None:
+    if (
+        span is None
+        or isinstance(span, StreamedSpan)
+        and span.end_timestamp is not None
+        or not isinstance(span, StreamedSpan)
+        and span.timestamp is not None
```
ericapisani: This is a bit difficult to read and I think could lead to an accidental bug because of orders or precedence with the `and` and `or` keywords (the former taking evaluation precedence over the latter).

Reading this, I'm  not sure if the intention was to have the statement evaluate as `isinstance(span, StreamedSpan) and span.end_timestamp is not None` rather than `(span is None or isinstance(span, StreamedSpan)) and ...`

sentry_sdk/integrations/openai_agents/patches/agent_run.py
```diff
@@ -188,7 +200,13 @@ async def _run_single_turn_streamed(
         is_streaming=True,
     )
 
-    if span is None or span.timestamp is not None:
+    if (
+        span is None
+        or isinstance(span, StreamedSpan)
+        and span.end_timestamp is not None
+        or not isinstance(span, StreamedSpan)
+        and span.timestamp is not None
```
ericapisani: Same situation here as above

sentry_sdk/integrations/openai_agents/spans/ai_client.py
```diff
@@ -28,13 +30,24 @@ def ai_client_span(
...
-        op=OP.GEN_AI_CHAT,
-        name=f"chat {model_name}",
-        origin=SPAN_ORIGIN,
-    )
-    # TODO-anton: remove hardcoded stuff and replace something that also works for embedding and so on
-    span.set_data(SPANDATA.GEN_AI_OPERATION_NAME, "chat")
+    span_streaming = has_span_streaming_enabled(sentry_sdk.get_client().options)
+    if span_streaming:
+        span = sentry_sdk.traces.start_span(
+            name=f"chat {model_name}",
```
ericapisani, line 36: Not something to change in this PR, but looking at this line - `model_name` seems like something we may want to eventually set as a span attribute 🤔

sentry_sdk/integrations/openai_agents/spans/ai_client.py
```diff
@@ -28,13 +30,24 @@ def ai_client_span(
...
+                SPANDATA.GEN_AI_OPERATION_NAME: "chat",
+            },
+        )
+    else:
+        span = sentry_sdk.start_span(
+            op=OP.GEN_AI_CHAT,
+            name=f"chat {model_name}",
+            origin=SPAN_ORIGIN,
+        )
+        # TODO-anton: remove hardcoded stuff and replace something that also works for embedding and so on
```
ericapisani, line 49: I'm not sure how useful this comment is anymore - any chance we can remove it?

sentry_sdk/integrations/openai_agents/spans/handoff.py
```diff
@@ -12,14 +13,29 @@
...
-    with sentry_sdk.start_span(
-        op=OP.GEN_AI_HANDOFF,
-        name=f"handoff from {from_agent.name} to {to_agent_name}",
-        origin=SPAN_ORIGIN,
-    ) as span:
-        span.set_data(SPANDATA.GEN_AI_OPERATION_NAME, "handoff")
+    span_streaming = has_span_streaming_enabled(sentry_sdk.get_client().options)
+    if span_streaming:
+        with sentry_sdk.traces.start_span(
+            name=f"handoff from {from_agent.name} to {to_agent_name}",
```
ericapisani, line 19: We have a `gen_ai.agent.name` [attribute](https://getsentry.github.io/sentry-conventions/attributes/gen_ai/#gen_ai-agent-name) and I'm thinking this should maybe be added as an attribute with a value of `from_agent.name` (since it's the originator of the span in terms of the action being taken)

sentry_sdk/integrations/openai_agents/spans/handoff.py
```diff
@@ -12,14 +13,29 @@
...
+            attributes={
+                "sentry.op": OP.GEN_AI_HANDOFF,
+                "sentry.origin": SPAN_ORIGIN,
+                SPANDATA.GEN_AI_OPERATION_NAME: "handoff",
```
ericapisani, line 23: Something for a future change - we should replace these hard coded operation strings with an enum so we can keep track of the possible options that this can be.

sentry_sdk/integrations/openai_agents/spans/invoke_agent.py
```diff
@@ -9,29 +9,45 @@
...
-        origin=SPAN_ORIGIN,
-    )
-    span.__enter__()
-
-    span.set_data(SPANDATA.GEN_AI_OPERATION_NAME, "invoke_agent")
+) -> "Union[sentry_sdk.tracing.Span, StreamedSpan]":
+    span_streaming = has_span_streaming_enabled(sentry_sdk.get_client().options)
+    if span_streaming:
+        span = sentry_sdk.traces.start_span(
+            name=f"invoke_agent {agent.name}",
```
ericapisani, line 34: This seems like a good use case for the [gen_ai-agent-name](https://getsentry.github.io/sentry-conventions/attributes/gen_ai/#gen_ai-agent-name) attribute - is there a reason that we're not setting it here?

tests/integrations/openai_agents/test_openai_agents.py
```diff
@@ -172,7 +175,71 @@ async def test_agent_invocation_span_no_pii(
...
+
+            result = await agents.Runner.run(
+                agent, "Test input", run_config=test_run_config
+            )
+
+            assert result is not None
+            assert result.final_output == "Hello, how can I help you?"
+
+        sentry_sdk.flush()
+        spans = [item.payload for item in items]
```
ericapisani, line 218: Even if we're only interested in the invoke agent span and ai client span, we should still assert the number of expected spans here since this should be stable. If there are more/less spans, we'd want this test to fail

tests/integrations/openai_agents/test_openai_agents.py
```diff
@@ -841,29 +1017,213 @@ async def test_agent_invocation_span(
...
+            )
+        elif instructions is None:  # type: ignore
+            assert ai_client_span["data"][
+                "gen_ai.system_instructions"
+            ] == safe_serialize(
+                [
+                    {"type": "text", "content": "You are a helpful assistant."},
+                    {"type": "text", "content": "Be concise and clear."},
+                ]
+            )
```
ericapisani: This is a large conditional block that's not the easiest to read because of there sometimes being only `param_id`, sometimes both `param_id` and `instructions`, and the one case with `instructions`. 

Can we look to re-organize this/break it up a bit? Maybe have a general conditional grouping to cover cases where `instructions` are None with the relevant combos with `param_id` in there, and the other cases involving `param_id` as a separate grouping?

tests/integrations/openai_agents/test_openai_agents.py
```diff
@@ -964,7 +1332,69 @@ def test_agent_invocation_span_sync_no_pii(
...
+        assert spans[2]["attributes"]["sentry.origin"] == "auto.ai.openai_agents"
+
+        invoke_agent_span = next(
+            span
+            for span in spans
+            if span["attributes"]["sentry.op"] == OP.GEN_AI_INVOKE_AGENT
+        )
+        ai_client_span = next(
+            span for span in spans if span["attributes"]["sentry.op"] == OP.GEN_AI_CHAT
+        )
```
ericapisani, line 948: As I'm reviewing, I'm noticing some test cases with this pattern, and other test cases with [this pattern](https://github.com/getsentry/sentry-python/pull/6404/changes#diff-3994ad06a0ebc1f49b0088b4c7210c4804d91e21c4886e98cf02b0da097dc06fR1656).

Is there a reason why some are implemented one way vs the other? Is it to avoid assigning spans to variables that won't be asserted on?

tests/integrations/openai_agents/test_openai_agents.py
```diff
@@ -4742,7 +6063,53 @@ def simple_tool(message: str) -> str:
...
+        for span in spans:
+            if span.get("name", "").startswith("execute_tool"):
+                tool_span = span
+                break
+
+        assert tool_span is not None
+        # Tool span should have the conversation_id passed to Runner.run()
+        assert tool_span["attributes"]["gen_ai.conversation.id"] == "conv_tool_test_456"
+
+        # Workflow span (transaction) should have the same conversation_id
```
ericapisani: Should this (and the comment below) be updated to read `segment`, not `transaction`?

ericapisani (changes requested):
Overall looking good but there's a few things around the potential for some bugs with a couple of the conditionals that were updated, and the inconsistent way that we're processing spans in tests (some aren't asserting the total amount of spans and the order  in which they appear, others are).

sentry_sdk/integrations/openai_agents/patches/agent_run.py
alexander-alderman-webb: Fair point about lack of intentionality.
The intention is for the condition to evaluate to true if the end timestamp has not yet been set, indicating that the span is still active. The complex conditional is used because `Span.timestamp` and `StreamedSpan.end_timestamp` represent the end timestamp (i.e., because the instance variable was renamed).
I've added brackets in https://github.com/getsentry/sentry-python/pull/6404/commits/e243b8617e5cd868fa6abfc08b1ed55d002ade24

sentry_sdk/integrations/openai_agents/patches/agent_run.py
alexander-alderman-webb: Replied in https://github.com/getsentry/sentry-python/pull/6404#discussion_r3304898706

sentry_sdk/integrations/openai_agents/spans/ai_client.py
alexander-alderman-webb: The logic for getting the request model name is duplicated and we already the set the attribute in https://github.com/getsentry/sentry-python/blob/53188b322fc96dab0c4b064945b62e773fada63c/sentry_sdk/integrations/openai_agents/utils.py#L65

alexander-alderman-webb: Here's an issue to track: https://github.com/getsentry/sentry-python/issues/6415

sentry_sdk/integrations/openai_agents/spans/handoff.py
alexander-alderman-webb: Yup, created https://github.com/getsentry/sentry-python/issues/6416

sentry_sdk/integrations/openai_agents/spans/invoke_agent.py
alexander-alderman-webb: It's set within `_set_agent_name()`, called inside `invoke_agent_span()`.

https://github.com/getsentry/sentry-python/blob/53188b322fc96dab0c4b064945b62e773fada63c/sentry_sdk/integrations/openai_agents/utils.py#L45-L50

I didn't directly pass it to the `attributes` parameter in `start_span()` because that would violate the expectation that `_set_agent_data()` sets all agent-related attributes on the span.

tests/integrations/openai_agents/test_openai_agents.py
alexander-alderman-webb: I have no idea what the authors of the integration were thinking. I have not made a conscious decision in favor of one or the other style.

tests/integrations/openai_agents/test_openai_agents.py
alexander-alderman-webb: nice find, there probably shouldn't be two duplicated comments to begin with 🤦 .

Deduplicated and dropped the transaction reference in https://github.com/getsentry/sentry-python/pull/6404/commits/931aa30e843f6c8848d13920508f89f7e403950f

sentry_sdk/integrations/openai_agents/spans/handoff.py
alexander-alderman-webb: I'd have the same instinct to add an attribute to avoid users matching on the name.

However, we're removing the workflow span entirely per [the RFC](https://github.com/getsentry/rfcs/blob/main/text/0153-decoupling-sentrys-generative-ai-conventions-from-open-telemetry.md).

sentry_sdk/integrations/openai_agents/spans/ai_client.py
alexander-alderman-webb: I think this is still a bug 😬.
Created https://github.com/getsentry/sentry-python/issues/6417 to track.

tests/integrations/openai_agents/test_openai_agents.py
alexander-alderman-webb: I agree that it's hard to read and needs improvement.
The underlying logic is complex because instructions can be provided both in the input array and in the dedicated `instructions` parameter when using the Responses API.
I would like to avoid increasing indentation though.

tests/integrations/openai_agents/test_openai_agents.py
alexander-alderman-webb: I'd consider this out of scope for this PR because the existing test did not assert the total number of spans the goal of the PR is to port the integration to support span streaming (and make the same test guarantees as before).

tests/integrations/openai_agents/test_openai_agents.py
alexander-alderman-webb: I'll just split up some tests and make better use of `pytest.mark.parametrize` (only tomorrow though).

tests/integrations/openai_agents/test_openai_agents.py
ericapisani: I'm not sure if it's always been the intention to open another pull request to introduce the stricter checks, but if so, please add that in the pull request descriptions going forward so it's clear if it's an intentional decision or an accidental oversight.<br><br>As we've been introducing these stricter checks as part of other streamed span integration migrations and code review comments have been left when it's not been done, it appears to me that these stricter checks are part of the migration process, not out of scope.

In the case of this change set it's especially noticeable because it's being applied in some places and not others.

alexander-alderman-webb: I was neither intending to open another PR nor was this an accidental oversight.

Going by the migration guide in https://github.com/getsentry/sentry-python/issues/5386#issue-3869822962 I'm adding test code equivalent to the tests for the transaction based traces. (emphasis mine below).

> Parametrize all tests that deal with tracing to execute both in legacy and in span streaming mode. **Add equivalent assertions in streaming mode**. You can use the capture_items fixture for unwrapping envelope item items automatically.

The two instances this has been brought up in reviews that I'm aware of (https://github.com/getsentry/sentry-python/pull/6206#discussion_r3194484390 and https://github.com/getsentry/sentry-python/pull/6123#discussion_r3146010068) were specifically pointing out looser assertions compared to the legacy path.

Based on the information I have seen, broader changes to the tests are not part of the span first migration. If they are let's update the integration guide.

100% agree that the change set is a pain to read (and it's frustrating that code written so recently has so much tech debt).

tests/integrations/openai_agents/test_openai_agents.py
alexander-alderman-webb: I've created https://github.com/getsentry/sentry-python/pull/6424 and made this PR dependent on the test cleanup.

sentry_sdk/integrations/openai_agents/patches/runner.py
```diff
@@ -61,8 +68,10 @@ async def wrapper(*args: "Any", **kwargs: "Any") -> "Any":
                                 context_wrapper, "_sentry_agent_span", None
                             )
 
-                            if (
-                                invoke_agent_span is not None
+                            if invoke_agent_span is not None and (
+                                isinstance(invoke_agent_span, StreamedSpan)
+                                and invoke_agent_span.end_timestamp is None
+                                or not isinstance(invoke_agent_span, StreamedSpan)
```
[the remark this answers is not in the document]
alexander-alderman-webb: and precedence is well-defined but just to make people happy: https://github.com/getsentry/sentry-python/pull/6404/changes/0b4ed5fe249efe2a93e05c1a693d5134d7857df0

tests/integrations/openai_agents/test_openai_agents.py
sentrivana: With regards to porting tests, if the existing tests don't assert on the number/order of spans currently, fine to also not do that in scope of the span first conversion.

What I want to avoid during the migration is _losing_ assertions compared to the legacy path. If they weren't there in the first place, we don't need to add them. (It's def nice to have though, and can be done low-effort with a clanker.)

sentrivana (approved):

sentry_sdk/integrations/openai_agents/spans/execute_tool.py
```diff
@@ -3,39 +3,58 @@
...
+        span = sentry_sdk.traces.start_span(
+            name=f"execute_tool {tool.name}",
+            attributes={
+                "sentry.op": OP.GEN_AI_EXECUTE_TOOL,
+                "sentry.origin": SPAN_ORIGIN,
+                SPANDATA.GEN_AI_OPERATION_NAME: "execute_tool",
+                SPANDATA.GEN_AI_TOOL_NAME: tool.name,
+                SPANDATA.GEN_AI_TOOL_DESCRIPTION: tool.description,
+            },
+        )
```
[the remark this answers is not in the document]
alexander-alderman-webb: no tests for this of course 😞 .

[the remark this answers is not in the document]
alexander-alderman-webb: https://github.com/getsentry/sentry-python/pull/6404/changes/d3f3d5bb4d2bb5239f0603430585ac088acb842d

sentry_sdk/integrations/openai_agents/spans/agent_workflow.py
```diff
@@ -2,15 +2,28 @@
...
+) -> "Union[sentry_sdk.tracing.Span, sentry_sdk.traces.StreamedSpan]":
     # Create a transaction or a span if an transaction is already active
+    span_streaming = has_span_streaming_enabled(sentry_sdk.get_client().options)
+    if span_streaming:
+        span = sentry_sdk.traces.start_span(
+            name=f"{agent.name} workflow", attributes={"sentry.origin": SPAN_ORIGIN}
+        )
+
+        return span
+
```
ericapisani, line 26: Taking a look at `get_start_span_function`, it returns one of: 
- `sentry_sdk.traces.start_span`
- `sentry_sdk.start_span`
- `sentry_sdk.start_transaction`

depending on if span streaming is enabled or not or if there's a transaction that's currently active.

What we can do here in order to leverage the existing streamed span awareness within that function is to, instead of invoking `start_span` above, change how we're invoking the `get_start_span_function` based on whether span streaming is enabled or not.

So the function body would look something like the following:

```suggestion
span_streaming = has_span_streaming_enabled(sentry_sdk.get_client().options)
span_func = get_start_span_function()

if span_streaming:
    return span_func(
        name=f"{agent.name} workflow", 
        attributes={"sentry.origin": SPAN_ORIGIN}
    )
else:
    return span_func(
        name=f"{agent.name} workflow", 
        origin=SPAN_ORIGIN
    )
```

ericapisani (approved):
Other than my one comment, LGTM otherwise, approving so as not to block

sentry_sdk/integrations/openai_agents/spans/agent_workflow.py
alexander-alderman-webb: We don't need `get_start_span_function()` in span first anymore, since spans are automatically promoted to segment spans if there is no active segment span.

I've opened a PR to stop `get_start_span_function()` from returning `sentry_sdk.traces.start_span`:

https://github.com/getsentry/sentry-python/pull/6560

PR #6410, opened 2026-05-26
test(django): Remove forking in ASGI tests

sl0thentr0py:
Drop `@pytest.mark.forked` from the Django ASGI test suite and stop mutating the module-level `asgi_application` instance. On Python 3.12 in GitHub Actions' multi-threaded runner, `pytest-forked` forks a multi-threaded process; the asyncio event loop in the child inherits dead threads and silently hangs, consuming the full CI budget.

**Removed `@pytest.mark.forked`**

Removed from all 10 affected tests. The `sentry_init` fixture already handles SDK state save/restore correctly without forking.

**Fresh ASGI app per test**

Three tests called `asgi_application.load_middleware(is_async=True)`, which mutates the shared module-level instance and leaks state into subsequent tests. They now use a new `make_asgi_application` fixture that returns `get_asgi_application` — letting tests build a fresh app *after* `sentry_init`, so middleware instrumentation is installed before Django builds the chain.

**Concurrent execution tests**

Replaced the deprecated `asyncio.get_event_loop()` + `create_task` pattern with `asyncio.gather()`.

### Issues

* part of #4538

### DIFF
--- a/tests/integrations/django/asgi/test_asgi.py
+++ b/tests/integrations/django/asgi/test_asgi.py
@@ -30,9 +30,17 @@ if django.VERSION >= (3, 0):
     APPS += [asgi_application]
 
 
+@pytest.fixture
+def make_asgi_application():
+    """Build a fresh ASGI application. Call AFTER sentry_init so middleware
+    instrumentation is installed before Django builds the middleware chain."""
+    from django.core.asgi import get_asgi_application
+
+    return get_asgi_application
+
+
 @pytest.mark.parametrize("application", APPS)
 @pytest.mark.asyncio
-@pytest.mark.forked
 @pytest.mark.skipif(
     django.VERSION < (3, 0), reason="Django ASGI support shipped in 3.0"
 )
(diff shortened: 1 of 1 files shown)


tests/integrations/django/asgi/test_asgi.py
```diff
@@ -428,26 +421,25 @@ async def test_async_middleware_spans(
...
-  - op="event.django": description="django.db.reset_queries"
-  - op="event.django": description="django.db.close_old_connections"
   - op="middleware.django": description="django.contrib.sessions.middleware.SessionMiddleware.__acall__"
     - op="middleware.django": description="django.contrib.auth.middleware.AuthenticationMiddleware.__acall__"
       - op="middleware.django": description="django.middleware.csrf.CsrfViewMiddleware.__acall__"
         - op="middleware.django": description="tests.integrations.django.myapp.settings.TestMiddleware.__acall__"
           - op="middleware.django": description="django.middleware.csrf.CsrfViewMiddleware.process_view"
-          - op="view.render": description="simple_async_view"
-  - op="event.django": description="django.db.close_old_connections"
-  - op="event.django": description="django.core.cache.close_caches"
```
sl0thentr0py, line 444: the ordering of these is now non-deterministic so I'm simply removing checking them now, not what this test is about and they're tested elsewhere

sl0thentr0py:
agents failure unrelated

alexander-alderman-webb (approved):

PR #6434, opened 2026-05-28
fix(span-first): Set `user.ip_address` on all streamed spans

sentrivana:
`user.ip_address` should be attached to all spans as per https://develop.sentry.dev/sdk/telemetry/spans/span-protocol/#common-attribute-keys, not just segments. Set it on the isolation scope so that it gets applied to all spans active within it. Had to move setting the attr in a couple integrations so that it's set early enough for the child spans to also get it.


sentry_sdk/integrations/_asgi_common.py
```diff
@@ -133,6 +133,5 @@ def _get_request_attributes(asgi_scope: "Any") -> "dict[str, Any]":
     if client and should_send_default_pii():
         ip = _get_ip(asgi_scope)
         attributes["client.address"] = ip
-        attributes["user.ip_address"] = ip
```
sentrivana, line 136: This was moved to `asgi.py`

alexander-alderman-webb (approved):

PR #6436, opened 2026-05-28
chore: Move mypy to uv

sl0thentr0py:
keeping this in a separate dependency group since we install lots of random packages just for typing


sl0thentr0py:
test failure is due to new redis 8 release for rq, will fix separately, ignore here

alexander-alderman-webb (approved):

sentrivana (approved):

Issue #6022, opened 2026-04-14
Migrate gcp to span first


sentrivana:
See [parent issue](<https://github.com/getsentry/sentry-python/issues/5386>) for details.

PR #6440, opened 2026-05-28
feat(gcp): Add span streaming support to GCP Cloud Functions integration

ericapisani:
Migrate GCP integration to support spans-first tracing via the span streaming API.
The integration now supports both:
- Span streaming (new trace lifecycle)
- Traditional transaction-based approach (fallback)

Add request attributes to spans including headers, HTTP method, and query string.
Include cloud provider context and segment source in span metadata.

Add comprehensive tests for span streaming functionality including error handling,
trace context propagation, and PII filtering for query strings.

Fixes PY-2324 and #6022

### DIFF
--- a/sentry_sdk/integrations/gcp.py
+++ b/sentry_sdk/integrations/gcp.py
@@ -116,17 +116,31 @@ def _wrap_func(func: "F") -> "F":
                 "gcp_event": gcp_event,
             }
 
+            function_name = environ.get("FUNCTION_NAME", "<unknown GCP function>")
+
+            if environ.get("GCP_PROJECT"):
+                additional_attributes["gcp.project.id"] = environ.get("GCP_PROJECT")
+
+            if environ.get("FUNCTION_IDENTITY"):
+                additional_attributes["faas.identity"] = environ.get(
+                    "FUNCTION_IDENTITY"
+                )
+
+            if environ.get("ENTRY_POINT"):
+                additional_attributes["faas.entry_point"] = environ.get("ENTRY_POINT")
+
             if has_span_streaming_enabled(client.options):
                 sentry_sdk.traces.continue_trace(headers)
                 Scope.set_custom_sampling_context(sampling_context)
                 span_ctx = sentry_sdk.traces.start_span(
-                    name=environ.get("FUNCTION_NAME", "<unknown GCP function>"),
+                    name=function_name,
                     parent_span=None,
                     attributes={
                         "sentry.op": OP.FUNCTION_GCP,
                         "sentry.origin": GcpIntegration.origin,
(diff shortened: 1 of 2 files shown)


ericapisani:
bugbot run
@sentry review

sentry_sdk/integrations/gcp.py
```diff
@@ -102,9 +115,36 @@ def sentry_func(
...
             }
-            with sentry_sdk.start_transaction(
-                transaction, custom_sampling_context=sampling_context
-            ):
+
+            if has_span_streaming_enabled(client.options):
+                sentry_sdk.traces.continue_trace(headers)
+                Scope.set_custom_sampling_context(sampling_context)
+                span_ctx = sentry_sdk.traces.start_span(
+                    name=environ.get("FUNCTION_NAME", ""),
```
sentrivana: I know this is how it was before, but since span names are important in span first, maybe we can have a fallback name, something like:

```suggestion
                    name=environ.get("FUNCTION_NAME", "<unknown GCP function>"),
```

sentry_sdk/integrations/gcp.py
sentrivana, line 223: Some of these (name, entry point, identity, region, project, maybe execution time?) look like they'd make sense as attributes in span first. Don't know if OTel/we already have conventions for them, but might be worth adding.

ericapisani: Most of these don't have an OTEL equivalent but I've added a number of them to our conventions as part of https://github.com/getsentry/sentry-conventions/pull/403

There's also some that I've dropped, which I've outlined in a thread [here](https://sentry.slack.com/archives/C09RR0KD2N7/p1779976553604619)

The addition of the execution time will be an interesting one - I think we'd need to handle it similar to another integration that had some trickiness around the setting of a span's end timestamp after that span had been completed 🤔 

I've held off on adding this value for now though because I'm not convinced it's worth adding at this time - there's nothing that I could find in hex that indicates that users are creating alerts, dashboards, etc. with this value.

sentry_sdk/integrations/gcp.py
ericapisani: Addressed in [8f4fba1](https://github.com/getsentry/sentry-python/pull/6440/commits/8f4fba1f7a83bbd96fcb155ea4873a159bd3b3d2)

sentry_sdk/integrations/gcp.py
sentrivana: In full agreement re: execution time (not worth it right now) and the dropped attrs.

I was mostly curious about the ones you've added to the conventions -- why not set them on spans in this integration?

sentrivana (approved):
🙏🏻 🚀

PR #6445, opened 2026-05-29
fix(sampling): Attribute `backpressure` as unsampling reason more accurately

sentrivana:
#### Context

If backpressure handling is active, it lowers the effective sample rate.

There are different outcomes for when a span is unsampled due to `sample_rate` or `backpressure`.

#### Behavior

Before this change, if under backpressure, we'd only report `backpressure` as the reason for unsampling if the monitor lowered the effective sample rate to a falsy value. Which only covers a small portion of unsample decisions made due to backpressure. The rest would be reported as a generic `sample_rate` drop, potentially leading to confusing reporting. If someone e.g. has a `traces_sample_rate` of `1.0`, they might see spans being dropped because of `sample_rate`, which is hard to justify.

In this PR, we try to do a better job at determining whether backpressure was actually the reason a span was dropped and emit the corresponding outcome.

The idea is, for an unsampled span:
- if `sample_rand` would've been higher than the original, non-backpressure `sample_rate`, the span would've been unsampled regardless of backpressure handling, so the outcome should be `sample_rate`
- if `sample_rand` actually would've been lower than the original `sample_rate` had the `sample_rate` not been lowered, the span would've been sampled. In that case backpressure handling is the direct cause and the outcome should be `backpressure`

#### Reminders

### DIFF
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -1577,6 +1577,7 @@ def _make_sampling_decision(
         return False, 0.0, None, "sample_rate"
 
     # Adjust sample rate if we're under backpressure
+    sample_rate_before_backpressure = sample_rate
     if client.monitor:
         sample_rate /= 2**client.monitor.downsample_factor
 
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb (approved):
Cool code!

PR #6486, opened 2026-06-02
feat(bottle): Add span streaming support to Bottle integration

### LINKED ISSUE #6008: Migrate bottle to span first

ericapisani:
Add span streaming support to Bottle integration.

Fixes PY-2310
Fixes #6008


sentry_sdk/integrations/bottle.py
```diff
@@ -100,7 +102,38 @@ def _patched_handle(self: "Bottle", environ: "Dict[str, Any]") -> "Any":
...
+                with sentry_sdk.traces.start_span(
+                    name="bottle",
+                    attributes={"sentry.span.source": SegmentSource.COMPONENT},
+                ) as span:
+                    res = old_handle(self, environ)
+
+                    try:
+                        if integration.transaction_style == "url":
+                            span.name = bottle_request.route.rule or "bottle"
+                            span.set_attribute("sentry.span.source", SegmentSource.URL)
```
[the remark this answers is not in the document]
ericapisani: This is an intentional change and is being discussed with the team. Will update this tomorrow should we decide to keep this consistent with the transaction span approach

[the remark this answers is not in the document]
ericapisani: Addressed in [4d7da81](https://github.com/getsentry/sentry-python/pull/6486/commits/4d7da81c7bca11d2b4588e4a4e5436a197b5fe04) - forgot we had a segment-specific mapping that we can use here

sentry_sdk/integrations/bottle.py
```diff
@@ -100,7 +102,38 @@ def _patched_handle(self: "Bottle", environ: "Dict[str, Any]") -> "Any":
...
+                                bottle_request.route.name
+                                or transaction_from_function(
+                                    bottle_request.route.callback
+                                )
+                                or "bottle"
+                            )
+                            span.name = name
+                            span.set_attribute(
+                                "sentry.span.source", SegmentSource.COMPONENT
+                            )
```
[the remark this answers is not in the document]
ericapisani: Switched to using `set_transaction_name` method on the current scope which will address this in [4d7da81](https://github.com/getsentry/sentry-python/pull/6486/commits/4d7da81c7bca11d2b4588e4a4e5436a197b5fe04)

sentry_sdk/integrations/bottle.py
```diff
@@ -100,7 +102,38 @@ def _patched_handle(self: "Bottle", environ: "Dict[str, Any]") -> "Any":
...
+                # Using SegmentSource.COMPONENT as the default since this is the also value for "handler_name"
+                # within SOURCE_FOR_STYLE.
+                # See https://github.com/getsentry/sentry-python/blob/c3aab3932f5a7e89ad3aff551a206db710acf0e6/sentry_sdk/tracing.py#L151-L161
+                with sentry_sdk.traces.start_span(
+                    name="bottle",
+                    attributes={"sentry.span.source": SegmentSource.COMPONENT},
+                ) as span:
+                    res = old_handle(self, environ)
+
+                    try:
```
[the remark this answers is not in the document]
ericapisani: The specific route information isn't set until the handler is run. Attempting to access `bottle_request.route.rule` will result in an error if we try to access it to set route-specific information beforehand

sentry_sdk/integrations/bottle.py
```diff
@@ -10,7 +10,9 @@
 )
 from sentry_sdk.integrations._wsgi_common import RequestExtractor
 from sentry_sdk.integrations.wsgi import SentryWsgiMiddleware
-from sentry_sdk.tracing import SOURCE_FOR_STYLE
+from sentry_sdk.traces import SOURCE_FOR_STYLE as SEGMENT_SOURCE_FOR_STYLE
+from sentry_sdk.tracing import SOURCE_FOR_STYLE as TRANSACTION_SOURCE_FOR_STYLE
+from sentry_sdk.tracing_utils import has_span_streaming_enabled
```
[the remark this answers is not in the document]
ericapisani: This is a false positive, the tests currently pass

sentry_sdk/integrations/bottle.py
```diff
@@ -119,17 +143,33 @@ def patched_make_callback(
...
-                    _capture_exception(exception, handled=False)
-                    raise exception
-
-                if (
-                    isinstance(res, HTTPResponse)
-                    and res.status_code in integration.failed_request_status_codes
-                ):
-                    _capture_exception(res, handled=True)
+                if has_span_streaming_enabled(sentry_sdk.get_client().options):
+                    with sentry_sdk.traces.start_span(name="bottle"):
```
sentrivana: Why a new span here? Currently, the Bottle integration relies on the WSGI integration to create the main segment span and doesn't add any spans on top -- I'd keep the overall span architecture the same

sentry_sdk/integrations/bottle.py
```diff
@@ -100,7 +102,29 @@ def _patched_handle(self: "Bottle", environ: "Dict[str, Any]") -> "Any":
...
+                            or transaction_from_function(bottle_request.route.callback)
+                            or "bottle"
+                        )
+
+                    sentry_sdk.get_current_scope().set_transaction_name(
+                        name,
+                        source=SEGMENT_SOURCE_FOR_STYLE[integration.transaction_style],
+                    )
+                except RuntimeError:
+                    pass
```
sentrivana: I think this is a nice place for this logic 👍🏻 

Couple questions/comments:
- Maybe `Bottle request` would be a bit more specific since just `bottle` doesn't specify what kind of span it is. We typically create different types of spans in WSGI context (requests, middlewares etc.) so it'd help to narrow it down for nicer categorization in Sentry.
- Nit, but maybe we can move this block to a helper func similar to `_set_transaction_name_and_source`, maybe `_set_segment_name_and_source`?

sentry_sdk/integrations/bottle.py
ericapisani: It was initially introduced because I hadn't understood that this was invoked as part of the normal request cycle and the WSGI middleware would automatically capture this within a span. Will remove in the next commit 🫡

sentry_sdk/integrations/bottle.py
ericapisani: Addressed these in https://github.com/getsentry/sentry-python/pull/6486/commits/0d92dd7e3a283992e41a4e4aa77566db1f9f78ad

sentry_sdk/integrations/bottle.py
ericapisani: Addressed this in https://github.com/getsentry/sentry-python/pull/6486/commits/0d92dd7e3a283992e41a4e4aa77566db1f9f78ad

sentry_sdk/integrations/bottle.py
```diff
@@ -100,7 +102,16 @@ def _patched_handle(self: "Bottle", environ: "Dict[str, Any]") -> "Any":
...
                 _make_request_event_processor(self, bottle_request, integration)
             )
-            res = old_handle(self, environ)
+
+            if has_span_streaming_enabled(sentry_sdk.get_client().options):
+                res = old_handle(self, environ)
+
+                _set_segment_name_and_source(
+                    transaction_style=integration.transaction_style
+                )
```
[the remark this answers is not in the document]
ericapisani: See https://github.com/getsentry/sentry-python/pull/6486#discussion_r3343821542

sentry_sdk/integrations/bottle.py
```diff
@@ -100,7 +102,16 @@ def _patched_handle(self: "Bottle", environ: "Dict[str, Any]") -> "Any":
             scope.add_event_processor(
                 _make_request_event_processor(self, bottle_request, integration)
             )
-            res = old_handle(self, environ)
+
+            if has_span_streaming_enabled(sentry_sdk.get_client().options):
+                res = old_handle(self, environ)
```
sentrivana: Looks like we can take this line out of the condition bodies since it's the same for both branches

sentrivana (approved):
One small suggestion, but lgtm!

sentry_sdk/integrations/bottle.py
ericapisani: Good eye, will make that fix!

PR #6495, opened 2026-06-03
fix(langchain): filter GraphBubbleUp control-flow exceptions in _handle_error

### LINKED ISSUE #6384: LangchainIntegration reports GraphInterrupt / GraphBubbleUp as errors via on_tool_error
### Environment
- `sentry-sdk` 2.59.0 (also confirmed present on `master` / 2.60.0)
- `langchain-core` 1.x, `langgraph` 0.x
- LangChain integration auto-enabled (not explicitly added)

### Problem
When `langgraph.types.interrupt()` is called from inside a LangChain `StructuredTool` — the standard pattern for exposing human-in-the-loop pauses as agent tools (e.g. `deepagents`) — it raises `GraphInterrupt`, a subclass of `langgraph.errors.GraphBubbleUp`. This is **normal control flow**, not an error: LangGraph's pregel executor catches it upstream and checkpoints the graph.

However:
1. `langchain_core`'s `BaseTool.arun` does not special-case `GraphBubbleUp`. It catches it in the generic `except (Exception, KeyboardInterrupt)` and calls `run_manager.on_tool_error(error)` before re-raising.
2. Sentry's `SentryLangchainCallback.on_tool_error` -> `_handle_error` calls `sentry_sdk.capture_exception(error, ...)` with **no exception-type filtering**:

```python
def _handle_error(self, run_id, error):
        if not run_id or run_id not in self.span_map:
            return
        span_data = self.span_map[run_id]
        span = span_data.span
        sentry_sdk.capture_exception(error, span.scope)   # captures GraphInterrupt as an error
        ...
```

Result: every trace-sampled HITL interrupt becomes a false-positive error issue in Sentry. (Note the dedicated `LanggraphIntegration` is tracing-only and does not have this problem — only the LangChain integration does.)

### Expected
`GraphBubbleUp` exceptions (`GraphInterrupt`, etc.) are LangGraph control-flow signals and should not be captured as errors. `_handle_error` (or `on_tool_error`) should skip them:

```python
try:
    from langgraph.errors import GraphBubbleUp
except ImportError:
    GraphBubbleUp = ()

def _handle_error(self, run_id, error):
    if isinstance(error, GraphBubbleUp):
        return  # langgraph control flow, not an error
    ...
```

### Workaround
Consumers can pass `ignore_errors=[GraphBubbleUp]` to `sentry_sdk.init()`, but the integration shipping a correct default would be preferable.

wahajahmed010:
## Summary

LangGraph raises GraphBubbleUp / GraphInterrupt for normal control flow (e.g. langgraph.types.interrupt() for human-in-the-loop). Sentry's SentryLangchainCallback._handle_error was capturing these as errors via on_tool_error, creating false-positive Sentry issues.

## Changes

- Added a lazy module-level import of GraphBubbleUp from langgraph.errors (graceful fallback if langgraph is not installed)
- Added an isinstance guard in _handle_error to skip capturing GraphBubbleUp instances

## Testing

- The change is a simple guard addition; existing behavior is preserved for all non-GraphBubbleUp exceptions
- When langgraph is not installed, _graph_bubble_up defaults to empty tuple and no filtering is applied

Fixes #6384

AI assistance disclosure: this PR was co-authored by an AI assistant (OpenClaw Buck) under human supervision.

### DIFF
--- a/sentry_sdk/integrations/langchain.py
+++ b/sentry_sdk/integrations/langchain.py
@@ -117,6 +117,16 @@ try:
 except ImportError:
     OllamaEmbeddings = None
 
+# LangGraph raises GraphBubbleUp / GraphInterrupt for normal control flow
+# (human-in-the-loop interrupts, etc.) — these should never be captured as errors.
+_graph_bubble_up: "type | tuple[()]" = ()
+try:
+    from langgraph.errors import GraphBubbleUp  # type: ignore[import-untyped]
+
+    _graph_bubble_up = GraphBubbleUp
+except ImportError:
+    pass
+
 
 def _get_ai_system(all_params: "Dict[str, Any]") -> "Optional[str]":
     ai_type = all_params.get("_type")
@@ -264,6 +274,12 @@ class SentryLangchainCallback(BaseCallbackHandler):  # type: ignore[misc]
     def _handle_error(self, run_id: "UUID", error: "Any") -> None:
         with capture_internal_exceptions():
             if not run_id or run_id not in self.span_map:
+                return
+
+            # LangGraph raises GraphBubbleUp/GraphInterrupt for normal
+            # control flow (e.g. human-in-the-loop interrupts). These are
+            # not errors and should not be captured as Sentry issues.
+            if isinstance(error, _graph_bubble_up):
                 return
 
             span = self.span_map[run_id]


alexander-alderman-webb:
Closing in favor of https://github.com/getsentry/sentry-python/pull/6608.
We shouldn't import langgraph stuff in the langchain integration.

PR #6502, opened 2026-06-04
feat(quart): Add span streaming support to Quart integration

### LINKED ISSUE #6050: Migrate quart to span first

ericapisani:
Add span streaming support for the Quart integration when the
trace_lifecycle stream experiment is enabled. Sets HTTP request
attributes (method, headers, URL, query, client IP) on the segment
span and uses the correct source constant from sentry_sdk.traces
for span-first mode.

Depends on https://github.com/getsentry/sentry-python/pull/6501
being merged first.

Fixes PY-2352
Fixes #6050


sentry_sdk/integrations/quart.py
sentrivana: We'll need a streaming counterpart for this piece of code, something like:

```python

    span = get_current_span()
    if span is not None:
        span._segment._update_active_thread()
```

sentry_sdk/integrations/quart.py
```diff
@@ -169,6 +179,39 @@ async def _request_websocket_started(app: "Quart", **kwargs: "Any") -> None:
...
+                    request_websocket.query_string.decode("utf-8", errors="replace"),
+                )
+
+                if len(request_websocket.access_route) >= 1:
+                    segment.set_attribute(
+                        "client.address", request_websocket.access_route[0]
+                    )
+                    segment.set_attribute(
+                        "user.ip_address", request_websocket.access_route[0]
+                    )
```
sentrivana: We should also set `user.id` if we have it (essentially port `_add_user_to_event`).

It's one of the attributes that should be set on [all spans](https://develop.sentry.dev/sdk/telemetry/spans/span-protocol/#common-attribute-keys), not just the segment, so it should be set on the scope.

Maybe extract the code that fetches the user id from `_add_user_to_event` to a helper func, get rid of `_add_user_to_event` and set the user id in the event processor directly, and in the streaming path set it on the current scope?

```python
def _fetch_user_id() -> None:
    if quart_auth is None:
        return

    user = quart_auth.current_user
    if user is None:
        return

    try:
        return quart_auth.current_user._auth_id
    except Exception:
        return None
```

and then here

```python
user_id = _fetch_user_id()
if user_id is not None:
    sentry_sdk.get_current_scope().set_attribute("user.id", user_id)
```

and in the event processor

```python
user_id = _fetch_user_id()
if user_id is not None:
    user_info = event.setdefault("user", {})
    user_info["id"] = user_id
```

sentrivana: Please also add a test for this if there is none or add to an existing one.

ericapisani: Sounds good. Regarding needing to set the `user.id` on the scope - would that mean that we also need to make an adjustment for the `user.ip_address` as well? I noticed it also appears in that list that you had linked 👀

sentrivana: That's a really good point that made me realize we had a bug in the attribute setting code, fix [here](https://github.com/getsentry/sentry-python/pull/6555). Yup we should absolutely also be setting `ip_address`.

I need to amend my previous comment though: we should port setting the user but not the way I wrote. There's an existing mechanism for it (`set_user`) that we should use -- it sets the user data on the scope and then [this code](https://github.com/getsentry/sentry-python/blob/1aa2e6d4ee4eeaca99bc7841f107c0097573f2c4/sentry_sdk/scope.py#L1729-L1746) materializes it on the span eventually. The only thing to be mindful of is that `set_user` accepts a dict where the keys are different from the final attribute names (see the linked code), so it's e.g. `name` instead of `user.name` etc.

sentrivana: Another thing to note: `set_user` will completely overwrite whatever was there previously (it doesn't merge the dicts), so we need to collect all the fields we have and update once.

.github/workflows/test-integrations-dbs.yml
```diff
@@ -33,17 +33,30 @@ jobs:
...
+        ports:
+          - 3306:3306
     env:
       # 3.6/3.7 run in the python:X.Y container; tell uv to use that system Python.
       TOX_UV_PYTHON_PREFERENCE: ${{ (matrix.python-version == '3.6' || matrix.python-version == '3.7') && 'only-system' || 'managed' }}
       SENTRY_PYTHON_TEST_POSTGRES_HOST: ${{ (matrix.python-version == '3.6' || matrix.python-version == '3.7') && 'postgres' || 'localhost' }}
       SENTRY_PYTHON_TEST_POSTGRES_USER: postgres
       SENTRY_PYTHON_TEST_POSTGRES_PASSWORD: sentry
+      SENTRY_PYTHON_TEST_MYSQL_USER: root
+      SENTRY_PYTHON_TEST_MYSQL_PASSWORD: sentry
```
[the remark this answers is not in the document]
ericapisani: I don't know where this came from 🤔

sentry_sdk/integrations/quart.py
```diff
@@ -169,6 +185,52 @@ async def _request_websocket_started(app: "Quart", **kwargs: "Any") -> None:
...
+                )
+
+                user_properties = {}
+                if len(request_websocket.access_route) >= 1:
+                    segment.set_attribute(
+                        "client.address", request_websocket.access_route[0]
+                    )
+                    segment.set_attribute(
+                        "user.ip_address", request_websocket.access_route[0]
+                    )
```
sentrivana: These should be set in `user_properties` instead, like `ip_address` is -- if they're set on the scope via `set_user`, they'll be set on all spans (including the segment) automatically

sentry_sdk/integrations/quart.py
```diff
@@ -169,6 +185,52 @@ async def _request_websocket_started(app: "Quart", **kwargs: "Any") -> None:
...
+                current_user_id = _get_current_user_id_from_quart()
+                if current_user_id:
+                    user_properties["id"] = current_user_id
+
+                if user_properties:
+                    current_scope = sentry_sdk.get_current_scope()
+                    existing_user_properties = current_scope._user or {}
+                    current_scope.set_user(
+                        {**existing_user_properties, **user_properties}
+                    )
```
sentrivana: If we want to merge like this, this should interact with the isolation scope. The top-level `set_user` API sets the user on the isolation scope, so if there is any user data set by the user, it'll be on the isolation scope.

```suggestion
                    isolation_scope = sentry_sdk.get_isolation_scope()
                    existing_user_properties = isolation_scope._user or {}
                    isolation_scope.set_user(
                        {**existing_user_properties, **user_properties}
                    )
```

sentrivana (approved):
Two small things, please see comments, otherwise LGTM

sentry_sdk/integrations/quart.py
ericapisani: The `client.address` didn't appear on the list of [common attribute keys](https://develop.sentry.dev/sdk/telemetry/spans/span-protocol/#common-attribute-keys) which is why I didn't add it (not sure if this is potentially a mistake?) but at the very least I can clean up lines 218-220 since it's being added on the scope.

sentry_sdk/integrations/quart.py
ericapisani: Sounds good! I realized looking at this again that there's a reference to the isolation scope above so I'll use that here.

PR #6505, opened 2026-06-04
fix(utils): match_regex_list crashes on an empty-string pattern

### LINKED ISSUE #6504: match_regex_list raises IndexError when a pattern list contains an empty string

```python
```

devteamaegis:
## What's broken

`match_regex_list` checks `item_matcher[-1] != "$"` to decide whether to anchor each pattern. With the default `substring_matching=False`, an empty-string element makes `""[-1]` raise:

```python
>>> from sentry_sdk.utils import match_regex_list
>>> match_regex_list("anything", [""])
IndexError: string index out of range
```

`exclude_beat_tasks` in the Celery integration flows straight into this with the default mode, so an empty string in that user-supplied list crashes the beat instrumentation.

## Why it happens

Indexing `item_matcher[-1]` on an empty string is out of range.

## Fix

Short-circuit on empty: `(not item_matcher or item_matcher[-1] != "$")`. An empty matcher then becomes `"$"`, which matches end-of-string — a sane result instead of a crash.

## Test

Added a case asserting `match_regex_list("anything", [""])` returns without raising; existing anchored/substring behavior is unchanged.

Fixes #6504

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -1740,7 +1740,7 @@ def match_regex_list(
         return False
 
     for item_matcher in regex_list:
-        if not substring_matching and item_matcher[-1] != "$":
+        if not substring_matching and (not item_matcher or item_matcher[-1] != "$"):
             item_matcher += "$"
 
         matched = re.search(item_matcher, item)
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -1740,7 +1740,10 @@ def match_regex_list(
         return False
 
     for item_matcher in regex_list:
-        if not substring_matching and (not item_matcher or item_matcher[-1] != "$"):
+        if not item_matcher:
+            return False
+
+        if not substring_matching and item_matcher[-1] != "$":
             item_matcher += "$"
 
         matched = re.search(item_matcher, item)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -562,6 +562,14 @@ def test_match_regex_list(item, regex_list, expected_result):
     assert match_regex_list(item, regex_list) == expected_result
 
 
+def test_match_regex_list_empty_string_pattern():
+    # An empty-string pattern must not raise IndexError (regression test).
+    result = match_regex_list("anything", [""])
+    assert isinstance(result, bool)
+    assert match_regex_list("foobar", ["foo"]) is False
+    assert match_regex_list("foo", ["foo"]) is True
+
+
 @pytest.mark.parametrize(
     "version,expected_result",
     [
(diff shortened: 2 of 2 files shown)


tests/test_utils.py
```diff
@@ -562,6 +562,14 @@ def test_match_regex_list(item, regex_list, expected_result):
     assert match_regex_list(item, regex_list) == expected_result
 
 
+def test_match_regex_list_empty_string_pattern():
+    # An empty-string pattern must not raise IndexError (regression test).
+    result = match_regex_list("anything", [""])
```
ericapisani: Rather than have this be a standalone test, this case can be incorporated into the test above (`test_match_regex_list`) as one of the parameterized test cases. You'll just need to add the expected outcome as the last element in the array.

This means that last 2 assertions around foo/foobar can also be removed since we already have test cases that check similar behaviour within the array (lines 553 and 554 above)

sentry_sdk/utils.py
```diff
@@ -1740,7 +1740,7 @@ def match_regex_list(
         return False
 
     for item_matcher in regex_list:
-        if not substring_matching and item_matcher[-1] != "$":
+        if not substring_matching and (not item_matcher or item_matcher[-1] != "$"):
```
ericapisani: While this fixes the index error, the result that's returned from this change means `match_regex_list("anything", [""])` will return `True`.

An empty string likely represents a typo/mistake on the developers part, and so we should treat this as if `[]` was passed in.

This would mean that this conditional should look like the following instead:

```python
    for item_matcher in regex_list:
        if not item_matcher:
            return False

        if not substring_matching and item_matcher[-1] != "$":
            item_matcher += "$"

```

ericapisani (changes requested):
Thanks for opening the issue and pull request to address this.

There's a couple of things we'll need to address before this gets merged - let me know if you have any questions about what I've mentioned below.

devteamaegis:
Thanks for the review @ericapisani — both points addressed.

1. **Empty-pattern handling**: adopted your suggested conditional, so an empty `item_matcher` now hits `return False` before the anchoring logic. `match_regex_list("anything", [""])` returns False (treated like `[]`) instead of matching everything, which also resolves the Bugbot finding about `re.search("$", ...)` always matching and the `exclude_beat_tasks` fallout.
2. **Test**: folded the case into the parameterized `test_match_regex_list` as `["some-string", [""], False]` and removed the standalone test plus the redundant foo/foobar assertions already covered by existing cases.

Verified the empty-string list now returns False while normal patterns still match. Let me know if you'd like anything else.

alexander-alderman-webb:

PR #6509, opened 2026-06-06
fix: handle empty string in match_regex_list to prevent IndexError

### LINKED ISSUE #6504: match_regex_list raises IndexError when a pattern list contains an empty string

wahajahmed010:
### Description

When `match_regex_list` receives a pattern list containing an empty string, `item_matcher[-1]` raises `IndexError` because empty strings have no characters. This is reachable from the Celery integration when `CeleryIntegration.exclude_beat_tasks` contains an empty string.

Fix: skip empty strings in the iteration loop before the `[-1]` access.

#### Issues

* resolves: #6504

#### Reminders

- [x] Added tests to validate changes
- [x] Conventional commit format in PR title

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -1740,6 +1740,8 @@ def match_regex_list(
         return False
 
     for item_matcher in regex_list:
+        if not item_matcher:
+            continue
         if not substring_matching and item_matcher[-1] != "$":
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb:
Closing in favor of https://github.com/getsentry/sentry-python/pull/6896.

PR #6550, opened 2026-06-10
feat(cloud_resource_context): Support span streaming

### LINKED ISSUE #6012: Migrate cloud_resource_context to span first

alexander-alderman-webb:
### Description

Set data currently set on context as attributes on the isolation scope.

All attributes are in `sentry-conventions`, except for `host.id` and `host.type`.
Conventions PRs for these are approved:

- https://github.com/getsentry/sentry-conventions/pull/416
- https://github.com/getsentry/sentry-conventions/pull/417

Tests only check integration internals, so no tests are required to maintain the same test coverage as before.

#### Issues

Closes https://github.com/getsentry/sentry-python/issues/6012

#### Reminders

### DIFF
--- a/sentry_sdk/integrations/cloud_resource_context.py
+++ b/sentry_sdk/integrations/cloud_resource_context.py
@@ -3,6 +3,7 @@ from typing import TYPE_CHECKING
 
 import urllib3
 
+import sentry_sdk
 from sentry_sdk.api import set_context
 from sentry_sdk.integrations import Integration
 from sentry_sdk.utils import logger
@@ -262,6 +263,8 @@ class CloudResourceContextIntegration(Integration):
         context = CloudResourceContextIntegration._get_cloud_resource_context()
         if context != {}:
             set_context(CONTEXT_TYPE, context)
+            for k, v in context.items():
+                sentry_sdk.get_isolation_scope().set_attribute(k, v)
 
 
 # Map with the currently supported cloud providers


sentry_sdk/integrations/cloud_resource_context.py
```diff
@@ -262,6 +263,8 @@ def setup_once() -> None:
         context = CloudResourceContextIntegration._get_cloud_resource_context()
         if context != {}:
             set_context(CONTEXT_TYPE, context)
+            for k, v in context.items():
+                sentry_sdk.get_isolation_scope().set_attribute(k, v)
```
sentrivana, line 267: This will set the attributes on all spans. Given that they're not in the [common attrs list](https://develop.sentry.dev/sdk/telemetry/spans/span-protocol/#common-attribute-keys), they should probably be set on the segment of the current span instead?

alexander-alderman-webb: Yes, the attributes should be on the segment span.
Now I look at this, I'm not sure how that would work because the context is set in `setup_once()`.
I guess we would need some mechanism to set attributes on a scope that are then only applied on the segment span.
Alternatively, we consider not migrating the integration to span first.

sentrivana: Right, missed that we're in `setup_once` here.

I would leave this integration out for now. I believe it'd make sense for this to live in the aws/gcp integrations eventually, but as is, I'd not migrate. Errors will still get the context and that's probably where it's most useful.

PR #6566, opened 2026-06-12
fix(flask): Set user data on scope at request start

### LINKED ISSUE #6565: Ensure user data is set on all spans within Flask
Related: getsentry/sentry-python#6201

ericapisani:
Previously, Flask user data (id, email, username) was only set via the `_add_user_to_event` event processor, which runs at event capture time. Under span streaming, spans are sent before an error event is captured, so user attributes were missing from those spans.

This refactors the user property logic into a shared `_get_flask_user_properties()` helper and calls it in `_request_started` to set user data directly on the isolation scope. The event processor continues to use the same helper to keep behavior consistent for non-streaming cases. Tests are extended to cover both span streaming and non-streaming paths.

Fixes GH-6565
Refs PY-2528

### DIFF
--- a/sentry_sdk/integrations/flask.py
+++ b/sentry_sdk/integrations/flask.py
@@ -155,6 +155,14 @@ def _request_started(app: "Flask", **kwargs: "Any") -> None:
     )
 
     scope = sentry_sdk.get_isolation_scope()
+
+    with capture_internal_exceptions():
+        if should_send_default_pii():
+            user_properties = _get_flask_user_properties()
+            if user_properties:
+                existing_user_properties = scope._user or {}
+                scope.set_user({**existing_user_properties, **user_properties})
+
     evt_processor = _make_request_event_processor(app, request, integration)
     scope.add_event_processor(evt_processor)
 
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/flask.py
```diff
@@ -223,43 +231,50 @@ def _capture_exception(
...
 
-        # The following attribute accesses are ineffective for the general
-        # Flask-Login case, because the User interface of Flask-Login does not
-        # care about anything but the ID. However, Flask-User (based on
-        # Flask-Login) documents a few optional extra attributes.
-        #
-        # https://github.com/lingthio/Flask-User/blob/a379fa0a281789618c484b459cb41236779b95b1/docs/source/data_models.rst#fixed-data-model-property-names
+    try:
+        properties["username"] = user.username
+    except Exception:
```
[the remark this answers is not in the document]
ericapisani: I wasn't able to find something from my google search on this to confirm if this is true or not (that the email and username could potentially be a None/null value) so I've added the guard out of an abundance of caution

sentry_sdk/integrations/flask.py
```diff
@@ -155,6 +155,14 @@ def _request_started(app: "Flask", **kwargs: "Any") -> None:
     )
 
     scope = sentry_sdk.get_isolation_scope()
+
+    with capture_internal_exceptions():
+        if should_send_default_pii():
```
alexander-alderman-webb: To match the ordering in `_make_request_event_processor()` and avoiding the `capture_internal_exceptions()` block from becoming too broad
```suggestion
    if should_send_default_pii():
        with capture_internal_exceptions():
```

sentry_sdk/integrations/flask.py
```diff
@@ -155,6 +155,14 @@ def _request_started(app: "Flask", **kwargs: "Any") -> None:
...
+            user_properties = _get_flask_user_properties()
+            if user_properties:
+                existing_user_properties = scope._user or {}
+                scope.set_user({**existing_user_properties, **user_properties})
```
alexander-alderman-webb: Since the event processor path also overwrites the user.
```suggestion
            scope.set_user(user_properties)
```

alexander-alderman-webb (approved):

PR #6570, opened 2026-06-15
fix: Remove 0000 trace_id fallbacks

### LINKED ISSUE #6558: Remove 0000 trace fallback
In metrics and logs, we currently have a fallback for when there is no trace ID in the current trace context. We [set](<https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/scope.py#L1864>) the trace_id since it's a required field to a dummy value.

This needs to be removed as it's causing issues (the nil trace ID telemetry is getting clustered into a huge trace) and there should never be a case where there is no trace ID.

We believe we shouldn't be hitting the fallback anymore in newer SDK versions due to the recent propagation context handling refactor but we need to verify if that's the case and then remove the fallback.

- [ ] find out when we hit the fallback and fix the code path so that it doesn't happen anymore
- [ ] remove the fallback

sl0thentr0py:
#### Issues

* resolves: #6558
* resolves: PY-2526

### DIFF
--- a/sentry_sdk/scope.py
+++ b/sentry_sdk/scope.py
@@ -622,7 +622,7 @@ class Scope:
         if (
             has_tracing_enabled(self.get_client().options)
             and self._span is not None
-            and not isinstance(self._span, NoOpStreamedSpan)
+            and not isinstance(self._span, (NoOpStreamedSpan, NoOpSpan))
         ):
             return self._span._get_trace_context()
 
@@ -693,7 +693,7 @@ class Scope:
         if (
             has_tracing_enabled(client.options)
             and span is not None
-            and not isinstance(span, NoOpStreamedSpan)
+            and not isinstance(span, (NoOpStreamedSpan, NoOpSpan))
         ):
             for header in span._iter_headers():
                 yield header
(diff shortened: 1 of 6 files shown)


sentry_sdk/scope.py
```diff
@@ -622,7 +622,7 @@ def get_trace_context(self) -> "Dict[str, Any]":
         if (
             has_tracing_enabled(self.get_client().options)
             and self._span is not None
-            and not isinstance(self._span, NoOpStreamedSpan)
+            and not isinstance(self._span, (NoOpStreamedSpan, NoOpSpan))
```
sl0thentr0py, line 625: generalized these instance checks just in case

alexander-alderman-webb (approved):

sentry_sdk/scope.py
```diff
@@ -1859,17 +1859,15 @@ def apply_to_telemetry(self, telemetry: "Union[Log, Metric, StreamedSpan]") -> N
         if not isinstance(telemetry, StreamedSpan):
             trace_context = self.get_trace_context()
             trace_id = trace_context.get("trace_id")
-            if telemetry.get("trace_id") is None:
-                telemetry["trace_id"] = (
-                    trace_id or "00000000-0000-0000-0000-000000000000"
-                )
+            if telemetry.get("trace_id") is None and trace_id is not None:
+                telemetry["trace_id"] = trace_id
```
alexander-alderman-webb, line 1863: Metrics batcher would crash here if the if condition is skipped:
https://github.com/getsentry/sentry-python/blob/cc802f6fcedcbd50f76afe2fc902d4d21b6691d6/sentry_sdk/_metrics_batcher.py#L24

sl0thentr0py: why do we have typedicts and not dataclasses like normal people...

PR #6572, opened 2026-06-15
fix(fastapi): use effective_route_context path for prefixed routers

### LINKED ISSUE #6571: Update `_set_transaction_name_and_source` in fastapi to handle changes in fastapi v0.137
tox failure: [https://github.com/getsentry/sentry-python/actions/runs/27528323860/job/81360282531?pr=6567](<https://github.com/getsentry/sentry-python/actions/runs/27528323860/job/81360282531?pr=6567>)

This is because router prefixes are stored in an EffectiveRouteContext, so we need to check if that exists and use it if it's available.

ericapisani:
FastAPI >= 0.137 stores the prefix-resolved path in `scope["fastapi"]["effective_route_context"]` instead of `scope["route"].path`, so transactions/segments were showing the unprefixed `/users/{user_id}` instead of the full `/api/users/{user_id}` for routers mounted with a prefix.

This fix prefers `effective_route_context.path` when available, falling back to `route.path` for older FastAPI versions.

Fixes PY-2530
Fixes #6571

### DIFF
--- a/sentry_sdk/integrations/fastapi.py
+++ b/sentry_sdk/integrations/fastapi.py
@@ -62,10 +62,23 @@ def _set_transaction_name_and_source(
 
     elif transaction_style == "url":
         route = request.scope.get("route")
+
         if route:
-            path = getattr(route, "path", None)
-            if path is not None:
-                name = path
+            # FastAPI >= 0.137 stores the prefix-resolved path on an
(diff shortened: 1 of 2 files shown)


ericapisani:
bugbot run

alexander-alderman-webb (approved):

Issue #5272, opened 2025-12-31
MCP tool calls break distributed tracing within same service


maxclaus:
### Problem Statement

MCP tool calls do not continue distributed traces from their parent HTTP requests within the same service. The MCP integration creates new transactions with new trace IDs instead of joining the existing trace from the incoming HTTP request.

**Environment:**
- **sentry-python version:** 2.48.0
- **MCP server:** FastMCP (HTTP/SSE transport) within FastAPI application
- **Service architecture:**
  - Service A (Go + Sentry) → Service B (Python/FastAPI + Sentry + MCP Server)
  - HTTP trace propagates correctly to Service B's FastAPI handler
  - MCP tool execution in Service B breaks the trace chain

**Current behavior:**
When `get_start_span_function()` executes during MCP tool handler wrapping, the HTTP transaction isn't accessible due to async context isolation. The function checks the current scope, but only finds unrelated spans (e.g., Redis, database operations) with different trace IDs, not the HTTP transaction that has the correct trace context from the incoming request.

**Expected behavior:**
MCP tool transaction should continue the HTTP request's trace (same trace_id, becomes child of HTTP span).

**Actual behavior:**
New transaction created with new trace_id from unrelated span in scope (e.g., Redis connection span), breaking the trace chain.

**Root cause:**
Async context isolation prevents the HTTP transaction from being accessible in the MCP tool handler's execution context.

### Solution Brainstorm

**Current workaround**

I patched sentry-sdk to get it working for my project:

```python
"""
Fix for MCP integration to properly continue distributed traces as transactions.

The issue: The MCP integration's get_start_span_function() doesn't check for
propagated trace context from incoming sentry-trace headers, so it creates new
transactions with new trace_ids instead of continuing the distributed trace.

The fix: When sentry-trace headers are present in the MCP request, use
continue_trace() to create a transaction that continues the existing trace
(same trace_id, but still a transaction for MCP Insights dashboard).

This maintains both:
1. Distributed tracing (same trace_id across services)
2. MCP Insights dashboard functionality (transactions, not spans)
"""
import logging
from typing import Any, Callable
import sentry_sdk
from sentry_sdk.tracing import Transaction

logger = logging.getLogger(__name__)


def create_transaction_from_trace_headers(
    op: str = None,
    name: str = None,
    origin: str = "manual",
    **kwargs: Any
) -> Transaction:
    """
    Creates a transaction that continues the distributed trace from MCP request headers.

[...]

PR #6583, opened 2026-06-16
fix(mcp): support MCP SDK v2 handler signature and removed request_ctx

### LINKED ISSUE #6574: Handle removal of `request_ctx` in `mcp.server.lowlevel.server`
v2 of the MCP Python SDK removes this property, and request context is now passed directly to handlers as the first argument (see [migration guide](<https://github.com/modelcontextprotocol/python-sdk/blob/main/docs/migration.md#lowlevel-server-request_context-property-removed>)).

Update our MCP integration to handle this.

Related tox failure: [https://github.com/getsentry/sentry-python/actions/runs/27557716423/job/81461057033?pr=6567](<https://github.com/getsentry/sentry-python/actions/runs/27557716423/job/81461057033?pr=6567>)

ericapisani:
MCP SDK v2 changed handler signatures from (name, arguments) to
(ctx, params) and removed the request_ctx ContextVar. Add version
detection to patch the correct entry-points (Server.__init__ and
add_request_handler for v2, decorator methods for v1), extract
request context from the ServerRequestContext argument directly,
and update tests to work with both SDK versions.

Migration notes for MCP SDK v2 can be found here:
- https://github.com/modelcontextprotocol/python-sdk/releases/tag/v2.0.0a1
- https://github.com/modelcontextprotocol/python-sdk/blob/main/docs/migration.md

What this PR does not do:
- address all breaking changes outlined in the migration guide. This is meant to fix the test failures on https://github.com/getsentry/sentry-python/pull/6567 while still being comprehensive enough to make sense for the next developer who continues the migration.

This was tested on the branch in https://github.com/getsentry/sentry-python/pull/6567 to confirm it doesn't introduce regressions while addressing the v2 breaking change.

Fixes PY-2532
Fixes https://github.com/getsentry/sentry-python/issues/6574

### DIFF
--- a/sentry_sdk/integrations/mcp.py
+++ b/sentry_sdk/integrations/mcp.py
@@ -19,14 +19,20 @@ from sentry_sdk.integrations._wsgi_common import nullcontext
 from sentry_sdk.scope import should_send_default_pii
 from sentry_sdk.traces import StreamedSpan
 from sentry_sdk.tracing_utils import has_span_streaming_enabled
-from sentry_sdk.utils import safe_serialize
+from sentry_sdk.utils import package_version, safe_serialize
+
+MCP_PACKAGE_VERSION = package_version("mcp")
 
 try:
     from mcp.server.lowlevel import Server  # type: ignore[import-not-found]
-    from mcp.server.lowlevel.server import request_ctx  # type: ignore[import-not-found]
     from mcp.server.streamable_http import (  # type: ignore[import-not-found]
         StreamableHTTPServerTransport,
     )
+
+    if MCP_PACKAGE_VERSION and MCP_PACKAGE_VERSION < (2, 0, 0):
+        from mcp.server.lowlevel.server import (  # type: ignore[import-not-found]
+            request_ctx,
+        )
 except ImportError:
     raise DidNotEnable("MCP SDK not installed")
 
@@ -34,6 +40,16 @@ try:
     from fastmcp import FastMCP  # type: ignore[import-not-found]
 except ImportError:
     FastMCP = None
+
+if MCP_PACKAGE_VERSION and MCP_PACKAGE_VERSION >= (2, 0, 0):
+    try:
+        from mcp.server.context import (  # type: ignore[import-not-found]
+            ServerRequestContext,
+        )
+    except ImportError:
+        ServerRequestContext = None
+else:
+    ServerRequestContext = None
 
 
 if TYPE_CHECKING:
@@ -71,13 +87,15 @@ class MCPIntegration(Integration):
             _patch_fastmcp()
 
 
-def _get_active_http_scopes() -> (
-    "Optional[Tuple[Optional[sentry_sdk.Scope], Optional[sentry_sdk.Scope]]]"
-):
-    try:
-        ctx = request_ctx.get()
-    except LookupError:
-        return None
+def _get_active_http_scopes(
+    ctx: "Optional[Any]" = None,
+) -> "Optional[Tuple[Optional[sentry_sdk.Scope], Optional[sentry_sdk.Scope]]]":
+    if MCP_PACKAGE_VERSION and MCP_PACKAGE_VERSION < (2, 0, 0):
+        if ctx is None:
+            try:
+                ctx = request_ctx.get()
+            except LookupError:
+                return None
 
     if (
         ctx is None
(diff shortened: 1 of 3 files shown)


sentry_sdk/integrations/mcp.py
```diff
@@ -569,10 +679,34 @@ async def wrapper(*args: "Any") -> "Any":
...
+
+
 def _patch_lowlevel_server() -> None:
     """
     Patches the mcp.server.lowlevel.Server class to instrument handler execution.
     """
+    if MCP_PACKAGE_VERSION and MCP_PACKAGE_VERSION >= (2, 0, 0):
+        _patch_lowlevel_server_v2()
+    else:
+        _patch_lowlevel_server_v1()
```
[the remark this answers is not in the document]
ericapisani: If it's falsy in the way it's being described (where `MCP_VERSION_PACKAGE` is `None` resulting in the `request_ctx` not being imported), that would mean the package isn't installed and we shouldn't be patching. 🤔

sentry_sdk/integrations/mcp.py
```diff
@@ -492,7 +599,10 @@ async def _handler_wrapper(
 
                 # For resources, extract and set protocol
                 if handler_type == "resource":
-                    if original_args:
+                    uri = None
+                    if params is not None:
+                        uri = getattr(params, "uri", None)
+                    elif original_args:
```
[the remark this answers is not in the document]
ericapisani: I think this one is valid, going to dig into this a bit more. Think this conditional needs to be updated to check if we're dealing with v2 vs. v1 and pull the uri from `original_args[0]` only when we know we're dealing with v1

sentry_sdk/integrations/mcp.py
```diff
@@ -492,10 +599,16 @@ async def _handler_wrapper(
...
+                    uri = None
+                    if params is not None:
+                        uri = getattr(params, "uri", None)
+
+                    # v1 scenario
+                    if ServerRequestContext is None:
+                        if original_args:
+                            uri = original_args[0]
+                        else:
+                            uri = original_kwargs.get("uri")
```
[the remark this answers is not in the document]
ericapisani: URI extraction for MCP v2 is done as part of the `params` check a few lines up:

```python
                    if params is not None:
                        uri = getattr(params, "uri", None)
```

tests/integrations/mcp/test_mcp.py
```diff
@@ -102,20 +130,149 @@ def __init__(self, text):
...
+
+    async def handler(ctx, params):
+        return None
+
+    wrapped = mcp_module._wrap_v2_handler("tool", handler)
+    assert wrapped is not handler
+    assert getattr(wrapped, "__sentry_mcp_wrapped__", False) is True
+
+    rewrapped = mcp_module._wrap_v2_handler("tool", wrapped)
+    assert rewrapped is wrapped
```
alexander-alderman-webb: This test is relying on private implementation details, making it brittle to internal changes such as renaming `_wrap_v2_handler` or `__sentry_mcp_wrapped__`.

I'd say we can just remove the test to be honest. If we want to keep it let's ensure we don't double wrap by checking behavior observable by the user. For example, ensuring that we don't emit the same span twice (I assume that's what happens when you double wrap).

tests/integrations/mcp/test_mcp.py
```diff
@@ -1432,3 +1806,190 @@ async def test_tool(tool_name, arguments):
...
+            ):
+                data = trace["data"]
+                break
+            for span in event.get("spans", []):
+                if span["data"].get(SPANDATA.MCP_METHOD_NAME) == "tools/call":
+                    data = span["data"]
+                    break
+            if data is not None:
+                break
+        assert data is not None
```
alexander-alderman-webb, line 1959: This means that we're capturing wrong data as we break traces. MCP spans always belong under the HTTP transaction/segment when using a HTTP-based transport. Users have pointed this out in the past: https://github.com/getsentry/sentry-python/issues/5272.

We'll need to figure how to propagate the context variables.

alexander-alderman-webb: This can be done as a follow-up so to be fair, so not blocking.

alexander-alderman-webb (approved):

tests/integrations/mcp/test_mcp.py
ericapisani: Created PY-2539 to tackle this

PR #6598, opened 2026-06-17
fix(litellm): Store span off-band, not in forwarded metadata

### LINKED ISSUE #6596: LiteLLMIntegration stores a non-JSON-serializable Span in request metadata, breaking outbound LLM requests
### How do you use Sentry?

Sentry SaaS (sentry.io)

### Version

`sentry-sdk` 2.52.0 (also present on current `master`); `litellm` 1.81.9; Python 3.13

### Steps to Reproduce

With the LiteLLM integration enabled, any call that passes caller `metadata` crashes during request serialization:

```python
import asyncio, litellm, sentry_sdk
from sentry_sdk.integrations.litellm import LiteLLMIntegration

sentry_sdk.init(
    dsn="<your-dsn>", traces_sample_rate=1.0,
    send_default_pii=True, integrations=[LiteLLMIntegration()],
)

async def main():
    await litellm.anthropic.messages.acreate(
        model="anthropic/claude-3-5-sonnet-latest",
        messages=[{"role": "user", "content": "hi"}],
        max_tokens=16, stream=True, api_key="sk-ant-...",
        metadata={"user_id": "my-org"},   # Anthropic abuse-isolation id
    )

asyncio.run(main())
```

**Root cause:** `_input_callback` stores the live span via `_get_metadata_dict(kwargs)["_sentry_span"] = span` (`sentry_sdk/integrations/litellm.py`). litellm's Anthropic `/v1/messages` passthrough forwards `metadata` into the request body, so the live `Span` lands at `request_body['metadata']['_sentry_span']` and `json.dumps(request_body)` fails — before the request is sent.

### Expected Result

Enabling the integration must not make outbound provider requests unserializable. The integration's bookkeeping span should be stashed off-band (e.g. a `WeakKeyDictionary` keyed on the logging object) rather than written into the caller's request `metadata` dict, which several providers (Anthropic here) forward into the wire payload.

### Actual Result

```
TypeError: Object of type Span is not JSON serializable
```

(a `Transaction` when there is no active parent span). Separately, with `send_default_pii=True` + `include_prompts=True` (the default), that span's `gen_ai.request.messages` holds the **verbatim prompt** — so any sink that serializes this span object can leak prompt content.

---

Related (defense-in-depth, downstream): litellm forwards/serializes this injected metadata into the provider body. Companion issue filed at BerriAI/litellm — see comment below. litellm has stripped such span objects in other logging paths: BerriAI/litellm#15728, BerriAI/litellm#12354.

jgreer013:
## Summary

Fixes #6596 (Linear PY-2540).

With `LiteLLMIntegration` enabled, any call that passes caller `metadata` could crash during request serialization:

```
TypeError: Object of type Span is not JSON serializable
```

### Root cause

`_input_callback` stored the live span via `_get_metadata_dict(kwargs)["_sentry_span"] = span`. litellm threads the caller's `metadata` dict through to `litellm_params["metadata"]` — it's the **same dict object** the caller passed — and some providers (e.g. Anthropic's `/v1/messages` passthrough) forward that dict into the outbound request body. So the live `Span` landed at `request_body["metadata"]["_sentry_span"]` and `json.dumps(request_body)` failed **before the request was sent**.

Separately, under `send_default_pii=True` + `include_prompts=True`, that span's `gen_ai.request.messages` holds the verbatim prompt, so any sink that serialized the injected metadata could leak prompt content to the provider.

### Fix

Store the bookkeeping span **off-band** in a module-level registry keyed by `litellm_call_id` (a per-request UUID that is stable across the input/success/failure callbacks), falling back to the identity of the shared callback `kwargs` dict for direct callback invocations that omit it. The span no longer lives in any litellm-visible dict, so it can't be forwarded, serialized, or deep-copied by litellm — fixing both the crash and the prompt-leak vector. The registry entry is removed by the terminal success/failure callback (streaming success peeks and pops only on the final call).

### Testing

- New regression test `test_caller_metadata_stays_json_serializable` asserts the forwarded `metadata` stays free of `_sentry_span` and JSON-serializable at the serialization moment, while the span is still recorded off-band. It **fails on `master`** and passes with this change.
- Full litellm suite: **160 passed** (litellm v1.81.16), including streaming/async/failure paths.
- Regression test also green on the **oldest (v1.77.7)** and **latest** litellm matrix entries.
- `mypy` clean; `ruff` check + format clean.

### Related (downstream, defense-in-depth)

litellm forwards/serializes this injected metadata into the provider body. Companion issue: BerriAI/litellm#30662. litellm has stripped such span objects in other logging paths (BerriAI/litellm#15728, BerriAI/litellm#12354).


sentry_sdk/integrations/litellm.py
```diff
@@ -31,16 +33,43 @@
...
+    return kwargs.get("litellm_call_id") or id(kwargs)
+
+
+def _store_span(kwargs: "Dict[str, Any]", span: "Any") -> None:
+    key = _span_key(kwargs)
+    with _spans_by_call_lock:
+        _spans_by_call[key] = span
+        _spans_by_call.move_to_end(key)
+        while len(_spans_by_call) > _MAX_TRACKED_SPANS:
+            _spans_by_call.popitem(last=False)
```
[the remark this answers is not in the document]
jgreer013: Fixed in 986f979 — evicted spans are now finished via `__exit__` (outside the lock), matching the `langchain` integration's eviction behavior, so an over-cap call still records a span instead of leaving it unfinished. Added a test asserting eviction finishes the oldest span while retained spans keep running.

sentry_sdk/integrations/litellm.py
```diff
@@ -31,16 +33,49 @@
...
+# through to `litellm_params`, and some providers (e.g. Anthropic's /v1/messages
+# passthrough) forward that dict into the outbound request body -- so a span
+# stored there would break `json.dumps(request_body)` and could leak prompt data
+# to the provider. `litellm_call_id` is a per-request UUID that stays stable
+# across the input/success/failure callbacks; the identity of the (shared)
+# callback kwargs dict is the fallback for direct callback invocations that omit
+# it. The terminal success/failure callback removes the entry; the registry is
+# capped (oldest evicted first) so calls abandoned before a terminal callback
+# fires -- e.g. a stream the caller stops iterating -- cannot grow it unbounded.
+_MAX_TRACKED_SPANS = 1000
```
alexander-alderman-webb: Thanks for the PR!
I'll think about approaches because the `_spans_by_call` will grow if spans aren't finished correctly and everything is dropped if `_MAX_TRACKED_SPANS` is reached.

Generally, if there's a public shared reference between input and output callbacks, we should use that. If not, your approach could be the best we can do.

jgreer013: It seems like there isn't one in this case, but perhaps there's an opportunity to add one with a bigger rework?

sentry_sdk/integrations/litellm.py
```diff
@@ -285,7 +319,7 @@ def _failure_callback(
     end_time: "datetime",
 ) -> None:
     """Handle request failure."""
-    span = _get_metadata_dict(kwargs).get("_sentry_span")
+    span = _pop_span(kwargs)
     if span is None:
```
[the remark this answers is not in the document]
jgreer013: Moot now — I dropped the registry and eviction entirely, so nothing gets evicted anymore. The span lives on the per-request dict litellm hands the callbacks, so the failure callback always finds it. Resolving.

sentry_sdk/integrations/litellm.py
jgreer013: Actually, turns out there is one — litellm threads the same `model_call_details` dict through all three callbacks, so I moved the span onto that (off the forwarded `metadata` path) instead of the module-level registry. No cap or eviction now, and the span's lifetime is tied to the request, like the clickhouse/dramatiq integrations do it. I double-checked the Anthropic passthrough doesn't forward it — the request body only gets the recognized params, not `model_call_details`. Lmk if you had a different reference in mind!

alexander-alderman-webb: Thanks for updating the PR! I'll review today

alexander-alderman-webb (commented):
Integration code looks good to me!

Could you remove the new tests @jgreer013?
They rely on private details such as `_input_callback` and `_sentry_span`. For a bugfix like this we really would prefer not to add tests that make future code changes more difficult (since the tests rely on private details).

jgreer013:
Updated to have just one test through the public api with mocks (damn you claude and your incessant need to write tests for private methods)

alexander-alderman-webb (approved):

alexander-alderman-webb:
The `httpx` import in your test additions fail because it's not installed in the environment. Happy to merge without the tests as before 😄.
Otherwise, you can find information about our test matrix below.
You would have to add `httpx` as a dependency to the [litellm `deps` list](https://github.com/getsentry/sentry-python/blob/10ee6040ddd51ae9e4609daffca213086101cda8/scripts/populate_tox/config.py#L265) and re-generate the test matrix.

https://github.com/getsentry/sentry-python/blob/master/scripts/populate_tox/README.md

jgreer013:
Gotcha, dropped it! CC @alexander-alderman-webb since the update needs workflow approvals to be re-run

sentry_sdk/integrations/litellm.py
```diff
@@ -31,16 +31,21 @@
...
+def _store_span(kwargs: "Dict[str, Any]", span: "Any") -> None:
+    kwargs[_SPAN_KEY] = span
+
+
+def _peek_span(kwargs: "Dict[str, Any]") -> "Any":
+    return kwargs.get(_SPAN_KEY)
+
+
+def _pop_span(kwargs: "Dict[str, Any]") -> "Any":
+    return kwargs.pop(_SPAN_KEY, None)
```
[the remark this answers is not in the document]
jgreer013: Nope, those come from get_requested_anthropic_messages_optional_param(local_vars)

PR #6599, opened 2026-06-17
fix(ai): handle Pydantic SDK objects in truncate_messages_by_size

### LINKED ISSUE #5350: OpenAI Integration: `truncate_messages_by_size` fails on Pydantic SDK objects (e.g., `ResponseFunctionToolCall`)

singlaamitesh:
## Summary

Fixes #5350.

OpenAI's Responses API returns Pydantic model instances (e.g. `ResponseFunctionToolCall`) that callers pass back as items in the `input` list of follow-up requests. `truncate_messages_by_size()` called `json.dumps(messages)` directly without any special handling, so a messages list containing such an object raised:

```
TypeError: Object of type ResponseFunctionToolCall is not JSON serializable
```

This is the same class of bug as the `Omit`/`NotGiven` handling done previously — the OpenAI integration assumes inputs are plain Python dicts, but the SDK passes Pydantic objects through the instrumentation path.

## Fix

Introduce `_PydanticEncoder`, a `JSONEncoder` subclass that serializes Pydantic models via `model_dump()`. It is used in the two `json.dumps()` calls inside `truncate_messages_by_size()` and `_find_truncation_index()` — only for byte-size measurement. The returned `messages` list is left unchanged (Pydantic objects are preserved as-is). This avoids the `TypeError` without altering the data the caller gets back.

The SDK already has `_normalize_data()` which handles Pydantic objects, but using it here would stringify `None` values to `"None"` (a pre-existing behaviour), which would break the existing `test_single_message_truncation_non_str_non_list_content` test. The encoder approach is more targeted.

## Test plan

- [x] New regression test `test_pydantic_objects_in_message_list_do_not_raise` — verifies a Pydantic-like object in the messages list does not raise `TypeError`
- [x] `TestTruncateMessagesBySize` — all 15 tests pass (including pre-existing `None` / int / bool content tests)

### DIFF
--- a/sentry_sdk/ai/utils.py
+++ b/sentry_sdk/ai/utils.py
@@ -588,13 +588,21 @@ class _PydanticEncoder(json.JSONEncoder):
     def default(self, obj: "Any") -> "Any":
         if not inspect.isclass(obj) and hasattr(obj, "model_dump"):
             try:
-                return obj.model_dump()
+                # mode="json" returns only JSON-native types (converts datetime,
+                # Decimal, etc.) — fall back to plain model_dump() for Pydantic v1
+                # and custom classes that don't accept the kwarg.
+                return obj.model_dump(mode="json")
+            except TypeError:
+                try:
+                    return obj.model_dump()
+                except Exception:
+                    pass
             except Exception:
                 pass
         return super().default(obj)
 
 
-def _find_truncation_index(messages: "List[Dict[str, Any]]", max_bytes: int) -> int:
+def _find_truncation_index(messages: "List[Any]", max_bytes: int) -> int:
     """
     Find the index of the first message that would exceed the max bytes limit.
     Compute the individual message sizes, and return the index of the first message from the back
@@ -721,10 +729,10 @@ def redact_blob_message_parts(
 
 
 def truncate_messages_by_size(
-    messages: "List[Dict[str, Any]]",
+    messages: "List[Any]",
     max_bytes: int = MAX_GEN_AI_MESSAGE_BYTES,
     max_single_message_chars: int = MAX_SINGLE_MESSAGE_CONTENT_CHARS,
-) -> "Tuple[List[Dict[str, Any]], int]":
+) -> "Tuple[List[Any], int]":
     """
     Returns a truncated messages list, consisting of
     - the last message, with its content truncated to `max_single_message_chars` characters,
--- a/sentry_sdk/ai/utils.py
+++ b/sentry_sdk/ai/utils.py
@@ -578,6 +578,22 @@ def _truncate_single_message_content_if_present(
     return message
 
 
+class _PydanticEncoder(json.JSONEncoder):
+    """JSON encoder that serializes Pydantic models via model_dump().
+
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb:
Stale issue, only one input item is kept these days.

PR #6600, opened 2026-06-17
fix(litellm): set gen_ai.operation.name accurately, not via fallback

### LINKED ISSUE #6442: Only set accurate operation name in `litellm`
Using a fallback leads to wrong data

https://github.com/getsentry/sentry-python/blob/b4e036730e7988f7a21b93b2024ee5c554510d7f/sentry_sdk/integrations/litellm.py#L90-L97

singlaamitesh:
## Summary

Fixes #6442.

The `_input_callback` previously used a binary if/else that mapped every call type except `embedding`/`aembedding` to `"chat"`. This caused wrong `gen_ai.operation.name` values and span names for call types like `text_completion`, `image_generation`, and others.

**Before:**
```python
call_type = kwargs.get("call_type", None)
if call_type == "embedding" or call_type == "aembedding":
    operation = "embeddings"
else:
    operation = "chat"   # ← wrong for text_completion, image_generation, etc.
```

**After:** an explicit mapping table covering the call types we can represent accurately:

| litellm `call_type` | `gen_ai.operation.name` | `sentry.op` |
|---|---|---|
| `completion` / `acompletion` | `chat` | `gen_ai.chat` |
| `embedding` / `aembedding` | `embeddings` | `gen_ai.embeddings` |
| `text_completion` / `atext_completion` | `text_completion` | `gen_ai.text_completion` |
| anything else (e.g. `image_generation`) | *(not set)* | `gen_ai.chat` (fallback for span routing) |

For unmapped call types the span is still created (so tracing works), but `gen_ai.operation.name` is intentionally omitted rather than recording a misleading `"chat"` value. The span name uses the raw `call_type` string so it's at least identifiable.

## Test plan

- [x] `test_input_callback_operation_name_accurate` — parametrised over all 6 known call types; verifies correct operation name and span op
- [x] `test_input_callback_unknown_call_type_no_operation_name` — verifies `gen_ai.operation.name` is absent for `image_generation`
- [x] All pre-existing litellm tests that assert `GEN_AI_OPERATION_NAME == "chat"` and `GEN_AI_OPERATION_NAME == "embeddings"` continue to pass

### DIFF
--- a/sentry_sdk/ai/utils.py
+++ b/sentry_sdk/ai/utils.py
@@ -578,6 +578,22 @@ def _truncate_single_message_content_if_present(
     return message
 
 
+class _PydanticEncoder(json.JSONEncoder):
+    """JSON encoder that serializes Pydantic models via model_dump().
+
+    This allows json.dumps() to handle OpenAI/LiteLLM SDK objects such as
+    ResponseFunctionToolCall that callers pass back in follow-up requests.
+    """
+
+    def default(self, obj: "Any") -> "Any":
+        if not inspect.isclass(obj) and hasattr(obj, "model_dump"):
+            try:
+                return obj.model_dump()
+            except Exception:
+                pass
+        return super().default(obj)
+
+
 def _find_truncation_index(messages: "List[Dict[str, Any]]", max_bytes: int) -> int:
     """
     Find the index of the first message that would exceed the max bytes limit.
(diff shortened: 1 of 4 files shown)


alexander-alderman-webb:
This changeset has unrelated stuff in it, closing.

PR #6601, opened 2026-06-17
test(aiomysql): Add span streaming variants to aiomysql tests

### LINKED ISSUE #6539: Migrate aiomysql to span first
See parent issue.

The integration was migrated by a contributor -- double check if all attributes etc. make sense are in conventions.

ericapisani:
Parametrize existing aiomysql tests with span_streaming to cover both
static and stream trace lifecycle modes.

Fixes PY-2523
Fixes #6539


alexander-alderman-webb (approved):
Didn't look too in-depth since it's test code

PR #6603, opened 2026-06-18
fix(openai-agents): set gen_ai.operation.name to "responses" for OpenAIResponsesModel

### LINKED ISSUE #6417: Set `gen_ai.operation.name` correctly in `openai-agents`
`openai-agents` can also call the responses API, for example: https://github.com/getsentry/sentry-python/blob/53188b322fc96dab0c4b064945b62e773fada63c/sentry_sdk/integrations/openai_agents/spans/ai_client.py#L37

singlaamitesh:
## Summary

Fixes #6417.

The `ai_client_span` function hardcoded `gen_ai.operation.name = "chat"` for every call, even when openai-agents was using `OpenAIResponsesModel`, which calls the OpenAI Responses API (not Chat Completions).

**Before:**
```python
# TODO-anton: implement other types of operations. Now "chat" is hardcoded.
SPANDATA.GEN_AI_OPERATION_NAME: "chat",  # ← always "chat"
name=f"chat {model_name}",               # ← always "chat ..."
```

**After:** the model class is detected in `_get_model` (where the model object is available) and passed to `ai_client_span`:
- `OpenAIResponsesModel` → `"responses"` (consistent with the standalone `openai` integration)
- Any other model class (e.g. `OpenAIChatCompletionsModel`) → `"chat"`

The `sentry.op` remains `OP.GEN_AI_CHAT` for both (same as the standalone openai integration does for Responses API spans).

## How it works

`_get_model` already has the full model object. A cheap `isinstance` check determines the operation name, which is captured in the closure and forwarded to both `wrapped_get_response` and `wrapped_stream_response` via the new `operation` parameter on `ai_client_span`.

## Test plan

- [x] All 144 existing openai-agents tests pass with updated assertions (`"chat"` → `"responses"`)
- [x] Bug reproduced: without fix `OpenAIResponsesModel` always produced `chat gpt-4o`; with fix it produces `responses gpt-4o`

### DIFF
--- a/sentry_sdk/integrations/openai_agents/patches/models.py
+++ b/sentry_sdk/integrations/openai_agents/patches/models.py
@@ -88,6 +88,16 @@ def _get_model(
     request_model_name = model.model if hasattr(model, "model") else str(model)
     agent._sentry_request_model = request_model_name
 
+    # Determine the gen_ai operation name from the model class.
+    # OpenAIResponsesModel uses the Responses API ("responses"); all other
+    # model classes (e.g. OpenAIChatCompletionsModel) use Chat Completions ("chat").
+    try:
+        from agents.models.openai_responses import OpenAIResponsesModel
+
+        _operation = "responses" if isinstance(model, OpenAIResponsesModel) else "chat"
+    except ImportError:
+        _operation = "chat"
+
     # Wrap _fetch_response if it exists (for OpenAI models) to capture response model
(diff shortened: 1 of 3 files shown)


alexander-alderman-webb:
No string matching on classes please, that's brittle.
Feel free to discuss the approach before PRing on the issue @singlaamitesh!

PR #6606, opened 2026-06-18
feat(pyramid): Set user.id on spans when PII is enabled

### LINKED ISSUE #6605: [Pyramid] Set user id on segment span

ericapisani:
When span streaming is enabled and `send_default_pii` is true, set the `user.id` attribute on all spans using `scope.set_user` using the authenticated user ID from the Pyramid request.

Fixes #6605
Fixes PY-2542


alexander-alderman-webb (approved):
Thanks!

PR #6609, opened 2026-06-19
fix(asyncpg): Use distinct span ops for cursor iteration and fetch to prevent N+1 false positives

### LINKED ISSUE #6576: Sentry detects N+1 queries in server side cursor
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.61.1

### Steps to Reproduce

We recently updated from sentry 2.57.0 to 2.61.1 and we started getting many reports of [N+1 queries](https://docs.sentry.io/product/issues/issue-details/performance-issues/n-one-queries/) in the Sentry UI. This was tracked to PR #6252.

In our particular case, we use sqlalchemy with asyncpg and the queries that are being flagged are all due to the use of server side cursors.

Our code is as follows:

```python
await conn.stream(query)
```

Is there any workaround? Should we just disable the AsyncPGIntegration if using sqlalchemy?

### Expected Result

No N+1 queries should be detected when using `conn.stream`.

### Actual Result

Lot of flagged N+1 queries in Sentry's UI

ericapisani:
When asyncpg iterates a server-side cursor (via CursorIterator.__anext__) or fetches rows in batches (via Cursor.fetch), each batch triggers BaseCursor._bind_exec or BaseCursor._exec, producing a burst of identical spans. Sentry's N+1 detector incorrectly flags these as N+1 queries.

Set `sentry.op` to `db.cursor.iter` for CursorIterator-driven spans and `db.cursor.fetch` for Cursor.fetch-driven spans so the backend can exclude them from N+1 detection.

Adds span_op_override_value parameter to record_sql_queries to propagate the override without changing existing callers.

This will require changes on the backend in order to not trigger n+1 alerts on these ops.

Fixes #6576
Fixes PY-2534


wahajahmed010 (commented):
## Review: getsentry/sentry-python#6609 — fix(asyncpg): distinct span ops for cursor iteration/fetch

**Logic:** ✅ Clean approach. Using ContextVar flags to distinguish cursor iteration vs fetch vs regular query execution, then overriding `sentry.op` accordingly. The N+1 detector can then filter out `db.cursor.iter`/`db.cursor.fetch` spans.

**Tests:** ✅ Comprehensive — tests both streaming and static span modes, both cursor iteration and fetch paths. Asserts exact span count (5 for prefetch=5, 2 for fetch(10)+fetch(10)).

**Edge case:** What if someone uses `CursorIterator.__anext__` directly without going through `async for`? The monkey-patch on `CursorIterator.__anext__` covers that. ✅

**Concern:** The ContextVar flags are module-level — if two cursors are active concurrently (e.g., in different tasks), the flags could cross-contaminate. However, ContextVar is task-local by design in Python, so this should be safe. ✅

**Performance:** Minimal overhead — just a ContextVar set/get per call. ✅

**Verdict:** Well-structured fix. Ship it.

sentry_sdk/integrations/asyncpg.py
```diff
@@ -161,13 +177,52 @@ async def _inner(*args: "Any", **kwargs: "Any") -> "T":
...
+
+
 def _wrap_cursor_method(
     f: "Callable[..., Awaitable[T]]",
 ) -> "Callable[..., Awaitable[T]]":
     async def _inner(*args: "Any", **kwargs: "Any") -> "T":
         if sentry_sdk.get_client().get_integration(AsyncPGIntegration) is None:
             return await f(*args, **kwargs)
 
+        if _asyncpg_cursor_iterator_is_invoked.get():
```
sl0thentr0py: preferable to set this as state on the cursor imo
```
cursor._iterator_is_invoked = True
```

sl0thentr0py (changes requested):

sentry_sdk/integrations/asyncpg.py
ericapisani: The various Cursor classes in the `asyncpg` integration prevent us from doing this through their use of `__slots__` (as their presence removes the `.__dict__` for setting dynamic values on each instance created).

sl0thentr0py: ok in that case, we need to be sure that each context var maps cleanly to only one running cursor instance at a time.

alexander-alderman-webb: Did a little snooping around in the library, both `Cursor` and `CursorIterator` inherit from `BaseCursor`.

So you should be able to tell which case you're in by checking `isinstance(self, ...)` in `BaseCursor._exec`, etc ...

ericapisani: That's a great idea 🚀 Give me a few to make that change

ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6609/commits/cd25110fc8ffc1500af30947a0c53f6b08011d33

sl0thentr0py (approved):

PR #6635, opened 2026-06-23
feat(clickhouse_driver): Add db.query.text to streamed query spans

### LINKED ISSUE #6634: Set db.query.text attribute on clickhouse_driver

ericapisani:
Adds the `db.query.text` attribute to ClickHouse driver spans when span
streaming is enabled. This brings ClickHouse driver spans in line with other
database integrations like asyncpg that already include query text in streamed
spans.

Fixes #6634
Fixes PY-2547


alexander-alderman-webb (approved):
🚀

PR #6650, opened 2026-06-24
fix(aiohttp): Gate url.full, url.path, url.query on send_default_pii

### LINKED ISSUE #6631: aiohttp missing `url.path`attribute

ericapisani:
URL attributes (`url.full`, `url.path`, `url.query`) contain potentially sensitive data — full request URLs, path segments, and query strings — that should not be captured unless the user has explicitly opted in to PII collection. Previously these were always recorded on both server and outgoing client spans.

This gates all three attributes behind `should_send_default_pii()` for both span types, matching the behavior of `client.address` and `user.ip_address` that were already gated. Tests for the two streaming span cases are parametrized to cover both PII-on and PII-off paths.

Fixes PY-2545
Fixes #6631

### DIFF
--- a/sentry_sdk/integrations/aiohttp.py
+++ b/sentry_sdk/integrations/aiohttp.py
@@ -159,12 +159,7 @@ class AioHttpIntegration(Integration):
                                 header_value
                             )
 
-                        url_attributes = (
-                            {"url.query": request.query_string}
-                            if request.query_string
-                            else {}
-                        )
-
+                        url_attributes = {}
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb (approved):
I'm not convinced that not gating URL attribute behind send_default_pii is considered a sufficiently severe bug that allows us to break existing behavior.
I'll stamp given precedent in https://github.com/getsentry/sentry-python/pull/6501.

Issue #6577, opened 2026-06-15
ASGI integration doubles the mount prefix in request.url for Starlette Mounted sub-apps (root_path + path)


pencil:
### Summary

`_get_url()` in `sentry_sdk/integrations/_asgi_common.py` builds the request URL as `root_path + path`:

```python
path = asgi_scope.get("root_path", "") + asgi_scope.get("path", "")
```

Under the current ASGI spec, `scope["path"]` **already includes** `scope["root_path"]`. So for any Starlette `Mount(...)`ed sub-app the mount prefix is counted twice — a request to `/mcp/` is reported as `/mcp/mcp/`. This affects `request.url` on both error and transaction events for every sub-mounted ASGI app.

### Version / environment

- sentry-sdk: `2.59.0` (also reproduces on `master` — `_get_url` is unchanged there)
- starlette: `1.0.1`
- uvicorn: `0.38.0`
- python: `3.14.5`

### Minimal reproduction

```python
from sentry_sdk.integrations._asgi_common import _get_url

# The scope a Starlette Mount("/mcp", sub_app) hands its sub-app when a client
# requests "/mcp/" — path includes root_path, per the current ASGI spec.
scope = {
    "type": "http", "scheme": "http",
    "server": ("api.example.com", 80),
    "headers": [(b"host", b"api.example.com")],
    "root_path": "/mcp",
    "path": "/mcp/",
}
print(_get_url(scope, "http", "api.example.com"))
# ->  http://api.example.com/mcp/mcp/        (expected: http://api.example.com/mcp/)
```

Confirming a real `Mount` produces exactly that scope:

```python
from starlette.applications import Starlette
from starlette.routing import Mount, Route
from starlette.responses import PlainTextResponse
from starlette.testclient import TestClient

seen = {}
async def ep(request):
    seen.update(path=request.scope["path"], root_path=request.scope["root_path"])
    return PlainTextResponse("ok")

app = Starlette(routes=[Mount("/mcp", app=Starlette(routes=[Route("/", endpoint=ep)]))])
TestClient(app).get("/mcp/")
print(seen)   # {'path': '/mcp/', 'root_path': '/mcp'}
```

### Root cause

[...]

PR #6652, opened 2026-06-24
fix(asgi): Add url.path to ASGI request span attributes

### LINKED ISSUE #6651: asgi missing `url.path` attribute

ericapisani:
The ASGI integration was not setting `url.path` on streamed request spans. Add it by combining `root_path` and `path` from the ASGI scope, which correctly handles sub-mounted apps where `root_path` is non-empty.

Fixes PY-2551
Fixes #6651

### DIFF
--- a/sentry_sdk/integrations/_asgi_common.py
+++ b/sentry_sdk/integrations/_asgi_common.py
@@ -135,6 +135,9 @@ def _get_request_attributes(asgi_scope: "Any") -> "dict[str, Any]":
                 if query_string is not None
                 else url_without_query_string
             )
+            attributes["url.path"] = asgi_scope.get("root_path", "") + asgi_scope.get(
+                "path", ""
+            )
 
     client = asgi_scope.get("client")
     if client and should_send_default_pii():
--- a/tests/integrations/asgi/test_asgi.py
+++ b/tests/integrations/asgi/test_asgi.py
@@ -221,6 +221,7 @@ async def test_capture_transaction(
                 span["attributes"]["url.full"]
                 == "http://localhost/some_url?somevalue=123"
             )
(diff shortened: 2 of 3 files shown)


sentry_sdk/integrations/_asgi_common.py
```diff
@@ -135,6 +135,9 @@ def _get_request_attributes(asgi_scope: "Any") -> "dict[str, Any]":
                 if query_string is not None
                 else url_without_query_string
             )
+            attributes["url.path"] = asgi_scope.get("root_path", "") + asgi_scope.get(
+                "path", ""
+            )
```
alexander-alderman-webb, line 140: Why can't we do this?

```suggestion
            attributes["url.path"] = url_without_query_string
```

ericapisani: Because the value in `url_without_query_string` is different from the path. 

For example, one can expect to see `https://cooldogstuff.com/api/v1/toys?id=123` in `url_without_query_string`, but the path would be `/api/v1/toys`

alexander-alderman-webb: Yes you're right!
It's version dependent whether `asgi_scope["path"]` includes `asgi_scope["root_path"]` with Starlette's mounts, so I'd suggest we tackle https://github.com/getsentry/sentry-python/issues/6577 before adding these attributes.

alexander-alderman-webb (approved):

sentry_sdk/integrations/_asgi_common.py
alexander-alderman-webb: Actually I think it's fine, we can merge this and fix both attributes together.

PR #6654, opened 2026-06-24
fix(wsgi): Gate url.full, url.path, and http.query behind send_default_pii

### LINKED ISSUE #6653: wsgi missing `url.path` attribute

ericapisani:
The WSGI integration now gates `url.full`, `url.path`, and `http.query` span attributes behind `send_default_pii`, consistent with how `client.address` is already handled. These attributes can contain user-provided query parameters and path segments that may include PII, so they should not be captured by default.

Fixes PY-2552
Fixes #6653

### DIFF
--- a/sentry_sdk/integrations/wsgi.py
+++ b/sentry_sdk/integrations/wsgi.py
@@ -394,12 +394,6 @@ def _get_request_attributes(
     for header, value in headers.items():
         attributes[f"http.request.header.{header.lower()}"] = value
 
-    query_string = environ.get("QUERY_STRING")
-    if query_string:
-        attributes["http.query"] = query_string
-
-    attributes["url.full"] = get_request_url(environ, use_x_forwarded_for)
-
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb (approved):
I'm not convinced that _not_ gating URL attribute behind `send_default_pii` is considered a sufficiently severe bug that allows us to break existing behavior.
I'll stamp given precedent in https://github.com/getsentry/sentry-python/pull/6501.

ericapisani:
> I'm not convinced that not gating URL attribute behind send_default_pii is considered a sufficiently severe bug that allows us to break existing behavior.

If we decide to talk about this more, we should make it a thread/convo with some other folks outside of this PR. One could argue that we were breaking the contract set out by our conventions/attributes by not doing so, and, in another case for a different attribute relatively recently, we considered it sufficiently severe that we called an incident over it to correct this.

PR #6676, opened 2026-06-26
feat(data-collection): foundation — option, resolution, accessors

ericapisani:
Introduces the `data_collection` init option, A/B/C/D resolution precedence, and the `should_collect_*` accessors plus the scrubbing primitives (with unit tests).

Behavior-neutral: in legacy mode the resolved `DataCollection` mirrors `send_default_pii`, and nothing consumes the new filtering yet — so merging this changes no observed behavior.

Base of the `data_collection` split. Consumers stack on top:
- frames (stack-frame variables + source context lines)
- http (request data collection)
- genai (AI input/output collection)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add the `data_collection` option, a structured configuration that supersedes `send_default_pii` for controlling what data integrations collect automatically (user identity, cookies, HTTP headers, query params, HTTP bodies, generative AI inputs/outputs, stack frame variables, source context). See the [Data Collection spec](https://develop.sentry.dev/sdk/foundations/client/data-collection/).
+  - Adds `sentry_sdk.DataCollection`, `KeyValueCollectionBehavior`, `HttpHeadersCollection`, and `GenAICollection`.
+  - When `data_collection` is not set, behavior is derived from `send_default_pii` (now deprecated), so upgrading without configuring `data_collection` changes nothing.
+  - `frame_context_lines` is now configurable (previously hardcoded to 5); AI integrations' `include_prompts` becomes a per-integration override of `data_collection.gen_ai`.
+
 ## 2.63.0
 
 ### Bug Fixes 🐛
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -23,6 +23,12 @@ from sentry_sdk.consts import (
     VERSION,
     ClientConstructor,
 )
+from sentry_sdk.data_collection import (
+    OFF_DATA_COLLECTION,
+    DataCollection,
+    _map_from_send_default_pii,
+    resolve_data_collection,
+)
 from sentry_sdk.envelope import Envelope, Item, PayloadRef
 from sentry_sdk.integrations import _DEFAULT_INTEGRATIONS, setup_integrations
 from sentry_sdk.integrations.dedupe import DedupeIntegration
@@ -345,11 +351,11 @@ def _get_options(*args: "Optional[str]", **kwargs: "Any") -> "Dict[str, Any]":
     if rv["enable_tracing"] is True and rv["traces_sample_rate"] is None:
         rv["traces_sample_rate"] = 1.0
 
+    rv["data_collection"] = resolve_data_collection(rv)
+
     if rv["event_scrubber"] is None:
         rv["event_scrubber"] = EventScrubber(
-            send_default_pii=(
-                False if rv["send_default_pii"] is None else rv["send_default_pii"]
-            )
+            send_default_pii=rv["data_collection"].user_info
         )
 
     if rv["socket_options"] and not isinstance(rv["socket_options"], list):
@@ -423,6 +429,23 @@ class BaseClient:
         return None
 
     def should_send_default_pii(self) -> bool:
+        return False
+
+    @property
+    def data_collection(self) -> "DataCollection":
+        return OFF_DATA_COLLECTION
+
+    def should_collect_user_info(self) -> bool:
+        return False
+
+    def should_collect_gen_ai_inputs(
+        self, include_prompts: "Optional[bool]" = None
+    ) -> bool:
+        return False
+
+    def should_collect_gen_ai_outputs(
+        self, include_prompts: "Optional[bool]" = None
+    ) -> bool:
         return False
 
     def is_active(self) -> bool:
@@ -614,6 +637,17 @@ class _Client(BaseClient):
                 self.options["error_sampler"] = sample_all
                 self.options["traces_sampler"] = sample_all
                 self.options["profiles_sampler"] = sample_all
+                # data_collection was resolved in _get_options() before this
+                # spotlight override flipped send_default_pii on. Re-derive it so
+                # the should_collect_* accessors agree with should_send_default_pii()
+                # in DSN-less spotlight mode (only when the user did not set
+                # data_collection explicitly).
+                if not self.options["data_collection"].explicit:
+                    self.options["data_collection"] = _map_from_send_default_pii(
+                        True,
+                        self.options["include_local_variables"] is not False,
+                        self.options["include_source_context"] is not False,
+                    )
 
             self.session_flusher = SessionFlusher(capture_func=_capture_envelope)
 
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
ericapisani, line 1: This would need to be removed

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,606 @@
...
+
+    return _map_from_send_default_pii(
+        bool(send_default_pii), include_local_variables, include_source_context
+    )
+
+
+#: Safe default used by non-recording clients: collect nothing PII-gated.
+#: This is a shared, process-wide singleton. Treat it as read-only — do not
+#: mutate the returned ``DataCollection`` or its nested config objects.
+OFF_DATA_COLLECTION = _map_from_send_default_pii(False, True, True)
```
ericapisani, line 606: Change `OFF_DATA_COLLECTION` name to something better - this could be clearer.

Use named args, not positional - this is hard to understand otherwise

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,606 @@
...
+]
+
+#: Default number of source lines captured above and below a stack frame.
+DEFAULT_FRAME_CONTEXT_LINES = 5
+
+#: Collection modes for key-value data (cookies, headers, query params).
+COLLECTION_OFF = "off"
+COLLECTION_DENYLIST = "denyList"
+COLLECTION_ALLOWLIST = "allowList"
+_VALID_MODES = (COLLECTION_OFF, COLLECTION_DENYLIST, COLLECTION_ALLOWLIST)
```
ericapisani, line 74: Improve the names of these variables, put the possible collection modes into an enum

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,606 @@
...
+BODY_TYPE_OUTGOING_RESPONSE = "outgoingResponse"
+
+#: All valid body types. ``http_bodies`` defaults to this (collect everything the
+#: platform supports); an empty list is the explicit opt-out.
+ALL_BODY_TYPES = [
+    BODY_TYPE_INCOMING_REQUEST,
+    BODY_TYPE_OUTGOING_REQUEST,
+    BODY_TYPE_INCOMING_RESPONSE,
+    BODY_TYPE_OUTGOING_RESPONSE,
+]
```
ericapisani, line 65: Clean this up as well:
- `ALL_BODY_TYPES` explore whether this should also be a tuple/set (should be consistent with `_VALID_MODES` below unless there's a good reason for this not to be
- `BODY_TYPE_*` could potentially become an enum. If it's not being used outside of this file, should prefix with an underscore

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,606 @@
...
+#: Collection modes for key-value data (cookies, headers, query params).
+COLLECTION_OFF = "off"
+COLLECTION_DENYLIST = "denyList"
+COLLECTION_ALLOWLIST = "allowList"
+_VALID_MODES = (COLLECTION_OFF, COLLECTION_DENYLIST, COLLECTION_ALLOWLIST)
+
+#: Canonical sensitive denylist from the spec. Values of keys that contain any of
+#: these terms (partial, case-insensitive) are always replaced with
+#: ``"[Filtered]"`` regardless of the configured collection mode.
+SENSITIVE_DENYLIST = [
```
ericapisani, line 79: See if this can be consolidated with the existing `DEFAULT_DENYLIST` in `scrubber/py`

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,606 @@
...
+    "xsrf",
+    "credentials",
+    "session",
+    "sid",
+    "identity",
+]
+
+#: Additional GDPR-sensitive terms users may opt into via custom deny terms.
+#: Not applied automatically; documented here for convenience.
+EXTENDED_GDPR_DENYLIST = ["forwarded", "-ip", "remote-", "via", "-user"]
```
ericapisani, line 101: I can see this was added because of [this section](https://develop.sentry.dev/sdk/foundations/client/data-collection/#extended-deny-terms) in the spec.

We don't currently have an extended GDPR denylist - maybe something worth considering exposing to users for convenience

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,606 @@
...
+
+    return DataCollection(**kwargs)
+
+
+def _kvcb_from_value(val: "Any") -> "KeyValueCollectionBehavior":
+    """Coerce a string or dict to :class:`KeyValueCollectionBehavior`."""
+    if isinstance(val, KeyValueCollectionBehavior):
+        return val
+    if isinstance(val, str):
+        return KeyValueCollectionBehavior(mode=val)
```
ericapisani, line 521: Reviewing the spec - I don't think this is a valid option. It looks like if it's provided, it needs at least a dict with a `mode` (see [here](https://develop.sentry.dev/sdk/foundations/client/data-collection/#key-value-collection-behavior))

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,606 @@
...
+
+
+def _resolve_explicit(
+    user_dc: "DataCollection",
+    include_local_variables: bool,
+    include_source_context: bool,
+) -> "DataCollection":
+    """
+    Fill in any omitted fields of a user-supplied ``DataCollection`` with their
+    spec defaults (resolution case A). Frame fields fall back to the legacy
```
ericapisani, line 453: no one has any context on what "resolution case X" is. remove this, and the other references of 'resolution case' within this PR

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,606 @@
...
+    resolved = DataCollection(
+        # These fields are always concrete on a constructed DataCollection.
+        user_info=user_dc.user_info,
+        cookies=user_dc.cookies,
+        http_headers=user_dc.http_headers,
+        query_params=user_dc.query_params,
+        gen_ai=user_dc.gen_ai,
+        # http_bodies: None means "all valid types"; materialize for clarity.
+        http_bodies=(
+            list(user_dc.http_bodies)
```
ericapisani, line 476: I think this `list` may be redundant as this should be coming in as list of strings when provided.

Same goes for `list(ALL_BODY_TYPES)` a couple of lines below

sentry_sdk/client.py
```diff
@@ -724,6 +758,59 @@ def should_send_default_pii(self) -> bool:
...
+        """
+        dc = self.options.get("data_collection")
+        return dc if dc is not None else OFF_DATA_COLLECTION
+
+    def should_collect_user_info(self) -> bool:
+        """
+        Returns whether the SDK should automatically populate ``user.*`` fields
+        (id, email, username, ip_address) from instrumentation.
+        """
+        return bool(self.data_collection.user_info)
```
ericapisani, line 775: `user_info` it initialized to a boolean value, so we shouldn't need to cast here.

Edit: should this, similar to `_should_collect_gen_ai_content`, also fall back to `should_send_default_pii` if `self.data_collection.explicit` is not `True` (i.e. fallback to legacy behaviour)

sentry_sdk/client.py
```diff
@@ -724,6 +758,59 @@ def should_send_default_pii(self) -> bool:
...
+        """
+        Returns whether the SDK should collect generative AI output content.
+
+        ``include_prompts`` is the integration-level override (if set, it takes
+        precedence over the global ``data_collection.gen_ai.outputs`` setting).
+        """
+        return self._should_collect_gen_ai_content("outputs", include_prompts)
+
+    def _should_collect_gen_ai_content(
+        self, direction: str, include_prompts: "Optional[bool]"
```
ericapisani, line 800: we should try to lock down the possible values of `direction` to be one of `inputs` or `output` - it's not necessary for us to have this type be so broad

sentry_sdk/client.py
```diff
@@ -724,6 +758,59 @@ def should_send_default_pii(self) -> bool:
...
+    ) -> bool:
+        dc = self.data_collection
+        if dc.explicit:
+            # Integration-level override wins over the global gen_ai setting.
+            if include_prompts is not None:
+                return include_prompts
+            return bool(getattr(dc.gen_ai, direction))
```
ericapisani, line 807: We also shouldn't need to cast to `bool` here, the values within the GenAICollection are boolean

sentry_sdk/client.py
```diff
@@ -425,6 +431,23 @@ def parsed_dsn(self) -> "Optional[Dsn]":
...
+
+    def should_collect_gen_ai_inputs(
+        self, include_prompts: "Optional[bool]" = None
+    ) -> bool:
+        return False
+
+    def should_collect_gen_ai_outputs(
+        self, include_prompts: "Optional[bool]" = None
+    ) -> bool:
+        return False
```
ericapisani, line 449: To reduce the API surface on the client, I think we should only expose the `data_collection` and put the `should_collect_X` methods on the `data_collection` class itself.

Edit: alternatively, we could introduce helper methods similar to what we do for streamed spans (`has_span_streaming_enabled`)

sentry_sdk/client.py
```diff
@@ -614,6 +637,17 @@ def _record_lost_event(
...
+                # data_collection was resolved in _get_options() before this
+                # spotlight override flipped send_default_pii on. Re-derive it so
+                # the should_collect_* accessors agree with should_send_default_pii()
+                # in DSN-less spotlight mode (only when the user did not set
+                # data_collection explicitly).
+                if not self.options["data_collection"].explicit:
+                    self.options["data_collection"] = _map_from_send_default_pii(
+                        True,
+                        self.options["include_local_variables"] is not False,
+                        self.options["include_source_context"] is not False,
```
ericapisani, line 649: Similar to another spot I mentioned this on, this is not easy to follow, should pass this in as named args, not as positional args

sentry_sdk/scope.py
```diff
@@ -2173,6 +2174,26 @@ def should_send_default_pii() -> bool:
...
+
+
+def should_collect_gen_ai_inputs(include_prompts: "Optional[bool]" = None) -> bool:
+    """Shortcut for `Scope.get_client().should_collect_gen_ai_inputs(...)`."""
+    return Scope.get_client().should_collect_gen_ai_inputs(include_prompts)
+
+
+def should_collect_gen_ai_outputs(include_prompts: "Optional[bool]" = None) -> bool:
+    """Shortcut for `Scope.get_client().should_collect_gen_ai_outputs(...)`."""
+    return Scope.get_client().should_collect_gen_ai_outputs(include_prompts)
```
ericapisani, line 2189: This mirrors what happens for `should_send_default_pii`, but as mentioned in another comment, I think this API surface should be reduced as if we do this for every type of configuration, this will become a large list (database queries, GQL documents/variables, etc.)

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,606 @@
...
+        if term in lowered:
+            return True
+    if extra_terms:
+        for term in extra_terms:
+            if term and term.lower() in lowered:
+                return True
+    return False
+
+
+def apply_key_value_collection(
```
ericapisani, line 286: This name could be better

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,606 @@
...
+        user_info=send_default_pii,
+        cookies=KeyValueCollectionBehavior(
+            COLLECTION_DENYLIST if send_default_pii else COLLECTION_OFF
+        ),
+        # Headers are collected in both PII modes today (sensitive ones filtered
+        # when PII is off), so this never maps to "off".
+        http_headers=HttpHeadersCollection(),
+        # Bodies are collected regardless of PII today, bounded by
+        # ``max_request_body_size``.
+        http_bodies=list(ALL_BODY_TYPES),
```
ericapisani, line 432: `ALL_BODY_TYPES` is already a list, it doesn't need to be cast as a list again

PR #6702, opened 2026-06-30
feat(data-collection): create DataCollection option in client

### LINKED ISSUE #6700: Create DataCollection option in client
Although this won't yet be respected by all integrations, update the client to accept the `DataCollection` option in the `init` method.

See [https://develop.sentry.dev/sdk/foundations/client/data-collection](<https://develop.sentry.dev/sdk/foundations/client/data-collection>) for expected behaviour regarding what data is collected, and how it interacts alongside the `send_default_pii` flag that currently exists to gate PII data.

ericapisani:
Support more granular PII controls via the `data_collection` option as seen in the [spec](https://develop.sentry.dev/sdk/foundations/client/data-collection).

These changes:
- introduce the option
- normalizes the user-provided options into a DataCollection dict
- handles populating the data_collection option based on what the user provided, and whether or not `send_default_pii` (which will eventually be legacy) was set.

They don't currently hook up to anything in order to keep the size of this pull request somewhat sane.

**Some high-level design decisions to call out:**

1) Introducing 2 types for the different config options - one that's more flexible for user input, the other more strict after we've received the input.

This was done so that, once we've received the config from the user and have set defaults for values that weren't provided, we can avoid needing to constantly check if the value is `None` within the SDK.

2) Including a `provided_by_user` property within the `DataCollection` type

This is mainly to deal with spotlight-related logic within the client, which you can see [here](https://github.com/getsentry/sentry-python/pull/6702/changes#diff-de17468b7fdc1d17ecfa0a4e5a26f61af3373576759e547c5c149e87b33c7bddR639).

Fixes PY-2570
Fixes #6700


ericapisani:
bugbot run

sentry_sdk/client.py
```diff
@@ -724,6 +752,14 @@ def should_send_default_pii(self) -> bool:
         """
         return self.options.get("send_default_pii") or False
 
+    @property
```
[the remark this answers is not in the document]
ericapisani: The callout regarding the EventScrubber is a good one - there were no changes made to the scrubber in this PR to factor in the granularity of the `dataCollection` config, so we shouldn't be passing in dataCollection values to it yet (we should instead be using the `send_default_pii` value).

Re: the `should_send_default_pii` helper method on the `Client` - this will be left alone as we'll want to update the various integrations to reference the specific dataCollection properties relevant to the data they may be collecting rather than use this helper which uses what will be a legacy value.

sentry_sdk/client.py
```diff
@@ -614,6 +629,19 @@ def _record_lost_event(
...
+                # DSN-less spotlight mode (only when the user did not set
+                # data_collection explicitly).
+                if not self.options["data_collection"]["provided_by_user"]:
+                    self.options["data_collection"] = _map_from_send_default_pii(
+                        send_default_pii=True,
+                        include_local_variables=self.options["include_local_variables"]
+                        is not False,
+                        include_source_context=self.options["include_source_context"]
+                        is not False,
+                    )
```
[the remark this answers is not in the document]
ericapisani: Related to the other bugbot comment - the change made involving the initialization of the EventScrubber in the Client will be reverted because there are multiple types of data handled within the scrubber that will need to reference specific configuration within the data collector, and these changes don't cover that functionality (to be done in future PRs).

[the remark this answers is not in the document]
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6702/commits/d7f823cc29b03d3cadbc1a59984b236181dcb65f

sentry_sdk/client.py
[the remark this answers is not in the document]
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6702/commits/d7f823cc29b03d3cadbc1a59984b236181dcb65f

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+
+def _database_from_value(
+    val: "dict[str, Any]",
+) -> "DatabaseCollectionBehaviour":
+    return {"query_params": val.get("query_params", True)}
+
+
+def resolve_data_collection(options: "Dict[str, Any]") -> "DataCollection":
```
alexander-alderman-webb:
```suggestion
def _resolve_data_collection(options: "Dict[str, Any]") -> "DataCollection":
```

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,272 @@
...
+
+#: Collection modes for key-value data (cookies, headers, query params).
+#: snake_case (Python-only deviation from the spec's camelCase); never
+#: serialized to Sentry.
+_VALID_KEY_VALUE_COLLECTION_BEHAVIOUR_MODES = ("off", "deny_list", "allow_list")
+
+#: Values of keys that contain any of
+#: these terms (partial, case-insensitive) are always replaced with
+#: ``"[Filtered]"`` regardless of the configured collection mode.
+SENSITIVE_DENYLIST = [
```
alexander-alderman-webb:
```suggestion
_SENSITIVE_DENYLIST = [
```

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,272 @@
...
+#: platform supports); an empty list is the explicit opt-out.
+ALL_HTTP_BODY_TYPES = [
+    "incoming_request",
+    "outgoing_request",
+    "incoming_response",
+    "outgoing_response",
+]
+
+#: Default number of source lines captured above and below a stack frame.
+DEFAULT_FRAME_CONTEXT_LINES = 5
```
alexander-alderman-webb:
```suggestion
_DEFAULT_FRAME_CONTEXT_LINES = 5
```

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,272 @@
...
+        DataCollection,
+        GenAICollectionBehaviour,
+        GraphQLCollectionBehaviour,
+        HttpHeadersCollectionBehaviour,
+        KeyValueCollectionBehaviour,
+    )
+
+#: ``http_bodies`` defaults to this (collect everything the
+#: platform supports); an empty list is the explicit opt-out.
+ALL_HTTP_BODY_TYPES = [
```
alexander-alderman-webb:
```suggestion
_ALL_HTTP_BODY_TYPES = [
```

sentry_sdk/client.py
```diff
@@ -724,6 +754,14 @@ def should_send_default_pii(self) -> bool:
         """
         return self.options.get("send_default_pii") or False
 
+    @property
+    def data_collection(self) -> "DataCollection":
+        """
+        Returns the resolved :class:`~sentry_sdk.data_collection.DataCollection`
+        config for this client.
+        """
+        return self.options["data_collection"]
```
alexander-alderman-webb: Do we need to add this accessor?

sentry_sdk/client.py
```diff
@@ -425,6 +438,10 @@ def parsed_dsn(self) -> "Optional[Dsn]":
     def should_send_default_pii(self) -> bool:
         return False
 
+    @property
+    def data_collection(self) -> "DataCollection":
+        return _DISABLED_DATA_COLLECTION_CONFIG
```
alexander-alderman-webb: Do we need to add this accessor?

alexander-alderman-webb (commented):
Just did a quick pass, not a proper review.
Can we add underscores to everything we don't want users to access? 🙏

sentry_sdk/client.py
ericapisani: Probably not, but had added it to be consistent with `dsn` and `should_send_default_pii`. While I can kind of see why we have something for `should_send_default_pii`, I'm not sure why we have `dsn` in the first place - any context?

sentry_sdk/client.py
ericapisani: We might not depending on our convo in the [other comment](https://github.com/getsentry/sentry-python/pull/6702#discussion_r3527412556) that's related.

This was added for consistency with `dsn` and `should_send_default_pii`, but otherwise I don't think we need this.

sentry_sdk/client.py
alexander-alderman-webb: There's also client options for which we do not provide accessors. We can always add the accessor later on if there's popular demand, but IMO it'd be good to avoid public API bloat.
If it's convenient for us in the integrations we can also keep a private `_data_collection()` accessor.

ericapisani:
Moving back to draft while I investigate the bot comments

sentry_sdk/client.py
ericapisani: Sounds good, will remove

sentry_sdk/_types.py
```diff
@@ -152,6 +152,63 @@ class SDKInfo(TypedDict):
         version: str
         packages: "Sequence[Mapping[str, str]]"
 
+    class KeyValueCollectionBehaviour(TypedDict):
+        mode: 'Literal["off", "deny_list", "allow_list"]'
```
sentrivana: I know in JS it's `denyList` and `allowList`, so technically two words, but I'm used to writing these as a single word. A short web search also seems to prefer the single word version.

```suggestion
        mode: 'Literal["off", "denylist", "allowlist"]'
```

sentry_sdk/_types.py
```diff
@@ -152,6 +152,63 @@ class SDKInfo(TypedDict):
...
+    class DatabaseCollectionBehaviour(TypedDict):
+        query_params: bool
+
+    class HttpHeadersCollectionUserOptions(TypedDict, total=False):
+        request: "KeyValueCollectionBehaviour"
+
+    class HttpHeadersCollectionBehaviour(TypedDict):
+        request: "KeyValueCollectionBehaviour"
+
+    class DataCollectionUserOptions(TypedDict, total=False):
```
sentrivana, line 187: For my understanding, we have these type pairs (`UserOptions` and `Behaviour`) so that users don't need to specify all options (`total=False` on the `UserOptions` types), but also so that the materialized `DataCollection` that we'll use internally won't nag us about an option potentially being `None`, because it has to be fully filled out with default values in case any are missing, right?

I see why we have it like this but I'm afraid the two might drift eventually because we forget to update one of them. 

It's not like I have better ideas 😀 The complexity we introduce with the config needs to be dealt with someplace, and I guess this is an ok place.

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,264 @@
...
+    from sentry_sdk._types import (
+        DatabaseCollectionBehaviour,
+        DataCollection,
+        GenAICollectionBehaviour,
+        GraphQLCollectionBehaviour,
+        HttpHeadersCollectionBehaviour,
+        KeyValueCollectionBehaviour,
+    )
+
+#: ``http_bodies`` defaults to this (collect everything the
```
sentrivana: What does the colon after the hash do? `#:`

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,264 @@
...
+    "outgoing_request",
+]
+
+#: Default number of source lines captured above and below a stack frame.
+_DEFAULT_FRAME_CONTEXT_LINES = 5
+
+#: Collection modes for key-value data (cookies, headers, query params).
+#: snake_case (Python-only deviation from the spec's camelCase); never
+#: serialized to Sentry.
+_VALID_KEY_VALUE_COLLECTION_BEHAVIOUR_MODES = ("off", "deny_list", "allow_list")
```
sentrivana:
```suggestion
_VALID_KEY_VALUE_COLLECTION_BEHAVIOUR_MODES = ("off", "denylist", "allowlist")
```

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,264 @@
...
+    send_default_pii: bool,
+    include_local_variables: bool,
+    include_source_context: bool,
+) -> "DataCollection":
+    """
+    Build a fully-resolved ``DataCollection`` dict that mirrors the data
+    ``send_default_pii`` collects today. Used when ``data_collection`` is not
+    provided explicitly.
+    """
+    kv_mode: "Literal['deny_list', 'off']" = "deny_list" if send_default_pii else "off"
```
sentrivana:
```suggestion
    kv_mode: "Literal['denylist', 'off']" = "denylist" if send_default_pii else "off"
```

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,264 @@
...
+    terms = [] if send_default_pii else ["forwarded", "-ip", "remote-", "via", "-user"]
+
+    return {
+        "provided_by_user": False,
+        "user_info": send_default_pii,
+        "cookies": {"mode": kv_mode, "terms": terms},
+        # Headers are collected in both PII modes today (sensitive ones filtered
+        # when PII is off), so this never maps to "off".
+        "http_headers": {
+            "request": {"mode": "deny_list", "terms": terms},
```
sentrivana:
```suggestion
            "request": {"mode": "denylist", "terms": terms},
```

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,264 @@
...
+        "database": _database_from_value(d.get("database") or {}),
+        "stack_frame_variables": stack_frame_variables,
+        "frame_context_lines": frame_context_lines,
+    }
+
+
+def _kvcb_from_value(
+    val: "dict[str, Any]",
+) -> "KeyValueCollectionBehaviour":
+    mode = val.get("mode", "deny_list")
```
sentrivana:
```suggestion
    mode = val.get("mode", "denylist")
```

I'll stop commenting on all the `deny_list`s now, you get the gist 😁

sentry_sdk/consts.py
```diff
@@ -1432,6 +1434,23 @@ def __init__(
             If you enable this option, be sure to manually remove what you don't want to send using our features for
             managing `Sensitive Data <https://docs.sentry.io/data-management/sensitive-data/>`_.
 
+        :param data_collection: (EXPERIMENTAL) Structured configuration controlling what data integrations collect automatically,
```
sentrivana, line 1437: Maybe mention https://docs.sentry.io/platforms/python/configuration/options/#data_collection somewhere in this docstring

CHANGELOG.md
```diff
@@ -66,7 +66,6 @@ Self-hosted Sentry users should opt out with `stream_gen_ai_spans=False`, since
 
 - (test): exclude django 6.1 alphas and betas from tox by @ericapisani in [#6690](https://github.com/getsentry/sentry-python/pull/6690)
 - (test): need to include the alpha tag for mcp package inclusion by @ericapisani in [#6688](https://github.com/getsentry/sentry-python/pull/6688)
-
```
sentrivana, line 69: Please revert so that the file doesn't display weird when rendered

sentrivana (commented):
LGTM. Left a few comments.

sentry_sdk/_types.py
ericapisani: > For my understanding, we have these type pairs (UserOptions and Behaviour) so that users don't need to specify all options (total=False on the UserOptions types), but also so that the materialized DataCollection that we'll use internally won't nag us about an option potentially being None, because it has to be fully filled out with default values in case any are missing, right?

Yup, you've got it! I wanted to spare us from all the extra code we'd have to write to satisfy the typechecker when performing the checks within the integrations 😅 

I agree with you about the risk of drift. I thought it'd be worth it in this case because after the data collection spec is implemented and settled, the amount of updates we'd need to do here would likely be minimal.

If we think of anything better throughout the implementation of the spec though, happy to come back and change the approach! 😄

sentry_sdk/data_collection.py
```diff
@@ -0,0 +1,264 @@
...
+def _http_headers_from_value(
+    val: "dict[str, Any]",
+) -> "HttpHeadersCollectionBehaviour":
+    return {
+        "request": (
+            _kvcb_from_value(val["request"])
+            if "request" in val
+            else _kvcb_from_value({"mode": "deny_list"})
+        ),
+    }
```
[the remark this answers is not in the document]
ericapisani: String inputs instead of dict inputs should not be provided - [the spec](https://develop.sentry.dev/sdk/foundations/client/data-collection/#key-value-collection-behavior) outlines KVCB inputs as the following:

CleanShot 2026-07-07 at 08 30 04@2x

so if someone provides "off", it should be provided as `{ mode: "off" }`

sentry_sdk/data_collection.py
ericapisani: Nothing 😅 This looks like a fun thing the AI did in the initial pass from the other week that I didn't notice when I was checking the comments

CHANGELOG.md
ericapisani: Addressed in [ff55bad](https://github.com/getsentry/sentry-python/pull/6702/commits/ff55badbfd181dcb71e1dafe637b04c8083c7fee)

sentry_sdk/consts.py
ericapisani: Done in [ff55bad](https://github.com/getsentry/sentry-python/pull/6702/commits/ff55badbfd181dcb71e1dafe637b04c8083c7fee)

alexander-alderman-webb (approved):

PR #6757, opened 2026-07-06
fix(tracing): Fix unsampled/deferred trace propagation in span streaming

### LINKED ISSUE #6755: Sampling decision is deferred when it should be unsampled
The SDK in stream mode falls back to the propagation context too eagerly. When a no op streamed span is active, use it to populate propagation headers. Only use propagation context if tracing is disabled.

sentrivana:
### Description
The SDK didn't behave correctly in some trace propagation scenarios in stream mode.

The current SDK design was emitting headers with a deferred sampling decision if the sampling decision was negative (whether inherited or not), because it was falling back to TwP (tracing without performance) behavior too eagerly.

Additionally, some edge cases concerning `ignore_spans` and its interaction with sampling were clarified and the necessary changes included in this PR.

## How it should work

In the following, _tracing enabled_ means either a `traces_sample_rate` or a `traces_sampler` is defined. _Tracing disabled_ means both are `None` and tracing without performance is active.

There are three kinds of sampling decisions: sampled (`1`), unsampled (`0`), and deferred (``), meaning no decision.

1. No incoming trace headers
    - tracing enabled: make a sampling decision and propagate it
    - tracing disabled: do not make a decision, propagate a deferred decision (no sampling decision)
2. Incoming trace headers with a sampled/unsampled decision
    - tracing enabled: respect and propagate incoming decision
    - tracing disabled: respect and propagate incoming decision
3. Incoming trace headers with a deferred sampling decision (essentially same as 1)
    - tracing enabled: make a sampling decision and propagate it
    - tracing disabled: propagate deferred sampling decision

## The problem and the fix

If a no-op span was active, we were falling back to the propagation context to generate propagation headers, which by default doesn't contain a sampling decision.

With this PR, `NoOpStreamedSpans` now carry their own trace and span ID information as well as the current propagation data (a snapshot of the propagation context, like `StreamedSpans`) and use it to emit the correct headers. This mirrors old behavior: spans would always be created, whether they'd be sampled or not, and they would be responsible for generating propagation headers; with the fallback to the propagation context only kicking in if there is no span.

I recommend checking out the new tests to see what behavior is now expected.

## Edge cases

The following cases regarding `ignore_spans` were clarified and will be documented soon:

- An outgoing request is made from an ignored segment: `sentry-trace` should have an unsampled decision (`-0`), `parent_span_id` is the ID of the segment (or any random span ID, since for all intents and purposes, the ignored segment doesn't actually exist for any downstream service)
- An outgoing request is made from an ignored child span of a non-ignored segment: `sentry-trace`'s sampling decision should be the segment's sampling decision; `parent_span_id` should be the first non-ignored span ancestor

#### Issues
Closes https://github.com/getsentry/sentry-python/issues/6755

#### Reminders


sentrivana:
Converting to draft to address bot comments

sentry_sdk/traces.py
```diff
@@ -610,15 +610,37 @@ def _to_json(self) -> "SpanJSON":
...
     )
 
     def __init__(
         self,
+        segment: "Optional[StreamedSpan]" = None,
+        trace_id: "Optional[str]" = None,
+        parent_span_id: "Optional[str]" = None,
+        parent_sampled: "Optional[bool]" = None,
+        baggage: "Optional[Baggage]" = None,
+        sampled: "Optional[bool]" = False,
```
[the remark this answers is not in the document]
sentrivana: I will verify but we might need to take the segment's `sampled` flag into account here, instead of the child. It used to not make a difference, but now it does.

sentry_sdk/scope.py
```diff
@@ -1295,13 +1295,18 @@ def start_streamed_span(
...
             propagation_context = self.get_active_propagation_context()
 
             if is_ignored_span(name, attributes):
                 return NoOpStreamedSpan(
                     scope=self,
+                    segment=None,
+                    trace_id=propagation_context.trace_id,
+                    parent_span_id=propagation_context.parent_span_id,
+                    parent_sampled=propagation_context.parent_sampled,
+                    baggage=propagation_context.baggage,
```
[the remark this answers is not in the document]
sentrivana: This is now by design.

sentry_sdk/traces.py
```diff
@@ -736,6 +749,14 @@ def start_timestamp(self) -> "Optional[datetime]":
     def end_timestamp(self) -> "Optional[datetime]":
         return None
 
+    def _get_trace_context(self) -> "dict[str, Any]":
+        return {
+            "trace_id": self.trace_id,
+            "span_id": self.span_id,
+            "parent_span_id": self._parent_span_id,
+            "dynamic_sampling_context": self._dynamic_sampling_context(),
+        }
```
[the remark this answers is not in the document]
sentrivana: That's fine

sentry_sdk/scope.py
```diff
@@ -1314,10 +1319,18 @@ def start_streamed_span(
             if sample_rate is not None:
                 self._update_sample_rate(sample_rate)
 
-            if sampled is False:
+            if sampled is False or sampled is None:
```
ericapisani, line 1322: This can also be written as to remove the extra conditional:

```suggestion
            if bool(sampled):
```

sentry_sdk/traces.py
```diff
@@ -610,15 +610,39 @@ def _to_json(self) -> "SpanJSON":
 
 class NoOpStreamedSpan(StreamedSpan):
     __slots__ = (
+        "_trace_id",
+        "_span_id",
+        "_sampled",
+        "_segment",
```
ericapisani: Of these, only `_sampled` needs to be added as the other 3 already exist on the parent 's `__slots__` and will still be accessible by instances of `NoOpStreamedSpan`.

As an example:

```bash
>>> class Foo():
...     __slots__ = ("x")
...
>>> class Bar(Foo):
...     __slots__ = ("y")
...
>>> b = Bar()
>>> b.x = "hi"
>>> b.y = "world"
>>> print(b.x)
hi
>>> b.z
  File "<python-input-7>", line 1, in <module>
    b.z
AttributeError: 'Bar' object has no attribute 'z'
```

One thing to be wary of that I found when looking into the slots docs - duplicating the variable name in `__slots__` as shown here creates a shadow of the parent value.

ericapisani (commented):
Haven't had a chance to look at the tests yet, but noticed a couple things on my first pass

ericapisani (approved):
Other than the __slots__ thing that I mentioned earlier, this LGTM

sentry_sdk/scope.py
sentrivana: I'd prefer to keep the original to keep it more explicit. It's sort of a reminder that `sampled` can also be `None`, which is a case that sometimes gets overlooked.

sentry_sdk/traces.py
sentrivana: Ty, removed in https://github.com/getsentry/sentry-python/pull/6757/commits/dca94467d01a713dcfff3471340d3467c4bc4ede

PR #6786, opened 2026-07-08
fix(serializer): avoid copying large Mapping locals before trimming

### LINKED ISSUE #6773: `serialize()` copies large Mapping locals in full before trimming to MAX_DATABAG_BREADTH

gitcommit90:
fix(serializer): avoid copying large Mapping locals before trimming

`_Serializer._serialize_node_impl()` now limits the temporary copy
to at most `remaining_breadth + 1` pairs when breadth is bounded,
while keeping the existing full-copy path for unbounded breadth.
This avoids O(len(mapping)) work on huge locals that will be trimmed
anyway (especially frame locals during exception capture).

- Added a regression test using a counting `Mapping` to prove
  the serializer only iterates/lookups the bounded prefix + 1.
- Preserves mutation safety during iteration via `dict()` copy.
- `_annotate(len=...)` now reports the original length, not the
  truncated copy size.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

Closes #6773

### DIFF
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -3,6 +3,7 @@ import sys
 from array import array
 from collections.abc import Mapping
 from datetime import datetime
+from itertools import islice
 from typing import TYPE_CHECKING
 
 from sentry_sdk.utils import (
@@ -301,14 +302,18 @@ class _Serializer:
         elif isinstance(obj, Mapping):
             # Create temporary copy here to avoid calling too much code that
             # might mutate our dictionary while we're still iterating over it.
-            obj = dict(obj.items())
+            obj_len = len(obj)
+            if isinstance(remaining_breadth, int):
+                obj = dict(islice(obj.items(), remaining_breadth + 1))
+            else:
+                obj = dict(obj.items())
 
             rv_dict: "Dict[str, Any]" = {}
             i = 0
 
             for k, v in obj.items():
                 if remaining_breadth is not None and i >= remaining_breadth:
-                    self._annotate(len=len(obj))
+                    self._annotate(len=obj_len)
                     break
 
                 str_k = str(k)
--- a/tests/test_serializer.py
+++ b/tests/test_serializer.py
@@ -1,6 +1,7 @@
 import gc
 import re
 from array import array
+from collections.abc import Mapping
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


alexander-alderman-webb:
There's already an in-progress PR for this: https://github.com/getsentry/sentry-python/pull/6783/changes

Issue #6741, opened 2026-07-03
Create helper for scrubbing request cookies


ericapisani:
Currently scrubbing for request cookies is centralized in the `EventScrubber`, and assumes that it's a dictionary (as most frameworks convert the cookies to dictionaries as part of handling the request).

In order to make it easier for us to apply both the granular filtering that's done on dictionaries when the `dataCollection` config is provided - and to handle the fallback of a string being provided in which case we replace the whole thing with `[Filtered]` - we should create a small helper function to handle both cases and use it within the `EventScrubber` .

PR #6788, opened 2026-07-08
ref(integrations): route HTTP header filtering through data collection config

### LINKED ISSUE #6744: Update `_filter_headers` to check data collection settings when available
`_filter_headers` handles the scrubbing of sensitive data that lives within the request headers, and is used across various integrations.

Update this function to respect the settings within the data collection object when it's available.

ericapisani:
`_filter_headers` follows the legacy `send_default_pii`/`use_annotated_value` behaviour when the data collection behaviour is not enabled in `_experiments`. 

When data collection is enabled, it now delegates to `_apply_key_value_collection_filtering` from `sentry_sdk.data_collection`, so header scrubbing respects the new `data_collection.http_headers.request` allowlist/denylist/off configuration when data collection.

Work to scrub cookies in a more granular way will be tackled as part of PY-2581/#6741.

Fixes PY-2584
Fixes #6744


sentry_sdk/integrations/_wsgi_common.py
```diff
@@ -4,8 +4,9 @@
 
 import sentry_sdk
 from sentry_sdk._types import SENSITIVE_DATA_SUBSTITUTE
+from sentry_sdk.data_collection import _apply_key_value_collection_filtering
```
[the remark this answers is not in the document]
ericapisani: Those terms now need to be added as part of an extended deny-list by the user when data collection is enabled. See https://develop.sentry.dev/sdk/foundations/client/data-collection/#sensitive-denylist

sentry_sdk/data_collection.py
```diff
@@ -77,6 +79,62 @@
...
+
+def _apply_key_value_collection_filtering(
+    items: "Mapping[str, Any]",
+    behaviour: "KeyValueCollectionBehaviour",
+    substitute: "Any" = SENSITIVE_DATA_SUBSTITUTE,
+) -> "Dict[str, Any]":
+
+    if behaviour["mode"] == "off":
+        return {}
+
```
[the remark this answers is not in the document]
ericapisani: This was actually a good find

sentry_sdk/integrations/_asgi_common.py
```diff
@@ -115,9 +115,13 @@ def _get_request_data(
     if ty in ("http", "websocket"):
         request_data["method"] = asgi_scope.get("method")
 
-        request_data["headers"] = headers = _filter_headers(
-            _get_headers(asgi_scope),
+        headers = _get_headers(asgi_scope)
+
+        request_data["headers"] = _filter_headers(
+            headers,
+            use_annotated_value=False,
```
ericapisani, line 122: This change was done because when if the headers are filtered with "allowlist" and no terms are provided, and the "host" header is present, then the constructed URL in `_get_url` below would contain `[Filtered]` within the URL.

Doing this ensures that it gets filtered from the headers to respect data collection, but still correctly appears in the url

sentry_sdk/data_collection.py
```diff
@@ -77,6 +79,62 @@
...
+    result: "Dict[str, Any]" = {}
```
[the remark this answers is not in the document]
ericapisani: In the data collection spec, it's specified that users wanting to filter out `X-Forwarded-For` need to provide extended deny terms to the configuration. See https://develop.sentry.dev/sdk/foundations/client/data-collection/#extended-deny-terms

sentry_sdk/data_collection.py
```diff
@@ -77,6 +79,62 @@
...
+        if isinstance(key, str) and _is_sensitive_key(
+            key=key, extra_terms=behaviour.get("terms", [])
+        ):
+            result[key] = substitute
+        else:
+            result[key] = value
+    return result
+
+
 def _map_from_send_default_pii(
```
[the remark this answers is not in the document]
ericapisani: We're okay with this behavior because the more granular filtering will happen in a subsequent pull request that's within the stack.

sentry_sdk/integrations/_wsgi_common.py
```diff
@@ -206,19 +207,39 @@ def _filter_headers(
...
+
+        for key in filtered:
+            if isinstance(key, str) and key.lower() in ("cookie", "set-cookie"):
+                filtered[key] = SENSITIVE_DATA_SUBSTITUTE
 
-    return {
-        k: (v if k.upper().replace("-", "_") not in SENSITIVE_HEADERS else substitute)
-        for k, v in headers.items()
-    }
+        return filtered
```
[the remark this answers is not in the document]
ericapisani: This behavior conforms to the new data collection spec, and so this is expected.

sentry_sdk/integrations/_wsgi_common.py
```diff
@@ -206,19 +207,39 @@ def _filter_headers(
...
+        data_collection_configuration = client_options["data_collection"]
+
+        filtered = _apply_key_value_collection_filtering(
+            items=headers,
+            behaviour=data_collection_configuration["http_headers"]["request"],
+        )
+
+        for key in filtered:
+            if isinstance(key, str) and key.lower() in ("cookie", "set-cookie"):
+                filtered[key] = SENSITIVE_DATA_SUBSTITUTE
```
sentrivana, line 222: > Work to scrub cookies in a more granular way will be tackled as part of PY-2581/https://github.com/getsentry/sentry-python/issues/6741.

This means this cookies logic will be updated later right?

tests/integrations/aiohttp/test_aiohttp.py
```diff
@@ -1254,8 +1254,185 @@ async def hello(request):
...
+    sentry_init, aiohttp_client, capture_items, options, expected, request
+):
+    sentry_init(
+        integrations=[AioHttpIntegration()],
+        traces_sample_rate=1.0,
+        send_default_pii=options["send_default_pii"],
+        _experiments={
+            "trace_lifecycle": "stream",
+            "data_collection": options["data_collection"],
+        },
```
sentrivana, line 1381:
```suggestion
        trace_lifecycle="stream",
        _experiments={
            "data_collection": options["data_collection"],
        },
```

tests/integrations/aiohttp/test_aiohttp.py
```diff
@@ -1254,8 +1254,185 @@ async def hello(request):
...
+            "x-custom-header": "foobar",
+            "Cookie": "sessionid=secret",
+        },
+    )
+
+    sentry_sdk.flush()
+
+    (server_span,) = [item.payload for item in items]
+
+    if request.node.callspec.id.endswith("data_collection_off_does_not_add_headers"):
```
sentrivana: Can we, instead of matching on the id here, just match on `expected is None`?
```suggestion
    if expected is None:
```

tests/integrations/aiohttp/test_aiohttp.py
```diff
@@ -1254,8 +1254,185 @@ async def hello(request):
...
+
+    # client.address and user.ip_address is captured under send_default_pii=True.
+    # TODO: This block will eventually need to be removed from this test into a separate
+    # test once data collection gating is introduced on these values
+    if options["send_default_pii"]:
+        assert server_span["attributes"]["client.address"] == "127.0.0.1"
+        assert server_span["attributes"]["user.ip_address"] == "127.0.0.1"
+    else:
+        assert "user.ip_address" not in server_span["attributes"]
+        assert "client.address" not in server_span["attributes"]
```
sentrivana, line 1431: Why have this here in the first place? It doesn't look related to sensitive header behavior

tests/integrations/asgi/test_asgi.py
```diff
@@ -831,6 +837,87 @@ def test_get_headers():
     }
 
 
+def test_get_request_data_url_with_filtered_host(sentry_init):
```
sentrivana, line 840: Could we change these new test cases to set up a proper ASGI app, like the tests below? Just to make them a little less unit-testy.

tests/integrations/quart/test_quart.py
```diff
@@ -861,8 +861,171 @@ async def test_span_streaming_request_attributes_with_pii(sentry_init, capture_i
...
+    )
+    assert response.status_code == 200
+
+    sentry_sdk.flush()
+
+    spans = [item.payload for item in items]
+    assert len(spans) == 1
+
+    segment = spans[0]
+    if request.node.callspec.id == "data_collection_off_does_not_add_headers":
```
sentrivana: Same suggestion as above, to make this a bit more future-proof if someone decides to change the ID.

```suggestion
    if expected is None:
```

sentrivana (approved):
Lgtm! Left some suggestions.

sentry_sdk/integrations/_asgi_common.py
sentrivana: As long as this doesn't affect current behavior should be ok 👍🏻

sentry_sdk/integrations/_wsgi_common.py
ericapisani: Correct, it's part of https://github.com/getsentry/sentry-python/pull/6797 😄

tests/integrations/aiohttp/test_aiohttp.py
ericapisani: Yeah, that works here 👍🏻

tests/integrations/aiohttp/test_aiohttp.py
ericapisani: It isn't related to sensitive header behaviour, but I had decided to add this in to make it clear that it isn't expected that user data would be present when data collection is enabled just yet.

Since assertions on these values appear in other tests within the file, I thought that not having assertions like these in a data collection-specific context would raise flags.

tests/integrations/asgi/test_asgi.py
ericapisani: I missed this earlier on, but I addressed this in https://github.com/getsentry/sentry-python/pull/6841/changes#diff-a4ca49dac9c5fcce27a4fb5fb53e5f8e764b807773ba7c5476d70d74a9789353 as 

I can look into pulling that change into this branch, but if we wanted to merge this stack 1 big feature branch, it won't make a difference as it'll eventually be applied.

sentrivana: Ahh the joys of stacked PRs 😄 No worries then -- as long as it's in the final changeset eventually all good!

PR #6792, opened 2026-07-09
fix(litellm): Set operation name from call type instead of chat fallback

### LINKED ISSUE #6442: Only set accurate operation name in `litellm`

VihaanAgarwal:
Fixes #6442.

The input callback mapped every call type except `embedding`/`aembedding` to `chat`, so a `litellm.image_generation()` call showed up in Sentry as `chat dall-e-3` with `gen_ai.operation.name: chat`. Same story for text completions, speech, transcription, rerank and the rest.

This replaces the binary fallback with an explicit table of the call types we can name accurately:

- `completion`/`acompletion` -> `chat`
- `text_completion`/`atext_completion` -> `text_completion`
- `embedding`/`aembedding` -> `embeddings`
- `responses`/`aresponses` -> `responses`

Each maps to its matching `OP.GEN_AI_*` span op (the constants for `text_completion` and `responses` already exist and are used by the langchain and openai integrations).

For everything else, `gen_ai.operation.name` is simply not set. Recording a guess is what this issue is about, and litellm has 140+ call types (batch, files, assistants, video...), most of which have no semconv equivalent. The raw call type still ends up in the span name (`image_generation dall-e-3`), so the spans stay identifiable. I kept `gen_ai.chat` as the span op fallback for unmapped types so those spans keep flowing into the AI views like before; happy to change that if you'd rather they drop out.

One test tweak: `test_response_without_usage` built callback kwargs without a `call_type` and asserted the span was named `chat ...`, which only held because of the fallback. Real callbacks always carry `call_type` (litellm's `Logging.pre_call` sets it), so I added `"call_type": "completion"` to those kwargs rather than special casing a missing key as chat.

Tests cover all eight mapped call types plus the unknown case. The litellm test module passes locally except `test_multiple_providers`, which errors at setup on current master too (missing optional provider package in my env).

### DIFF
--- a/sentry_sdk/integrations/litellm.py
+++ b/sentry_sdk/integrations/litellm.py
@@ -22,7 +22,7 @@ from sentry_sdk.utils import event_from_exception
 
 if TYPE_CHECKING:
     from datetime import datetime
-    from typing import Any, Dict, List, Optional, Tuple
+    from typing import Any, Dict, List, Tuple
 
 try:
     import litellm  # type: ignore[import-not-found]
@@ -36,8 +36,8 @@ except ImportError:
 _SPAN_KEY = "_sentry_span"
 
 # Call types whose gen_ai operation name we can determine accurately. Everything
-# else gets no gen_ai.operation.name attribute, since guessing records wrong data.
-_CALL_TYPE_OPERATIONS: "Dict[Any, Tuple[Optional[str], str]]" = {
+# else is not instrumented, since guessing records wrong data.
+_CALL_TYPE_OPERATIONS: "Dict[Any, Tuple[str, str]]" = {
     "completion": ("chat", consts.OP.GEN_AI_CHAT),
     "acompletion": ("chat", consts.OP.GEN_AI_CHAT),
     "text_completion": ("text_completion", consts.OP.GEN_AI_TEXT_COMPLETION),
@@ -96,6 +96,12 @@ def _input_callback(kwargs: "Dict[str, Any]") -> None:
     if integration is None:
         return
 
+    call_type = kwargs.get("call_type", None)
+    if call_type not in _CALL_TYPE_OPERATIONS:
+        return
+
+    operation, span_op = _CALL_TYPE_OPERATIONS[call_type]
+
     # Get key parameters
     full_model = kwargs.get("model", "")
     try:
@@ -104,11 +110,7 @@ def _input_callback(kwargs: "Dict[str, Any]") -> None:
         model = full_model
         provider = "unknown"
 
-    call_type = kwargs.get("call_type", None)
-    operation, span_op = _CALL_TYPE_OPERATIONS.get(
-        call_type, (None, consts.OP.GEN_AI_CHAT)
-    )
-    span_name = f"{operation or call_type or 'unknown'} {model}"
+    span_name = f"{operation} {model}"
 
     # Start a new span/transaction
     if has_span_streaming_enabled(client.options):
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/litellm.py
```diff
@@ -92,32 +105,24 @@ def _input_callback(kwargs: "Dict[str, Any]") -> None:
...
 
     call_type = kwargs.get("call_type", None)
-    if call_type == "embedding" or call_type == "aembedding":
-        operation = "embeddings"
-    else:
-        operation = "chat"
+    operation, span_op = _CALL_TYPE_OPERATIONS.get(
+        call_type, (None, consts.OP.GEN_AI_CHAT)
+    )
+    span_name = f"{operation or call_type or 'unknown'} {model}"
```
alexander-alderman-webb: No data is better than wrong data, so please exit early if the key is not in the dictionary.

tests/integrations/litellm/test_litellm.py
```diff
@@ -3415,3 +3417,81 @@ def test_convert_message_parts_image_url_missing_url():
...
+            "model": "gpt-3.5-turbo",
+            "call_type": call_type,
+        }
+
+        _input_callback(kwargs)
+        _success_callback(
+            kwargs,
+            MockCompletionResponse(),
+            datetime.now(),
+            datetime.now(),
```
alexander-alderman-webb: For tests, the expectation is that you use the library like a user would and assert that the correct telemetry was emitted.
The existing tests have prior art for how to achieve this.

alexander-alderman-webb (changes requested):

VihaanAgarwal:
Fair points, both addressed. Unknown call types now bail before any span is created, so nothing gets recorded for them. Also replaced the callback-level tests with ones that go through litellm.text_completion / litellm.responses / litellm.image_generation with mocked transports and assert on the captured spans.

tests/integrations/litellm/test_litellm.py
```diff
@@ -3415,3 +3419,147 @@ def test_convert_message_parts_image_url_missing_url():
...
+    assert span["data"][SPANDATA.GEN_AI_OPERATION_NAME] == "responses"
+
+
+def test_unknown_call_type_is_not_instrumented(
+    sentry_init,
+    capture_events,
+    get_model_response,
+    reset_litellm_executor,
+):
+    """Call types with no accurate operation name emit no span at all."""
```
alexander-alderman-webb: Please do not test a negative.
We test that telemetry is emitted for certain operations, and **not** that telemetry doesn't exist sometimes.

VihaanAgarwal: Makes sense, dropped it (and the now-unused image imports). The text_completion and responses tests still cover the mapping.

VihaanAgarwal:
Also extended the two new tests to assert the prompt is recorded, since the review bots flagged that `text_completion` and `responses` might lose it. They don't: litellm hands the input to the callback as `messages` in both cases, and the tests now pin that.

Anything else you'd like changed here?

VihaanAgarwal:
Rebased on master, PR is mergeable again. Only conflict was the litellm test file, where master and this branch both appended tests at the end; kept both. Squashed the four commits into one while I was in there.

litellm suite on the rebased branch: 165 passed, plus the 8 `test_multiple_providers` setup errors that also happen on clean master here (missing optional provider package).

VihaanAgarwal:
@alexander-alderman-webb this has been ready since the last push. The negative test is gone, and what's left drives `text_completion` and `responses` through litellm the way a user would, then asserts the span each one emits. Anything else you want changed?

PR #6797, opened 2026-07-10
feat(integrations): apply data_collection cookie filtering to wsgi, starlette, litestar, starlite

### LINKED ISSUE #6741: Create helper for scrubbing request cookies

ericapisani:
Extends the granular cookie collection controls (data_collection.cookies)
to _wsgi_common, starlette, litestar, and starlite, matching the behavior
already used elsewhere. Falls back to should_send_default_pii() when
data_collection is not configured for cookies.

HTTP "Cookie" and "set-cookie" headers will continue to be completely filtered
with the "[Filtered]" value.

Fixes PY-2581
Fixes https://github.com/getsentry/sentry-python/issues/6741


sentry_sdk/integrations/fastapi.py
```diff
@@ -118,7 +117,7 @@ def event_processor(event: "Event", hint: "Dict[str, Any]") -> "Event":
             # Extract information from request
             request_info = event.get("request", {})
             if info:
-                if "cookies" in info and should_send_default_pii():
+                if "cookies" in info:
```
ericapisani, line 120: the `should_send_default_pii` check here is no longer needed since this check happens within `StarletteRequestExtractor`

sentry_sdk/integrations/starlette.py
```diff
@@ -716,8 +718,15 @@ def __init__(self: "StarletteRequestExtractor", request: "Request") -> None:
     def extract_cookies_from_request(
         self: "StarletteRequestExtractor",
     ) -> "Optional[Dict[str, Any]]":
+        client_options = sentry_sdk.get_client().options
         cookies: "Optional[Dict[str, Any]]" = None
-        if should_send_default_pii():
```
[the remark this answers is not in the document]
ericapisani: Addressed in https://github.com/getsentry/sentry-python/pull/6797/commits/620e222a6f2bc1960d1775ec1cc886d063001b59

tests/integrations/django/test_data_scrubbing.py
```diff
@@ -99,3 +102,118 @@ def test_scrub_django_custom_session_cookies_filtered(
...
+
+@pytest.mark.forked
+@pytest_mark_django_db_decorator()
+@pytest.mark.parametrize(
+    "cookies_to_set, data_collection, expected_cookies",
+    [
+        pytest.param(
+            {"sessionid": "123", "csrftoken": "456", "foo": "bar"},
+            {"cookies": {"mode": "off"}},
+            NO_COOKIES,
```
sentrivana: Why not just `None`?

sentry_sdk/integrations/fastapi.py
sentrivana: Been some time since I last looked at the extractor code -- so the Starlette extractor runs after this and overwrites the cookies?

Just wanted to double-check that we have the precedence right (and that the cookies are really missing if `should_send_default_pii=False` after this change)

sentrivana (approved):
Looks good! Had a small question regarding the event processor and the NO_COOKIES sentinel in the tests.

sentry_sdk/integrations/fastapi.py
ericapisani: > Been some time since I last looked at the extractor code -- so the Starlette extractor runs after this and overwrites the cookies?

It runs before this, and then, yes, overwrites the cookies. 

If `should_send_default_pii` is false or if the data collection configuration filters out cookies, then the `cookies` key/value doesn't get set on `info` within the Starlette extractor's `extract_request_info` (and line 121 doesn't run).

tests/integrations/django/test_data_scrubbing.py
ericapisani: I wanted to try it out to see if it read a bit better, but happy to change this to `None` to match what's happened on the previous branch if we'd rather not have an extra variable lying around in the test file:)

sentrivana: It just caught my eye since I'm not used to seeing `object()` sentinels outside of situations where you need a default value different from `None`.

I think in this case a `None` would be simpler, but either way is fine, don't want to nitpick on this too much. 😄

ericapisani: > I think in this case a None would be simpler, but either way is fine, don't want to nitpick on this too much. 😄

Don't worry about it, it's all good! 😄 I'll switch it back to `None` for consistency 🚀

ericapisani: Addressed in [19d0b12](https://github.com/getsentry/sentry-python/pull/6797/commits/19d0b1248bb978fc4e62d81b193ba68e43a9841d)

PR #6802, opened 2026-07-13
fix(match_regex_list): handle empty string patterns without IndexError

### LINKED ISSUE #6504: match_regex_list raises IndexError when a pattern list contains an empty string

hiro-nikaitou:
An empty string pattern in the regex list causes an `IndexError: string index out of range` when `item_matcher[-1]` is accessed. This is reachable through integrations like Celery (`CeleryIntegration.exclude_beat_tasks`) when users accidentally configure an empty pattern.

Skip empty string patterns in the loop to prevent the crash. A pattern with no characters is a no-op anyway.

Closes #6504

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -1736,6 +1736,8 @@ def match_regex_list(
         return False
 
     for item_matcher in regex_list:
+        if item_matcher == "":
+            continue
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb:

PR #6808, opened 2026-07-13
fix(tracing): Skip child span creation in streaming path when no current span (databases)

sentrivana:
## Summary
- When span streaming is enabled and there is no current span, database integrations should not create new root segments for child-span operations
- Adds a `sentry_sdk.traces.get_current_span() is None` guard before creating spans in the streaming path
- Affected integrations: aiomysql, asyncpg, clickhouse_driver, pymongo, redis

Part of https://github.com/getsentry/sentry-python/issues/6819


sentry_sdk/integrations/aiomysql.py
ericapisani: With the new change below where we may return early, `span_attributes` may be created unnecessarily.

As a cleanup, could consider moving this below the `if sentry_sdk.traces.get_current_span() is None:` block

ericapisani (approved):
LGTM! 🚀

PR #6811, opened 2026-07-13
fix(tracing): Skip child span creation in streaming path when no current span (HTTP clients)

sentrivana:
## Summary
- When span streaming is enabled and there is no current span, HTTP client integrations should not create new root segments for child-span operations
- Adds a `sentry_sdk.traces.get_current_span() is None` guard before creating spans in the streaming path
- Affected integrations: httpx, httpx2, pyreqwest, boto3, stdlib (urllib)

## Test plan
- [ ] Existing tests pass
- [ ] Verify that in streaming mode, no orphan root segments are created for outgoing HTTP requests when there's no active span

### DIFF
--- a/sentry_sdk/integrations/httpx.py
+++ b/sentry_sdk/integrations/httpx.py
@@ -4,18 +4,15 @@ import sentry_sdk
 from sentry_sdk.consts import OP, SPANDATA
 from sentry_sdk.integrations import DidNotEnable, Integration
 from sentry_sdk.scope import should_send_default_pii
-from sentry_sdk.tracing import BAGGAGE_HEADER_NAME
 from sentry_sdk.tracing_utils import (
     add_http_request_source,
-    add_sentry_baggage_to_headers,
     has_span_streaming_enabled,
-    should_propagate_trace,
+    propagate_trace_headers,
 )
 from sentry_sdk.utils import (
     SENSITIVE_DATA_SUBSTITUTE,
     capture_internal_exceptions,
     ensure_integration_enabled,
-    logger,
     parse_url,
 )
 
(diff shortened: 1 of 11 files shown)


ericapisani (dismissed):
Blocking because we're losing trace propagation within some integrations with these changes. Will work on fixes now

sentry_sdk/integrations/httpx.py
```diff
@@ -60,6 +57,9 @@ def send(self: "Client", request: "Request", **kwargs: "Any") -> "Response":
             parsed_url = parse_url(str(request.url), sanitize=False)
 
         if is_span_streaming_enabled:
+            if sentry_sdk.traces.get_current_span() is None:
+                return real_send(self, request, **kwargs)
```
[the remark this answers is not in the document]
ericapisani: Fixes in [c8c9924](https://github.com/getsentry/sentry-python/pull/6811/commits/c8c992431ac7fd4376fb74fd899540f956927819)

sentry_sdk/tracing_utils.py
```diff
@@ -973,6 +973,31 @@ def should_propagate_trace(client: "sentry_sdk.client.BaseClient", url: str) ->
     return match_regex_list(url, trace_propagation_targets, substring_matching=True)
 
 
+def propagate_trace_headers(
+    client: "sentry_sdk.client.BaseClient", request: "Any"
```
sl0thentr0py, line 977: `request` is `Any` here and since this is a shared helper, I'd make it future proof by enforcing the contract of what it needs, so please add `hasattr` checks for `url` and `headers` below.

sentry_sdk/integrations/pyreqwest.py
```diff
@@ -88,18 +89,22 @@ def _sentry_pyreqwest_span(request: "Request") -> "Generator[Any, None, None]":
...
-                "sentry.op": OP.HTTP_CLIENT,
-                "sentry.origin": PyreqwestIntegration.origin,
-                SPANDATA.HTTP_REQUEST_METHOD: request.method,
-            },
-        ) as span:
-            if parsed_url is not None and should_send_default_pii():
-                span.set_attribute(SPANDATA.URL_FULL, parsed_url.url)
-                span.set_attribute(SPANDATA.URL_QUERY, parsed_url.query)
-                span.set_attribute(SPANDATA.URL_FRAGMENT, parsed_url.fragment)
+        span_ctx = nullcontext()
```
sl0thentr0py: can we not early return here like all the integrations above

sl0thentr0py (changes requested):

sl0thentr0py (approved):

sentry_sdk/integrations/pyreqwest.py
ericapisani: It's not visible in this diff, but we still want to propagate the trace which happens within the context manager a little further down.

PR #6828, opened 2026-07-15
docs: Remove AI commit attribution requirement from AGENTS.md

ericapisani:
Drop the rule requiring a Co-Authored-By line for AI-generated commits.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,14 +26,6 @@ Use **ruff** for linting and formatting:
 - `uv run ruff check --fix tests sentry_sdk`
 - `uv run ruff format tests sentry_sdk`
 
-## Commit Attribution
-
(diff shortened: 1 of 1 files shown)


cleptric:
We actually would prefer if we can attribute commits to a certain model. If you use `sentry-skill:commit` this works out of the box.

ericapisani:
> If you use sentry-skill:commit this works out of the box.

I think my version of this skill may have been out of date so I'll see what happens now that I've refreshed it 👀 

> We actually would prefer if we can attribute commits to a certain model.

Can you provide context on why? What are we hoping to do with this information? Do we have any dashboards, etc. that's tied to this info?

alexander-alderman-webb (approved):

PR #6832, opened 2026-07-16
feat(asgi): Apply data_collection filtering to URL query strings

ericapisani:
Filter query string parameters in the ASGI request data and span
attributes according to the data_collection.url_query_params config,
mirroring the existing WSGI behaviour. Falls back to the legacy
send_default_pii check when data_collection is not enabled.

Refs PY-2583
Refs #6743


tests/integrations/asgi/test_asgi.py
sentrivana, line 1: Same q as on one of the PRs before: can we make these a bit less unit-testy and set up an ASGI app instead, like the tests after these? (I suspect this might also already be part of a future PR, feel free to disregard in that case.)

sentrivana (approved):
See one comment.

tests/integrations/asgi/test_asgi.py
ericapisani: I'll confirm if I've already taken care of it in that later PR in the stack I mentioned elsewhere and address it there if I haven't already

ericapisani: Confirmed that as of the most recent branch in the stack (pr 13), these were all updated to use ASGI apps in their tests

tests/integrations/asgi/test_asgi.py
```diff
@@ -918,6 +918,215 @@ def test_get_request_attributes_url_with_headers_off(sentry_init):
...
+        ),
+        pytest.param(
+            {},
+            QUERY_STRING,
+            id="defaults",
+        ),
+        pytest.param(
+            {"_experiments": {"data_collection": {}}},
+            "token=[Filtered]&theme=dark&lang=en&session=[Filtered]",
+            id="data_collection_denylist_default",
```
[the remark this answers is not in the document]
ericapisani: Fixed in [657c186](https://github.com/getsentry/sentry-python/pull/6832/commits/657c1860aca59fb9de8ed9affe925bc47204a505)

PR #6833, opened 2026-07-16
feat(aiohttp): Apply data_collection filtering to URL query strings

ericapisani:
Use the new data_collection settings to filter url.query and related<br>span attributes for both the aiohttp server and client integrations,<br>falling back to should_send_default_pii() when data_collection is not<br>configured.

Refs getsentry/sentry-python#6743<br>Refs getsentry/sentry-python#6743


sentry_sdk/integrations/aiohttp.py
```diff
@@ -161,7 +165,26 @@ async def sentry_app_handle(
                             )
 
                         url_attributes = {}
-                        if should_send_default_pii():
+                        if has_data_collection_enabled(client.options):
+                            url_attributes["url.full"] = "%s://%s%s" % (
```
sentrivana, line 169: I know this is just copied from the `should_send_default_pii` branch, but I've just realized it's wrong: `url.full` should contain the query and it doesn't.

Probably makes more sense to fix this afterwards, when this stack has been merged. I'll make an issue for it.

sentrivana (approved):
Nice!

sentry_sdk/integrations/aiohttp.py
sentrivana: https://github.com/getsentry/sentry-python/issues/6860

ericapisani: Thanks! 🙏🏻

PR #6835, opened 2026-07-17
feat(tracing): Send `sentry.segment.name.source` instead of `sentry.span.source` attribute

Lms24:
Following the [deprecation](https://getsentry.github.io/sentry-conventions/attributes/sentry/#sentry-span-source) of `sentry.span.source`, this PR renames the attribute from `sentry.span.source` to [`sentry.segment.name.source`](https://getsentry.github.io/sentry-conventions/attributes/sentry/#sentry-segment-name-source). 

Background: We currently decided not do do segment name clustering for streamed spans but should we ever choose to do it, `sentry.segment.name.source` will be the attribute carrying the name source for it. Either way, `sentry.span.source` has no effect anymore. 

Reviewers, please let me know about your preferences:
- Technically, the renaming might break user callbacks relying on the old attribute. Since this only affects streamed spans I _think_ it's fine but if you have other policies for span first, we can also double-write.
- I decided to rename `SegmentSource` to `SegmentNameSource` for consistency. Is this considered breaking? If so, happy to revert as it's purely a cosmetic change IMHO.


sentry_sdk/integrations/celery/__init__.py
```diff
@@ -361,7 +362,7 @@ def _inner(*args: "Any", **kwargs: "Any") -> "Any":
                         parent_span=None,  # make this a segment
                         attributes={
                             "sentry.origin": CeleryIntegration.origin,
-                            "sentry.span.source": TransactionSource.TASK.value,
+                            "sentry.segment.name.source": SegmentNameSource.TASK.value,
```
Lms24, line 364: Seer pointed out that the "wrong" enum was used here (no big deal as the values were identical). So I changed it over to `SegmentNameSource`

sentry_sdk/integrations/rq.py
```diff
@@ -91,7 +91,7 @@ def sentry_patched_perform_job(
                         attributes={
                             "sentry.op": OP.QUEUE_TASK_RQ,
                             "sentry.origin": RqIntegration.origin,
-                            "sentry.span.source": SegmentSource.TASK,
+                            "sentry.segment.name.source": SegmentNameSource.TASK,
                             SPANDATA.MESSAGING_MESSAGE_ID: job.id,
                             SPANDATA.MESSAGING_DESTINATION_NAME: queue.name,
                         },
```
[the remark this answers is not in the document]
Lms24: I don't think this is a real issue, given that we set `SegmentSource.TASK` before. But will leave open for reviewers. Happy to adjust if I'm missing something.

ericapisani:
Still need to take a look at the changes themselves but to initially answer your questions in the pull request summary:

> Technically, the renaming might break user callbacks relying on the old attribute. Since this only affects streamed spans I think it's fine but if you have other policies for span first, we can also double-write.

As of right now I don't believe there are callbacks relying on this attribute so I think this should be a safe to do right now.

> I decided to rename SegmentSource to SegmentNameSource for consistency. Is this considered breaking? If so, happy to revert as it's purely a cosmetic change IMHO.

Since the underlying value hasn't changed, this isn't considered this breaking. 🚀

tests/integrations/aiohttp/test_aiohttp.py
```diff
@@ -1609,6 +1609,11 @@ async def hello(request):
 
     child_span, server_span = [item.payload for item in items]
 
+    assert server_span["attributes"]["sentry.segment.name.source"] == "component"
+    assert "sentry.segment.name.source" not in child_span["attributes"]
+    assert "sentry.span.source" not in server_span["attributes"]
+    assert "sentry.span.source" not in child_span["attributes"]
```
ericapisani: Since `sentry.span.source` is no longer an attribute we're emitting, these assertions are somewhat meaningless and should be removed

sentry_sdk/integrations/rq.py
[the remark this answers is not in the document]
ericapisani: Nope, it's not a real issue, you're good to leave this as is.

sentry_sdk/traces.py
```diff
@@ -65,9 +65,9 @@ def __str__(self) -> str:
 _VALID_SPAN_STATUSES = frozenset(e.value for e in SpanStatus)
 
 
-# Segment source, see
-# https://getsentry.github.io/sentry-conventions/generated/attributes/sentry.html#sentryspansource
-class SegmentSource(str, Enum):
+# Segment name source, see
```
ericapisani:
```suggestion
# The source of a segment name, see
```

ericapisani (approved):
🚀

tests/integrations/aiohttp/test_aiohttp.py
Lms24: yup agreed, thanks for flagging!

PR #6842, opened 2026-07-19
fix: skip empty patterns in match_regex_list to prevent IndexError

### LINKED ISSUE #6504: match_regex_list raises IndexError when a pattern list contains an empty string

mangodxd:
Fixes #6504

`match_regex_list` indexes `item_matcher[-1]` to check for `$` suffix. When the pattern list contains an empty string, `""[-1]` raises `IndexError: string index out of range`.

This is reachable from `CeleryIntegration.exclude_beat_tasks` when user configures an empty string in the pattern list.

Skip empty patterns with `if not item_matcher: continue` before the index check. Empty strings cannot match anything, so this is the correct behavior.

Three test cases added:
- `[""]` single empty pattern → no match
- `["", "some-string"]` empty then valid → matches valid
- `["", ""]` all empty → no match

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -1736,6 +1736,9 @@ def match_regex_list(
         return False
 
     for item_matcher in regex_list:
+        if not item_matcher:
+            continue
+
         if not substring_matching and item_matcher[-1] != "$":
             item_matcher += "$"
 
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb:

PR #6853, opened 2026-07-20
fix(tracing): handle exceptions raised within traces_sampler and other callbacks

ericapisani:
When a user-provided traces_sampler callback raises, catch the exception,
log a warning, and fall back to the parent sample rate or the configured
traces_sample_rate instead of propagating the error.

Wrap the before_send callback invocation in capture_internal_exceptions()
so that exceptions raised by user-provided before_send_log,
before_send_metric, and before_send_span callbacks do not crash the
application. When a callback raises, the original, unmodified telemetry
item is sent.

Fixes PY-2617
Fixes #6850

### DIFF
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -1244,7 +1244,8 @@ class _Client(BaseClient):
             serialized = telemetry._to_json()  # type: ignore[union-attr]
 
         if before_send is not None:
-            serialized = before_send(serialized, {})  # type: ignore[arg-type]
+            with capture_internal_exceptions():
+                serialized = before_send(serialized, {})  # type: ignore[arg-type]
 
             if ty in ("log", "metric"):
                 # Logs and metrics can be dropped in their respective
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -1244,8 +1244,16 @@ class _Client(BaseClient):
             serialized = telemetry._to_json()  # type: ignore[union-attr]
 
         if before_send is not None:
+            exception_raised_in_before_send_func = False
             with capture_internal_exceptions():
-                serialized = before_send(serialized, {})  # type: ignore[arg-type]
+                try:
+                    serialized = before_send(serialized, {})  # type: ignore[arg-type]
+                except Exception:
+                    exception_raised_in_before_send_func = True
+                    raise
+
+            if exception_raised_in_before_send_func:
+                return
 
             if ty in ("log", "metric"):
                 # Logs and metrics can be dropped in their respective
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -1264,6 +1264,11 @@ class _Client(BaseClient):
                     return
 
             elif ty == "span" and isinstance(telemetry, StreamedSpan):
+                # Reset the span to its original value before we attempted
+                # to call the `before_send_span` callback
+                if exception_raised_in_before_send_func:
+                    serialized = telemetry._to_json()
+
                 # Spans can't be dropped in before_send_span by design. They can
                 # be altered though (e.g. to sanitize). Only allow changes to
                 # name and attributes.
--- a/sentry_sdk/client.py
+++ b/sentry_sdk/client.py
@@ -1252,10 +1252,12 @@ class _Client(BaseClient):
                     exception_raised_in_before_send_func = True
                     raise
 
-            if exception_raised_in_before_send_func:
-                return
-
             if ty in ("log", "metric"):
+                # We are ok with dropping metrics and logs when an exception is raised
+                # because we allow users to drop them in their respect before_send_*
+                # functions.
+                if exception_raised_in_before_send_func:
+                    return
                 # Logs and metrics can be dropped in their respective
                 # before_send, so if we get None, don't queue them for sending.
                 if serialized is None:
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -1575,7 +1575,10 @@ def add_sentry_baggage_to_headers(
         stripped_existing_baggage + separator + sentry_baggage
     )
 
-def _get_fallback_sample_rate(client: "BaseClient", propagation_context: "PropagationContext") -> "Union[float, bool]":
+
+def _get_fallback_sample_rate(
+    client: "BaseClient", propagation_context: "PropagationContext"
+) -> "Union[float, bool]":
     if propagation_context.parent_sampled is not None:
         propagation_context_sample_rate = propagation_context._sample_rate()
 
@@ -1585,6 +1588,7 @@ def _get_fallback_sample_rate(client: "BaseClient", propagation_context: "Propag
             return propagation_context.parent_sampled
     else:
         return client.options["traces_sample_rate"]
+
 
 def _make_sampling_decision(
     name: str,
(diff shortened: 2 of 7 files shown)


alexander-alderman-webb (approved):
Nice !

sentry_sdk/tracing_utils.py
```diff
@@ -1630,23 +1645,28 @@ def _make_sampling_decision(
...
+            )
     else:
-        if propagation_context.parent_sampled is not None:
-            if propagation_context._sample_rate() is not None:
-                sample_rate = propagation_context._sample_rate()
-            else:
-                sample_rate = propagation_context.parent_sampled
-        else:
-            sample_rate = client.options["traces_sample_rate"]
+        sample_rate = _get_fallback_sample_rate(
```
sentrivana: I'd change the name of the helper function since in the context of this `else` branch, it's not really a fallback.

Maybe something like `_get_effective_sample_rate`?

tests/tracing/test_sampling.py
```diff
@@ -204,6 +204,96 @@ def test_tolerates_traces_sampler_returning_a_boolean_span_streaming(
...
+    with sentry_sdk.traces.start_span(name="dogpark") as span:
+        assert span.sampled is expected_decision
+
+
+def test_traces_sampler_raising_falls_back_to_propagated_sample_rate_span_streaming(
+    sentry_init,
+):
+    # parent said "don't sample", but its propagated sample rate of 0.75
+    # combined with sample_rand 0.5 should win over both the flag and
+    # traces_sample_rate=0.0
```
sentrivana: This is actually not correct. If the parent sample decision is 0, that should take precedence over anything else.

Can we remove this test? I think I know where this needs fixing (it's not something you've introduced in this PR), and I can follow up [with a fix](https://github.com/getsentry/sentry-python/issues/6856) after we've merged this. (I expect some other tests to break so don't want to tack this onto your PR.)

tests/tracing/test_sampling.py
```diff
@@ -204,6 +204,96 @@ def test_tolerates_traces_sampler_returning_a_boolean_span_streaming(
...
+    sentry_init(
+        traces_sampler=mock.Mock(side_effect=ValueError("boom")),
+        traces_sample_rate=0.0,
+        _experiments={
+            "trace_lifecycle": "stream",
+        },
```
sentrivana:
```suggestion
        trace_lifecycle="stream",
```

tests/tracing/test_sampling.py
```diff
@@ -204,6 +204,96 @@ def test_tolerates_traces_sampler_returning_a_boolean_span_streaming(
...
+    sentry_init,
+    traces_sample_rate,
+    expected_decision,
+):
+    sentry_init(
+        traces_sampler=mock.Mock(side_effect=ValueError("boom")),
+        traces_sample_rate=traces_sample_rate,
+        _experiments={
+            "trace_lifecycle": "stream",
+        },
```
sentrivana:
```suggestion
        trace_lifecycle="stream",
```

tests/test_logs.py
```diff
@@ -162,6 +162,32 @@ def _before_log(record, hint):
...
+    sentry_sdk.logger.error("This is an error log...")
+
+    get_client().flush()
+    logs = [item.payload for item in items]
+
+    # The exception in before_send_log is swallowed and the original,
+    # unmodified log is sent.
+    assert len(logs) == 1
+    assert logs[0]["body"] == "This is an error log..."
+    assert logs[0]["attributes"]["sentry.severity_text"] == "error"
```
sentrivana: We should drop the log in this case

tests/test_metrics.py
```diff
@@ -247,6 +247,29 @@ def _before_metric(record, hint):
...
+
+    sentry_sdk.metrics.count("test.keep", 1)
+
+    get_client().flush()
+
+    # The exception in before_send_metric is swallowed and the original,
+    # unmodified metric is sent.
+    metrics = [item.payload for item in items]
+    assert len(metrics) == 1
+    assert metrics[0]["name"] == "test.keep"
```
sentrivana: Same here -- the metric should not be sent

tests/tracing/test_sampling.py
```diff
@@ -204,6 +204,96 @@ def test_tolerates_traces_sampler_returning_a_boolean_span_streaming(
...
+        name="dogpark", baggage=baggage, parent_sampled=parent_sampling_decision
+    )
+    assert transaction.sampled is parent_sampling_decision
+
+
+@pytest.mark.parametrize(
+    "traces_sample_rate,expected_decision",
+    [(0.0, False), (0.25, False), (0.75, True), (1.00, True)],
+)
+def test_traces_sampler_raising_falls_back_to_traces_sample_rate_span_streaming(
```
sentrivana, line 246: Can we also add a span streaming test case with no incoming trace propagation headers/no continue_trace and check that we fall back to `traces_sample_rate`?

tests/tracing/test_sampling.py
```diff
@@ -204,6 +204,96 @@ def test_tolerates_traces_sampler_returning_a_boolean_span_streaming(
         assert span.sampled is traces_sampler_return_value
 
 
+@pytest.mark.parametrize(
```
sentrivana, line 205: [not related to this specific line or testcase] One case I'm not quite sure of: there is no incoming trace, `traces_sampler` is defined and throws, and there is no `traces_sample_rate` defined. We should have a test case (or two: streaming and non streaming) for that too.

sentrivana (changes requested):
Looks good, just a couple places where the behavior or tests need some tweaking.

Re: dropping/keeping the metrics/logs/spans that went through a faulty `before_send_*`: might need more alignment on this. Spans likely shouldn't be dropped, but metrics and logs we might want to drop. There's a convo on slack about this.

tests/tracing/test_sampling.py
ericapisani: Will add a couple, where the assertion is that the span/transaction isn't sampled

sentrivana (approved):
Approving to not block -- as long as the comments have been addressed feel free to merge 🙏🏻

sentry_sdk/client.py
```diff
@@ -1244,9 +1244,20 @@ def _capture_telemetry(
...
 
             if ty in ("log", "metric"):
+                # We are ok with dropping metrics and logs when an exception is raised
+                # because we allow users to drop them in their respect before_send_*
+                # functions.
+                if exception_raised_in_before_send_func:
+                    return
                 # Logs and metrics can be dropped in their respective
                 # before_send, so if we get None, don't queue them for sending.
                 if serialized is None:
```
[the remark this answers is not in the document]
ericapisani: Fixed in [227a4c4](https://github.com/getsentry/sentry-python/pull/6853/commits/227a4c4a8382439f2a3c90f55c09df7d81fb48b8)

sentry_sdk/tracing_utils.py
```diff
@@ -1630,23 +1644,28 @@ def _make_sampling_decision(
...
+        )
 
     # Validate whether the sample_rate we got is actually valid. Since
     # traces_sampler is user-provided, it could return anything.
     if not is_valid_sample_rate(sample_rate, source="Tracing"):
         logger.warning(f"[Tracing] Discarding {name} because of invalid sample rate.")
         return False, None, None, "sample_rate"
 
-    sample_rate = float(sample_rate)  # type: ignore[arg-type]
+    sample_rate = float(sample_rate)
```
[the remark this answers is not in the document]
ericapisani: Fixed in [f3e440e](https://github.com/getsentry/sentry-python/pull/6853/commits/f3e440eecb926fc411e0986816276adb17b3f203)

PR #6861, opened 2026-07-22
fix: Include query string and fragment in `url.full` span attribute

### LINKED ISSUE #6860: Audit if the `url.full` attribute contains the full URL
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

latest

### Steps to Reproduce

Look at our web framework and client integrations and what we're setting `url.full` to there. 

### Expected Result

`url.full` should contain the whole url, including the query string.

### Actual Result

Sometimes it doesn't (e.g. aiohttp).

sentrivana:
The `url.full` span attribute was only being set to the base URL (scheme + host + path), omitting query string and fragment.

Closes https://github.com/getsentry/sentry-python/issues/6860

### DIFF
--- a/sentry_sdk/integrations/aiohttp.py
+++ b/sentry_sdk/integrations/aiohttp.py
@@ -161,11 +161,15 @@ class AioHttpIntegration(Integration):
 
                         url_attributes = {}
                         if should_send_default_pii():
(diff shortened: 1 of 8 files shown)


alexander-alderman-webb (approved):
Nice!

PR #6863, opened 2026-07-23
fix(sampling): `sentry-trace` should take precedence over baggage

### LINKED ISSUE #6856: Incorrect sampling decision propagated in some cases in span streaming
If the incoming decision is `0`, we should always respect that and propagate a negative sampling decision (not talking about custom `traces_sampler`s here). It looks like in some cases, when the `sample_rand` and `sample_rate` found in incoming baggage are inconsistent with the sampling decision, we instead take them into account.

sentrivana:
### Description

We were making the wrong sampling decision in a specific case: when the incoming `sentry-trace` ended in `-0`, but the incoming baggage had a combination of `sentry-sample_rate` and `sentry-sample_rand` that would've actually produced a positive sampling decision, we would prioritize the information from the baggage over the parent sampling decision from `sentry-trace`.

This PR aligns the behavior with the legacy codepath, where we'd set `parent_sampled` as the new `sample_rate`.

#### Issues
* closes https://linear.app/getsentry/issue/PY-2618/incorrect-sampling-decision-propagated-in-some-cases-in-span-streaming
* closes https://github.com/getsentry/sentry-python/issues/6856

#### Reminders

### DIFF
--- a/tests/tracing/test_span_streaming.py
+++ b/tests/tracing/test_span_streaming.py
@@ -1016,20 +1016,15 @@ def test_outgoing_traceparent_and_baggage_inconsistent_incoming_trace(
     trace_id = "0af7651916cd43dd8448eb211c80319c"
     parent_span_id = "b7ad6b7169203331"
 
-    incoming_baggage = {
-        "sentry-trace_id": trace_id,
-    }
-
     # Baggage says sampled, sentry-trace says unsampled. sentry-trace takes
     # precedence
     incoming_sentry_trace = f"{trace_id}-{parent_span_id}-0"
-    incoming_baggage.update(
-        {
-            "sentry-sample_rate": "0.75",
-            "sentry-sample_rand": "0.500000",
-            "sentry-sampled": "true",
-        }
-    )
+    incoming_baggage = {
+        "sentry-trace_id": trace_id,
+        "sentry-sample_rate": "0.75",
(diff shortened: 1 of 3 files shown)


tests/tracing/test_span_streaming.py
```diff
@@ -988,10 +988,72 @@ def test_outgoing_traceparent_and_baggage_incoming_trace(
...
+    }
+
+    # Baggage says sampled, sentry-trace says unsampled. sentry-trace takes
+    # precedence
+    incoming_sentry_trace = f"{trace_id}-{parent_span_id}-0"
+    incoming_baggage.update(
+        {
+            "sentry-sample_rate": "0.75",
+            "sentry-sample_rand": "0.500000",
+            "sentry-sampled": "true",
```
ericapisani: Why not add these to the initial definition of `incoming_baggage` above?

ericapisani (approved):
LGTM, just noticed a small thing

alexander-alderman-webb (approved):

tests/tracing/test_span_streaming.py
sentrivana: Thanks for noticing! Copied this test from another one that only did this conditionally and missed this.

Fixed in https://github.com/getsentry/sentry-python/pull/6863/changes/ddc7386b5a118180cca13b11b532659d4f1912dd

PR #6864, opened 2026-07-23
fix(tracing): Respect parent_sampled=False in span streaming sampling decision (#6856)

### LINKED ISSUE #6856: Incorrect sampling decision propagated in some cases in span streaming

mangodxd:
Respect incoming `parent_sampled=False` in the span streaming model — do not override it with baggage `sample_rate`/`sample_rand`.

**WHY**

When `sampled=0` arrives in the `sentry-trace` header but baggage contains a positive `sample_rate` with a `sample_rand` that happens to fall below it, `_get_effective_sample_rate` ignored the explicit `parent_sampled=False` and used the baggage rate instead. The downstream `sampled = sample_rand < sample_rate` check could then incorrectly mark the span as sampled.

The existing `test_continue_trace_unsampled` was masked: it used `sample_rand=0.999999` with `sample_rate=0.5`, so `0.999999 < 0.5 = False` — the bug didn't surface.

**HOW**

`_get_effective_sample_rate` now checks `parent_sampled` before consulting baggage. If `parent_sampled` is `False`, it returns `False` immediately. The baggage `sample_rate` is only used when the parent was explicitly sampled (`True`).

**Verify**
```
python -m pytest tests/tracing/test_span_streaming.py -k "test_continue_trace_unsampled" -v
```

Fixes #6856

### DIFF
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -1579,12 +1579,15 @@ def _get_effective_sample_rate(
     client: "Any", propagation_context: "PropagationContext"
 ) -> "Union[float, bool]":
     if propagation_context.parent_sampled is not None:
-        propagation_context_sample_rate = propagation_context._sample_rate()
+        if propagation_context.parent_sampled:
+            propagation_context_sample_rate = propagation_context._sample_rate()
 
-        if propagation_context_sample_rate is not None:
-            return propagation_context_sample_rate
+            if propagation_context_sample_rate is not None:
+                return propagation_context_sample_rate
+            else:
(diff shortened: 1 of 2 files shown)


sentrivana:
Thanks @mangodxd, we're already taking care of this ourselves. Please don't open PRs for issues that have someone assigned to them to avoid double work.

Issue #6769, opened 2026-07-07
`sentry_sdk.init()` slowed by module-level `package_version("mcp")` scanning all installed distributions (2.64.0)


tejas7777:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.64.0

### Steps to Reproduce

In any environment with a few hundred installed packages and `mcp` **not** installed:

```python
import time
from sentry_sdk.integrations import DidNotEnable

t = time.perf_counter()
try:
    import sentry_sdk.integrations.mcp
except DidNotEnable:
    pass
print(f'{time.perf_counter() - t:.3f}s')
```

With ~200 installed distributions: **0.202s** on 2.64.0 vs **0.052s** on 2.63.0. The cost scales with the number of installed distributions.

`sentry_sdk.init()` imports this module through auto-enabling integrations, so every `init()` pays this even when `mcp` is not installed. In a process where all other imports are already warm (e.g. a forked worker), `init()` goes from ~0.04s to ~0.28s.

### Expected Result

When `mcp` is not installed, the MCP integration raises `DidNotEnable` without doing any expensive work, as in 2.63.0.

### Actual Result

#6583 added a module-level call at the top of `sentry_sdk/integrations/mcp.py`, above the `try/except ImportError` guard:

```python
MCP_PACKAGE_VERSION = package_version("mcp")
```

`package_version()` -> `_get_installed_modules()` parses the metadata of every installed distribution. Profiling `init()` attributes nearly all of the added time to this line.

This mainly hurts short-lived processes that call `init()` per process (test suites forking workers, CLIs, cron jobs) - our CI worker-boot timing assertions started failing on the 2.63.0 -> 2.64.0 bump.

Suggested fix: move the `package_version("mcp")` call below the `import mcp` guard (or compute it lazily in `setup_once()`), so environments without `mcp` skip the scan.

alexander-alderman-webb:
Thanks @tejas7777 I'll address

alexander-alderman-webb:
This was resolved in https://github.com/getsentry/sentry-python/commit/5bd85b442a53322dadb5987cb54a81c62cff03c3, released in 2.67.0 of the SDK.

PR #6865, opened 2026-07-23
perf: Avoid scanning installed packages when MCP SDK is not installed

mangodxd:
## WHY

When the MCP SDK is not installed, module-level `package_version("mcp")` at the top of `sentry_sdk/integrations/mcp.py` scans every installed distribution via `_get_installed_modules()`. This slows `sentry_sdk.init()` by ~0.2s in environments with hundreds of packages.

Issue #6769 reports that this regression was introduced by #6583, which added a module-level `MCP_PACKAGE_VERSION = package_version("mcp")` above the `try/except ImportError` guard.

## HOW

Move `MCP_PACKAGE_VERSION = package_version("mcp")` from module-level (above the guard) to inside the `try` block, after the successful `import mcp`. When `import mcp` fails and `DidNotEnable` is raised, `package_version()` is never called — so environments without MCP skip the expensive distribution scan entirely.

### DIFF
--- a/sentry_sdk/integrations/mcp.py
+++ b/sentry_sdk/integrations/mcp.py
@@ -20,13 +20,13 @@ from sentry_sdk.traces import StreamedSpan
 from sentry_sdk.tracing_utils import has_span_streaming_enabled
 from sentry_sdk.utils import nullcontext, package_version, safe_serialize
 
-MCP_PACKAGE_VERSION = package_version("mcp")
-
 try:
     from mcp.server.lowlevel import Server
     from mcp.server.streamable_http import (
         StreamableHTTPServerTransport,
     )
+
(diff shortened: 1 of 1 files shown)


Chenghao999 (commented):
LGTM — moving `package_version("mcp")` inside the try block skips the import-time package scan whenever MCP is not installed, which is a nice small perf win for the common case.

I checked the safety of the move: `MCP_PACKAGE_VERSION` is still referenced at module level outside the try (the version-gated imports and `_patch_*` selection), but the module either fully loads (assignment runs) or aborts with `DidNotEnable`, so there is no path where it is read while undefined. Clean change.

alexander-alderman-webb (approved):

PR #6868, opened 2026-07-23
feat(django): Gate user identity behind data_collection config

ericapisani:
Apply the data_collection.user_info setting to Django user identity (id, email, username) collection, giving it precedence over the legacy send_default_pii boolean when explicitly configured. This mirrors the existing behavior for client IP/user info already shipped for Flask, WSGI, and ASGI.

The user extraction logic in _after_get_response is pulled into a shared helper used by both the WSGI and ASGI event processors, and the parametrized send_default_pii/data_collection test matrix is extracted into tests/integrations/django/utils.py for reuse across the WSGI and ASGI test suites.

Refs PY-2583
Refs #6746

### DIFF
--- a/tests/integrations/django/asgi/test_asgi.py
+++ b/tests/integrations/django/asgi/test_asgi.py
@@ -16,6 +16,10 @@ from sentry_sdk.consts import SPANDATA
 from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.django.asgi import _asgi_middleware_mixin_factory
 from tests.integrations.django.myapp.asgi import channels_application
+from tests.integrations.django.utils import (
+    USER_INFO_INIT_KWARGS,
+    pytest_mark_django_db_decorator,
+)
 
 try:
     from django.urls import reverse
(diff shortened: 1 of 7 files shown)


tests/integrations/django/asgi/test_asgi.py
```diff
@@ -1096,3 +1100,33 @@ async def test_async_middleware_process_exception_is_awaited(
 
     assert response["status"] == 200
     assert response["body"] == b"handled by async process_exception"
+
+
+@pytest.mark.forked
```
sentrivana, line 1105: Do we need the `@pytest.mark.forked` here? It tends to take longer and it sometimes causes hard to debug problems with tests, so if it's not necessary, would be better to omit it.

sentrivana (approved):
🚀 Just one comment/question.

tests/integrations/django/asgi/test_asgi.py
ericapisani: > Do we need the @pytest.mark.forked here?

I think we might - this is a copy/paste from database-related tests in the [test_data_scrubbing](https://github.com/getsentry/sentry-python/blob/d3ca0b1c015cc09769d6985830c271e360c68875/tests/integrations/django/test_data_scrubbing.py) file and my understanding from it is that any time a test touches a database, we need this decorator to isolate state.

I might've come to an incorrect understanding though, so please let me know if that's not quite right.

Issue #6085, opened 2026-04-16
`asyncio.iscoroutinefunction` usage in integrations raises deprecation warning on 3.14


cooperoptigrid:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.58.0

### Steps to Reproduce

On Python 3.14.4, `asyncio.iscoroutinefunction` is deprecated in favour of `inspect.iscoroutinefunction`.

https://github.com/python/cpython/blob/2fdccb3f88603cb72d311ca304ab7fcf9df12d00/Lib/asyncio/coroutines.py#L20-L27

This raises a warning when using FastAPI's integration, in both `sentry_sdk/integrations/starlette.py` and `sentry_sdk/integrations/fastapi.py`

I manually converted them to use `inspect` in my `venv`s site package and the warning disappeared as expected. I have not checked for all usage of `asyncio.iscoroutinefunction`.

Reproduction using `pytest`

```python
# main.py
from fastapi import FastAPI
import sentry_sdk
from sentry_sdk.integrations.fastapi import FastApiIntegration

sentry_sdk.init(
    dsn="valid-dsn",
    integrations=[FastApiIntegration()],
)

app = FastAPI()

@app.get("/")
async def root():
    return {"message": "Hello World"}

# test_main.py
from main import app
from fastapi.testclient import TestClient

client = TestClient(app)

def test_app():
    response = client.get("/")
    assert response.status_code == 200
```

```
Package              Version
-------------------- ---------
annotated-doc        0.0.4
annotated-types      0.7.0
anyio                4.13.0
certifi              2026.2.25
click                8.3.2
dnspython            2.8.0
email-validator      2.3.0
fastapi              0.136.0
fastapi-cli          0.0.24
fastapi-cloud-cli    0.17.0
fastar               0.11.0
h11                  0.16.0
httpcore             1.0.9
httptools            0.7.1
httpx                0.28.1
idna                 3.11
iniconfig            2.3.0
jinja2               3.1.6
markdown-it-py       4.0.0
markupsafe           3.0.3
mdurl                0.1.2
packaging            26.1
pluggy               1.6.0
pydantic             2.13.1
pydantic-core        2.46.1
pydantic-extra-types 2.11.1
pydantic-settings    2.13.1
pygments             2.20.0
pytest               9.0.3
python-dotenv        1.2.2
python-multipart     0.0.26
pyyaml               6.0.3
rich                 15.0.0
rich-toolkit         0.19.7
[...]

PR #6869, opened 2026-07-23
fix(django): Use inspect.iscoroutinefunction on Python 3.14+

Digenis:
### Description

Fix asyncio.iscoroutinefunction deprecation warning as mentioned in #6085.

This PR touches code that had Python < 3.5 compatibility leftovers
so I dropped them in a preceding commit to reduce churn.

#### Issues

* contributes to: #6085

#### Notes

Sibling to #3027

### DIFF
--- a/sentry_sdk/integrations/django/views.py
+++ b/sentry_sdk/integrations/django/views.py
@@ -10,15 +10,12 @@ if TYPE_CHECKING:
     from typing import Any
 
 
-try:
-    from asyncio import iscoroutinefunction
-except ImportError:
-    iscoroutinefunction = None  # type: ignore
+from asyncio import iscoroutinefunction
 
 
 try:
     from sentry_sdk.integrations.django.asgi import wrap_async_view
-except (ImportError, SyntaxError):
+except ImportError:  # Django < 3.0
     wrap_async_view = None  # type: ignore
 
 
@@ -63,12 +60,7 @@ def patch_views() -> None:
 
         integration = sentry_sdk.get_client().get_integration(DjangoIntegration)
         if integration is not None:
-            is_async_view = (
-                iscoroutinefunction is not None
-                and wrap_async_view is not None
-                and iscoroutinefunction(callback)
-            )
-            if is_async_view:
+            if wrap_async_view is not None and iscoroutinefunction(callback):
                 sentry_wrapped_callback = wrap_async_view(callback)
             else:
                 sentry_wrapped_callback = _wrap_sync_view(callback)
--- a/sentry_sdk/integrations/django/views.py
+++ b/sentry_sdk/integrations/django/views.py
@@ -1,4 +1,5 @@
 import functools
+import sys
 from typing import TYPE_CHECKING
 
 import sentry_sdk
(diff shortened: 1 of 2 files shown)


Digenis:
I misunderstood the contribution guidelines
and thought the workflows' success is needed before removing draft status.

It turns out that maintainers don't see the drafts' workflows queued for approval.

Removing draft status.

sentry_sdk/integrations/django/views.py
```diff
@@ -10,15 +11,15 @@
     from typing import Any
 
 
-try:
+if sys.version_info >= (3, 14):
+    from inspect import iscoroutinefunction
+else:
```
alexander-alderman-webb, line 16: Any reason to not re-use the shim we already have in Django?

https://github.com/getsentry/sentry-python/blob/10b4f5b1e1ea1a5a16ba15b9e3b239c2ecac6a11/sentry_sdk/integrations/django/asgi.py#L39-L52

alexander-alderman-webb (changes requested):

sentry_sdk/integrations/django/views.py
Digenis: asgi.py imports django.core.handlers.asgi which doesn't exist in django < 3

sentry supports django ≥ 1.8

views.py wraps the asgi import in a try/except block
suppressing the inner import error
caused by the missing module in django  < 3
to keep django ≥ 1.8 compatibility

To rely on the shim
I have to redefine the other half of the conditional iscoroutinefunction definition
inside the try/except block
which defeats the purpose of using the shim.

The "conditional import" phrase in the asgi module docstring
[was, and still is, misleading](https://github.com/getsentry/sentry-python/pull/6869/changes/80b5328877b1570982aa5f29f16a99d894d545b9#diff-3c90144d4b456e8faf99c21a00cb2cb88f2fe8cdee56cf7f9a9b9a943e2f447a) people
into imagining an import within an if block.

Maybe I should rephrase it, since I'm touching this sentence in the chore commit anyway.
Examples:

> This module imports `django.core.handlers.asgi` (Django >= 3), so
> `sentry_sdk.integrations.django` imports it via a guarded import that
> handles `ImportError` on older Django versions.

> This module is imported from `sentry_sdk.integrations.django` using
> `try`/`except` to maintain compatibility with Django versions that do
> not provide `django.core.handlers.asgi`.

Separate commit or fixup on [chore(django): Drop Python < 3.5 compatibility](https://github.com/getsentry/sentry-python/pull/6869/commits/80b5328877b1570982aa5f29f16a99d894d545b9) ?

But this alone will not be enough to address the same misunderstanding in the future
so maybe a code comment too in views.py is needed.
It's obvious what the code does but not why it's here and not there.

> Duplicated from `asgi.py` because that module is not importable on Django < 3.

alexander-alderman-webb: Thanks for the thorough reply. We're not actually concerned about the 

```python
from django.core.handlers.asgi import ASGIRequest
```

import because it's in a type checking block. Our typing environment has newer package versions so there's no risk of `ImportError` for `ASGIRequest`.

Also, the `except:` block exists for Python 2 compatibility (see https://github.com/getsentry/sentry-python/pull/851). We're not concerned about Python 2 any more, so we can remove the `except :` 😄

alexander-alderman-webb:
Made you co-author on this commit: https://github.com/getsentry/sentry-python/commit/3e96fc3640350af89089067857c7dd1fa60f69b3.
Closing this PR now.

PR #6876, opened 2026-07-23
feat(integrations): Gate user info behind data_collection config

### LINKED ISSUE #6746: Support redacting/removing user data based on the data collection configuration

ericapisani:
This is a clean up pass from areas that I missed initially.

Refs getsentry/sentry-python#6743  <br>Fixes getsentry/sentry-python#6746


sentry_sdk/integrations/aws_lambda.py
```diff
@@ -437,7 +437,23 @@ def event_processor(
...
+                    identity = {}
+
+                id = identity.get("userArn")
+                if id is not None:
+                    user_info.setdefault("id", id)
+
+                ip = identity.get("sourceIp")
+                if ip is not None:
+                    user_info.setdefault("ip_address", ip)
+        elif should_send_default_pii():
```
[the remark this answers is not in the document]
ericapisani: This is intentional right now, the data collection configuration gating this isn't available yet and will be done when that's introduced

tests/integrations/aws_lambda/test_aws_lambda.py
```diff
@@ -427,11 +427,9 @@ def test_request_data_with_send_default_pii_true(lambda_client, test_environment
...
           "pathParameters": null,
           "stageVariables": null,
-          "requestContext": {
-            "identity": {
-              "sourceIp": "213.47.147.207",
-              "userArn": "42"
-            }
+          "identity": {
+            "sourceIp": "213.47.147.207",
+            "userArn": "42"
```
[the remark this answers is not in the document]
ericapisani: Fixed in [2fbf67a](https://github.com/getsentry/sentry-python/pull/6876/commits/2fbf67a0799f85c2d4062b487a74ae0510f234e0)

sentry_sdk/integrations/aws_lambda.py
```diff
@@ -437,10 +437,26 @@ def event_processor(
...
             user_info = sentry_event.setdefault("user", {})
 
-            identity = aws_event.get("identity")
+            identity = aws_event.get("requestContext", {}).get("identity")
```
ericapisani, line 459: This was a bug that was found in adding test coverage for this. Docs for the expected payload can be found [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html) but the tl;dr image from the page:

Image

sentrivana: Nice catch! I'm wondering it's always been like this or if the format of the aws event was changed at some point. Because if it's the latter, then we should support both possible locations.

sentrivana (approved):

sentry_sdk/integrations/aws_lambda.py
ericapisani: I think it has always been like this 😅 Below is a screenshot from May 2018, which predates the creation of this file by a few months:


May_2018_lambda_proxy

PR #6972, opened 2026-07-31
test: reduce common test suite while not reducing coverage

ericapisani:
May try this on individual integrations later, but used the autoresearch skill to collapse tests where possible, remove dead ones, etc.

Went from 2720 -> 2289

Refs PY-2641


tests/tracing/test_misc.py
```diff
@@ -350,19 +350,13 @@ def test_set_meaurement_compared_to_set_data(sentry_init, capture_events):
...
         ([r"^/api"], "/backend/api/envelopes", False),
         ([r"myApi.com/v[2-4]"], "myApi.com/v2/projects", True),
         ([r"myApi.com/v[2-4]"], "myApi.com/v1/projects", False),
-        ([r"https:\/\/.*"], "https://example.com", True),
-        (
-            [r"https://.*"],
-            "https://example.com",
-            True,
-        ),  # to show escaping is not needed
+        ([r"https://.*"], "https://example.com", True),
```
ericapisani, line 359: The 2nd test case removed here served more as documentation rather than exercising a specific code path.

Could add it back in for that reason, but the LLM removed it because it considered it a duplicate test.

sl0thentr0py (approved):

PR #6980, opened 2026-07-31
test(aws_lambda): reduce suite from 39 to 14 tests without losing coverage

ericapisani:
Follow-up to #6972, applying the same reduce-tests-without-losing-coverage experiment to the AWS Lambda integration suite — the most expensive suite in the repo to run locally, since every test round-trips through dockerized AWS SAM containers.

The suite goes from **39 to 14 test cases (-64%)** with no loss of assertions or behavior coverage. Reductions come from three sources:

Result: ~210s → ~115s per run on local machine, and 18 consecutive green runs (previously 5–9 flaky failures per run). Verified on py3.13 and py3.11 (full docker runs), py3.8-compatible syntax, ruff check + format clean.

### Additional notes

`test_timeout_error` was dropped because its assertions are a subset of `test_timeout_error_scope_modified`, which is staying. The timeout-warning behaviour is still fully covered through the latter test.

However, the two used different init paths - the removed one ran through the Lambda-layer auto-init path (`init_serverless_sdk`), while the one we kept uses an explicit `sentry_sdk.init` call. 

As a result, we lose timeout-specific coverage on the layer path. However, that path is still exercised by the other tests (BasicException, InitError, BasicOk), and the warning logic is identical.

I think it's worth the trade-off here, but if there are strong opinions about adding `test_timeout_error` back in to get that timeout-specific coverage, let me know.

### DIFF
--- a/tests/integrations/aws_lambda/lambda_functions/TimeoutError/index.py
+++ b/tests/integrations/aws_lambda/lambda_functions/TimeoutError/index.py
@@ -1,8 +1,0 @@
-import time
-
-
-def handler(event, context):
-    time.sleep(15)
-    return {
-        "event": event,
-    }
--- a/tests/integrations/aws_lambda/test_aws_lambda.py
+++ b/tests/integrations/aws_lambda/test_aws_lambda.py
@@ -330,10 +330,6 @@ def test_timeout_error_scope_modified(lambda_client, test_environment):
 @pytest.mark.parametrize(
     "aws_event, has_request_data, batch_size",
     [
-        # Scalar events (int/float/string/bool) are one equivalence class:
-        # not a list, not a dict, so request_data falls back to {}.
-        # (An empty list hits the identical path: len < 1 -> else arm ->
-        # non-dict reset.)
         (b"1231", False, 1),
         (
             b"""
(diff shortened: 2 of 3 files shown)


sentrivana (commented):
TBH I don't know how to approach reviewing this. Actually going through all existing test cases, understanding what they're testing, and then comparing with the end result would take a long time.

If you already did that work, I'm willing to stamp this because I trust your judgment.

But in general, I'm wary of these test compacting changes. I don't know if a slight speedup is worth it compared to making the tests less structured and harder to understand. Maybe it is! But the problem is, while the speedup is easily measurable, the quality of the test cases themselves isn't (coverage is just part of the story). So I can just go by what I know about our human vs AI written/assisted test cases, which is that I'd pick human-written test cases over AI anytime.

ericapisani:
I did do a check of the changes but, on further reflection, given @pabloDeputter is actively working in this area of code, it's probably better to not try and compact the tests at this time. I'll leave any potential refactoring/test restructuring opportunities as part the work that he's doing.

That being said - there's a couple of changes that are in this branch that will make a meaningful difference that I'll open a separate PR for:

#### Addition of "architecture" to `utils.py`

This boosts running these tests in a local environment (1m 35.7s -> 94.3s)

#### Removing `test_timeout_error`

As noted above in the pull request description, this is pretty much a subset of another test, and removing this drops the time down to 45.5s in my local environment.

ericapisani:
Opened https://github.com/getsentry/sentry-python/pull/7053 to replace this one

PR #6981, opened 2026-07-31
feat(openai): Make include_prompts override data collection when set

ericapisani:
Change the default of OpenAIIntegration's include_prompts from True to None (unset). When unset, prompt collection follows the global data_collection.gen_ai.inputs setting; when explicitly set, it takes precedence over the global setting in both directions.

The unset default is resolved lazily at call time rather than in __init__, so integrations constructed before sentry_sdk.init() still see the active client's options. Pre-data-collection behavior (defaults to True, subject to send_default_pii) is preserved.

Refs PY-2588

### DIFF
--- a/sentry_sdk/integrations/openai.py
+++ b/sentry_sdk/integrations/openai.py
@@ -610,31 +610,36 @@ def _set_embeddings_input_data(
     kwargs: "dict[str, Any]",
     integration: "OpenAIIntegration",
 ) -> None:
-    messages: "Union[str, SequenceNotStr[str], Iterable[int], Iterable[Iterable[int]]]" = kwargs.get(
-        "input"
-    )
+
+    set_data_normalized(span, SPANDATA.GEN_AI_OPERATION_NAME, "embeddings")
 
     set_on_span = (
         span.set_attribute if isinstance(span, StreamedSpan) else span.set_data
     )
+
     model = kwargs.get("model")
     if model is not None:
         set_on_span(SPANDATA.GEN_AI_REQUEST_MODEL, model)
 
-    if (
-        not should_send_default_pii()
-        or not integration.include_prompts
-        or messages is None
-    ):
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb (changes requested):
Can we just deprecate `include_prompts` instead (once data collection has landed)? 😄 

The way it was implemented, the `include_prompts` parameter lets users opt-out from collecting GenAI attributes, even if `include_default_pii=True`. You can opt-out of GenAI attributes while collecting other data with the data collection options though!

ericapisani:
> Can we just deprecate include_prompts instead (once data collection has landed)? 😄
>
> The way it was implemented, the include_prompts parameter lets users opt-out from collecting GenAI attributes, even if include_default_pii=True. You can already opt-out of GenAI attributes while collecting other data with data collection though!

We could but I don't think we should in this case. For users who want all the inputs/outputs or who want to opt-out altogether, data collection settings are fine and they won't need to worry about setting `include_prompts`. 

For those who want to exercise some granularity when they have multiple AI integrations (and there are a large number of projects that have 2+) they still need an override mechanism of some kind.

For what it's worth, I don't think that `include_prompts` is the best long-term override setting because it doesn't allow folks to opt-out of outputs, but for simplicity's sake on the initial implementation of the data collection spec, I left it as-is.

ericapisani:
As a follow up to my earlier comment - the JS folks have overrides for inputs/outputs on an integration level as well alongside data collection, so we wouldn't be "re-inventing the wheel" here

alexander-alderman-webb:
There's always `before_send_span()` in the case that a user wants data from one integration and not the other, so there is a migration path for deprecating the option.
Currently `include_prompts=True` doesn't override `send_default_pii`, so I'm wondering why the option would override the data collection options that supersede `send_default_pii`. 🤔

ericapisani:
closing in favour of https://github.com/getsentry/sentry-python/pull/7034

PR #6984, opened 2026-08-02
feat(integrations): Add span for DRF authentication

### LINKED ISSUE #3006: Add a span around Django rest framework authentication
### Problem Statement

In Sentry Performance, when I look at the queries list in a performance event of an API view, the `view.render` event always begins with a few of the same queries, for looking up the auth token, and are rarely relevant. It would be better if those auth queries were in a separate span, to reduce visual noise.

### Solution Brainstorm

I currently subclass the authentication class and add a span myself, like this:

```python
class TracedTokenAuthentication(rest_framework.authentication.TokenAuthentication):
    def authenticate(self, request):
        with sentry_sdk.start_span(op='authenticate'):
            return super().authenticate(request)
```

The disadvantage of this is you can use multiple authentications, and each would need modifying. A better place for the span is probably `rest_framework.request.Request._authenticate`, which loops over all configured authentications.

mgaligniana:
### Description
In this PR I applied the suggested changes adding a new span in `rest_framework.request.Request._authenticate` which loops over all configured authentications and making auth queries being in a separate span to reduce visual noise.

- [x] **AI tools were used**, I have disclosed which ones, and fully reviewed and verified their output.

#### Issues
* resolves: #3006

### DIFF
--- a/sentry_sdk/consts.py
+++ b/sentry_sdk/consts.py
@@ -1257,6 +1257,7 @@ class OP:
     SUBPROCESS_WAIT = "subprocess.wait"
     SUBPROCESS_COMMUNICATE = "subprocess.communicate"
     TEMPLATE_RENDER = "template.render"
+    VIEW_AUTHENTICATE = "view.authenticate"
     VIEW_RENDER = "view.render"
     VIEW_RESPONSE_RENDER = "view.response.render"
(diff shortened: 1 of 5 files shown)


alexander-alderman-webb (approved):

mgaligniana:
Thank you!!

PR #7002, opened 2026-08-03
ref: Drop Flask < 2

sentrivana:


sentrivana:
I think I removed some stuff from the toxgen cache. I'll regenerate it but the PR should not change otherwise so can already be reviewed.

alexander-alderman-webb (approved):

PR #7010, opened 2026-08-03
fix(pydantic-ai): Capture reasoning and OTEL output messages

vgrozdanic:
Pydantic AI spans now record assistant output only as `gen_ai.output.messages` (text, reasoning, and tool calls on chat spans; final agent text on invoke_agent). `ThinkingPart` is surfaced as reasoning in request history and model responses, and provider reasoning token usage is attached when the count is greater than zero.

Deprecated `gen_ai.response.text` and `gen_ai.response.tool_calls` are not set. Request message parts keep the existing `text` key so truncation and consumers of `gen_ai.request.messages` stay compatible.

### DIFF
--- a/tests/integrations/pydantic_ai/test_pydantic_ai.py
+++ b/tests/integrations/pydantic_ai/test_pydantic_ai.py
@@ -129,11 +129,11 @@ async def test_agent_run_async(
                 "content": [
                     {
                         "type": "text",
-                        "content": "Message demonstrating the absence of truncation.",
+                        "text": "Message demonstrating the absence of truncation.",
                     },
                     {
                         "type": "text",
-                        "content": "Test input",
+                        "text": "Test input",
                     },
                 ],
             }
@@ -179,11 +179,11 @@ async def test_agent_run_async(
                 "content": [
                     {
                         "type": "text",
-                        "content": "Message demonstrating the absence of truncation.",
+                        "text": "Message demonstrating the absence of truncation.",
                     },
                     {
                         "type": "text",
-                        "content": "Test input",
+                        "text": "Test input",
                     },
                 ],
             }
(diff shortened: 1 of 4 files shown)


tests/integrations/pydantic_ai/test_pydantic_ai.py
```diff
@@ -4287,6 +4424,64 @@ async def test_set_usage_data_with_cache_tokens(
...
+    events = capture_events()
+
+    with sentry_sdk.start_transaction(op="test", name="test"):
+        span = sentry_sdk.start_span(op="test_span")
+        usage = RequestUsage(
+            input_tokens=100,
+            output_tokens=50,
+            details={details_key: 12},
+        )
+        _set_usage_data(span, usage)
```
alexander-alderman-webb, line 4454: I cannot find a single new test that **uses the `pydantic-ai` package as a user would**. All the tests are full of unnecessary mocks.
I want to know that your changes work with `pydantic-ai`, not some mocks you made up.
In it's current state, it doesn't make sense to review the PR further.

Imports in the middle of function w/o justification, most things `Any`-typed, old style type annotation which the repo doesn't use anymore should all be removed as well.

alexander-alderman-webb (changes requested):

tests/integrations/pydantic_ai/test_pydantic_ai.py
vgrozdanic: This is following existing test patterns, unless you are proposing refactor of all the tests for this package, which is IMO out of the scope for this small bug fix.

If you prefer, I can close this PR, and open an issue in `sentry-python` to support reasoning and proper tool calls format

alexander-alderman-webb:
Closing in favor of https://github.com/getsentry/sentry-python/pull/7037

PR #7043, opened 2026-08-05
ref(tornado): Remove unused compat code

sentrivana:
We're now enforcing Tornado > 6, so compat code for older versions is not needed anymore

### DIFF
--- a/sentry_sdk/integrations/tornado.py
+++ b/sentry_sdk/integrations/tornado.py
@@ -1,6 +1,5 @@
 import contextlib
 import weakref
-from inspect import iscoroutinefunction
 
 import sentry_sdk
 from sentry_sdk.api import continue_trace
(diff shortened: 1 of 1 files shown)


alexander-alderman-webb (approved):

Chenghao999 (commented):
Verified this removal is safe and consistent:

- The `tornado` extra in `setup.py` is pinned to `tornado>=6`, and the runtime floor check already enforces it via `_MIN_VERSIONS["tornado"] = (6, 0)` in `sentry_sdk/integrations/__init__.py`. So `RequestHandler._execute` is guaranteed to be a native coroutine whenever this integration can run: the `iscoroutinefunction` branch was always True and the generator path was dead code.
- `tests/integrations/tornado/test_tornado.py` contains no references to the removed path, which is consistent with it being unreachable, and patch coverage is 100%.
- The import removals are complete — nothing else in the file references `iscoroutinefunction` or `tornado.gen`.

Optional nit: the "Starting Tornado 6 ..." comment disappeared together with the branch. A one-liner on the now-unconditional `async def` wrapper explaining that Tornado >= 6 (enforced both in the extras and at runtime) makes the compat path unnecessary would help future readers.

LGTM — nice small cleanup.

PR #7050, opened 2026-08-05
fix(boto3): Fix botocore SigV4 failures caused by post-sign trace propagation

### LINKED ISSUE #7031: S3 HeadObject returns 403 when sentry_sdk.start_span and Datadog botocore instrumentation are both active
### Environment

- Python 3.14
- Linux aarch64
- `sentry-sdk==2.66.0`
- `ddtrace==4.12.1` (Datadog admission-controller library injection)
- FastAPI
- boto3/botocore with AWS SigV4

### Description

An S3 `HeadObject` request against an existing, accessible object returns `403 Forbidden` only when it executes inside `sentry_sdk.start_span(...)` while Datadog's botocore integration is active. The same pod identity, bucket, key, client configuration, and credentials return `200 OK` when either the Sentry span is removed or Datadog's botocore integration is disabled.

This also reproduces through our real FastAPI endpoint. The example below is read-only and reproduces on its first request.

### Minimal reproducer

```python
import boto3
import sentry_sdk
from botocore.config import Config
from fastapi import FastAPI
from fastapi.testclient import TestClient

sentry_sdk.init(
    dsn=None,
    traces_sample_rate=1.0,
    integrations=[FastApiIntegration()],
)

s3 = boto3.client(
    "s3",
    config=Config(
        signature_version="s3v4",
        s3={
            "addressing_style": "virtual",
            "us_east_1_regional_endpoint": "regional",
        },
    ),
)

app = FastAPI()

@app.post("/reproduce")
def reproduce():
    with sentry_sdk.start_span(op="dependency.resolve", name="FileStorage"):
        response = s3.head_object(
            Bucket="YOUR_BUCKET",
            Key="YOUR_EXISTING_KEY",
        )
    return {"status": response["ResponseMetadata"]["HTTPStatusCode"]}

response = TestClient(app).post("/reproduce")
print(response.status_code, response.text)
```

Run under Datadog auto-instrumentation, for example:

```shell
DD_TRACE_BOTOCORE_ENABLED=true ddtrace-run python reproducer.py
```

### Actual result

```text
botocore.exceptions.ClientError: An error occurred (403) when calling the HeadObject operation: Forbidden
```

### Control matrix

| Datadog botocore integration | Nested `sentry_sdk.start_span` | Result |
|---|---|---|
| enabled | present | `403 Forbidden` |
| enabled | absent | `200 OK` |
| disabled | present | `200 OK` |

The confirmed workaround is:

```text
DD_TRACE_BOTOCORE_ENABLED=false
```

A corresponding Datadog report exists at https://github.com/DataDog/dd-trace-py/issues/19477. This appears to be an interoperability issue; we are reporting it to both projects because the trigger requires both instrumentations.

pabloDeputter:
#### Summary of issue

* pre-4.12.0 ddtrace basically behaved like Sentry before: propagation headers were added by the HTTP client after botocore already had calculated the SigV4 signature. So `baggage` was not included in `SignedHeaders`. Any later modifications to the value did not invalidate the request.
* 4.12.0 moves the header injection to botocore's `before-sign` event. It adds `baggage`, ... and `x-datadog-*` before signing. Any later modifications to the value DO invalidate the request, thus later HTTP-client injection is suppressed to avoid duplicate headers.
* How failure happens:
  1. incoming FastAPI request contains Sentry or other W3C baggage
  2. datadog extracts that baggage into its active context
  3. boto3 request is created
  4. datadog’s `before-sign` handler writes the baggage to the AWS request
  5. botocore includes `baggage` in the SigV4 signature
  6. sentry’s httplib runs afterwards and updates or adds another `baggage` value
  7. AWS receives a different baggage value from the one that was signed. Because request was "tampered" with after signing, AWS rejects the request with `403 Forbidden` or `SignatureDoesNotMatch`.

#### Changes

* inject sentry propagation through `before-sign` handler, so final `baggage` and `sentry-trace` values are created before SigV4 signing.
* existing third-party baggage is merged into one header.
* `http.client` propagation is delayed until `endheaders()`, when the complete request headers and SigV4 `SignedHeaders` are available. Existing `baggage` header is never mutated after it already was signed.
* propagation is skipped for [presigned requests](<https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html>), since adding headers could require a future caller to reproduce these exact headers.

#### Issues

Resolves: getsentry/sentry-python#7031 & getsentry/sentry-python#7031  <br>Related issues in [dd-trace-py](<https://github.com/DataDog/dd-trace-py>): [#19477](<https://github.com/DataDog/dd-trace-py/issues/19477>) & [#19358](<https://github.com/DataDog/dd-trace-py/issues/19358>)

#### Reminders

* Please add tests to validate your changes, and lint your code using `uv run ruff`.
* Add GH Issue ID *&* Linear ID (if applicable)
* PR title should use [conventional commit](<https://develop.sentry.dev/engineering-practices/commit-messages/#type>) style (`feat:`, `fix:`, `ref:`, `meta:`)
* For external contributors: [CONTRIBUTING.md](<https://github.com/getsentry/sentry-python/blob/master/CONTRIBUTING.md>), [Sentry SDK development docs](<https://develop.sentry.dev/sdk/>), [Discord community](<https://discord.gg/Ww9hbqr>)


sentry_sdk/integrations/stdlib.py
```diff
@@ -157,26 +193,57 @@ def putrequest(
...
+        else:
+            self._sentrysdk_trace_headers = ()  # type: ignore[attr-defined]
+            self._sentrysdk_trace_url = None  # type: ignore[attr-defined]
 
         self._sentrysdk_span = span  # type: ignore[attr-defined]
 
         return rv
 
+    def endheaders(self: "HTTPConnection", *args: "Any", **kwargs: "Any") -> "Any":
+        trace_headers = getattr(self, "_sentrysdk_trace_headers", ())
```
alexander-alderman-webb: Does calling `sentry_sdk.get_current_scope().iter_trace_propagation_headers()` here instead of in `putrequest()` work?
It would be best to avoid stashing stuff on the `HTTPConnection` instance if we can help it.

pabloDeputter: Good suggestion :) header generation is moved into `endheaders()` instead of being stored on the connection.

sentry_sdk/integrations/stdlib.py
```diff
@@ -157,26 +193,57 @@ def putrequest(
...
+                signed_headers = _aws_sigv4_signed_headers(request_buffer)
+
+                for header_name, header_value in trace_headers:
+                    normalized_header = header_name.lower()
+                    # do not mutate headers already signed by SigV4.
+                    if normalized_header in existing_headers and (
+                        normalized_header != BAGGAGE_HEADER_NAME
+                        or normalized_header in signed_headers
+                    ):
+                        continue
```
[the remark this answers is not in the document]
pabloDeputter: It checks 1. any existing header (e.g. `sentry-trace`) is provided by the caller and should be preserved; 2. existing `baggage` should only be appended when it is not yet signed, otherwise preserve. Boto3 is responsible for replacing stale values before signing (sentry_sdk/integrations/boto3.py#L152-L156).

sentry_sdk/integrations/stdlib.py
```diff
@@ -61,6 +61,41 @@ def add_python_runtime_context(
...
+    if buffer is None:
+        return set()
+    for line in buffer:
+        name, separator, value = line.partition(b":")
+        if not separator or name.lower() != b"authorization":
+            continue
+
+        value = value.lstrip()
+        if not value.startswith((b"AWS4-HMAC-SHA256", b"AWS4-ECDSA-P256-SHA256")):
+            return set()
```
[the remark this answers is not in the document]
pabloDeputter: There shouldn't be multiple `Authorization` headers, nevertheless replaced `return set()` with `continue`.

sentry_sdk/integrations/stdlib.py
```diff
@@ -233,6 +299,7 @@ def close(self: "HTTPResponse") -> None:
                 _complete_span(span)
 
     HTTPConnection.putrequest = putrequest  # type: ignore[method-assign]
+    HTTPConnection.endheaders = endheaders  # type: ignore[method-assign]
```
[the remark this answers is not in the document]
pabloDeputter: No problem, only braces in the template are parsed. So for example
```python
>>> "{value}".format(value="http://getsentry.net/api/{id}")
'http://getsentry.net/api/{id}'
```
works just fine.

alexander-alderman-webb:
I haven't forgot about this, it's just complex so I'll likely only re-review fully at the start of next week.

tests/integrations/boto3/test_trace_propagation.py
```diff
@@ -0,0 +1,207 @@
...
+
+        def capture_headers_after_instrumentation(request, **kwargs):
+            for header_name in ("baggage", "sentry-trace"):
+                signed_request_headers[header_name] = request.headers.get_all(
+                    header_name
+                )
+
+        # register `before-sign` handler that adds third-party baggage.
+        client.meta.events.register("before-sign", _inject_third_party_baggage)
+        client.meta.events.register_last(
```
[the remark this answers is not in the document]
pabloDeputter: `boto3.client()` reuses the global session, event emitters are copied. See below for example:

```python
import boto3
c1 = boto3.client("dynamodb")
c2 = boto3.client("dynamodb")
calls = []

def handler(**kwargs):
    calls.append(kwargs)
    
c1.meta.events.register("before-sign", handler)
# handler registered on c1 is not called for c2.
c2.meta.events.emit("before-sign", request=None, signature_version="v4")

assert calls == []
assert c1.meta.events._emitter is not c2.meta.events._emitter
```

sentry_sdk/integrations/stdlib.py
```diff
@@ -157,26 +170,58 @@ def putrequest(
...
+                        normalized_header = header_name.lower()
+                        # preserve signed headers and avoid duplicate `sentry-trace`.
+                        if normalized_header in existing_headers and (
+                            normalized_header != BAGGAGE_HEADER_NAME
+                            or normalized_header in signed_headers
+                        ):
+                            continue
+
+                        logger.debug(
+                            "[Tracing] Adding `{key}` header {value} to outgoing request to {real_url}.".format(
```
alexander-alderman-webb: `real_url` is constructed from the `host` and `port`, which are available in `endheaders`.

Attaching `_sentrysdk_trace_url` to the `HTTPConnection` instance when it's available in the local scope is unnecessary complexity. You can re-use the logic for creating the `real_url`, or just change the log message 😄 .

```suggestion
                            "[Tracing] Adding `{key}` header {value} to outgoing request to port {self.port} of host {self.host}.".format(
```

sentry_sdk/integrations/stdlib.py
```diff
@@ -157,26 +170,58 @@ def putrequest(
...
         return rv
 
+    def endheaders(self: "HTTPConnection", *args: "Any", **kwargs: "Any") -> "Any":
+        real_url = getattr(self, "_sentrysdk_trace_url", None)
+        span = getattr(self, "_sentrysdk_span", None)
+
+        try:
+            if real_url is not None:
+                with capture_internal_exceptions():
+                    request_buffer = getattr(self, "_buffer", None)
```
alexander-alderman-webb: I'm uneasy about parsing raw bytes from a the private `_buffer` variable on every request.
Is it possible to patch `HTTPConnection.putheader()` instead to build a list of signed headers?

alexander-alderman-webb (commented):
This looks really good broadly as fixes the issue without disrupting existing users (including those that not using `boto3`).

My comments are mainly concerned with the patching approach. It looks more complex than needed to achieve the result.

sentry_sdk/integrations/boto3.py
```diff
@@ -114,6 +120,53 @@ def _sentry_request_created(
...
+        ):
+            return
+
+        if request.url is None or not should_propagate_trace(client, request.url):
+            return
+
+        def _replace_header(request: "AWSRequest", key: str, value: str) -> None:
+            if key in request.headers:
+                del request.headers[key]
+            request.headers[key] = value
```
alexander-alderman-webb, line 181: This looks weird. Why do we delete the entry before assigning the value?
Can we not just assign the value directly?

pabloDeputter: `request.headers` is of type `HTTPHeaders`. If you do `request.headers[key] = value` it will append another header instead of overwriting the existing one. So baggage won't be merged before signing. Existing tests also fail if you remove the deletion part.

```python
a = [
    "old baggage",
    "third party baggage",
    "combined baggage",
]
# instead of
a = ["combined baggage"]
```

alexander-alderman-webb: lol okay that's surprising behavior. All good then.

sentry_sdk/integrations/stdlib.py
pabloDeputter: I think this way is better, since the full URL is needed. The query string is only available in either `putrequest()` or `_buffer` which we don't want to rely on.  e.g. `X-Amz-SignedHeaders` value is needed in `get_aws_sigv4_signed_headers()` to avoid modifying headers in presigned requests. Basically you can't reconstruct the request path and query string which are both needed.

sentry_sdk/utils.py
```diff
@@ -1697,6 +1697,43 @@ def parse_url(url: str, sanitize: bool = True) -> "ParsedUrl":
     )
 
 
+def get_aws_sigv4_signed_headers(
```
alexander-alderman-webb: This should be marked private. A lot of old stuff isn't but we do this more recently. 

```suggestion
def _get_aws_sigv4_signed_headers(
```

tests/test_utils.py
```diff
@@ -665,6 +666,33 @@ def test_default_release_empty_string():
...
+            "X-Amz-SignedHeaders=host%3Bx-amz-date&"
+            "X-Amz-Signature=sixtyseven",
+            {"host", "x-amz-date"},
+        ),
+    ],
+)
+def test_get_aws_sigv4_signed_headers(authorization, url, expected):
+    assert (
+        get_aws_sigv4_signed_headers(authorization=authorization, url=url) == expected
+    )
```
alexander-alderman-webb: After `get_aws_sigv4_signed_headers()` is marked private you can remove this test.
There's no point testing private helpers, just test the intended behavior with the real library.

tests/integrations/stdlib/test_httplib.py
```diff
@@ -76,6 +77,43 @@ def create_chunked_server():
...
+    connection.putrequest("POST", path)
+
+    for key, value in headers:
+        connection.putheader(key, value)
+
+    connection.endheaders()
+
+    response = connection.getresponse()
+    response.read()
+    connection.close()
```
alexander-alderman-webb: You can revert new tests from `test_httplib.py` and add tests that use `AWSHTTPConnection` instead.

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +80,154 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
...
+    """
+    Responsible for initializing `_sentrysdk_signed_headers` on the instance.
+    """
+
+    def putrequest(
+        self: "HTTPConnection", method: str, url: str, *args: "Any", **kwargs: "Any"
+    ) -> "Any":
+        # track existing headers separately from signed so thatexisting unsigned
+        # baggage header can be extended while signed baggage stays unchanged.
+        self._sentrysdk_request_headers = set()  # type: ignore[attr-defined]
```
alexander-alderman-webb: Can we do without tracking all request headers (i.e., only tracking the signed headers?).

alexander-alderman-webb (commented):
Next round 😃.
I've also pushed directly to your branch to limit the behavior to the affected `HTTPConnection` subclasses. I remove the changes to `aiohttp` as well, they can be in a separate PR when we get around to it (this one is big enough).

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +80,154 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
...
+        AWSHTTPSConnection.endheaders = _get_wrapped_endheaders(  # type: ignore[method-assign]
+            AWSHTTPSConnection.endheaders
+        )
+        AWSHTTPSConnection.putrequest = _get_wrapped_putrequest(  # type: ignore[method-assign]
+            AWSHTTPSConnection.putrequest
+        )
+
+
 def _install_httplib() -> None:
+    _patch_aws_connection()
```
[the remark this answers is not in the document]
alexander-alderman-webb: If someone signs headers that's on them.
We're special casing AWS because that's popular.

sentry_sdk/integrations/stdlib.py
```diff
@@ -202,6 +360,8 @@ def putrequest(
                     )
                 )
                 self.putheader(key, value)
+        elif should_propagate_trace(client, real_url):
+            self._sentrysdk_trace_url = real_url  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
alexander-alderman-webb: my bad: https://github.com/getsentry/sentry-python/pull/7050/commits/628543e81ac6a4026feb19a1c7641d59ed12f94a

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +80,157 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
...
+        )
+        AWSHTTPSConnection.endheaders = _get_wrapped_endheaders(  # type: ignore[method-assign]
+            AWSHTTPSConnection.endheaders
+        )
+        AWSHTTPSConnection.putrequest = _get_wrapped_putrequest(  # type: ignore[method-assign]
+            AWSHTTPSConnection.putrequest
+        )
+
+
 def _install_httplib() -> None:
```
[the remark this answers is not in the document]
alexander-alderman-webb: no they call the superclass

sentry_sdk/integrations/stdlib.py
pabloDeputter: We do need it to keep avoid duplicating headers, preserving already signed `baggage` and knowing when existing`baggage`is not yet signed so we can add our own.

However we only track `sentry-trace` and `baggage`. So I removed `_sentrysdk_signed_headers`, and `_sentrysdk_request_headers` now tracks whether a header is present and signed like so: `{"sentry-trace": (is_present, is_signed)}`

alexander-alderman-webb: tracking only what we need is good

sentry_sdk/integrations/boto3.py
```diff
@@ -143,6 +150,54 @@ def _sentry_request_created(
...
+        if isinstance(signature_version, str) and signature_version.endswith(
+            ("-query", "-presign-post")
+        ):
+            return
+
+        if request.url is None or not should_propagate_trace(client, request.url):
+            return
+
+        def _replace_header(request: "AWSRequest", key: str, value: str) -> None:
+            # HTTPHeaders appends on assignment, so delete existing values first.
```
ericapisani: Because there's subtly different ways that headers can be handled when multiple values are set on the same header key, I think we should add a bit more context to this comment.

It'd be worth clarifying what "append" (ideally with a concrete example) means as it could be interpreted as:
- adding to a list (e.g: [foo, bar, baz])
- appending to a comma separated string (e.g. foo, bar, baz)

And I think it may also be worth pointing out that this is coming from [`email.message.Message`](https://docs.python.org/3/library/email.compat32-message.html#compat32-message) from the standard library as this is what the botocore package is using under the hood.

sentry_sdk/integrations/stdlib.py
```diff
@@ -40,6 +41,14 @@
     "build": sys.version,
 }
 
+_SENTRY_HEADER_NAMES = frozenset((BAGGAGE_HEADER_NAME, "sentry-trace"))
```
ericapisani: We also have a constant for `sentry-trace` that you can use here


```suggestion
_SENTRY_HEADER_NAMES = frozenset((BAGGAGE_HEADER_NAME, SENTRY_TRACE_HEADER_NAME))
```

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +82,171 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
             add_http_request_source(span)
 
 
+def _get_wrapped_putheader(
+    original_putheader: "Callable[..., Any]",
+) -> "Callable[..., Any]":
+    """
+    Responsible for tracking request and signed headers.
```
ericapisani: Similar to the other comment, I think we should point future devs reading this to the specific "signed headers" that are being referred to here (in this case botocore's SignedHeaders).

```suggestion
    Responsible for tracking request and `SignedHeaders`.
```

sentry_sdk/utils.py
```diff
@@ -1697,6 +1697,43 @@ def parse_url(url: str, sanitize: bool = True) -> "ParsedUrl":
     )
 
 
+def _get_aws_sigv4_signed_headers(
```
ericapisani: Assuming I'm understanding this correctly, we've got 2 distinct code paths here:
- searching for the signed headers in the Authorization header (lines 1705-1716)
- searching for the signed headers in the URL's query string (lines 1718-1734)

If this were a public API and we wanted to optimize for user convenience, I can see the value of supporting the passing in both `authorization` and `url` and then searching in the correct place based on if the value is present or not, but because this is strictly for internal use, I think it'd be cleaner to have two distinct helper functions for each path:
- `_get_aws_sigv4_signed_headers_from_authorization_header`
- `_get_aws_sigv4_signed_headers_from_url_query_string`

since at the calling sites, we already know where these headers are.

sentry_sdk/utils.py
```diff
@@ -1697,6 +1697,43 @@ def parse_url(url: str, sanitize: bool = True) -> "ParsedUrl":
...
+    authorization: "Optional[str]", url: "Optional[str]" = None
+) -> "Set[str]":
+
+    signed_headers: "Set[str]" = set()
+    if authorization is not None:
+        # only AWS SigV4 authorization has the SignedHeaders parameter.
+        value = authorization.lstrip()
+        if value.startswith(("AWS4-HMAC-SHA256", "AWS4-ECDSA-P256-SHA256")):
```
ericapisani: Although these aren't used very widely in the broader SDK codebase, what are your thoughts on doing something similar to the `_SENTRY_HEADER_NAMES` constant with the SigV4 algorithms?

```
_AWS_SIGV4_SIGNING_ALGORITHMS = frozenset(("AWS4-HMAC-SHA256", "AWS4-ECDSA-P256-SHA256"))
```

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +82,171 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
...
+
+        if normalized_header in _SENTRY_HEADER_NAMES:
+            _, is_signed = request_headers.get(normalized_header, (False, False))
+            request_headers[normalized_header] = (True, is_signed)
+
+        if normalized_header == "authorization" and values:
+            with capture_internal_exceptions():
+                authorization = values[0]
+                if isinstance(authorization, bytes):
+                    authorization = authorization.decode("ascii", "ignore")
```
ericapisani: To mirror the codecs used in the `putheader` method - it looks like [we want `latin-1`](https://github.com/python/cpython/blob/ee521e8ac19ad012ebc4e1b3e71b369988a9b9f8/Lib/http/client.py#L1369-L1373) decoding instead of `ascii` here.
```suggestion
                    authorization = authorization.decode("latin-1", "ignore")
```

ericapisani (commented):
Still working through the changes but have some initial comments

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +82,171 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
...
+                        ):
+                            normalized_header = header_name.lower()
+                            # preserve signed headers and avoid duplicate `sentry-trace`.
+                            is_present, is_signed = request_headers.get(
+                                normalized_header, (False, False)
+                            )
+                            if is_present and (
+                                normalized_header != BAGGAGE_HEADER_NAME or is_signed
+                            ):
+                                continue
```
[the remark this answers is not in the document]
pabloDeputter: I guess this is true, but normally shouldn't happen since `SignedHeaders` are derived from the present headers when signing. I'll switch it to 
```python
if is_signed or (
    is_present and normalized_header != BAGGAGE_HEADER_NAME
):
```
so that it is handled, but existing logic is preserved.

sentry_sdk/integrations/boto3.py
pabloDeputter: Updated the comment to clarify that it creates another header field rather than extending, @alexander-alderman-webb was also confused about this 😄

sentry_sdk/integrations/stdlib.py
pabloDeputter: Yup, good idea; didn't know this existed.

sentry_sdk/utils.py
pabloDeputter: agreed, I left it like that because originally we didn't read from the query string. I split it into 2 separate functions.

sentry_sdk/utils.py
pabloDeputter: I think that's okay. I'm doing it in `stdlib.py` with `_SENTRY_HEADER_NAMES` as well. Not sure whether we have a separate file for such constants, but I put it at the top of `utils.py`.

sentry_sdk/integrations/stdlib.py
pabloDeputter: Yup, changed it to "Responsible for tracking which sentry headers are present and whether they are listed in AWS SigV4 `SignedHeaders`."

sentry_sdk/integrations/stdlib.py
pabloDeputter: Yup good idea, didn't know this existed

sentry_sdk/integrations/stdlib.py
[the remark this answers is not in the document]
ericapisani: From looking at the botocore code, I think this is legit. Here's a sample script of what it's describing. If you reverse the order of the `Sub`/`Base` patches, you'll see that when `Sub` is called that it also calls the patch on `Base` as well (right now it doesn't).

Edit: the solution would be to move the `_patch_aws_connection` call to below the `HTTPConnection.*` patches

[demo_shadowing_bug.py](https://github.com/user-attachments/files/31420549/demo_shadowing_bug.py)

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +83,178 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
...
+def _get_wrapped_endheaders(
+    original_endheaders: "Callable[..., Any]",
+) -> "Callable[..., Any]":
+    """
+    Responsible for injecting trace propagation headers, ensuring that the request is not invalidated
+    by honoring signed headers.
+    """
+
+    def endheaders(self: "HTTPConnection", *args: "Any", **kwargs: "Any") -> "Any":
+        real_url = getattr(self, "_sentrysdk_trace_url", None)
```
ericapisani: Does `real_url` represent the original URL in the request? Or should this instead be called `trace_url`?

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +83,178 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
...
+        span = getattr(self, "_sentrysdk_span", None)
+
+        try:
+            if real_url is not None:
+                with capture_internal_exceptions():
+                    request_headers: "Optional[Dict[str, Tuple[bool, bool]]]" = getattr(
+                        self, "_sentrysdk_request_headers", {}
+                    )
+                    if request_headers is not None:
```
ericapisani: If `_sentrysdk_request_headers` is not set on `self` and the empty dict default is returned, the `if request_headers is not None` conditional will still pass. 

If this is the intent, it'd be clearer to have the default value for `getattr` be `None` as the way that this is currently written, this looks like a potential bug 😅

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +83,178 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
...
+                        for (
+                            signed_header
+                        ) in _get_aws_sigv4_signed_headers_from_url_query_string(
+                            real_url
+                        ):
+                            if signed_header in _SENTRY_HEADER_NAMES:
+                                is_present, _ = request_headers.get(
+                                    signed_header, (False, False)
+                                )
+                                request_headers[signed_header] = (is_present, True)
```
ericapisani: I'm finding this block of code (and the related one further up that deals with the authorization header) a bit confusing.

In what circumstances would the `signed_header` not be found in `request_headers` but is present on the query string, such that `is_present` would be `False` (lines 160-62)? 

Related to that - do we need the `request_headers.get()` part that sets `is_present`? If the signed header is in the query string, would that not mean that, at this point in the code, `is_present` is always true?

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +83,178 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
...
+                        ) in sentry_sdk.get_current_scope().iter_trace_propagation_headers(
+                            span=span
+                        ):
+                            normalized_header = header_name.lower()
+                            # preserve signed headers and avoid duplicate `sentry-trace`.
+                            is_present, is_signed = request_headers.get(
+                                normalized_header, (False, False)
+                            )
+                            if is_signed or (
+                                is_present and normalized_header != BAGGAGE_HEADER_NAME
```
ericapisani: If we're looking to avoid a duplicate `sentry-trace` (the comment on line 172), should `BAGGAGE_HEADER_NAME` instead be `SENTRY_TRACE_HEADER_NAME`? Do we potentially need to avoid duplicating both of these headers?

sentry_sdk/integrations/stdlib.py
```diff
@@ -73,7 +83,178 @@ def _complete_span(span: "Union[Span, StreamedSpan]") -> None:
...
+    """
+    Responsible for initializing request and signed header tracking on the instance.
+    """
+
+    def putrequest(
+        self: "HTTPConnection", method: str, url: str, *args: "Any", **kwargs: "Any"
+    ) -> "Any":
+        # track which propagation headers are present and signed e.g. {"sentry-trace": (is_present, is_signed)}
+        request_headers: "Optional[Dict[str, Tuple[bool, bool]]]" = {}
+        self._sentrysdk_request_headers = request_headers  # type: ignore[attr-defined]
```
ericapisani: Was there code that was previously here that used `request_headers`? This looks like we could remove `request_headers` and assign `self._sentrysdk_request_headers` to an empty dict instead


```suggestion
        self._sentrysdk_request_headers = {}
```

ericapisani (commented):
I'm going to hold off on reviewing the tests for now until we've addressed all the questions around the core logic.

Overall lots of great stuff here -  I particularly appreciate the docstrings/comments explaining what's happening.

Don't hesitate to reach out if there's any notes that I've left here that you want to chat about! 😄

sentry_sdk/integrations/stdlib.py
pabloDeputter: I was trying to fix a mypy error, but just did `type: ignore[attr-defined]` in the end and forgot to remove that part.

sentry_sdk/integrations/stdlib.py
pabloDeputter: yes, it represents the whole URL, but its mainly used for trace propagation. `_sentrysdk_real_url` is better though.

sentry_sdk/integrations/stdlib.py
pabloDeputter: `{}` represents the state whenever tracking was initialized, but no headers were observed; `None` if tracking of headers was never initialized. So default should indeed be `None`.

sentry_sdk/integrations/stdlib.py
pabloDeputter: The only tracked headers are `baggage` and `sentry-trace`, so `normalized_header != BAGGAGE_HEADER_NAME` is basically `normalized_header == 'sentry-trace'` and prevents a duplicate `sentry-trace`. I can rewrite the condition to make this clearer.

For both `baggage` and `sentry-trace`, we preserve them when they are signed because changing either would invalidate the signature. When unsigned, we never duplicate an existing `sentry-trace` because from my understanding it's just the trace-parent value, so multiple values would be ambiguous. We do allow another unsigned `baggage` field since multiple baggage fields are allowed and are treated as one combined list (https://www.w3.org/TR/baggage/#baggage-http-header-format).

sentry_sdk/integrations/stdlib.py
pabloDeputter: So originally I tracked both seen and signed headers separately, but combined them to avoid storing 2 vars on the `HTTPConnection` instance (see @alexander-alderman-webb comment).  Also `x-amz-signedheaders` only contains signed header names, so finding a header there does not mean that it was observed in `putheader()`. For a valid request, however, I think should normally also be present.

Apart from that, I think we can just use a single set that contains the headers we need to protect (avoid adding/modyfing):
- both `baggage` and `sentry-trace` are added when signed
- `sentry-trace` is added when seen, to avoid any duplication
- as explained below `baggage` is ignored, since another field is allowed

sentry_sdk/integrations/stdlib.py
```diff
@@ -473,6 +460,9 @@ def close(self: "HTTPResponse") -> None:
                 _complete_span(span)
 
     HTTPConnection.putrequest = putrequest  # type: ignore[method-assign]
+    # patch `HTTPConnection.putrequest()` first. urllib3 < 1.25.9 inherits it, so
```
ericapisani, line 463: I'd also add a note that once `boto3-v1.16.63` support is dropped we don't need to worry about this

tests/integrations/boto3/test_aws_http_connection.py
```diff
@@ -0,0 +1,230 @@
...
+    response = connection.getresponse()
+    response.read()
+    connection.close()
+
+
+@pytest.mark.parametrize("span_streaming", [False, True])
+def test_aws_http_connection_adds_missing_unsigned_propagation_headers(
+    sentry_init, local_http_server, span_streaming
+):
+    """Add missing unsigned `sentry-trace` and `baggage`."""
```
ericapisani, line 52: The test name here is descriptive enough that this comment is not necessary and can be removed.

tests/integrations/boto3/test_aws_http_connection.py
```diff
@@ -0,0 +1,230 @@
...
+
+    # existing `sentry-trace`: leave as-is since it is a single trace context.
+    assert headers.get_all("sentry-trace") == ["existing-trace"]
+
+
+@pytest.mark.parametrize("span_streaming", [False, True])
+def test_aws_http_connection_preserves_signed_propagation_headers(
+    sentry_init, local_http_server, span_streaming
+):
+    """Leave signed `sentry-trace` and `baggage` as-is."""
```
ericapisani, line 130: Same as [comment above](https://github.com/getsentry/sentry-python/pull/7050/changes#r3874830855)

tests/integrations/boto3/test_trace_propagation.py
```diff
@@ -0,0 +1,217 @@
...
+    _AwsRequestHandler.requests = []
+    server = HTTPServer(("127.0.0.1", 0), _AwsRequestHandler)
+    thread = Thread(target=server.serve_forever, daemon=True)
+    thread.start()
+    return server, thread
+
+
+@pytest.mark.parametrize("span_streaming", [False, True])
+def test_botocore_merges_propagation_before_sigv4_signing(sentry_init, span_streaming):
+    """Merge `sentry-trace` and `baggage` before SigV4 signs them."""
```
ericapisani: Same as [other comment](https://github.com/getsentry/sentry-python/pull/7050/changes#r3874830855)

ericapisani (approved):
LGTM 🚀 

There's some small nitpicks as I went through the tests but feel free to disregard if you feel they don't add value. 

The test coverage also looks great, the descriptive/specific test names are 👌🏻

adilnaimi:
Field confirmation: **2.69.1 verified in production** alongside ddtrace 4.13.2, with botocore instrumentation and distributed tracing fully enabled and no `trace_propagation_targets` restriction.

S3 `PutObject`/`HeadObject`/`CopyObject`/`DeleteObject` all sign correctly under an active sampled transaction — the same operations that failed for us on pre-#7050 sentry-sdk with ddtrace ≥ 4.12 (the two-writer baggage collision, dd-trace-py#19477, which forced a production rollback in August). ~40 hours in production since upgrading with zero `SignatureDoesNotMatch` errors, and we were able to remove all the workaround configuration on both SDKs.

Thanks for the fix.

PR #7054, opened 2026-08-05
feat(anthropic): Gate prompt collection on data_collection option

ericapisani:
Replace include_prompts and send_default_pii checks with the new data_collection configuration for controlling whether messages and system instructions are captured. Maintain backwards compatibility: when data_collection is not configured, fall back to the legacy pii/include_prompts behavior. Tools are always collected regardless of the message collection setting.

Refs PY-2588

### DIFF
--- a/sentry_sdk/integrations/anthropic.py
+++ b/sentry_sdk/integrations/anthropic.py
@@ -420,73 +420,14 @@ def _set_common_input_data(
     if messages is None or len(messages) == 0:  # type: ignore
         return
 
+    record_inputs = False
     if has_data_collection_enabled(client.options):
         if client.options["data_collection"]["gen_ai"]["inputs"]:
-            if isinstance(system, str) or isinstance(system, Iterable):
-                set_on_span(
-                    SPANDATA.GEN_AI_SYSTEM_INSTRUCTIONS,
-                    json.dumps(_transform_system_instructions(system)),
-                )
-
-            normalized_messages = []
-            for message in messages:
-                if (
-                    message.get("role") == GEN_AI_ALLOWED_MESSAGE_ROLES.USER
-                    and "content" in message
-                    and isinstance(message["content"], (list, tuple))
-                ):
-                    transformed_content = []
-                    for item in message["content"]:
-                        # Skip tool_result items - they can contain images/documents
-                        # with nested structures that are difficult to redact properly
-                        if isinstance(item, dict) and item.get("type") == "tool_result":
-                            continue
-
-                        # Transform content blocks (images, documents, etc.)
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb:
This is currently failing tests so I'll review when they're green

ericapisani:
@alexander-alderman-webb Got it sorted - the original tests I had written didn't work on older versions of the Anthropic SDK because the `tool` property wasn't supported at that point in `messages.create`.

Should be ready for 👀 now

sentry_sdk/integrations/anthropic.py
```diff
@@ -470,8 +403,154 @@ def _set_common_input_data(
...
+                    SPANDATA.GEN_AI_REQUEST_AVAILABLE_TOOLS, safe_serialize(tools)
+                )
+    else:
+        # Tools were unconditionally added pre-data collection configuration.
+        # This can be removed once data collection is fully rolled out
+        if tools is not None and _is_given(tools) and len(tools) > 0:  # type: ignore
+            set_on_span(SPANDATA.GEN_AI_REQUEST_AVAILABLE_TOOLS, safe_serialize(tools))
+
+    if messages is not None and len(messages) > 0:  # type: ignore
+        if has_data_collection_enabled(client.options):
```
alexander-alderman-webb: Would an early return along these lines make sense here?

```suggestion
    if messages is None or len(messages) == 0:  # type: ignore
    	return
 
    if has_data_collection_enabled(client.options):
```

alexander-alderman-webb (approved):

sentry_sdk/integrations/anthropic.py
```diff
@@ -390,74 +391,6 @@ def _set_common_input_data(
...
-
-        normalized_messages = []
-        for message in messages:
-            if (
-                message.get("role") == GEN_AI_ALLOWED_MESSAGE_ROLES.USER
-                and "content" in message
-                and isinstance(message["content"], (list, tuple))
-            ):
-                transformed_content = []
-                for item in message["content"]:
```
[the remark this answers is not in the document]
ericapisani: We gate them when data collection is enabled. When it's not enabled (legacy) we always collect them.

sentry_sdk/integrations/anthropic.py
ericapisani: 100%, nice catch! Feels good to remove remove an indent 🧹

ericapisani: Done in [c16adff](https://github.com/getsentry/sentry-python/pull/7054/commits/c16adff640b35e0e3405ad34cf20300818f2c2f2)

PR #7135, opened 2026-08-07
ref(aiohttp): Move crumbs to integration

sentrivana:
### Description
Create HTTP client breadcrumbs directly in the aiohttp integration instead of using `maybe_create_breadcrumbs_from_span`.

👀 **Best reviewed with Hide whitespace on.**

This turned out to be a bigger change than expected:
- Breadcrumb data should be on par with what we're sending now. When we were taking it from the span, we had all the data readily available when we needed to create the breadcrumb. When moving the logic to the integration, we now need to remember information from the start request hook because it's not available in the finish request hook, where we create the breadcrumb now. (We need to create the breadcrumb at the end, because it contains stuff like HTTP status code.)
- Breadcrumb data should be subject to PII filtering from `send_default_pii`/`data_collection`. So e.g. if parts of the query have been redacted on a span, they should not appear verbatim in the breadcrumb.
- Since the PII redacting logic is different between transaction-based tracing and span streaming, the existing crumb tests needed to test both paths, as well as different PII filtering settings.

Additionally:
- Rename `span` on the aiohttp trace context to `_sentry_span`

#### Reminders


sentry_sdk/integrations/aiohttp.py
```diff
@@ -388,60 +389,69 @@ async def on_request_start(
...
+
         span_name = "%s %s" % (
             method,
             parsed_url.url if parsed_url else SENSITIVE_DATA_SUBSTITUTE,
         )
 
         span: "Union[Span, StreamedSpan, None]"
         if has_span_streaming_enabled(client.options):
-            if sentry_sdk.traces.get_current_span() is None:
-                span = None
```
sentrivana, line 399: This had to be moved after the PII redaction, because even if we don't want to create a span, we do want to create a breadcrumb, and we need to apply the same PII redacting logic to breadcrumbs.

sentry_sdk/tracing_utils.py
```diff
@@ -210,10 +210,26 @@ def record_sql_queries(
...
+        kwargs["level"] = level
+
+    sentry_sdk.add_breadcrumb(**kwargs)
+
+
 def maybe_create_breadcrumbs_from_span(
     scope: "sentry_sdk.Scope", span: "sentry_sdk.tracing.Span"
 ) -> None:
-    if span.op == OP.HTTP_CLIENT:
+    if span.op == OP.HTTP_CLIENT and span.origin not in ("auto.http.aiohttp",):
```
sentrivana, line 231: This is just here to make sure we're not creating breadcrumbs the old way in transaction-based tracing anymore. Once all HTTP client integrations have been migrated, the whole function will go away

sentry_sdk/tracing_utils.py
```diff
@@ -210,10 +210,26 @@ def record_sql_queries(
             yield span
 
 
+def add_http_breadcrumb(status_code, data):
+    # type: (Optional[int], dict[str, Any]) -> None
```
ericapisani: Is there something preventing this type definition from living inline like the following?

```suggestion
def add_http_breadcrumb(status_code: "Optional[int]", data: "dict[str,Any]") -> "None":
```

Or is it in a comment as a matter of personal preference?

ericapisani (approved):
Non-blocking question, otherwise LGTM 🚀

sentry_sdk/tracing_utils.py
sentrivana: No, thanks for catching this, fixed in https://github.com/getsentry/sentry-python/pull/7135/commits/689d9d1848605d0275fe9c32ef7154c3cbb981df

sentry_sdk/integrations/aiohttp.py
```diff
@@ -475,18 +490,35 @@ async def on_request_start(
...
 
     async def on_request_end(
         session: "ClientSession",
         trace_config_ctx: "SimpleNamespace",
         params: "TraceRequestEndParams",
     ) -> None:
-        if trace_config_ctx.span is None:
+        status = int(params.response.status)
+
+        breadcrumb = trace_config_ctx._sentry_breadcrumb
```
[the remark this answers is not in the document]
sentrivana: I cannot think of a usecase for disabling the integration mid-request, but being extra defensive doesn't hurt: https://github.com/getsentry/sentry-python/pull/7135/commits/8bfb87981307c2a29a915c58a969b347299d045a

sentry_sdk/integrations/aiohttp.py
```diff
@@ -453,6 +461,13 @@ async def on_request_start(
                 legacy_span.set_data("url", parsed_url.url)
                 legacy_span.set_data(SPANDATA.HTTP_QUERY, parsed_url.query)
                 legacy_span.set_data(SPANDATA.HTTP_FRAGMENT, parsed_url.fragment)
+                breadcrumb.update(
+                    {
+                        SPANDATA.HTTP_QUERY: parsed_url.query,
+                        SPANDATA.HTTP_FRAGMENT: parsed_url.fragment,
+                        "url": parsed_url.url,
+                    }
```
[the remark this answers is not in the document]
sentrivana: This just follows current behavior

PR #7140, opened 2026-08-10
feat(django): Add `failed_request_status_codes`

mgaligniana:
### Description

Hi! In this PR I've added `failed_request_status_code` to the Django integration

- [X] I used AI and reviewed the changes.

Since English isn't my first language, I tried to keep the comments as simple and clear as possible, so they're easy to understand even for beginners like me who don't know the full Sentry product. Feel free to make any changes or suggestions!

#### Issues

* resolves 1 item of: getsentry/sentry-python#3134

### DIFF
--- a/sentry_sdk/integrations/django/__init__.py
+++ b/sentry_sdk/integrations/django/__init__.py
@@ -1,3 +1,4 @@
+import functools
 import inspect
 import sys
 import threading
@@ -6,7 +7,12 @@ from importlib import import_module
 
 import sentry_sdk
 from sentry_sdk.consts import OP, SPANDATA, SPANNAME
-from sentry_sdk.integrations import DidNotEnable, Integration, _check_minimum_version
+from sentry_sdk.integrations import (
+    _DEFAULT_FAILED_REQUEST_STATUS_CODES,
+    DidNotEnable,
+    Integration,
+    _check_minimum_version,
+)
 from sentry_sdk.integrations._wsgi_common import (
     DEFAULT_HTTP_METHODS_TO_CAPTURE,
     RequestExtractor,
@@ -80,6 +86,7 @@ else:
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
+    from collections.abc import Set
     from typing import Any, Callable, Dict, List, Optional, Union
 
     from django.core.handlers.wsgi import WSGIRequest
@@ -87,7 +94,13 @@ if TYPE_CHECKING:
     from django.http.response import HttpResponse
     from django.utils.datastructures import MultiValueDict
 
-    from sentry_sdk._types import Event, EventProcessor, Hint, NotImplementedType
+    from sentry_sdk._types import (
+        Event,
+        EventProcessor,
+        ExcInfo,
+        Hint,
+        NotImplementedType,
+    )
     from sentry_sdk.integrations.wsgi import _ScopedResponse
     from sentry_sdk.traces import StreamedSpan
     from sentry_sdk.tracing import Span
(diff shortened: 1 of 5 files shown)


sentry_sdk/integrations/django/__init__.py
```diff
@@ -614,18 +632,89 @@ def _got_request_exception(request: "WSGIRequest" = None, **kwargs: "Any") -> No
...
+    # Record that this exception is reported, so `_patch_response_for_exception`
+    # doesn't report it a second time.
+    with capture_internal_exceptions():
+        request._sentry_exception_reported = True
+
+    _capture_exception(sys.exc_info(), request, integration, handled=False)
+
+
+def _capture_exception(
+    exc_info: "Any",
```
ericapisani: We can tighten this typing a bit by copying what we do within the `event_from_exception` method that's being invoked within this function:


```suggestion
    exc_info: "Union[BaseException, ExcInfo]",
```

This requires updating the import from `sentry_sdk._types` at the top to import `ExcInfo`:

```
    from sentry_sdk._types import (
        Event,
        EventProcessor,
        ExcInfo,
        Hint,
        NotImplementedType,
    )
```

tests/integrations/django/test_basic.py
```diff
@@ -1761,6 +1761,96 @@ def test_does_not_capture_403(
...
+    events = capture_events()
+
+    _, status, _ = unpack_werkzeug_response(client.get(reverse("view_exc")))
+    assert status.lower() == "500 internal server error"
+
+    (event,) = events
+    (exception,) = event["exception"]["values"]
+    assert exception["type"] == "ZeroDivisionError"
+    assert exception["mechanism"]["type"] == "django"
+    assert exception["mechanism"]["handled"] is False
```
ericapisani, line 1851: Although I don't imagine there are many cases where someone would want to do this, if a user provides an empty set for `failed_request_status_codes`, or a non-empty set that doesn't include 5xx response codes, they will expect that no events are sent for these exceptions.

This will mean we need some additional changes to `_got_request_exception` to account for this, using a similar check to what you've added in `_patch_response_for_exception`. 

Once that's done, we can look to remove this test case in favour of another parameterized case to the test you've added above (`test_failed_request_status_codes`), ideally one that confirms the behaviour by making a request against the `view_exc` endpoint to trigger a 500 error and ensure that no events are created.

ericapisani (changes requested):
Thanks for opening this PR @mgaligniana , we really appreciate you taking the time to do so!

Overall the changes are looking great. However, I don't think we can exclude the 5xx codes from the `failed_request_status_codes` (left a more detailed comment below) and we'll need to address that before we can merge.

Don't hesitate to reach out with any questions!

tests/integrations/django/test_basic.py
alexander-alderman-webb: Other integrations still capture exception raised by user code in the route.
This behavior is independent of the `failed_request_status_codes` parameter in the integrations.
IMO, it'd be good to keep the behavior consistent with other integrations. This means
- An unexpected exception that occurs in the route is always captured (even if it materialized in a 500 response)
- An explicit `django.http.HttpResponse (status=500)` is not captured if 500 is not in `failed_request_status_codes`

ericapisani: @mgaligniana Ignore my earlier comment, I followed up with @alexander-alderman-webb to get some additional clarification and he's correct here. Marking this thread as resolved.

mgaligniana: Perfect!! Thanks for the explanation! I was reading the rest of the integrations to understand them better and respond haha

sentry_sdk/integrations/django/__init__.py
mgaligniana: Done!

ericapisani (approved):

ericapisani:
@mgaligniana Thanks so much again for your contribution! :shipit: 🚀

mgaligniana:
Wohoo!! Thanks to you for the review! Should I create a PR to sentry-docs or you handle that part?

ericapisani:
We'd love a PR if you have the time. If you don't though, don't sweat it, we can take care of it.

PR #7154, opened 2026-08-10
feat(mcp): Add MCP data collection gating

ericapisani:
Refs PY-2588
Refs #6748

### DIFF
--- a/sentry_sdk/integrations/mcp.py
+++ b/sentry_sdk/integrations/mcp.py
@@ -20,7 +20,12 @@ from sentry_sdk.integrations import DidNotEnable, Integration
 from sentry_sdk.scope import should_send_default_pii
 from sentry_sdk.traces import StreamedSpan
 from sentry_sdk.tracing_utils import has_span_streaming_enabled
-from sentry_sdk.utils import nullcontext, package_version, safe_serialize
+from sentry_sdk.utils import (
+    has_data_collection_enabled,
+    nullcontext,
+    package_version,
+    safe_serialize,
+)
 
 try:
     from mcp.server.lowlevel import Server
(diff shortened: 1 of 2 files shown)


tests/integrations/mcp/test_mcp.py
```diff
@@ -2015,3 +2015,429 @@ def failing_tool(tool_name, arguments):
...
+            return {"result": "success"}
+
+    params = {
+        "name": "calculate",
+        "arguments": {"x": 10, "y": 5},
+    }
+
+    if span_streaming:
+        items = capture_items("span")
+        with sentry_sdk.traces.start_span(name="mcp tx"):
```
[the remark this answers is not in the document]
ericapisani: Not adding `send_default_pii` gating here because data collection will be out soon

sentry_sdk/integrations/mcp.py
```diff
@@ -322,13 +327,21 @@ async def _tool_handler_wrapper(
...
+
     if original_kwargs is None:
         original_kwargs = {}
 
     handler_name, arguments = _extract_handler_data_from_args(
         "tool", original_args, original_kwargs
     )
 
+    if has_data_collection_enabled(client.options):
+        if not client.options["data_collection"]["gen_ai"]["inputs"]:
```
alexander-alderman-webb, line 340: The `gen_ai.output.messages` attribute contains tool call arguments and is gated by `data_collection["gen_ai"]["outputs"]`. Since MCP handler arguments are the output of a model, should we gate them by `data_collection["gen_ai"]["outputs"]`?

ericapisani: > Since MCP handler arguments are the output of a model...

I don't think so in this context given this is a tool call argument (defined on the DC spec as input data), even if it is the result of the output of a model.

I think I'd expect, in terms of what gets set on a span, that if I had opt-ed out of outputs and wanted inputs collected, I'd expect that the output data attribute (the model result) would be omitted, but the input to a tool call (also the model result) would get set on the related "input" span attribute.

alexander-alderman-webb: Cool, this seems to align with Vercel AI intrepretation, which uses similar input and output gates.

alexander-alderman-webb (approved):

Issue #6975, opened 2026-07-31
Investigate reducing number of tests without test coverage


ericapisani:
We have a pretty sizable test suite. Explore if there are opportunities for removing dead/duplicate tests without losing coverage

PR #7157, opened 2026-08-10
test(django): Drop span_streaming parametrize where the arm is inert

ericapisani:
### Description

These tests are event-only: they capture events, never spans, and the `span_streaming` argument was only ever fed to `trace_lifecycle`. Both arms exercised identical code paths and asserted identical payloads, so the parametrize doubled the case count without adding coverage.

Tests that inspect spans or transactions keep their parametrize.

391 -> 356 cases.

Refs PY-2641
Refs #6975


sentrivana (commented):
I think some of these go through different code paths depending on whether span streaming is enabled or not, even if they arrive at the same result

So maybe we can simplify the asserts (remove the `if span_streaming` branching and just leave one branch), but it might still make sense to parametrize so that we exercise both paths

I'll have a closer look to see if I can identify tests where we'd want to keep the parametrize

tests/integrations/django/test_basic.py
```diff
@@ -170,56 +132,30 @@ def test_middleware_exceptions(sentry_init, client, capture_exceptions):
     assert isinstance(error, ZeroDivisionError)
 
 
-@pytest.mark.parametrize("span_streaming", [True, False])
 def test_request_captured(
```
sentrivana, line 136: This one I'd keep parametrized, because it tests that `transaction` is put on events correctly, and that happens differently in streaming/not streaming

tests/integrations/django/test_basic.py
```diff
@@ -1291,201 +1092,115 @@ def test_transaction_style(
...
 
-            (event,) = events
+        (event,) = (item.payload for item in items)
 
     assert event["message"] == "hi"
     assert "data" not in event["request"]
 
 
-@pytest.mark.parametrize("span_streaming", [True, False])
 def test_template_tracing_meta(
```
sentrivana, line 1250: This one tests outgoing tracing headers, so it should also be parametrized on `span_streaming` or not

tests/integrations/django/test_basic.py
```diff
@@ -1726,36 +1427,24 @@ def test_rest_framework_authentication_span_without_authenticators(
 @pytest.mark.parametrize(
     "endpoint", ["rest_permission_denied_exc", "permission_denied_exc"]
 )
-@pytest.mark.parametrize("span_streaming", [True, False])
 def test_does_not_capture_403(
```
sentrivana, line 1497: I'm not 100% clear if this only aims to tests errors or also spans/transactions, but it seems like both according to the `capture_items` filter. So we should keep the parametrize, and accept `span` and `transaction` in `capture_items`

sentrivana (approved):
Found 3 tests that use a different code path in span streaming -- we should keep the `parametrize` on those. Otherwise lgtm

tests/integrations/django/test_basic.py
```diff
@@ -1345,72 +1151,40 @@ def test_transaction_style_tracing_disabled(
 def test_request_body(
     sentry_init,
     client,
-    capture_events,
     capture_items,
     span_streaming,
```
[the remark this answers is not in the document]
ericapisani: We need it in this case

Issue #7067, opened 2026-08-06
Set breadcrumbs in HTTP client integrations in the streaming trace lifecycle


alexander-alderman-webb:
Do what https://github.com/getsentry/sentry-python/blob/f670a1beed57735b3a956dd49eb72e4c106559d3/sentry_sdk/tracing_utils.py#L213 previously did.

- [x] aiohttp https://github.com/getsentry/sentry-python/pull/7135
- [x] pyreqwest https://github.com/getsentry/sentry-python/pull/7148
- [x] httpx https://github.com/getsentry/sentry-python/pull/7149
- [x] httpx2 https://github.com/getsentry/sentry-python/pull/7149
- [x] stdlib https://github.com/getsentry/sentry-python/pull/7161
- [x] boto3 https://github.com/getsentry/sentry-python/pull/7165

PR #7161, opened 2026-08-11
ref(stdlib): Move crumbs to integration

sentrivana:
### Description
Migrate `http.client` breadcrumbs from `maybe_create_breadcrumbs_from_span` directly to the `StdlibIntegration`.

As the flow is scattered (one point where a request starts, but multiple points where it might end), we save the breadcrumb data alongside the span and emit a breadcrumb whenever we detect a response is finished.

Also:
- Add tests to both the standard library test suite and requests.

#### Issues
Part of https://github.com/getsentry/sentry-python/issues/7067


sentry_sdk/integrations/stdlib.py
```diff
@@ -174,28 +204,42 @@ def putrequest(
...
         try:
             rv = real_getresponse(self, *args, **kwargs)
-        except BaseException:
-            _complete_span(span)
+        except BaseException as ex:
+            if span:
+                _complete_span(span)
+            if (
+                breadcrumb
+                and "getresponse() got an unexpected keyword argument 'buffering'"
```
sentrivana, line 222: This exception is basically used for control flow/compatibility in old urllib3. `getresponse()` will be called again afterwards without the extra arg, which is when we'll emit the breadcrumb. If this extra check were not here, we'd emit one crumb too many.

sentry_sdk/tracing.py
```diff
@@ -1495,5 +1493,4 @@ def calculate_interest_rate(amount, rate, years):
     extract_sentrytrace_data,
     has_span_streaming_enabled,
     has_tracing_enabled,
-    maybe_create_breadcrumbs_from_span,
 )
```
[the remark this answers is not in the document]
sentrivana: Right, forgot about boto3.

sentry_sdk/integrations/stdlib.py
sentrivana: (Also, we can get rid of this once we drop 3.6, it's only really a problem in super old requests on Python 3.6.)

alexander-alderman-webb (approved):

PR #7170, opened 2026-08-11
ref: Flush trace buckets when segment spans finish

alexander-alderman-webb:
### Description

Mark the corresponding bucket as pending in the span buffer when a segment span is added.

This aims to keep memory pressure low. Segment spans typically finish after their children. There may be multiple segments in a trace, and flushing more frequently leads to more outbound network requests in these cases.

Adapt span batcher tests by adding an outer segment in most tests. As a result, the various flush conditions are still exercised as the assertions run before the segment span has finished (finishing the segment span otherwise flushes the buffer as well).

The tests that exercise multiple buckets are changed to use `traces.new_trace()` in combination with a segment span that's left open

Without changes to the `sentry_init` fixture, tests fail on free-threading. This is because the background flusher collects pending items before `sentry_sdk.flush()` can synchronously flush all everything. The background flush is asynchronous and creates a race conditions.

#### Reminders

### DIFF
--- a/sentry_sdk/_span_batcher.py
+++ b/sentry_sdk/_span_batcher.py
@@ -100,7 +100,7 @@ class SpanBatcher(Batcher["SpanJSON"]):
                 self._flush()
                 self._last_full_flush = time.monotonic()
 
-    def add(self, span: "SpanJSON") -> None:
+    def add(self, span: "SpanJSON", flush_trace_bucket: "bool" = False) -> None:
         # Bail out if the current thread is already executing batcher code.
         # This prevents deadlocks when code running inside the batcher (e.g.
         # _add_to_envelope during flush, or _flush_event.wait/set) triggers
@@ -129,7 +129,8 @@ class SpanBatcher(Batcher["SpanJSON"]):
                 self._running_size[span["trace_id"]] += self._estimate_size(span)
 
                 if (
-                    size + 1 >= self.MAX_BEFORE_FLUSH
+                    flush_trace_bucket
+                    or size + 1 >= self.MAX_BEFORE_FLUSH
                     or self._running_size[span["trace_id"]]
                     >= self.MAX_BYTES_BEFORE_FLUSH
                 ):
--- a/sentry_sdk/_span_batcher.py
+++ b/sentry_sdk/_span_batcher.py
@@ -100,7 +100,7 @@ class SpanBatcher(Batcher["SpanJSON"]):
                 self._flush()
                 self._last_full_flush = time.monotonic()
 
-    def add(self, span: "SpanJSON", flush_trace_bucket: "bool" = False) -> None:
+    def add(self, span: "SpanJSON") -> None:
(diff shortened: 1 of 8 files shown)


sentry_sdk/_span_batcher.py
```diff
@@ -138,7 +139,9 @@ def add(self, span: "SpanJSON") -> None:
                 else:
                     notify = False
 
-            if notify:
+            if span["is_segment"] is True:
+                self._flush(only_pending=True)
```
alexander-alderman-webb: Maybe this should just set the `_flush_event`, thinking about it.

tests/conftest.py
```diff
@@ -293,11 +294,71 @@ def inner(identifier):
...
+            target = drained_count
+
+        batcher._flush_event.set()
+        while True:
+            with lock:
+                if drained_count > target:
+                    break
+            wake.wait()
+            wake.clear()
+        orig_flush()
```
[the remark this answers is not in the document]
alexander-alderman-webb: the full flush won't occur in tests because it takes 5 seconds to reach (bar a few tests).

sentrivana (approved):

tests/conftest.py
```diff
@@ -293,11 +294,71 @@ def inner(identifier):
     return inner
 
 
+def _install_flush_completion_handshake(client: "sentry_sdk.Client") -> None:
+    """Make batcher.flush() wait for the flusher thread to drain.
+
+    Otherwise, test assertions can run before envelopes are captured.
+    The span batcher flushes pending items asynchronously with the main thread.
+    Flushes triggered by segments finishing are asynchronous, and can collect buckets
+    that would have otherwise been flushed synchronously by `sentry_sdk.flush()`.
```
ericapisani, line 303: I'm finding this comment a little tricky to follow. My understanding of this section is:

- The span batcher sends out pending items in the background, separately from the main thread. 
- When a segment finishes, that triggers a background flush too
- The background flush in point 2 can end up sending buckets that sentry_sdk.flush() would otherwise have sent immediately.

Do I have this right? And why is point 3 problematic? Is it because we need those buckets immediately when invoking `sentry_sdk.flush()` and we may not get them because those buckets are in the process of being flushed by the background process?

alexander-alderman-webb: Yes that's correct! To add some links:

Each batcher has a flushing thread that runs the `_flush_loop()` method.
When a segment span finishes, `_flush()` is triggered inside the thread here:

https://github.com/getsentry/sentry-python/blob/4e4ea83b4fafaae3d53e68d756c55c74fee37545/sentry_sdk/_span_batcher.py#L94

This is what I called "asynchronous" in the comment.
When you call `sentry_sdk.flush()`, it triggers the flush here

https://github.com/getsentry/sentry-python/blob/4e4ea83b4fafaae3d53e68d756c55c74fee37545/sentry_sdk/client.py#L1381

which is **not in the flushing thread**. This runs synchronously with the user code.

PR #7174, opened 2026-08-11
fix(ai): Gate response tool calls on the gen_ai inputs setting

ericapisani:
`gen_ai.response.tool_calls` was gated on `data_collection.gen_ai.outputs` in the OpenAI, Anthropic, OpenAI Agents, and Google GenAI integrations. Tool calls belong to the inputs category, so they are now gated on `data_collection.gen_ai.inputs` while response text stays on `outputs`.

Where both attributes previously shared a single gate, the block is split so each is evaluated independently. The legacy `send_default_pii`/`include_prompts` path is unchanged.

Refs PY-2588
Refs #6748


alexander-alderman-webb (approved):
Thanks this makes sense!

PR #7175, opened 2026-08-11
feat(langgraph): Gate prompt/response collection on data_collection option

ericapisani:
Modify the LangGraph integration to respect the data_collection config for controlling whether prompts, responses, tool calls, and available tools are captured in spans.

When data collection is enabled, the gen_ai.inputs flag controls request messages, tool calls, and available tools, while gen_ai.outputs controls the response text. Tool calls are gated on inputs because they are fed back to the model as input. Available tools are only gated once data collection is configured, since they were never gated on the legacy PII settings.

When data collection is not configured, falls back to legacy send_default_pii and include_prompts settings for compatibility.

Refs getsentry/sentry-python#6748  <br>Refs getsentry/sentry-python#6748

### DIFF
--- a/sentry_sdk/integrations/langgraph.py
+++ b/sentry_sdk/integrations/langgraph.py
@@ -212,7 +212,7 @@ def _wrap_pregel_invoke(f: "Callable[..., Any]") -> "Callable[..., Any]":
 
                 # Store input messages to later compare with output
                 input_messages = None
-                if len(args) > 0 and _should_record_inputs(integration):
+                if len(args) > 0:
                     input_messages = _parse_langgraph_messages(args[0])
                     if input_messages:
                         normalized_input_messages = normalize_message_roles(
@@ -227,7 +227,9 @@ def _wrap_pregel_invoke(f: "Callable[..., Any]") -> "Callable[..., Any]":
                             if should_truncate_gen_ai_input(client.options)
                             else normalized_input_messages
                         )
-                        if messages_data is not None:
+                        if messages_data is not None and _should_record_inputs(
+                            integration
+                        ):
                             set_data_normalized(
                                 span,
                                 SPANDATA.GEN_AI_REQUEST_MESSAGES,
@@ -254,7 +256,7 @@ def _wrap_pregel_invoke(f: "Callable[..., Any]") -> "Callable[..., Any]":
 
                 # Store input messages to later compare with output
                 input_messages = None
-                if len(args) > 0 and _should_record_inputs(integration):
+                if len(args) > 0:
                     input_messages = _parse_langgraph_messages(args[0])
                     if input_messages:
                         normalized_input_messages = normalize_message_roles(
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb:
This integration sucks 😞 .
Could you take a look at the bot comments @ericapisani 🙏

ericapisani:
@alexander-alderman-webb Got it sorted - the changes I had introduced broke the message deltas. Have moved where the input gate is so it doesn't affect that and added regression tests for this scenario 🚀

sentry_sdk/integrations/langgraph.py
```diff
@@ -188,33 +212,29 @@ def new_invoke(self: "Any", *args: "Any", **kwargs: "Any") -> "Any":
...
-                            if should_truncate_gen_ai_input(client.options)
-                            else normalized_input_messages
-                        )
-                        if messages_data is not None:
-                            set_data_normalized(
-                                span,
-                                SPANDATA.GEN_AI_REQUEST_MESSAGES,
-                                messages_data,
-                                unpack=False,
+                        if _should_record_inputs(integration):
```
sentrivana: We could move this check above, to line 215, and combine it with the `len(args)` check so that we only do the input messages manipulation if necessary. Cheap way to save some cycles and remove one level of nesting.

sentry_sdk/integrations/langgraph.py
```diff
@@ -235,33 +255,29 @@ def new_invoke(self: "Any", *args: "Any", **kwargs: "Any") -> "Any":
...
```
sentrivana: Same here, we could combine this with the `args` check on L258.

sentry_sdk/integrations/langgraph.py
```diff
@@ -299,33 +315,29 @@ async def new_ainvoke(self: "Any", *args: "Any", **kwargs: "Any") -> "Any":
...
```
sentrivana: And here

sentry_sdk/integrations/langgraph.py
```diff
@@ -345,31 +357,27 @@ async def new_ainvoke(self: "Any", *args: "Any", **kwargs: "Any") -> "Any":
...
-                        if should_truncate_gen_ai_input(client.options)
-                        else normalized_input_messages
-                    )
-                    if messages_data is not None:
-                        set_data_normalized(
-                            span,
-                            SPANDATA.GEN_AI_REQUEST_MESSAGES,
-                            messages_data,
-                            unpack=False,
+                    if _should_record_inputs(integration):
```
sentrivana: And here (goodness this integration is a mess)

sentrivana (approved):
Looks good, left some suggestions

sentry_sdk/integrations/langgraph.py
ericapisani: > goodness this integration is a mess

Yeah, it needs a good overhaul 😭 😅

sentry_sdk/integrations/langgraph.py
ericapisani: It was originally there but I had to move it a little further down because `input_messages` are used in the creation of metadata information (such as token usage) for output messages that aren't gated on either flag (original bugbot error [here](https://github.com/getsentry/sentry-python/pull/7175#discussion_r3761464918)) 😞 

Looking at it again, I could add it move it up to the `if input_messages` conditional (saves us from the `normalize_message_roles` call), but not any higher.

PR #7176, opened 2026-08-12
fix(serializer): Avoid building full repr of large objects (#6649)

### LINKED ISSUE #6649: Logging integration blocks the event loop for seconds while serializing frame locals on FastAPI >= 0.137
## Steps to reproduce

No artificial `__repr__` — plain FastAPI routes are enough. The delay scales with the number of routes in the app:

```python
import os
import time
from typing import Never
import logging

from fastapi import Query, Depends, FastAPI, Request, APIRouter
from pydantic import BaseModel
import sentry_sdk
from fastapi.responses import JSONResponse
from fastapi.testclient import TestClient

sentry_sdk.init(
    dsn="https://examplekey@o0.ingest.sentry.io/0",  # fake DSN, nothing is sent
    include_local_variables=os.getenv("LOCALS", "1") == "1",  # default True
)


class Model(BaseModel):  # a typical response model
    a: int
    b: str
    c: float
    d: bool
    e: list[int]
    f: dict[str, int]


async def dep(q: str = Query("x")):
    return q


async def boom_dep() -> Never:
    raise KeyError("boom")


leaf = APIRouter()
for k in range(800):  # a non-trivial app: real services have hundreds/thousands of routes

    @leaf.get(f"/r{k}", response_model=Model)
    async def ep(d=Depends(dep)):
        return {}


@leaf.get("/boom", response_model=Model)
async def boom(_: object = Depends(boom_dep)) -> None: ...


router = leaf
for i in range(6):  # nested include_router()
    parent = APIRouter(prefix=f"/l{i}")
    parent.include_router(router)
    router = parent

app = FastAPI()
app.include_router(router)


@app.exception_handler(Exception)
async def on_error(request: Request, exc: Exception):
    t = time.perf_counter()
    logging.getLogger("x").error("err", exc_info=exc)  # ERROR + exc_info -> Sentry capture
    print(f"logger.error took {time.perf_counter() - t:.3f}s")
    return JSONResponse(status_code=500, content={})


path = "".join(f"/l{i}" for i in range(5, -1, -1)) + "/boom"
TestClient(app, raise_server_exceptions=False).get(path)
```

Measured locally with the repro above:

```text
FastAPI 0.137.2:

LOCALS=1  python repro.py   # logger.error took 0.530s
LOCALS=0  python repro.py   # logger.error took 0.073s


FastAPI 0.136.3:

LOCALS=1  python repro.py   # logger.error took 0.067s
LOCALS=0  python repro.py   # logger.error took 0.084s
```

So the regression is specifically that `include_local_variables=True` used to be cheap on FastAPI 0.136.3, but becomes much slower on FastAPI >= 0.137.

**Expected:** capturing a single exception from a normal FastAPI route should not require materializing a representation of the whole router graph, and should not block the event loop for hundreds of milliseconds or seconds.

**Actual:** the `logging` call blocks while Sentry serializes the locals of every traceback frame. Each nested routing frame holds an `_IncludedRouter` (FastAPI >= 0.137) whose `repr()` walks the whole router graph, so the cost grows with app size — ~0.53s at 800 routes here, and seconds in a real app with thousands of routes and richer models. On FastAPI 0.136.3 this object is not in the frame locals, and the same capture is ~0.07s even with `include_local_variables=True`.

This was observed in production: a service was upgraded from FastAPI 0.136.3 to 0.137.2, and any unhandled exception started taking ~20s to log. Running under gunicorn's `UvicornWorker`, this blocks the event loop long enough to trip `WORKER TIMEOUT` → `SIGABRT` → worker restart, so a normal 500 turns into a worker kill. Reverting to FastAPI 0.136.3 restors ~0.1s.

## Environment

* `sentry-sdk` 2.63.0
* `fastapi` 0.137.2 (fast on 0.136.3)
* `starlette` 1.3.1, `uvicorn` 0.49.0, `gunicorn` 26.0.0 (`-k uvicorn.workers.UvicornWorker`)
* Python 3.14

## Measurements (production)

Instrumenting the exception path showed the whole delay is inside one logging call, and only when it goes through Sentry's `callHandlers` patch.

Walking the traceback frame-by-frame and timing exactly what Sentry does for each frame (`serialize(dict(frame.f_locals), is_vars=True)`) shows the cost is entirely in FastAPI's own router objects, not in any application object:

```text
frame routing.py:handle           self / included_router  <_IncludedRouter>  ~117 MiB  ~0.96s
frame routing.py:_handle_selected  self / route            <_IncludedRouter>  ~117 MiB  ~0.95s
... repeated for every level of the nested include_router() tree (~8 frames) ...
frame <our endpoint / domain frames>  <plain app objects>                    < 1 KiB   0.000s
```

Each `_IncludedRouter` local causes a huge transient representation to be materialized before truncation, and one appears as `self` / `route` / `included_router` in every nested routing frame, so the total can reach ~15-20s. The application's own frame locals are negligible (`0.000s`) — this is not caused by a user object with a heavy `__repr__`; serializing FastAPI's standard per-request router graph is enough on its own.

Not the same as getsentry/sentry-python#6568 / [#6569](<https://github.com/getsentry/sentry-python/issues/6569>) (sync-handler `_sentry_call` accumulation → `RecursionError` after ~987 requests, fixed in 2.63.0): this affects async endpoints too, triggers on the first error, and the cost is time spent serializing locals.

## Possible root cause

`LoggingIntegration` patches `logging.Logger.callHandlers`; for an `ERROR` record with `exc_info` it builds an event via `event_from_exception(..., include_local_variables=True)`, serializing each frame's locals. A `_IncludedRouter` is not a Mapping/Sequence/Set, so the serializer does not appear to apply its depth/breadth limits to it — it goes to `safe_repr`:

```python
# sentry_sdk/utils.py
def safe_repr(value):
    return repr(value)          # full repr is materialized here
# ...then strip_string() truncates the already-built string
```

FastAPI 0.137 ([fastapi/fastapi#15745](<https://github.com/fastapi/fastapi/issues/15745>)) stopped flattening routes on `include_router()`. Instead, each routing frame now holds an `_IncludedRouter`, which is a `@dataclass`:

```python
# fastapi/routing.py
@dataclass
class _IncludedRouter(BaseRoute):
    original_router: "APIRouter"
    include_context: _RouterIncludeContext
    _effective_candidates: list["_EffectiveRouteContext | _IncludedRouter"] = field(default_factory=list)
    ...
```

The auto-generated dataclass `__repr__` recurses through every field — `original_router` (and its `.routes`) plus the lazily-built `_effective_candidates` tree — i.e. effectively the whole router graph of the app. `safe_repr` materializes that entire string before `strip_string` truncates it. The same `_IncludedRouter` then appears as `self` / `route` / `included_router` in each nested routing frame, so this whole-app `repr()` is paid once per frame. On 0.136 routes were flattened and no such object was in the frame locals, so the capture stayed cheap.

## Workarounds

`sentry_sdk.init(..., include_local_variables=False)` removes the delay entirely, but drops locals from every frame.

A more targeted workaround is `custom_repr`, which keeps locals everywhere else and only short-circuits the router objects:

```python
def custom_repr(value):
    cls = type(value)
    module = getattr(cls, "__module__", None)

    if module in {"fastapi.routing", "starlette.routing"}:
        return f"<{module}.{cls.__name__}>"

    return None  # fall back to the default repr for everything else


sentry_sdk.init(..., custom_repr=custom_repr)
```

---

> **Disclosure:** This issue was researched and written with the assistance of an AI (Claude) and reviewed by a human before submission. The reproduction and timings have been manually confirmed; I did not go deep into Sentry's serializer internals to verify the exact root cause or the best fix.

zkasuran:
When capturing frame locals, the serializer builds `repr()` of any non-container object in full and only truncates the resulting string afterwards. `repr()` of a large container or dataclass walks the whole object graph, so the string is built entirely and then mostly discarded. On FastAPI >= 0.137 every nested routing frame holds an `_IncludedRouter`, a dataclass whose auto-generated `__repr__` recurses through the full router tree, so serializing one frame local becomes a multi-megabyte repr. The same object recurs in each nested frame, so logging a single exception blocks the event loop for hundreds of milliseconds to seconds and can trip `UvicornWorker` timeouts.

This adds `bounded_repr()` in `utils.py`, which renders dataclass fields and the standard container types itself and stops once the output reaches the limit, returning a prefix marked with `...`. When the full repr fits, the result is identical to `repr()`. Leaf values including strings are rendered in full, so string values are never shortened. The serializer uses it capped at `max_value_length` when that is set, otherwise at a generous default so only pathologically large graphs are cut.

Behavior for a user-supplied `custom_repr` is unchanged. The call site is `repr_value or bounded_repr(value, ...)`, so a `custom_repr` result is returned verbatim exactly as before. `bounded_repr` only replaces the old `safe_repr` fallback when no `custom_repr` value is produced. The existing `test_custom_repr` and `test_custom_repr_graceful_fallback_to_safe_repr` still pass.

Fixes #6649

AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally on 3.12: the serializer, utils and logging tests pass (230), the new tests failing before and passing after the fix, `ruff check` and `ruff format --check` clean on the changed files, `mypy sentry_sdk` with no new errors, plus a before/after on the reporter's FastAPI repro dropping the stall from over a second to milliseconds.

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -1,5 +1,6 @@
 import base64
 import copy
+import dataclasses
 import json
 import linecache
 import logging
@@ -544,6 +545,115 @@ def safe_repr(value: "Any") -> str:
         return repr(value)
     except Exception:
         return "<broken repr>"
+
+
+# Maximum recursion depth when building a length-bounded repr of an object
+# graph. Anything deeper is rendered as "...". This mirrors the databag depth
+# limit and guards against pathologically deep (or self-referential) objects.
+_MAX_BOUNDED_REPR_DEPTH = 100
+
+
+class _BoundedReprLimit(Exception):
+    """Raised internally by bounded_repr() once the length budget is spent."""
+
+
+def bounded_repr(value: "Any", max_length: "Optional[int]") -> str:
+    """``repr(value)`` that stops once the output would exceed ``max_length``.
+
+    ``repr()`` of a container or a dataclass walks the whole object graph and
+    builds the entire string up front. When that string is then truncated by
+    the serializer, everything past the limit was built for nothing. For a
+    large graph the cost is significant: FastAPI's ``_IncludedRouter`` is a
+    dataclass whose auto-generated ``__repr__`` recurses through the full
+    router tree, so a single frame local can turn into a multi-megabyte repr
+    that blocks the event loop for hundreds of milliseconds before truncation
+    (getsentry/sentry-python#6649).
+
+    This renders dataclass fields and the standard container types itself and
+    stops as soon as the accumulated output reaches ``max_length``, returning a
+    prefix of ``repr(value)`` followed by ``"..."``. When the full repr fits
+    within ``max_length`` the result is byte-for-byte identical to
+    ``repr(value)``. Leaf values (strings, numbers, arbitrary objects with a
+    custom ``__repr__``, ...) are always rendered in full, so string values are
+    never shortened here; only over-large container/dataclass graphs are cut.
+
+    Objects that are neither dataclasses nor standard containers fall back to
+    ``safe_repr()``.
+    """
+    if max_length is None:
+        return safe_repr(value)
+
+    chunks: "List[str]" = []
+    total = 0
+
+    def emit(text: str) -> None:
+        nonlocal total
+        chunks.append(text)
+        total += len(text)
+
+    def render(obj: "Any", depth: int) -> None:
+        if depth > _MAX_BOUNDED_REPR_DEPTH:
+            emit("...")
+            return
+
+        if dataclasses.is_dataclass(obj) and not isinstance(obj, type):
+            items = (
+                (field.name + "=", getattr(obj, field.name))
+                for field in dataclasses.fields(obj)
+                if field.repr
+            )
+            _render_items(type(obj).__qualname__ + "(", ")", items, depth)
+            return
+
+        obj_type = type(obj)
+        if obj_type is dict:
+            _render_items(
+                "{", "}", ((safe_repr(k) + ": ", v) for k, v in obj.items()), depth
+            )
+        elif obj_type is list:
+            _render_items("[", "]", (("", v) for v in obj), depth)
+        elif obj_type is tuple:
+            if len(obj) == 1:
+                emit("(")
+                render(obj[0], depth + 1)
+                emit(",)")
+            else:
+                _render_items("(", ")", (("", v) for v in obj), depth)
+        elif obj_type is set:
+            if obj:
+                _render_items("{", "}", (("", v) for v in obj), depth)
+            else:
+                emit("set()")
+        elif obj_type is frozenset:
+            if obj:
+                _render_items("frozenset({", "})", (("", v) for v in obj), depth)
+            else:
+                emit("frozenset()")
+        else:
+            emit(safe_repr(obj))
+
+    def _render_items(opening: str, closing: str, items: "Any", depth: int) -> None:
+        emit(opening)
+        first = True
+        for prefix, child in items:
+            if total >= max_length:
+                # The graph is bigger than the budget. Stop here and let the
+                # caller mark the value as truncated. What we have emitted so
+                # far is a literal prefix of repr(value).
+                raise _BoundedReprLimit
+            if not first:
+                emit(", ")
+            first = False
+            emit(prefix)
+            render(child, depth + 1)
+        emit(closing)
+
+    try:
+        render(value, 0)
+    except _BoundedReprLimit:
+        chunks.append("...")
+
+    return "".join(chunks)
 
 
 def filename_for_module(
--- a/sentry_sdk/serializer.py
+++ b/sentry_sdk/serializer.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING
 
 from sentry_sdk.utils import (
     AnnotatedValue,
+    bounded_repr,
     capture_internal_exception,
     disable_capture_event,
     format_timestamp,
@@ -46,6 +47,14 @@ MAX_EVENT_BYTES = 10**6
 MAX_DATABAG_DEPTH = 5
 MAX_DATABAG_BREADTH = 10
 CYCLE_MARKER = "<cyclic>"
+
+# Upper bound on the length of a single object's repr when max_value_length is
+# not set (string truncation is disabled by default, see #6290). This does not
+# shorten string values; it only stops the repr of a large container/dataclass
+# graph from being fully materialized. A value repr larger than the maximum
+# event size would be trimmed away later anyway, so building it is pure waste
+# and, worse, can block the event loop (#6649).
+MAX_REPR_LENGTH = 100_000
 
 
 global_repr_processors: "List[ReprProcessor]" = []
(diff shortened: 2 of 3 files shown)


sentry_sdk/utils.py
```diff
@@ -546,6 +547,115 @@ def safe_repr(value: "Any") -> str:
...
+# graph. Anything deeper is rendered as "...". This mirrors the databag depth
+# limit and guards against pathologically deep (or self-referential) objects.
+_MAX_BOUNDED_REPR_DEPTH = 100
+
+
+class _BoundedReprLimit(Exception):
+    """Raised internally by bounded_repr() once the length budget is spent."""
+
+
+def bounded_repr(value: "Any", max_length: "Optional[int]") -> str:
```
alexander-alderman-webb, line 560: We can't do this in a minor version.
The `repr()` implementation for built-in types is a Python implementation detail, and enforcing our own logic is a disruptive change.

alexander-alderman-webb (changes requested):

zkasuran:
Makes sense, thanks for the quick look. Agreed that reimplementing repr for built-in types is too disruptive for a minor release. The two Bugbot points are fair too: the unconditional `dataclasses` import breaking 3.6, plus dataclass `__repr__`/`repr=False` overrides being ignored. Both fall out of rendering the repr ourselves rather than calling it.

I did try to find a non-disruptive way to bound the cost without reimplementing repr. The catch in this issue is that the expense lives inside the object's own `__repr__` recursion (the nested-router tree), not in the serializer's traversal, so the serializer's existing depth limit never sees it. I checked stdlib `reprlib` as the one option that isn't our own logic: for a recursive dataclass it still calls the object's full `__repr__` and only truncates the result, so it changes output without avoiding the cost.

Given that, I'll leave the reasonable-limits design to you rather than push this approach. Happy to test a candidate against the repro in the issue if that helps.

zkasuran:
You are right. There is a second reason this diff should not land as written. Taking your point first, then two findings, then a question, because I do not think I should pick the answer to it.

**Reimplementing `repr()` for the built-in containers.** Agreed, so I am dropping it. Rendering `dict`, `list`, `tuple` and `set` ourselves means our output can drift from CPython's on any version. The docstring's "byte-for-byte identical when it fits" is a promise across every container subtlety rather than a fact. That is not a minor-version change.

**A hard blocker I missed.** The unconditional `import dataclasses` in `sentry_sdk/utils.py` breaks Python 3.6. `dataclasses` arrived in 3.7 and `setup.py` still says `python_requires=">=3.6"`, so `import sentry_sdk` would fail outright there. Cursor flagged it and it is correct. That alone means the current head is not shippable.

**The dataclass finding is real too, and it is the worse of the two.** `bounded_repr` walked `dataclasses.fields()` for anything `is_dataclass` reports. A class with `@dataclass(repr=False)` or its own `__repr__` therefore got its fields rendered instead of the repr it chose. A value someone deliberately redacted would reappear in frame locals. Measured on 3.12, these three cases are distinguishable:

| | `type(obj).__repr__.__wrapped__.__qualname__` | `__dataclass_params__.repr` |
| --- | --- | --- |
| generated repr | `__create_fn__.<locals>.__repr__` | True |
| `@dataclass(repr=False)` | no `__wrapped__` | False |
| own `__repr__` | the user's own qualname | True |

So the fix is to take the bounded path only when the generated `__repr__` is the one actually in use, falling back to `safe_repr` otherwise. It fails closed, which matters: a detector that guesses wrong costs current behaviour, not a leak.

**Now the part I want your call on.** I checked whether a dataclass-only bound is enough to close #6649. It is not. `_IncludedRouter` is a dataclass, but its repr-included fields are `original_router` plus two `list` fields, `_effective_candidates` and `_effective_low_priority_routes`, and those lists hold the nested routers. So the recursion that stalls is dataclass into list into dataclass. Bounding only the dataclass shell still calls `repr()` on the first list and rebuilds the whole tree, which is the thing the issue is about. Bounding the list is what you just vetoed, and I think you are right to.

Three shapes I can see, none of them reimplementing a built-in repr:

1. **Refuse rather than render.** For a generated-repr dataclass, walk its fields, and when a field holds a sized container over some threshold emit `<list of 12043 items>` in place of calling `repr()` on it. That is not our version of list repr but a visible refusal to build one. Output changes only in the pathological case, and never silently.
2. **Opt in.** Same mechanism behind a new client option, off by default, so a minor release changes nothing for anyone who does not ask for it.
3. **Close this and leave #6649 to the team**, since you already said you wanted reasonable limits applied to object serialization more broadly and that is a bigger design than this PR.

I am happy with any of the three, including the third. My instinct is 1, on the grounds that the current default behaviour on a big FastAPI app is a multi-second stall rather than a correct repr anybody reads. That is a product judgement rather than mine to make. Say which and I will send it, or close it and I will not take it personally.

PR #7178, opened 2026-08-12
chore: Drop `enable_logs`, `enable_metrics` options in new major

sentrivana:
The options were deprecated with https://github.com/getsentry/sentry-python/commit/2fef9bce3cd58d7f2993bb2d1c7169845e154c81.


MIGRATION_GUIDE.md
```diff
@@ -45,6 +45,22 @@ Looking to upgrade from Sentry SDK 2.x to 3.x? Here's a comprehensive list of wh
 - Removed the RedisIntegration `max_data_size` option.
 - Removed the possibility to supply a specific client to the LaunchDarklyIntegration.
 - The `enable_tracing` option was removed. Use `traces_sample_rate=1.0` instead.
+- The `enable_logs` option was removed. Using Sentry's logging API now works without requiring setting `enable_logs=True`. Automatic capture of logs emitted by the `logging` standard library module or Loguru can be turned on by providing the `capture_sentry_logs=True` option to either `LoggingIntegration` or `LoguruIntegration`:
```
ericapisani, line 48: > Automatic capture of logs emitted by the `logging` standard library module or Loguru can be turned on by providing the `capture_sentry_logs=True` option to either `LoggingIntegration` or `LoguruIntegration`

Given`LoggingIntegration` and `LoguruIntegration` are both auto-enabling integrations, I wonder if it might be more useful to give an example for opting out rather than opting in.

ericapisani (approved):
Other than the one note for our guidance in the migration guide, this LGTM

MIGRATION_GUIDE.md
sentrivana: They're auto-enabling for errors, but they won't start capturing logs unless you explicitly opt in. So no logs is default behavior, and if you want logs, you need to explicitly enable that.

sentrivana: Also, status update 🤦🏻‍♀️ but this will change in a later PR. Will leave this as is as it captures the current state and will update it in the follow up PR.

PR #7180, opened 2026-08-12
feat(litellm): Gate prompt/response collection on data_collection option

ericapisani:
Modify the LiteLLM integration to respect the data_collection config for controlling whether chat messages, embeddings inputs, and response text are captured in spans.

When data collection is enabled, the new gen_ai.inputs and gen_ai.outputs flags control what data is collected. When data collection is not configured, falls back to legacy send_default_pii and include_prompts settings for compatibility.

Refs PY-2588
Refs #6748

### DIFF
--- a/sentry_sdk/integrations/litellm.py
+++ b/sentry_sdk/integrations/litellm.py
@@ -18,7 +18,7 @@ from sentry_sdk.tracing_utils import (
     has_span_streaming_enabled,
     should_truncate_gen_ai_input,
 )
-from sentry_sdk.utils import event_from_exception
+from sentry_sdk.utils import event_from_exception, has_data_collection_enabled
 
 if TYPE_CHECKING:
     from datetime import datetime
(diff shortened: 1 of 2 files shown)


ericapisani:
bugbot run

sentry_sdk/integrations/litellm.py
```diff
@@ -129,7 +129,13 @@ def _input_callback(kwargs: "Dict[str, Any]") -> None:
     set_data_normalized(span, SPANDATA.GEN_AI_OPERATION_NAME, operation)
 
     # Record input/messages if allowed
-    if should_send_default_pii() and integration.include_prompts:
+    record_inputs = False
+    if has_data_collection_enabled(client.options):
```
sentrivana, line 133: This might've already been discussed somewhere, in that case disregard, but: Should we also take `integration.include_prompts` into account here, so that users have more fine-grained control over which prompts will be sent?

sentrivana (approved):

sentry_sdk/integrations/litellm.py
alexander-alderman-webb: We've agreed that niche use cases like filtering by integration should be done in `before_send()`.
The option will be deprecated once data collection has fully landed.

alexander-alderman-webb (approved):

PR #7181, opened 2026-08-12
ref(pydantic_ai): Split prompt gating into separate inputs/outputs controls

ericapisani:
Replace the single \_should_send_prompts check with \_should_send_inputs and \_should_send_outputs, gated by the new data_collection.gen_ai.inputs/outputs experimental options when set, falling back to the existing send_default_pii

* include_prompts behavior otherwise. Available tools are now gated on inputs specifically. This lets callers control request and response data collection independently.

Refs getsentry/sentry-python#6748<br>Refs getsentry/sentry-python#6748


ericapisani:
bugbot run

ericapisani:
@cursor review

sentrivana (approved):

sentry_sdk/integrations/pydantic_ai/spans/ai_client.py
```diff
@@ -247,28 +248,38 @@ def _set_output_data(
...
                         {
                             "type": "reasoning",
                             "content": part.content,
                         }
                     )
 
-                elif BaseToolCallPart is not None and isinstance(
-                    part, BaseToolCallPart
+                elif (
+                    record_inputs
```
alexander-alderman-webb: I don't think we should be gating different parts of one attribute by different data collection options. 
Since we're setting the `gen_ai.output.messages` attribute can we gate everything by `record_outputs`?

```suggestion
                    record_outputs
```

alexander-alderman-webb (dismissed):

sentry_sdk/integrations/pydantic_ai/spans/ai_client.py
alexander-alderman-webb: I know we set `gen_ai.response.tool_calls` to be gated by `data_collection["gen_ai"]["inputs"]` in other AI integrations, which makes sense since these are the input to the tool calls.

Maybe you can check what the JS SDK does here?

ericapisani: Follow up from offline convo - tool calls in this context will be gated by "outputs" rather than "inputs". 

I've created a distinct [ticket](https://github.com/getsentry/sentry-python/issues/7200) to do an audit of all the tool call sites to make sure that tool calls that are added to the responses span attribute are gated by outputs instead of outputs, and ensuring that, where relevant, tool call arguments/responses are set on the non-deprecated `gen_ai.tool.call.arguments` and `gen_ai.tool.call.result` instead of `ai.response.toolCalls` (which is what I've been seeing a lot of across the SDK codebase)

ericapisani: Addressed for this integration in [5d83199](https://github.com/getsentry/sentry-python/pull/7181/commits/5d831998645678a4fd510c68a4c325091e3b45eb)

PR #7184, opened 2026-08-13
fix: Allow to correctly compare parsed versions

sentrivana:
Originally raised by a bot [here](https://github.com/getsentry/sentry-python/pull/5005#discussion_r3760275153): the `parse_version` function parses version strings as is (e.g. 3.1 becomes `(3, 1)`). We use these parsed version tuples in integrations to compare the installed version against the minimum (defined in `integrations/__init__.py`). The minimum versions are often three-part, e.g. `(3, 1, 0)`.

This means that we can mistakenly consider a valid version to be below the minimum, because in pure tuple terms, `(3, 1) < (3, 1, 0)` is true.

This can also happen in reverse (package version has three parts, while our min version boundary has two).

In this PR, we make the internal version comparison work as expected regardless of mismatches in the length of the version strings/tuples.


ericapisani (approved):
The comment left by bugbot regarding sanic looks legit. Looks like a quick fix though so approving in order to not block because the changes LGTM otherwise

sentry_sdk/integrations/celery/__init__.py
```diff
@@ -36,7 +37,7 @@
 
 
 try:
-    from celery import VERSION as CELERY_VERSION  # type: ignore
+    from celery import __version__ as CELERY_VERSION  # type: ignore
```
sentrivana, line 40: `VERSION` is celery's own `namedtuple` that they parse out of `__version__`. Use the raw version instead now -- the namedtuple won't compare natively anymore.

sentrivana:
Re-tagging you for a review @ericapisani since I've changed this PR substantially 🙏🏻

ericapisani (approved):

PR #7186, opened 2026-08-13
fix(batcher): Do not let a failed flush kill the flusher thread (#7138)

### LINKED ISSUE #7138: Tolerate exceptions in `_flush_loop()`
We saw this exception

```
  File "/app/.venv/lib/python3.14/site-packages/sentry_sdk/_log_batcher.py", line 33, in _to_transport_format
    k: serialize_attribute(v) for (k, v) in item["attributes"].items()
RuntimeError: _core::PyFeatureContext is unsendable, but is being dropped on another thread"
```

zkasuran:
Closes #7138.

## What

The batcher's `_flush_loop` calls `_flush()` unguarded. Any exception there (the `RuntimeError` in the issue, but really anything raised while serializing or sending a batch) propagates out of the loop and ends the daemon flusher thread. After that `_ensure_thread()` still sees `_flusher_pid == pid` and returns `True`, so `add()` keeps appending to a buffer that nothing drains. Once it reaches `MAX_BEFORE_DROP` every later log, metric or span is dropped silently for the rest of the process lifetime.

## Fix

Following the issue title, wrap the flush in each loop in `capture_internal_exceptions()`, the SDK's own helper for internal errors that should be logged rather than propagated:

- `Batcher._flush_loop` (logs and metrics)
- `SpanBatcher._flush_loop` (both the pending-buckets flush and the periodic full flush)

A single bad batch is now logged and swallowed. The flusher thread stays alive and the next flush delivers everything that queued up in the meantime.

## Tests

One regression test per loop (`test_flush_loop_swallows_flush_exception` in `tests/test_logs.py` and `tests/tracing/test_span_batcher.py`) drives one loop iteration where `_flush` raises, then asserts `_flush_loop` returns instead of propagating. They fail on unpatched source (the `RuntimeError` escapes the loop) and pass with the guard. Both carry the `tests_internal_exceptions` marker since they intentionally exercise the internal-exception path.

## Verification

- `pytest tests/tracing/test_span_batcher.py tests/test_logs.py tests/test_metrics.py`: 68 passed.
- The two new tests fail on unpatched source and pass with the fix.
- `ruff check` and `ruff format --check`: clean on the changed files.
- `mypy sentry_sdk`: no new errors (the pre-existing ones are all in an unrelated integration).

AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: the tests, ruff and mypy above.


sentrivana (approved):
Thanks @zkasuran, makes sense to me.

PR #7201, opened 2026-08-13
fix(starlette): Set transaction name on current scope in sync handler

mjq:
The sync request/response handler passed the _isolation_ scope to `_set_transaction_name_and_source`, but the transaction/segment span lives on the _current_ scope. As a result the route-resolved name never reached the span for sync endpoints, which were instead named by the raw URL from the ASGI middleware (`transaction_info.source` of `url` rather than `route`). Async handlers already used the current scope and were unaffected.

Pass the current scope (already computed above) so sync and async handlers behave identically:
- streaming: the segment name / `sentry.segment.name.source` are route-based
- static: the transaction event name / source are route-based

For parametrized routes this also removes high-cardinality URL transaction names for sync endpoints.

Found while working on https://github.com/getsentry/sentry-python/pull/7183.

### DIFF
--- a/tests/integrations/starlette/test_starlette.py
+++ b/tests/integrations/starlette/test_starlette.py
@@ -1499,6 +1499,47 @@ def test_active_thread_id_span_streaming(sentry_init, capture_items, endpoint):
     assert str(data["active"]) == segments[0]["attributes"]["thread.id"]
 
 
+@pytest.mark.parametrize("endpoint", ["/sync/thread_ids", "/async/thread_ids"])
+def test_transaction_name_span_streaming(sentry_init, capture_items, endpoint):
+    sentry_init(
+        auto_enabling_integrations=False,
+        integrations=[StarletteIntegration(transaction_style="url")],
+        traces_sample_rate=1.0,
+        trace_lifecycle="stream",
+    )
+    app = starlette_app_factory()
+
+    items = capture_items("span")
+
+    client = TestClient(app)
+    response = client.get(endpoint)
+    assert response.status_code == 200
+
+    sentry_sdk.flush()
+
+    segments = [item.payload for item in items if item.payload.get("is_segment")]
+    assert len(segments) == 1
+    assert segments[0]["name"] == endpoint
+    assert segments[0]["attributes"]["sentry.segment.name.source"] == "route"
+
+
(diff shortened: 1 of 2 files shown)


tests/integrations/starlette/test_starlette.py
```diff
@@ -1499,6 +1499,47 @@ def test_active_thread_id_span_streaming(sentry_init, capture_items, endpoint):
     assert str(data["active"]) == segments[0]["attributes"]["thread.id"]
 
 
+@pytest.mark.parametrize("endpoint", ["/sync/thread_ids", "/async/thread_ids"])
+def test_transaction_name_span_streaming(sentry_init, capture_items, endpoint):
```
ericapisani: In the context of span streaming, "transactions" no longer exist in favour of segments, so to avoid the possibility of misleading a future reader once we've removed the transaction code (which is happening in the next major), I'd replace "transaction" with "segment".

This test name more generally is also a bit vague in terms of what's being tested/result expectations  - what are your thoughts on the following?

```suggestion
def test_segment_name_is_route_resolved_name_span_streaming(sentry_init, capture_items, endpoint):
```

tests/integrations/starlette/test_starlette.py
```diff
@@ -1499,6 +1499,47 @@ def test_active_thread_id_span_streaming(sentry_init, capture_items, endpoint):
...
+    sentry_sdk.flush()
+
+    segments = [item.payload for item in items if item.payload.get("is_segment")]
+    assert len(segments) == 1
+    assert segments[0]["name"] == endpoint
+    assert segments[0]["attributes"]["sentry.segment.name.source"] == "route"
+
+
+@pytest.mark.parametrize("endpoint", ["/sync/thread_ids", "/async/thread_ids"])
+def test_transaction_name_static(sentry_init, capture_events, endpoint):
```
ericapisani: Related to my comment above - don't worry so much about this test name as this will be removed when we remove the transaction-related code in this integration.

ericapisani (approved):
Great catch, and thanks for making these changes. Very minor suggestion for the span streaming test, but at the core this LGTM.

Thanks for also separating span streaming enabled and the transactions approach into 2 separate tests - will make it easier when we remove the transactions code 🔥

Issue #4853, opened 2025-09-24
asyncio.CancelledError exceptions are swallowed by capture_internal_exceptions()


aconchillo:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.38.0

### Steps to Reproduce

This issue occurs when using OpenAI SDK and Sentry. If you try to cancel a task that is in the middle of processing OpenAI completion iterators, the task will never be cancelled because Sentry SDK is swallowing the exception.

This occurs in the following code:
https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/openai.py#L337-L370

The following code of course fixes the issue but it defeats the purpose of suppressing internal exceptions:

```python
class CaptureInternalException:
    __slots__ = ()

    def __enter__(self):
        # type: () -> ContextManager[Any]
        return self

    def __exit__(self, ty, value, tb):
        # type: (Optional[Type[BaseException]], Optional[BaseException], Optional[TracebackType]) -> bool
        if ty is not None and value is not None:
            capture_internal_exception((ty, value, tb))

        if isinstance(ty, asyncio.CancelledError):          <-------------------- ADDED
            return False

        return True
```

However, I believe `asyncio.CancelledError` is an special case that should be allowed.

### Expected Result

Task cancellations should work.

### Actual Result

Tasks are never cancelled leaving applications in an invalid state.

sl0thentr0py:
thx for the report @aconchillo that's indeed incorrect wrapping on our end, will fix!

aconchillo:
> thx for the report [@aconchillo](https://github.com/aconchillo) that's indeed incorrect wrapping on our end, will fix!

Thank you for the quick fix! 🙌

PR #4612, opened 2025-07-23
OpenAI integration update

antonpirker:
Update our OpenAI integration to support new APIs and be OTel compatible.

Contains:
- https://github.com/getsentry/sentry-python/pull/4563
- https://github.com/getsentry/sentry-python/pull/4564
- https://github.com/getsentry/sentry-python/pull/4628

PR #7223, opened 2026-08-18
fix(openai): Don't let instrumentation raise into user code

### LINKED ISSUE #7222: OpenAI integration raises `TypeError: 'NoneType' object is not iterable` into user code when `response.output` is `None`
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.68.0 (also reproduces on 2.35.1 and 2.40.0)

### Steps to Reproduce

The OpenAI integration raises out of the instrumented `create()` call when a response object has a `None` collection where the integration expects an iterable. The exception is not contained by the SDK, so it surfaces as an application error and replaces whatever the API actually returned.

This is reachable in practice through OpenAI-compatible gateways. When a gateway has already flushed `200 OK` (for example after sending keep-alive padding while it buffers a large request) and only then learns the upstream call failed, it can no longer change the status code, so it reports the failure in the body:

```json
{"error": {"message": "... At least one of the image dimensions exceed max allowed size: 8000 pixels", "code": 400}}
```

The `openai` client parses that into a model object with no `output` / `choices` populated, and instrumenting it raises.

Minimal reproduction, no network needed:

```python
from unittest import mock

import sentry_sdk
from openai import OpenAI
from openai.types.responses import Response
from sentry_sdk.integrations.openai import OpenAIIntegration

sentry_sdk.init(
    dsn="https://public@o1.ingest.sentry.io/1",
    traces_sample_rate=1.0,
    integrations=[OpenAIIntegration()],
    transport=lambda event: None,
)

# What the client parses out of a 200 response that only carries an error
gateway_error = Response.construct(
    error={"message": "upstream failed", "code": 400},
    output=None,
)

client = OpenAI(api_key="z")
client.responses._post = mock.Mock(return_value=gateway_error)

with sentry_sdk.start_transaction(name="tx"):
    response = client.responses.create(model="gpt-4o", input="hello")
    print(response.error)
```

### Expected Result

`create()` returns the response object, and the application can read `response.error` to see what actually went wrong. Instrumentation problems stay inside the SDK.

### Actual Result

```
  File "sentry_sdk/integrations/openai.py", line 1532, in _new_sync_responses_create
    _set_responses_api_output_data(
  File "sentry_sdk/integrations/openai.py", line 1275, in _set_responses_api_output_data
    _set_common_output_data(
  File "sentry_sdk/integrations/openai.py", line 746, in _set_common_output_data
    for output in response.output:
                  ^^^^^^^^^^^^^^^
```

The application never sees the real error message, so any retry/fallback logic keyed on it stops working.

There appear to be two separate causes:

1. **`response.output` is not checked for `None`.** `response.choices` was given an `is not None` check after #5071, but the Responses API branch still does a bare `hasattr(response, "output")` in `_set_common_output_data` and `_calculate_responses_token_usage`.

2. **The non-streaming paths lost their `capture_internal_exceptions()` wrapper.** Before #4612 the whole response-handling block ran inside `with capture_internal_exceptions():`. The restructuring into `_set_*_input_data` / `_set_*_output_data` moved that code out from under the wrapper, so any error there now propagates into user code. The streaming iterators kept their protection; the non-streaming paths did not. This is why `<2.34` is unaffected.

The second point is the more general one — with it fixed, this class of bug degrades to a missing span attribute rather than an application crash, which matches the SDK contract in CONTRIBUTING.md ("Users do not expect their application to crash").

I have a patch with tests for both points and am happy to open a PR if this looks like the right direction.

FromCSUZhou:
Fixes #7222

## What

The OpenAI integration can raise out of the instrumented `create()` call, so an SDK problem surfaces as an application error and replaces whatever the API actually returned:

```
  File "sentry_sdk/integrations/openai.py", line 1532, in _new_sync_responses_create
    _set_responses_api_output_data(
  File "sentry_sdk/integrations/openai.py", line 746, in _set_common_output_data
    for output in response.output:
                  ^^^^^^^^^^^^^^^
```

## Why

Two independent causes, both addressed here.

**1. `response.output` is never checked for `None`.** `response.choices` got an `is not None` check after #5071, but the Responses API branch still does a bare `hasattr(response, "output")` in `_set_common_output_data` and `_calculate_responses_token_usage`.

**2. The non-streaming paths lost their `capture_internal_exceptions()` wrapper.** Before #4612 the whole response-handling block ran inside `with capture_internal_exceptions():`. Splitting it into `_set_*_input_data` / `_set_*_output_data` moved that code out from under the wrapper, so errors there now propagate into user code. The streaming iterators kept their protection; the non-streaming paths did not — which is why `<2.34` is unaffected.

(1) fixes the crash that was actually hit; (2) makes this class of bug degrade to a missing span attribute rather than an application crash, per the SDK contract in `CONTRIBUTING.md` ("Users do not expect their application to crash").

## How it is reached in practice

Through an OpenAI-compatible gateway. When a gateway has already flushed `200 OK` — for example after sending keep-alive padding while it buffers a large request — and only then learns the upstream call failed, it cannot change the status code, so it reports the failure in the body:

```json
{"error": {"message": "... At least one of the image dimensions exceed max allowed size: 8000 pixels", "code": 400}}
```

The `openai` client parses this into a model object with `output` / `choices` unset. Instrumenting it raises, and the application sees `TypeError: 'NoneType' object is not iterable` instead of the real message — so retry and fallback logic keyed on the upstream error stops working.

Minimal reproduction, no network needed:

```python
from unittest import mock

import sentry_sdk
from openai import OpenAI
from openai.types.responses import Response
from sentry_sdk.integrations.openai import OpenAIIntegration

sentry_sdk.init(
    dsn="https://public@o1.ingest.sentry.io/1",
    traces_sample_rate=1.0,
    integrations=[OpenAIIntegration()],
    transport=lambda event: None,
)

gateway_error = Response.construct(
    error={"message": "upstream failed", "code": 400},
    output=None,
)

client = OpenAI(api_key="z")
client.responses._post = mock.Mock(return_value=gateway_error)

with sentry_sdk.start_transaction(name="tx"):
    response = client.responses.create(model="gpt-4o", input="hello")
    print(response.error)  # raises TypeError before this line on master
```

## Changes

- `_calculate_responses_token_usage` and `_set_common_output_data`: require `response.output is not None`, mirroring the existing `response.choices is not None` check. A response with `output=None` now falls through to the same branch a response with `choices=None` already does.
- `_set_responses_api_input_data`, `_set_completions_api_input_data`, `_set_embeddings_input_data` and `_set_common_output_data` are now thin wrappers that run the recording work inside `capture_internal_exceptions()`. The bodies moved unchanged into `_record_*` helpers.
- `_set_common_output_data` finishes the span *after* the guarded block rather than at the end of each branch, so a failure while recording cannot leave an unfinished span behind.

Regarding #4853 (`asyncio.CancelledError` swallowed by `capture_internal_exceptions()`): that report concerned a wrapper spanning the user's iterator consumption. The blocks wrapped here contain only synchronous instrumentation code with no `await` points and no calls back into user code, so a task cancellation cannot be absorbed by them.

## Testing

Added to `tests/integrations/openai/test_openai.py`:

- `test_responses_api_none_output_does_not_crash` / `..._async` — a Responses object with `output=None` is returned to the caller, and the span is still recorded and finished.
- `test_chat_completion_none_choices_does_not_crash` — regression guard for the `choices=None` path fixed in #5081.
- `test_instrumentation_output_error_does_not_propagate` / `test_instrumentation_input_error_does_not_propagate` — an unexpected error while recording request or response data does not reach the caller, and the span is still finished.

4 of the 5 fail on `master` and pass with this change; the fifth guards the already-fixed `choices=None` path.

```
$ pytest tests/integrations/openai -q
699 passed
```
(694 before this change, 5 new.) `ruff format --check` and `ruff check` are clean on both touched files, and `mypy` reports no new errors relative to `master`.

### DIFF
--- a/sentry_sdk/integrations/openai.py
+++ b/sentry_sdk/integrations/openai.py
@@ -301,7 +301,7 @@ def _calculate_responses_token_usage(
         if streaming_message_responses is not None:
             for message in streaming_message_responses:
                 output_tokens += count_tokens(message)
-        elif hasattr(response, "output"):
+        elif hasattr(response, "output") and response.output is not None:
             for output_item in response.output:
                 if hasattr(output_item, "content"):
                     for content_item in output_item.content:
@@ -326,6 +326,16 @@ def _calculate_responses_token_usage(
 
 
 def _set_responses_api_input_data(
+    span: "Union[Span, StreamedSpan]",
+    kwargs: "dict[str, Any]",
+    integration: "OpenAIIntegration",
+) -> None:
+    # Instrumentation must never raise into the calling application.
+    with capture_internal_exceptions():
+        _record_responses_api_input_data(span, kwargs, integration)
+
+
+def _record_responses_api_input_data(
     span: "Union[Span, StreamedSpan]",
     kwargs: "dict[str, Any]",
     integration: "OpenAIIntegration",
@@ -477,6 +487,16 @@ def _set_completions_api_input_data(
     kwargs: "dict[str, Any]",
     integration: "OpenAIIntegration",
 ) -> None:
+    # Instrumentation must never raise into the calling application.
+    with capture_internal_exceptions():
+        _record_completions_api_input_data(span, kwargs, integration)
+
+
+def _record_completions_api_input_data(
+    span: "Union[Span, StreamedSpan]",
+    kwargs: "dict[str, Any]",
+    integration: "OpenAIIntegration",
+) -> None:
     set_on_span = (
         span.set_attribute if isinstance(span, StreamedSpan) else span.set_data
     )
@@ -598,6 +618,16 @@ def _set_embeddings_input_data(
     kwargs: "dict[str, Any]",
     integration: "OpenAIIntegration",
 ) -> None:
+    # Instrumentation must never raise into the calling application.
+    with capture_internal_exceptions():
+        _record_embeddings_input_data(span, kwargs, integration)
+
+
+def _record_embeddings_input_data(
+    span: "Union[Span, StreamedSpan]",
+    kwargs: "dict[str, Any]",
+    integration: "OpenAIIntegration",
+) -> None:
 
     set_data_normalized(span, SPANDATA.GEN_AI_OPERATION_NAME, "embeddings")
 
@@ -666,6 +696,22 @@ def _set_common_output_data(
     integration: "OpenAIIntegration",
     finish_span: bool = True,
 ) -> None:
+    # Instrumentation must never raise into the calling application, so any
+    # error while reading the response is swallowed here. The span is still
+    # finished afterwards so that a failure cannot leak an unfinished span.
+    with capture_internal_exceptions():
+        _record_common_output_data(span, response, input, integration)
+
+    if finish_span:
+        span.__exit__(None, None, None)
+
+
+def _record_common_output_data(
+    span: "Union[Span, StreamedSpan]",
+    response: "Any",
+    input: "Any",
+    integration: "OpenAIIntegration",
+) -> None:
     client = sentry_sdk.get_client()
     if hasattr(response, "model"):
         set_data_normalized(span, SPANDATA.GEN_AI_RESPONSE_MODEL, response.model)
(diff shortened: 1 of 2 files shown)


alexander-alderman-webb:
Closing because the issue is completed with https://github.com/getsentry/sentry-python/commit/cf426adb428dd693089c745f51ffadacd9cf98cf.

It's fair to say that we should include broader `capture_internal_exceptions` blocks.
We'd probably do this across GenAI integrations. If that happens I'll re-open / attribute you.

PR #7224, opened 2026-08-20
fix(mcp): don't report stateless StreamableHTTP servers as stdio

### LINKED ISSUE #7225: MCPIntegration reports stateless StreamableHTTP servers as mcp.transport "stdio"
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.64.0

### Steps to Reproduce

1. Serve an MCP server over StreamableHTTP with StreamableHTTPSessionManager(app=server, json_response=True, stateless=True), mounted in a Starlette app.

 2. Initialise the SDK with MCPIntegration().

 3. POST a tools/call request to the mount path and let the tool return normally.

 4. Inspect the resulting mcp.server span's attributes.

 Versions: sentry-sdk==2.64.0, mcp==1.28.1, starlette==0.42.0, Python 3.13. The same result holds on mcp 2.x.

### Expected Result

`mcp.transport: "http"`. The server is reachable only over HTTP, so the transport should say so.

### Actual Result

mcp.transport: "stdio" and network.transport: "pipe", with no mcp.session.id.

on our hosted MCP server, 100% of spans claim to be local stdio processes, and any query splitting local from hosted traffic on this attribute is silently meaningless.

Snaylaker:
### Description

`MCPIntegration` reports `mcp.transport: "stdio"` and `network.transport: "pipe"` for a StreamableHTTP server running stateless. `_get_request_context_data` only sets `"http"` when the request carries an `mcp-session-id` header, and a stateless server issues no session id, so no client sends that header back and the `"stdio"` default stands.

The header is a session signal, not a transport signal: per the spec a server *"MAY assign a session ID at initialization time"*, so its presence is a deployment choice. This drops it from the transport check and leaves it supplying `mcp.session.id` when the server is stateful. stdio is unaffected: it attaches no request object, so the outer guard stops it before this branch.

| server | before | after |
| --- | --- | --- |
| stdio | `stdio` | unchanged |
| SSE | `sse` + session id from `?session_id=` | unchanged |
| StreamableHTTP, stateful | `http` + session id from header | unchanged |
| StreamableHTTP, stateless | `stdio` | `http`, no session id |

`test_tool_handler_stateless_streamable_http` drives a real stateless `StreamableHTTPSessionManager` app and fails on `master` with `assert 'stdio' == 'http'`. Green on mcp 1.x and 2.x.

#### Issues

* resolves: #7225

### DIFF
--- a/sentry_sdk/integrations/mcp.py
+++ b/sentry_sdk/integrations/mcp.py
@@ -162,7 +162,7 @@ def _get_request_context_data(
                 # SSE transport uses query parameter
                 mcp_transport = "sse"
                 session_id = request.query_params.get("session_id")
-            elif hasattr(request, "headers") and request.headers.get("mcp-session-id"):
+            elif hasattr(request, "headers"):
                 # StreamableHTTP transport uses header
                 mcp_transport = "http"
                 session_id = request.headers.get("mcp-session-id")
--- a/tests/integrations/mcp/test_mcp.py
+++ b/tests/integrations/mcp/test_mcp.py
@@ -70,6 +70,7 @@ from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
 from starlette.applications import Starlette
 from starlette.responses import Response
 from starlette.routing import Mount, Route
+from starlette.testclient import TestClient
 
 import sentry_sdk
 from sentry_sdk import start_transaction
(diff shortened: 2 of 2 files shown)


ericapisani:
Thanks for opening a PR with us @Snaylaker !

I've read the conversation so far with @alexander-alderman-webb  in the issue that you originally opened and the changes you have here LGTM. 

I've looked into the failing tests and it looks due to an assumption on our test fixture/helper, not your changes specifically (they just uncovered the assumption). I'm opening a PR shortly to address this and once that's merged, we'll look to get your changes in.

ericapisani (approved):
:shipit:

PR #7227, opened 2026-08-23
fix(openai_agents): correctly patch turn_preparation.get_model to capture model spans

x-tahosin:
## Description

Fixes a typo in the `openai_agents` integration where the `get_model` function was incorrectly assigned to the wrong namespace.

In `sentry_sdk/integrations/openai_agents/__init__.py`, the documentation states that for `openai-agents >= 0.8.0`, `AgentRunner._get_model()` was refactored to `agents.run_internal.turn_preparation.get_model()`. The integration correctly wraps `turn_preparation.get_model`, but then incorrectly assigns the wrapped function back to `agents.run_internal.run_loop.get_model`. 

Because `run_loop.get_model` is not used by the `openai-agents` SDK internally to fetch the model, the wrapped method is never called, and `gen_ai.model.invoke` spans may be missed or misattributed. 

This PR fixes the assignment target to `agents.run_internal.turn_preparation.get_model`, ensuring the correct function is patched and the span is successfully tracked.

## Solution
Changed the assignment from `run_loop.get_model` to `turn_preparation.get_model`.

### DIFF
--- a/sentry_sdk/integrations/openai_agents/__init__.py
+++ b/sentry_sdk/integrations/openai_agents/__init__.py
@@ -143,7 +143,7 @@ class OpenAIAgentsIntegration(Integration):
 
                 agents.run_internal.run_loop.get_model = new_wrapped_get_model
 
-                if hasattr(run_loop, "get_all_tools"):
+                if not use_run_hooks and hasattr(run_loop, "get_all_tools"):
                     original_get_all_tools = run_loop.get_all_tools
                     @wraps(original_get_all_tools)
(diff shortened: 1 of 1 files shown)


alexander-alderman-webb:
Please open an issue describing the problem before opening a PR.
The run hooks defined here already create tool spans FWIW:
https://github.com/getsentry/sentry-python/blob/e32031816394208bce58c050ff563bfde8d7b228/sentry_sdk/integrations/openai_agents/patches/runner.py#L40

Issue #7095, opened 2026-08-06
Remove transaction-based tracing from huey


sentrivana:
See parent issue.

pnagre05:
@sentrivana am I free to pick up this one? My approach will the same as the one I used for the graphene integration work that I did, unless I run into any blockers while scoping out the code.

sentrivana:
@pnagre05 Yes, go ahead.

PR #7243, opened 2026-08-24
chore(huey): drop transaction based tracing

pnagre05:
### Description
 I removed the transaction-based tracing fallback from the Huey integration - I did this by doing the following:
- Removed all has_span_streaming_enabled branching and enforced the use of sentry_sdk.traces.start_span for both patch_enqueue and patch_execute
- Removed legacy compatibility utilities like the else: fallback blocks and the legacy continue_trace/set_status calls on scope.transaction
- Deleted transaction-based tests (like the has_span_streaming parameterization) and updated the remaining test suite to unconditionally validate span streaming payloads instead of legacy transactions

#### Issues
* resolves: #7095 

#### Reminders

### DIFF
--- a/sentry_sdk/integrations/huey.py
+++ b/sentry_sdk/integrations/huey.py
@@ -3,17 +3,15 @@ from datetime import datetime
 from typing import TYPE_CHECKING
 
 import sentry_sdk
-from sentry_sdk.api import continue_trace, get_baggage, get_traceparent
-from sentry_sdk.consts import OP, SPANDATA, SPANSTATUS
+from sentry_sdk.api import get_baggage, get_traceparent
+from sentry_sdk.consts import OP, SPANDATA
 from sentry_sdk.integrations import DidNotEnable, Integration, _check_minimum_version
 from sentry_sdk.scope import should_send_default_pii
 from sentry_sdk.traces import SegmentNameSource, SpanStatus, StreamedSpan
 from sentry_sdk.tracing import (
     BAGGAGE_HEADER_NAME,
     SENTRY_TRACE_HEADER_NAME,
-    TransactionSource,
 )
-from sentry_sdk.tracing_utils import has_span_streaming_enabled
 from sentry_sdk.utils import (
     SENSITIVE_DATA_SUBSTITUTE,
     _register_control_flow_exception,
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/huey.py
```diff
@@ -81,38 +79,26 @@ def _sentry_enqueue(
...
-                },
-            )
-        else:
-            span_ctx = sentry_sdk.start_span(
-                op=OP.QUEUE_SUBMIT_HUEY,
-                name=span_name,
-                origin=HueyIntegration.origin,
-            )
-            span_ctx.set_data(SPANDATA.MESSAGING_DESTINATION_NAME, self.name)
+        span_ctx = sentry_sdk.traces.start_span(
```
sentrivana: Can we remove this extra variable and just inline the `sentry_sdk.traces.start_span` call into the context manager on L103?

tests/integrations/huey/test_huey.py
```diff
@@ -76,75 +75,35 @@ def increase(num):
...
 
 @pytest.mark.parametrize("task_fails", [True, False], ids=["error", "success"])
-@pytest.mark.parametrize(
-    "has_span_streaming", [True, False], ids=["streaming", "no_streaming"]
-)
-def test_task_transaction_or_segment(
-    capture_events, capture_items, init_huey, task_fails, has_span_streaming
-):
-    huey = init_huey(has_span_streaming=has_span_streaming)
+def test_task_transaction_or_segment(capture_items, init_huey, task_fails):
```
sentrivana:
```suggestion
def test_task_segment(capture_items, init_huey, task_fails):
```

The "transaction" part was only relevant in non-span-streaming, and we're removing that here. In span streaming, the service-level span is internally called a segment.

sentrivana (commented):
Hey @pnagre05, thanks for this one, too. It looks great, just two minor comments. I'll also run CI on it.

Thanks especially for migrating the tests that we apparently had forgotten about when adding span streaming to this integration, like `test_huey_enqueue`.

tests/integrations/huey/test_huey.py
pnagre05: Just resolved this, as well as the previous comment you pointed out.

sentrivana (approved):
Ty! As soon as CI passes this is good to merge from my POV.

PR #7266, opened 2026-08-26
fix(starlette): Use `url` transaction source for host routing

mjq:
### Description

When host routing, we don't have a route and instead fall back to the request path for the transaction name. This should be given transaction source `url` (for raw URLs) instead of the current value of `route` (for parameterized routes).

This matters for #7183, where the mis-categorized transaction source would lead to bad data in `http.route`. (Bug found by Cursor in https://github.com/getsentry/sentry-python/pull/7183#discussion_r3866410662 when reviewing that PR).

### DIFF
--- a/sentry_sdk/integrations/starlette.py
+++ b/sentry_sdk/integrations/starlette.py
@@ -843,21 +843,24 @@ class StarletteRequestExtractor:
             return None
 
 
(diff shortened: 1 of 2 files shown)


sentrivana (approved):
Ty!

PR #7287, opened 2026-08-28
feat(httpx2): Support data_collection filtering for URL query params

### LINKED ISSUE #7278: Update httpx2 integration to support data collection

ericapisani:
Mirror the httpx integration: apply the data_collection.url_query_params
behaviour to httpx2 span attributes (url.full, url.query, url.fragment) and
breadcrumbs when span streaming is enabled, filtering sensitive query params
instead of relying solely on send_default_pii.

Fixes PY-2743
Fixes https://github.com/getsentry/sentry-python/issues/7278

### DIFF
--- a/sentry_sdk/integrations/httpx2.py
+++ b/sentry_sdk/integrations/httpx2.py
@@ -2,6 +2,9 @@ from typing import TYPE_CHECKING
 
 import sentry_sdk
 from sentry_sdk.consts import OP, SPANDATA
+from sentry_sdk.data_collection import (
+    _apply_data_collection_filtering_to_query_string,
+)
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/httpx2.py
```diff
@@ -45,6 +51,45 @@ def setup_once() -> None:
         _install_httpx2_async_client()
 
 
+def _get_url_attributes(
```
sentrivana, line 54: This looks identical to the one in `httpx`. Would it make sense to extract this into a more generic function? One that could maybe be used in other http client integrations as well?

sentrivana (commented):
LGTM, left one comment/question

sentrivana (approved):

PR #7288, opened 2026-08-28
feat(pyreqwest): Support data_collection filtering for URL query params

### LINKED ISSUE #7281: Update pyreqwest integration to support data collection

ericapisani:
Previously the pyreqwest integration only redacted URL query params
and fragments based on send_default_pii, always leaking the full
query string when PII collection was enabled. Now the
data_collection experiment's url_query_params allow/deny list is
applied to spans and breadcrumbs, matching the behaviour already
shipped for httpx.

Legacy (non span-streaming) breadcrumbs still report the bare URL
without the query string, but the query value itself is filtered
consistently with span streaming.

Fixes PY-2746
Fixes https://github.com/getsentry/sentry-python/issues/7281

### DIFF
--- a/sentry_sdk/integrations/pyreqwest.py
+++ b/sentry_sdk/integrations/pyreqwest.py
@@ -1,9 +1,12 @@
 from contextlib import contextmanager
-from typing import Any, Generator
+from typing import TYPE_CHECKING, Any, Generator
 
 import sentry_sdk
 from sentry_sdk import start_span
 from sentry_sdk.consts import OP, SPANDATA
+from sentry_sdk.data_collection import (
+    _apply_data_collection_filtering_to_query_string,
+)
 from sentry_sdk.integrations import DidNotEnable, Integration
 from sentry_sdk.scope import should_send_default_pii
 from sentry_sdk.traces import StreamedSpan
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/pyreqwest.py
```diff
@@ -86,8 +97,63 @@ def sentry_patched_method(self: "Any", *args: "Any", **kwargs: "Any") -> "Any":
     setattr(cls, method_name, sentry_patched_method)
 
 
+def _get_url_attributes(
```
sentrivana, line 100: Same q as on the 2nd PR. Looks like these `_get_url_attributes` functions always take the `parsed_url`, so it looks like they all get the data in the same format already, so the attribute extraction could follow the same recipe everywhere? If there's currently differences, we can look into consolidating them. IMO would be worth it in the long run as well for everything to be unified.

sentrivana (approved):

sentry_sdk/integrations/pyreqwest.py
ericapisani: I'll do an audit of all the spots to double check they all behave the same. Assuming they do, I'll make a separate PR for the changes 🚀

ericapisani: It was small enough so it's on the last PR in the stack, commit can be found [here](https://github.com/getsentry/sentry-python/pull/7291/commits/af80e45d8b93af7351a037e8d362bd229b3ab943)

PR #7290, opened 2026-08-28
feat(boto3): Support data_collection filtering for URL query params

### LINKED ISSUE #7280: Update boto3 integration to support data collection

ericapisani:
Previously the boto3 integration only redacted URL query params and
fragments based on send_default_pii, always leaking the full query string
when PII collection was enabled. Now the data_collection experiment's
url_query_params allow/deny list is applied to the span-streaming span
attributes and to breadcrumbs, matching the behaviour already shipped for
httpx and pyreqwest.

url.full now includes the filtered query string and fragment, and
url.query / url.fragment are omitted rather than reported as empty
strings when the request URL has none.

The legacy (non span-streaming) span path is left as-is.

Fixes PY-2745
Fixes https://github.com/getsentry/sentry-python/issues/7280

### DIFF
--- a/sentry_sdk/integrations/boto3.py
+++ b/sentry_sdk/integrations/boto3.py
@@ -3,6 +3,9 @@ from typing import TYPE_CHECKING
 
 import sentry_sdk
 from sentry_sdk.consts import OP, SPANDATA
+from sentry_sdk.data_collection import (
+    _apply_data_collection_filtering_to_query_string,
+)
 from sentry_sdk.integrations import DidNotEnable, Integration, _check_minimum_version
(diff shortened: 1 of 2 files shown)


sentry_sdk/integrations/boto3.py
```diff
@@ -62,6 +70,40 @@ def sentry_patched_init(
         BaseClient.__init__ = sentry_patched_init  # type: ignore
 
 
+def _get_url_attributes(
```
sentrivana, line 73: Same q as in the other http client integrations re this function

sentrivana (approved):

PR #7295, opened 2026-08-28
feat(tracing): Gate create_span gen_ai attributes via data_collection

### LINKED ISSUE #7293: Update underlying helper methods for `create_span_decorator` in tracing_utils.py to respect data collection settings
This was missed in the initial data collection pass because this indirectly affects what gets set on a span based on a given span template (which may be an AI agent/chat, etc. span).

ericapisani:
Route input/output attribute collection for AI_AGENT, AI_CHAT, and AI_TOOL span templates through the `data_collection.gen_ai` option when enabled, falling back to `send_default_pii` otherwise. Prompts recorded via `prompt`/`system_prompt` kwargs stay ungated when `data_collection` is not configured, preserving pre-existing behavior.

Fixes PY-2749
Fixes #7293

### DIFF
--- a/sentry_sdk/tracing_utils.py
+++ b/sentry_sdk/tracing_utils.py
@@ -40,7 +40,16 @@ from sentry_sdk.utils import (
 
 if TYPE_CHECKING:
     from types import FrameType
-    from typing import Any, Dict, Generator, Iterator, Optional, Tuple, Union
+    from typing import (
+        Any,
+        Dict,
+        Generator,
+        Iterator,
+        Literal,
+        Optional,
+        Tuple,
+        Union,
+    )
 
     from sentry_sdk._types import Attributes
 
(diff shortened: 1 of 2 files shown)


sentry_sdk/tracing_utils.py
```diff
@@ -1066,20 +1075,21 @@ async def async_wrapper(*args: "Any", **kwargs: "Any") -> "Any":
             span_op = op or _get_span_op(template)
             function_name = name or qualname_from_function(f) or ""
             span_name = _get_span_name(template, function_name, kwargs)
-            send_pii = should_send_default_pii()
+            collect_inputs = is_ai_template and _should_collect_gen_ai("inputs")
+            collect_outputs = is_ai_template and _should_collect_gen_ai("outputs")
```
sentrivana: Why do we now additionally check `is_ai_template` here? Wouldn't `_set_input/_output_attributes` anyway no-op internally  if `not is_ai_template`?

sentrivana (approved):

sentry_sdk/tracing_utils.py
ericapisani: You're right - it would no-op internally. If i recall correctly from when i wrote this originally a few days ago, I think I was trying to create an early return 🤔 

But regardless, you're right that this is unnecessary, will remove 🔥

ericapisani: Addressed in [7fc9210](https://github.com/getsentry/sentry-python/pull/7295/commits/7fc9210c588e534491258dbf0c0233d5c999d01a)

Issue #6990, opened 2026-08-03
Test with Python 3.15rc1 (out Aug 4)


sentrivana:
The first 3.15 release candidate [comes out on Aug 4](<https://peps.python.org/pep-0790/>). Add it to our CI to see whether everything works out of the box.

raulparada:
Hi. It will be needed I think cause seeing some issues:
- when using the new lazy-imports
- only with the sentry-sdk (out of 100 dependencies in the little CLI I tested on)

```
  File "/Users/<redacted>/scripts/cli.py", line 60, in cli
    sentry_sdk.init(
    ~~~~~~~~~~~~~~~^
        dsn=SENTRY_DSN,
        ^^^^^^^^^^^^^^^
    ...<9 lines>...
        add_full_stack=True,
        ^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/<redacted>/.venv/lib/python3.15/site-packages/sentry_sdk/_init_implementation.py", line 56, in _init
    client = sentry_sdk.Client(*args, **kwargs)
  File "/Users/<redacted>/.venv/lib/python3.15/site-packages/sentry_sdk/client.py", line 524, in __init__
    self._init_impl()
    ~~~~~~~~~~~~~~~^^
  File "/Users/<redacted>/.venv/lib/python3.15/site-packages/sentry_sdk/client.py", line 618, in _init_impl
    self.transport = make_transport(self.options)
                     ^^^^^^^^^^^^^^
  File "/Users/<redacted>/.venv/lib/python3.15/site-packages/sentry_sdk/transport.py", line 22, in <module>
    import httpcore
  File "/Users/<redacted>/.venv/lib/python3.15/site-packages/httpcore/__init__.py", line 141, in <module>
    setattr(__locals[__name], "__module__", "httpcore")  # noqa
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'lazy_import' object has no attribute '__module__' and no __dict__ for setting new attributes. Did you mean '.__reduce__' instead of '.__module__'?
```
Good luck, cheers

sentrivana:
Thanks for the heads up @raulparada, I've linked your comment in https://github.com/getsentry/sentry-python/issues/6991 too -- we'll pick up the second RC for testing this week.

PR #7326, opened 2026-09-01
feat: Add Python 3.15 to test matrix

sentrivana:
### 1. Lazy imports

Tried manually testing with a lazy import of `sentry_sdk`, looked ok.

#### `PYTHON_LAZY_IMPORTS=all`

Guarding the optional `httpcore` import with an additional `AttributeError` now, since `httpcore` doesn't work with `PYTHON_LAZY_IMPORTS=all`.

Notified other frameworks/libraries:
- [httpcore](https://github.com/encode/httpcore/discussions/1106)
- [sqlalchemy](https://github.com/sqlalchemy/sqlalchemy/discussions/13557)

Other libraries/frameworks that are problematic under `PYTHON_LAZY_IMPORTS=all` but already have a tracking issue for 3.15 support:
- [fastapi](https://github.com/pydantic/pydantic/issues/13173)

**Caveat:** It's not 100% clear to me whether this forcing of `PYTHON_LAZY_IMPORTS=all` is something libraries are supposed to support, see also https://peps.python.org/pep-0810/#global-lazy-imports-control. Probably fine to wait for the ecosystem to stabilize a bit here, rather than patching everything to be ready on day one.

### 2. CI

Marked some test suites as not python 3.15 ready yet:
  - `langgraph`, `cohere`: fail to install on 3.15
  - `statsig`: hangs on 3.15
  - anything `otlp`/`otel`: hangs on 3.15
  - `integration_deactivation`: fails to install on 3.15

### 3. Misc

Ran the common test suite with `PYTHON_LAZY_IMPORTS=all`. Everything looked fine.

## Issues

Closes https://linear.app/getsentry/issue/PY-2649/test-with-python-315
Closes https://linear.app/getsentry/issue/PY-2651/declare-support-for-315


sentrivana:
While testing manually, I encountered the issue first pointed out [here](https://github.com/getsentry/sentry-python/issues/6990#issuecomment-5426196903). Opened an httpcore [discussion](https://github.com/encode/httpcore/discussions/1106) for it. Looking into a workaround or at least a way to fail gracefully if httpcore is installed for now.

alexander-alderman-webb (approved):

PR #7345, opened 2026-09-02
test: Run tracing tests in stream mode

sentrivana:
Adapt tests of integrations we've already fully ported to span streaming so that they actually use span streaming everytime they define a `traces_sampler` or `traces_sample_rate`. Otherwise, they're implicitly using static tracing, and they'll break the moment we remove it.

Note: there are at least two tests in `test_threading.py` that also need this change but I'll submit them separately because those will need additional changes.


sentrivana:
I'll just merge this since I don't see why anyone would object to this 😆

PR #7397, opened 2026-09-04
fix: match Sentry URLs by parsed hostname

### LINKED ISSUE #7332: DSN netloc substring matching misclassifies unrelated destinations as SDK-internal
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

2.68.1 (latest release; also reproduced on current `master`)

### Steps to Reproduce

`sentry_sdk.utils.is_sentry_url()` currently classifies a candidate using:

```python
client.transport.parsed_dsn.netloc in url
```

This reproduction uses the installed SDK and makes no network requests:

```python
from http.client import HTTPConnection

from sentry_sdk.client import Client
from sentry_sdk.tracing_utils import should_propagate_trace
from sentry_sdk.utils import is_sentry_url

client = Client(
    dsn="https://key@abcd1234.ingest.sentry.io/1",
    trace_propagation_targets=[".*"],
    default_integrations=False,
)

cases = [
    (
        "hostname suffix",
        "https://abcd1234.ingest.sentry.io.evil.test/api/1",
        False,
    ),
    (
        "userinfo",
        "https://abcd1234.ingest.sentry.io@attacker.test/api/1",
        False,
    ),
    (
        "query",
        "https://attacker.test/?next=abcd1234.ingest.sentry.io",
        False,
    ),
    (
        "case-only real host",
        "https://ABCD1234.INGEST.SENTRY.IO/api/1/envelope/",
        True,
    ),
]

for label, url, expected in cases:
    print(
        label,
        expected,
        is_sentry_url(client, url),
        should_propagate_trace(client, url),
    )

custom_port_client = Client(
    dsn="http://key@localhost:9000/1",
    default_integrations=False,
)
connection = HTTPConnection("localhost", 9000)
print(
    "custom-port raw host",
    True,
    is_sentry_url(custom_port_client, connection.host),
)
```

Current output (`expected`, `actual`, `should_propagate_trace`) is:

```text
hostname suffix False True False
userinfo False True False
query False True False
case-only real host True False True
custom-port raw host True False
```

The same behavior reproduces in `sentry-sdk==2.68.1` and current `master` at
`0aa3f2bd4a27f40c7e828cb0f6f92bcf677df852`. The focused existing suites remain
green (`3` classifier tests and `16` trace-target tests), but do not cover these
boundaries.

### Expected Result

Internal-request classification should depend on the candidate's actual
hostname, not an arbitrary occurrence of DSN text.

Proposed narrow contract for maintainer confirmation:

- compare the candidate hostname with the DSN hostname using case-insensitive
  exact equality;
- support both absolute URLs and raw-host input;
- preserve the current scheme-independent behavior;
- exclude port from identity for this focused change;
- do not include child subdomains by default; and
- return `False`, without raising, when a hostname cannot be safely obtained.

This deliberately leaves same-host/different-port requests classified as
internal. Would this focused contract be acceptable? I will wait for a
maintainer response before implementing anything and will follow a different
port or subdomain contract if requested.

### Actual Result

The substring comparison produces seven false positives and two false negatives
in the extended standalone matrix. The additional cases cover an embedded
hostname, path text, a raw-host suffix, and a neighboring IPv6 literal.

The common verified downstream effect is suppression of otherwise configured
`sentry-trace` and `baggage` headers for an unrelated destination. In the
stdlib raw-host path, a lookalike host can also take the early return that skips
the outgoing HTTP span and breadcrumb path. Conversely, a case-only real host
or the raw host of a custom-port DSN is not recognized as internal.

This is an instrumentation-correctness report. I have not demonstrated traffic
redirection, credential disclosure, account takeover, or another security
exploit. I found no matching public Sentry Python issue or open pull request in
bounded symbol and behavior searches.

YusefSyed:
### Description

Determine whether a destination is the configured Sentry endpoint by comparing
its parsed hostname with the DSN hostname instead of searching for the DSN
network-location text anywhere in the candidate string.

This prevents hostname suffixes, URL user-info, paths, and query strings from
being classified as SDK-internal. It preserves scheme-independent and raw-host
input, case-insensitive matching, IPv6 handling, and the agreed
same-host/different-port behavior. The exact contract was confirmed by the
maintainer in #7332.

Regression coverage includes the classifier and downstream trace-header
propagation decisions.

#### Issues

* resolves: #7332

#### Validation

- `uv run tox -e py3.14-common`: 2,431 total cases, 173 skipped, 0 failures/errors
- `uv run --group typing mypy sentry_sdk`: no issues in 193 source files
- full `tests/test_utils.py`: 224 passed, 2 skipped
- full `tests/tracing/test_misc.py`: 43 passed
- full `tests/integrations/stdlib/test_httplib.py`: 91 passed
- `uv run ruff check` and `uv run ruff format --check`: passed
- `uv run pre-commit run --all-files`: passed
- `git diff --check`: passed

#### Review and tooling

AI tools assisted with repository navigation, implementation drafting, test
execution, and diff analysis. The account holder reviewed and approved the
final three-file diff and validation evidence before submission, understands
the hostname-matching contract, and accepts responsibility for the
contribution. The commands above were executed through Codex in the account
holder's local checkout; they are not presented as commands manually typed by
the account holder.

### DIFF
--- a/sentry_sdk/utils.py
+++ b/sentry_sdk/utils.py
@@ -1823,14 +1823,31 @@ def match_regex_list(
 
 def is_sentry_url(client: "sentry_sdk.client.BaseClient", url: str) -> bool:
     """
-    Determines whether the given URL matches the Sentry DSN.
+    Determines whether the given URL's hostname matches the Sentry DSN.
+
+    ``url`` may be an absolute URL or a raw host with an optional port.
     """
-    return (
-        client is not None
-        and client.transport is not None
-        and client.transport.parsed_dsn is not None
-        and client.transport.parsed_dsn.netloc in url
-    )
+    if (
+        client is None
+        or client.transport is None
+        or client.transport.parsed_dsn is None
+    ):
+        return False
+
+    dsn_host = client.transport.parsed_dsn.host
+
+    # ``HTTPConnection.host`` is a raw hostname without brackets for IPv6.
+    if url.lower() == dsn_host.lower():
+        return True
+
+    candidate = url if "://" in url or url.startswith("//") else "//" + url
+
+    try:
+        hostname = urlsplit(candidate).hostname
+    except ValueError:
+        return False
+
+    return hostname is not None and hostname.lower() == dsn_host.lower()
 
 
 def _generate_installed_modules() -> "Iterator[Tuple[str, str]]":
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -882,15 +882,15 @@ def test_check_minimum_version(monkeypatch, version, min_version, expected_pass)
 
 
 @pytest.fixture
-def mock_client_with_dsn_netloc():
+def mock_client_with_dsn():
     """
-    Returns a mocked Client with a DSN netloc of "abcd1234.ingest.sentry.io".
+    Returns a mocked Client with a DSN host of "abcd1234.ingest.sentry.io".
     """
     mock_client = mock.Mock(spec=sentry_sdk.Client)
     mock_client.transport = mock.Mock(spec=sentry_sdk.Transport)
     mock_client.transport.parsed_dsn = mock.Mock(spec=Dsn)
 
-    mock_client.transport.parsed_dsn.netloc = "abcd1234.ingest.sentry.io"
+    mock_client.transport.parsed_dsn.host = "abcd1234.ingest.sentry.io"
 
     return mock_client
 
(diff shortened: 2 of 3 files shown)


YusefSyed:
Following up on the hostname-matching fix for #7332. The change compares parsed hostnames so suffixes, user-info, paths, and query strings do not classify a destination as SDK-internal, while retaining the existing same-host/different-port behavior. Could someone familiar with this area review the intended matching contract? Happy to make changes.

PR #7398, opened 2026-09-05
ci: Isolate `httpx` test suite from `httpx2`

### LINKED ISSUE #7151: httpx CI group also runs httpx2 test suite
Looks like the `httpx` target in the Networks group in CI is also testing httpx2.

Open `httpx` here, for instance: https://github.com/getsentry/sentry-python/actions/runs/31385230658/job/93444058506 It only contains httpx2 runs. Both httpx and httpx2 are run in the httpx group (I assume because we do/tox does some prefix matching). In Python 3.14, specifically, there are no httpx versions to run the tests against, so we only run them against httpx2.

This is not a big deal, but we could check whether we can isolate the test suites.

ZaafirRizwan:
Closes #7151

### Problem

`scripts/runtox.sh` selects tox environments with an unanchored `grep` over `tox -l`:

```sh
ENV="$(uv run tox -l | grep -- "$searchstring" | grep -v -- '-latest$' | tr $'\n' ',')"
```

The Network workflow's `Test httpx` step passes `py<version>-httpx`, which therefore also matches the `py<version>-httpx2-*` environments.

The result:

- On versions where both suites exist (e.g. 3.12), the `httpx` job ran the httpx2 suite as well as its own, duplicating the work already done by the `httpx2` job.
- On 3.14 / 3.14t / 3.15, where there are no `httpx` environments at all, the `httpx` job ran **only** httpx2 environments — which is what the issue reports.

### Fix

Rename the tox test suite from `httpx` to `httpx0` (httpx is still on 0.x), so the `httpx0` and `httpx2` selectors can no longer collide. This follows the precedent set for `openai-base` in #4730, where the suite was renamed rather than making the selector boundary-aware.

`integration_name: "httpx"` is added alongside the rename so the suite keeps resolving to the existing integration:

- `_TESTPATH` stays `tests/integrations/httpx` (the test directory is unchanged)
- the minimum supported version lookup against `_MIN_VERSIONS` in `sentry_sdk/integrations/__init__.py` keeps working, so the suite still floors at httpx 0.16.0

Without `integration_name` the suite would point at a nonexistent `tests/integrations/httpx0` directory and lose its version floor.

### Changes

- `scripts/populate_tox/config.py` — rename the suite, add `integration_name`
- `scripts/split_tox_gh_actions/split_tox_gh_actions.py` — update the `Network` group
- `scripts/populate_tox/README.md` — the `python` key example documents this suite; updated to match
- `tox.ini` and `.github/workflows/test-integrations-network.yml` — regenerated

The `tox.ini` diff is a pure rename: 207 lines removed, 207 added, no version or dependency changes.

### Note

The same class of collision still exists for `redis`, which is a prefix of `redis_py_cluster_legacy`, so the `Test redis` step also runs the cluster-legacy suite. Left out of this PR to keep it scoped to #7151 — happy to open a separate issue for it.

### DIFF
--- a/.github/workflows/test-integrations-network.yml
+++ b/.github/workflows/test-integrations-network.yml
@@ -41,10 +41,10 @@ jobs:
         run: |
           set -x # print commands that are executed
           ./scripts/runtox.sh "py${{ matrix.python-version }}-grpc"
-      - name: Test httpx
+      - name: Test httpx0
         run: |
           set -x # print commands that are executed
-          ./scripts/runtox.sh "py${{ matrix.python-version }}-httpx"
+          ./scripts/runtox.sh "py${{ matrix.python-version }}-httpx0"
       - name: Test httpx2
         run: |
           set -x # print commands that are executed
--- a/scripts/populate_tox/README.md
+++ b/scripts/populate_tox/README.md
@@ -146,7 +146,7 @@ a Python version older than 3.9 if the HTTPX version is 0.28 or higher, you can
 say:
 
 ```python
-"httpx": {
+"httpx0": {
     "python": {
         # run the test suite for httpx v0.28+ on Python 3.9+ only
         ">=0.28": ">=3.9",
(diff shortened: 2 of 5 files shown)


sentrivana (approved):
Thanks @ZaafirRizwan.

PR #7418, opened 2026-09-07
chore(aws_lambda): Remove transaction-based tracing

sentrivana:
User setting was missing in the span streaming branch.

Duplicated tests -- removed the non-span-streaming versions.

Closes https://linear.app/getsentry/issue/PY-2682/remove-transaction-based-tracing-from-aws-lambda


alexander-alderman-webb:
Could do the user stuff on master as well.

alexander-alderman-webb (approved):

PR #7422, opened 2026-09-08
Skip formatting log records when Sentry logs are off

### LINKED ISSUE #7402: `SentryLogsHandler.emit` formats every log record before checking `has_logs_enabled`
### Summary

`SentryLogsHandler.emit` calls `self.format(record)` before it checks whether Sentry Logs are enabled, so every `logging` record in the process pays a full `Formatter.format` even when the feature is off. Measured at **~1.9 µs per log record** on CPython 3.14 / Apple M2.

### Where

`sentry_sdk/integrations/logging.py`:

```python
class SentryLogsHandler(_BaseHandler):
    def emit(self, record):
            self.format(record)              # <-- unconditional
            if not self._can_record(record):
                return
            if not client.is_active():
                return
            if not has_logs_enabled(client.options):   # <-- the cheap check comes last
                return
            self._capture_log_from_record(client, record)
```

`LoggingIntegration.__init__` defaults `sentry_logs_level=DEFAULT_LEVEL` (INFO), so the handler is installed for every user, and `LoggingIntegration` is a default integration. Applications that never set `enable_logs` - the default - format every INFO+ record for nothing.

`BreadcrumbHandler.emit` has the same shape, but there the formatting is actually used to build the breadcrumb, so only this one is wasted work.

### Suggested fix

Move the three guards above the `format()` call:

```python
def emit(self, record):
        if not self._can_record(record):
            return
        if not client.is_active() or not has_logs_enabled(client.options):
            return
        self.format(record)
        self._capture_log_from_record(client, record)
```

`_capture_log_from_record` reads `record.message`, which `format()` sets, so it has to stay before that call - just not before the guards.

### Impact

Measured on a FastAPI endpoint emitting three log records per request, in-process, SDK otherwise at defaults:

| config | added µs/req |
|---|---:|
| defaults | +99.7 |
| `LoggingIntegration(sentry_logs_level=None)` | +92.9 |
| `LoggingIntegration(level=None, sentry_logs_level=None)` | +73.3 |

The first delta (6.8 µs over three records) is this bug. Small per record, but it scales with log volume and applies to every application that has not opted into Sentry Logs.

Environment: CPython 3.14.7, sentry-sdk 2.67.1, fastapi 0.141.1, Apple M2.

gyanu2507:
`SentryLogsHandler.emit` called `format()` before checking whether logs are enabled. Every stdlib record paid for a Formatter even with the feature off.

Format after the capture check.

Fixes #7402

### DIFF
--- a/tests/integrations/logging/test_logging.py
+++ b/tests/integrations/logging/test_logging.py
@@ -8,6 +8,7 @@ from sentry_sdk import get_client
 from sentry_sdk.consts import VERSION
 from sentry_sdk.integrations.logging import (
     LoggingIntegration,
+    SentryLogsHandler,
     ignore_logger,
     ignore_logger_for_sentry_logs,
     unignore_logger,
@@ -242,6 +243,23 @@ def test_sentry_logs_collection_off_by_default(sentry_init, capture_items, reque
     get_client().flush()
 
     assert not items
+
+
+def test_sentry_logs_handler_skips_format_when_disabled(sentry_init):
+    sentry_init()
+    handler = SentryLogsHandler()
+    record = logging.LogRecord(
+        name="test-logger",
+        level=logging.INFO,
+        pathname=__file__,
+        lineno=1,
+        msg="hello %s",
+        args=("world",),
+        exc_info=None,
+    )
+    with mock.patch.object(handler, "format") as formatted:
+        handler.emit(record)
+    formatted.assert_not_called()
 
 
 def test_sentry_logs_collection_opt_in(sentry_init, capture_items, request):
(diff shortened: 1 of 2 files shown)


tests/integrations/logging/test_logging.py
```diff
@@ -244,6 +245,23 @@ def test_sentry_logs_collection_off_by_default(sentry_init, capture_items, reque
     assert not items
 
 
+def test_sentry_logs_handler_skips_format_when_disabled(sentry_init):
+    sentry_init()
```
sentrivana: Better to be explicit here to not have this break if the default changes:

```suggestion
    sentry_init(integrations=[LoggingIntegration(capture_sentry_logs=False)])
```

sentrivana (commented):
Thanks @gyanu2507, the change looks good to me, please have a look at the suggestion I left.

tests/integrations/logging/test_logging.py
gyanu2507: Done. The test now passes `LoggingIntegration(capture_sentry_logs=False)` so it still holds if the default flips.

sentrivana (approved):

PR #7431, opened 2026-09-08
chore(grpc): Remove transaction-based tracing

sentrivana:
Closes https://linear.app/getsentry/issue/PY-2696/remove-transaction-based-tracing-from-grpc


alexander-alderman-webb (approved):
Thanks, completely missed these

sentrivana:
Sure, np. Going over everything and double-checking now 👀

PR #7442, opened 2026-09-09
chore: Remove GenAI input truncation functions

alexander-alderman-webb:
### Description

Remove dead truncation code.
Both `truncate_and_annotate_messages()` and `truncate_and_annotate_embedding_inputs()` were only used in the static trace lifecycle. This lifecycle is dropped in the major.

#### Reminders
- Please add tests to validate your changes, and lint your code using `uv run ruff`.
- Add GH Issue ID _&_ Linear ID (if applicable)
- PR title should use [conventional commit](https://develop.sentry.dev/engineering-practices/commit-messages/#type) style (`feat:`, `fix:`, `ref:`, `meta:`)
- For external contributors: [CONTRIBUTING.md](https://github.com/getsentry/sentry-python/blob/master/CONTRIBUTING.md), [Sentry SDK development docs](https://develop.sentry.dev/sdk/), [Discord community](https://discord.gg/Ww9hbqr)

PR #7449, opened 2026-09-09
chore: Remove unnecessary `None` checks

sentrivana:
All of these should be lists, never `None`.


tests/integrations/openai_agents/test_openai_agents.py
```diff
@@ -3935,49 +3932,6 @@ def calculator(a: int, b: int) -> int:
     )
 
 
-def test_openai_agents_message_truncation(
```
sentrivana: Removed because there is no message truncation anymore (correct me if that's incorrect)

alexander-alderman-webb: correct, I have this open to remove all remnants: https://github.com/getsentry/sentry-python/pull/7442

alexander-alderman-webb (approved):

alexander-alderman-webb:
I've removed unnecessary checks in the other GenAI integrations as well.

PR #7452, opened 2026-09-10
chore: Remove has_span_streaming from core

sentrivana:
WIP.


sentrivana:
will do this in https://github.com/getsentry/sentry-python/pull/7468

PR #7462, opened 2026-09-11
chore: Adapt even more tests

sentrivana:
Remove tests testing implementation details of the old tracing system. Adapt tests that still make sense in span streaming. Remove tests that already have a streaming version.


sentry_sdk/traces.py
```diff
@@ -668,7 +668,7 @@ def __init__(
         self._start()
 
     def __repr__(self) -> str:
-        return f"<{self.__class__.__name__}(sampled={self.sampled})>"
+        return f"<{self.__class__.__name__}(name={self.name}, sampled={self.sampled})>"
```
sentrivana, line 671: Now that no-op spans optionally accept a name, adding it to `__repr__` makes debugging easier.

alexander-alderman-webb (approved):

PR #7481, opened 2026-09-11
feat(boto3): Add common OTel AWS client attributes

### LINKED ISSUE #7474: ref(boto3): Own one logical botocore client-call lifecycle
### Problem

* spans are currently created in `request-created`. Botocore may emit this during retries; it's also too late to cover any failures during operation lookup, parameter construction, serialization, or other work before an HTTP request exists.

### Proposed Solution

* patch `BaseClient._make_api_call` as [OTel](<https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py>) does; one logical client span should cover all retries until response, error, or cancellation.
* wrapper is the only owner of the aws client span's start and completion.
* `_AwsCallContext` can be used to contain generic data such botocore client, service name, operation name, region, endpoint URL, etc

#### Sources

* [https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#rpc-client-span](<https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#rpc-client-span>)
* [https://github.com/boto/botocore/blob/develop/botocore/client.py](<https://github.com/boto/botocore/blob/develop/botocore/client.py>)
* [https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py](<https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py>)

### LINKED ISSUE #7475: feat(boto3): Add generic attributes that are common across services
* add generic attributes that are reused across different AWS services.
* request attributes must be created before span creation so streamed-span sampling can consider them.

| Field | Value |
| -- | -- |
| span name | `<modeled ServiceId>.<Operation>`, for example `S3.GetObject` |
| `sentry.op` | `http.client` |
| `sentry.origin` | `auto.http.boto3` |
| `rpc.system.name` | `aws-api` |
| `rpc.service` | modeled service ID |
| `rpc.method` | modeled operation name |
| `cloud.provider` | `aws` |
| `cloud.region` | client region, when available |
| `server.address` | parsed endpoint hostname, when valid |
| `server.port` | endpoint port according to existing URL conventions |

#### Sources

* [https://opentelemetry.io/docs/specs/semconv/cloud-providers/aws-sdk/](<https://opentelemetry.io/docs/specs/semconv/cloud-providers/aws-sdk/>)
* [https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py](<https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py>)

pabloDeputter:
### Description
Implements #7475  by adding common OTel attributes to boto client spans.

### Changes
- add the following attributes: `rpc.system.name`, `rpc.service`, `rpc.method`, `cloud.region`, `server.address`, `server.port`.
- set `rpc.system.name` to `aws-api`
- changing span name to `Service.Operation` (now it's `aws.<service>.<Operation>`), e.g. `S3.HeadObject` following OTel is a breaking change and will be done in major.
- use modeled AWS service for `rpc.service`; not supported anymore by OTel, but JS keeps this too.
- use boto operation name for `rpc.method`.
- extend `AwsCallContext` to carry extra client metadata for instrumentation.

#### Issues
Resolves #7475


sentry_sdk/consts.py
```diff
@@ -977,12 +983,24 @@ class SPANDATA:
...
     Status code of the RPC returned by the RPC server or generated by the client.
     Example: "DEADLINE_EXCEEDED"
     """
 
+    RPC_SYSTEM_NAME = "rpc.system.name"
+    """
+    A string identifying the remoting system.
+    Example: "aws-api"
+    """
+
```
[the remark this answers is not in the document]
pabloDeputter: `rpc.system` is deprecated (https://opentelemetry.io/docs/specs/semconv/cloud-providers/aws-sdk/#aws-sdk-spans) and replaced by `rpc.system.name`.

tests/integrations/boto3/test_s3.py
```diff
@@ -161,7 +162,7 @@ def test_streaming(
 
         span1 = event["spans"][0]
         assert span1["op"] == "http.client"
-        assert span1["description"] == "aws.s3.GetObject"
+        assert span1["description"] == "S3.GetObject"
```
alexander-alderman-webb: Changing span descriptions should go in the major.

pabloDeputter: Does this also apply to span name, since these represent the same parameter?

alexander-alderman-webb: Yes it does!
Splitting up `rpc.method` into `rpc.service` and `rpc.method` is fine to do against master, since we can consider the attribute to be "wrong" and therefore can change the behavior.
However, I wouldn't consider the span name or description to be "wrong" in the same sense, hence it should go in the major.

sentry_sdk/integrations/boto3/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Boto3Integration(Integration):
+    identifier = "boto3"
+    origin = f"auto.http.{identifier}"
+
+    @staticmethod
+    def setup_once() -> None:
+        version = parse_version(BOTOCORE_VERSION)
+        _check_minimum_version(Boto3Integration, version, "botocore")
```
alexander-alderman-webb: This strict version boundary should also live in the major.
Since the integration is auto-enabling, it may suddenly stop working for users.

sentry_sdk/integrations/boto3/_client.py
```diff
@@ -0,0 +1,74 @@
...
+        span: "Optional[Union[Span, StreamedSpan]]" = None
+
+        with capture_internal_exceptions():
+            ctx = AwsCallContext(self, operation_name, api_params)
+
+        if ctx is not None:
+            with capture_internal_exceptions():
+                span = _start_client_span(ctx)
+                if span is not None:
+                    span.__enter__()
```
alexander-alderman-webb: If part of the `AwsCallContext` initialization fails, we don't create the span at all.
Would it make sense to only omit the attributes that couldn't be extracted and still create the span?

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -0,0 +1,343 @@
...
+    if not endpoint_url:
+        return {}
+
+    default_ports = {
+        "http": 80,
+        "https": 443,
+    }
+
+    try:
+        parsed_url = urlsplit(endpoint_url)
```
alexander-alderman-webb, line 60: What's the difference between `urlsplit()` and `sentry_sdk.utils.parse_url()`?

sentry_sdk/integrations/boto3/_client.py
```diff
@@ -0,0 +1,74 @@
...
+
+        try:
+            parsed = orig_make_api_call(self, operation_name, api_params)
+        except BaseException as exc:
+            if span is not None:
+                with capture_internal_exceptions():
+                    _finish_client_span_with_error(span, exc)
+            raise
```
alexander-alderman-webb, line 71: Please remove the `_finish_client_span_with_error()` function. This is just needless indirection.

```suggestion
			exc_info = sys.exc_info()
            with capture_internal_exceptions():
            	if span is not None:
                    span.__exit__(*exc_info)
            reraise(*exc_info)
```

alexander-alderman-webb: You can probably use the span as a context manager here tbh

pabloDeputter: yeah `_finish_client_span_with_error()` has more logic further in the pr stack, but it does make sense to remove it for now.

sentry_sdk/integrations/boto3/_client.py
pabloDeputter: yeah makes sense. `AwsCallContext` now always contains the data already supplied to `_make_api_call()` (so `operation_name` and `params` with fallback to empty dict); the other metadata is extracted independently; Missing metadata such as `rpc.service` is replaced with `unknown` such that span name can still be created, e.g. `aws.unknown.GetObject`.

sentry_sdk/integrations/boto3/_client.py
```diff
@@ -0,0 +1,80 @@
...
+            span = _start_client_span(ctx)
+            if span is not None:
+                span.__enter__()
+
+        try:
+            parsed = orig_make_api_call(self, operation_name, api_params)
+        except BaseException as exc:
+            if span is not None:
+                with capture_internal_exceptions():
+                    _finish_active_http_child_span(span)
```
alexander-alderman-webb, line 69: Why are we ending the stdlib span here?

pabloDeputter: We finish the HTTP span before the parent boto3 span so that the correct order is preserved. Otherwise the HTTP span would be kept active until the whole response is read, while the parent boto3 span finishes first; afterwards when the HTTP span finishes, it will restore the already-finished parent as current, causing subsequent calls to be parented incorrectly.

pabloDeputter: I can send a screenshot from sentry maybe to show against latest release

pabloDeputter: old:
Screenshot 2026-09-15 at 14 18 55
new:
Screenshot 2026-09-15 at 14 18 46

sentry_sdk/integrations/boto3/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+class Boto3Integration(Integration):
+    identifier = "boto3"
+    origin = f"auto.http.{identifier}"
+
+    @staticmethod
+    def setup_once() -> None:
+        # local import to avoid import cycle
+        from sentry_sdk.integrations.boto3._client import _patch_botocore_client
+
+        _patch_botocore_client()
```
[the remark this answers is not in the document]
alexander-alderman-webb: bro really

sentry_sdk/integrations/boto3/_instrumentation.py
pabloDeputter: `ursplit()` gives us exactly what we need (port and hostname); `sentry_sdk.utils.parse_url()` additionally removes sensitive information and doesn't give us port or hostname directly (we'll need to call `urlsplit()` additionally).

```
>>> from urllib.parse import urlsplit
>>> urlsplit("https://user:pass@example.com:8443/foo?token=secret#section")
SplitResult(scheme='https', netloc='user:pass@example.com:8443', path='/foo', query='token=secret', fragment='section')
>>> from sentry_sdk.utils import parse_url
>>> parse_url("https://user:pass@example.com:8443/foo?token=secret#section")
ParsedUrl(url='https://[Filtered]:[Filtered]@example.com:8443/foo', query='token=[Filtered]', fragment='section')
```

sentry_sdk/integrations/boto3/__init__.py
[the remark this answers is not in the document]
pabloDeputter: cursor is based

PR #7493, opened 2026-09-14
ref!: Stop disabling client integrations when an agent integration is present

alexander-alderman-webb:
### Description

Use a boolean scope member to suppress chat spans from client libraries that are inside a chat span from an agent framework. Add the `traces._AgentFrameworkChatGenerationContext` wrapper to ensure the scope member is cleaned up correctly when the agent framework span exits.

Remove logic that disables client library integrations when an agent library is active. Delete the `integration_deactivation` tox environment that ran the corresponding tests.

Use the context manager in all agent frameworks that create chat generation spans (`langchain`, `pydantic-ai`, and `openai-agents`). Add early returns for chat generation spans in the `openai`, `anthropic`, and `google-genai` client libraries to prevent duplicate generation spans.

Remove embedding patches from the `langchain` integration. These must now be provided by client libraries like `openai`, `anthropic`, etc.

#### Issues

Closes https://github.com/getsentry/sentry-python/issues/5515

#### Reminders


sentry_sdk/traces.py
```diff
@@ -93,6 +93,53 @@ def __str__(self) -> str:
...
+            active=active,
+        )
+        if type(self._span) is not StreamedSpan:
+            return
+
+        self.span._scope._agent_framework_generation_entered = True
+
+    @property
+    def span(self) -> "StreamedSpan":
+        return self._span
```
sentrivana, line 126: Why have a special property for this and not return the span directly in `__enter__`? Might be wrong, but from a cursory look we seem to only ever be interacting with the span when using the context manager.

sentrivana (commented):
I guess this would work.

Why do we want to go via this route in the first place? What do we gain over the current way of doing things?

alexander-alderman-webb:
Users that use both agent and client frameworks in the same process would benefit.
So, e.g., Seer does a `google-genai` call to search the docs inside a tool. Seer doesn't use an off-the-shelf framework but generally this seems to be a common pattern that I've seen a few times in users' traces.

sentry_sdk/traces.py
alexander-alderman-webb: The idea would be for the agent frameworks to use this wrapper with the context manager protocol since `__exit__()` sets

```python
self._span._scope._agent_framework_generation_entered = False
```

sentrivana (approved):
OK sounds good!

Issue #7346, opened 2026-09-03
`_experiments={"data_collection": None}` disables the event scrubber without enabling data collection


JOhnsonKC201:
`data_collection` is typed as `Optional[DataCollectionUserOptions]` in the `Experiments` TypedDict, so `None` type checks. But the two functions that read it disagree about what `None` means:

- `has_data_collection_enabled()` (`sentry_sdk/utils.py`) checks for **key presence**: `"data_collection" in options.get("_experiments", {})`
- `_resolve_data_collection()` (`sentry_sdk/data_collection.py`) only treats a **non-None** value as user provided: `if user_dc is not None`

So with `data_collection=None` the client lands in a state where neither side of the option is active. `data_collection` resolves to the legacy `send_default_pii` mapping with `provided_by_user=False`, but every caller of `has_data_collection_enabled()` takes the data collection branch.

Two consequences I was able to reproduce:

**1. The default event scrubber is never installed.**

In `_get_options()`, `has_data_collection_enabled(rv)` is true, so the `EventScrubber` is skipped. An explicitly configured `event_scrubber` is also discarded, with a warning saying data collection configuration was provided, which is not what happened.

```python
sentry_sdk.init(_experiments={"data_collection": None})
sentry_sdk.get_isolation_scope().set_extra("password", "hunter2")
sentry_sdk.capture_message("hi")
```

| config | `extra["password"]` in the event |
|---|---|
| default (no `_experiments`) | `"[Filtered]"` |
| `data_collection={}` | `"[Filtered]"` |
| `data_collection=None` | `"hunter2"` |

Same for `token` and `x_forwarded_for`.

**2. WSGI request attributes gain URL query params.**

`_get_request_attributes()` takes the data collection branch and attaches `http.query`, `url.path` and `url.full`, using the denylist derived from `send_default_pii=False`. Without `data_collection` set at all, `should_send_default_pii()` is false and none of those are attached.

```python
# QUERY_STRING = "email=a%40b.com&coupon=SAVE10", no data_collection set
attributes.get("http.query")  # None

# same request, _experiments={"data_collection": None}
attributes.get("http.query")  # 'email=a%40b.com&coupon=SAVE10'
```

### Which behaviour is intended?

[...]

PR #7502, opened 2026-09-14
Reproduction for sentry-python#7346

ericapisani:
Reproduction for #7346.

`_experiments={"data_collection": None}` type checks, but the two functions reading the option disagree about what `None` means:

- `has_data_collection_enabled()` (`sentry_sdk/utils.py:2108`) gates on **key presence** → `True`
- `_resolve_data_collection()` (`sentry_sdk/data_collection.py:287`) only treats a **non-`None`** value as user provided → `provided_by_user=False`

So the client lands between the two: every caller of `has_data_collection_enabled()` takes the data collection branch, while the resolved config is the legacy `send_default_pii=False` mapping. `None` is the only config where the two disagree — `sentry-python/7346/repro.py` prints that gate table first, then the three observable symptoms:

1. the default `EventScrubber` is skipped, so `extra["password"]` is sent as `hunter2`
2. WSGI request attributes gain `http.query`, `url.path` and `url.full`
3. an explicitly configured `event_scrubber` is discarded, with a warning claiming data collection configuration was provided

## Running it

```bash
cd sentry-python/7346
uv run repro.py
```

No DSN needed — events are inspected in `before_send`. It runs against this checkout, since `data_collection` is an unreleased `_experiments` option, and exits `0` while the bug reproduces, `1` once it is fixed.

## One correction to the issue

The issue's table reports `[Filtered]` for `data_collection={}`. On master that config also yields `hunter2`, because skipping the `EventScrubber` is intended whenever data collection is configured (`sentry_sdk/client.py:357`). The report still holds — the divergence is `has_data_collection_enabled()` vs `provided_by_user`.

https://claude.ai/code/session_01K1iSm2akLNZAgPpzBkrbBe


ericapisani:
I closed this because I was trying to create a reproduction locally and the bot decided to go off on its own...

PR #7525, opened 2026-09-16
tests: add real coverage for _generate_installed_modules

### LINKED ISSUE #4594: Test `_generate_installed_modules`
We removed the test for this method in getsentry/sentry-python#4593, as the test was pointless (essentially, just comparing the method to a slight refactor of itself). 

We should figure out a way to test the method properly. Perhaps, this would involve running the test inside a `tox` environment where all the packages are pinned.

hikmetba-bit:
## Summary
Fixes #4594.

The previous test for `_generate_installed_modules` was removed in #4593 because it was pointless — it just re-implemented the function's own loop and compared the two, so it never exercised any behavior independent of the implementation.

## Change
Added `test_generate_installed_modules` to `tests/test_utils.py`, which mocks `importlib.metadata.distributions()` with a small fixed set of fake distributions and asserts on the actual contract:
- names are normalized case-insensitively
- an entry with no `Name` in its metadata is skipped
- an entry with no version is skipped
- a case-insensitive duplicate name is skipped in favor of the first occurrence

This is independent of whatever packages happen to be installed in the test environment, so it doesn't need the `tox`-pinned-environment approach suggested in the issue.

## Test plan
- `python -m pytest tests/test_utils.py -k installed_modules -v` — both `test_installed_modules_caching` (existing) and the new `test_generate_installed_modules` pass.
- Verified the new test actually catches a regression: locally removed the `normalized_name not in yielded` dedup check from `_generate_installed_modules`, confirmed the test fails with a clear diff, then reverted.


alexander-alderman-webb:
We don't need tests for private helpers. I'll close the PR and the issue.

PR #7537, opened 2026-09-18
ref(boto3): Split integration into internal modules

pabloDeputter:
### Description
Move `Boto3Integration` from single module into package with smaller internal modules.

- botocore client patching is moved into `_client.py`.
- request instrumentation and event handlers into `_instrumentation.py`.
- existing public import is kept unchanged: `from sentry_sdk.integrations.boto3 import Boto3Integration`.


sentry_sdk/integrations/boto3/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Boto3Integration(Integration):
+    identifier = "boto3"
+    origin = f"auto.http.{identifier}"
+
+    @staticmethod
+    def setup_once() -> None:
+        # local import to avoid import cycle
+        from sentry_sdk.integrations.boto3._client import _patch_botocore_client
```
ericapisani, line 11: @pabloDeputter and I chatted a bit about this offline, leaving notes here for other reviewers and so we don't forget about this when we come back to this on Monday:

We're going to look to reorganize things a little bit in order to remove this dynamic import and the one on line 24 in the `_client.py` file below .

The import cycle can be mitigated by moving the values for `identifier` and `origin` that currently live on the `Boto3Integration` to a `consts.py` file, and having the `_instrumentation.py` file import those values from there.

We can also look to pull up the minimum version check from within `_client.py` to here.

So this file would look like the following after these changes:

```python
from sentry_sdk.integrations import DidNotEnable, Integration, _check_minimum_version
from sentry_sdk.integrations.boto3._client import _patch_botocore_client
from sentry_sdk.integrations.boto3.consts import IDENTIFIER, ORIGIN
from sentry_sdk.utils import parse_version

try:
    from botocore import __version__ as BOTOCORE_VERSION
except ImportError:
    raise DidNotEnable("botocore is not installed")


class Boto3Integration(Integration):
    identifier = IDENTIFIER
    origin = ORIGIN

    @staticmethod
    def setup_once() -> None:
        version = parse_version(BOTOCORE_VERSION)
        _check_minimum_version(Boto3Integration, version, "botocore")

        _patch_botocore_client()
```

ericapisani (changes requested):

sentry_sdk/integrations/boto3/__init__.py
pabloDeputter: Good idea; like I said: I thought it was not really justified to have a separate `consts.py` file for only 2 constants for now, but it will probably be handy to have this in the future as the integration expands.

alexander-alderman-webb: If it's too tricky we can just keep the single file structure as well

pabloDeputter: I'll think there also might be some overlap in constants between different services; so it would be nice to have it for that as well.

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -109,6 +111,8 @@ def _sentry_request_created(
 def _sentry_before_sign(
     request: "AWSRequest", signature_version: "Any", **kwargs: "Any"
 ) -> None:
+    from sentry_sdk.integrations.boto3 import Boto3Integration
```
ericapisani, line 114: A fun thing that you can do to try and get around the need for this dynamic import - you can provide the string rather than the class to `client.get_integration`.

So I think lines 117 and 118 below would look like this:

```
if client.get_integration("Boto3Integration") is None:
        return
```

ericapisani (approved):
Thanks for making the earlier changes. There's one more change that we can try and make to avoid the need for dynamic imports altogether, have left a comment within one of the methods explaining.

Otherwise, this LGTM, approving in order to not block once that's been addressed

sentry_sdk/integrations/boto3/_instrumentation.py
pabloDeputter: Okay good to know, I didn't knew this was possible. Then we'll probably use `client.get_integration("boto3")` instead, right? Since we use `boto3` as identifier for the integration.

pabloDeputter: I'll do this probably in a later PR, cause I don't want to do merge hell once again 😭

ericapisani: Sounds good, totally understandable 👍🏻

PR #7538, opened 2026-09-18
fix(boto3): Trace the complete botocore client-call lifecycle

### LINKED ISSUE #7474: ref(boto3): Own one logical botocore client-call lifecycle

pabloDeputter:
### Description

Move boto3 span creation from individual HTTP request attempts to full botocore client-call lifecycle.

Previously, the client span was created from the botocore `request-created` event; since botocore creates a new `AwsRequest` event for every retry, these spans represent individual retries rather than the full event. Following OTel ([https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#rpc-client-span](<https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#rpc-client-span>)) the span should cover the entire call lifecycle including all retries. Wrapping `_make_api_call()` covers all retries performed by botocore, including serialization, endpoint resolution, the final response or failures ([https://github.com/boto/botocore/blob/develop/botocore/client.py:999](<https://github.com/boto/botocore/blob/develop/botocore/client.py:999>)).

* patch `BaseClient._make_api_call()` so one span represents a single boto operation; this span is also kept active across every retry attempt.
* `request-created` is kept for breadcrumbs, HTTP request attributes, and trace propagation.
* client spans that fail will preserve the original exception.
* responses that return a `StreamingBody` are kept open until body is consumed or closed.
* fix the correct parent/child relationship between boto, HTTP, and streaming spans.<br><br>

#### Issues

Resolves #7474


sentry_sdk/integrations/boto3/_client.py
```diff
@@ -1,44 +1,108 @@
...
 except ImportError:
-    raise DidNotEnable("botocore is not installed")
+    raise DidNotEnable("botocore not installed")
 
 
-def _patch_botocore_client() -> None:
-    from sentry_sdk.integrations.boto3 import Boto3Integration
+@contextmanager
+def _activate_client_span(span: "StreamedSpan") -> "Iterator[StreamedSpan]":
+    """Temporarily activate an inactive boto span without ending it."""
```
ericapisani, line 31: Worth adding context on why we're doing this.

sentry_sdk/integrations/boto3/_client.py
```diff
@@ -1,44 +1,108 @@
...
-        meta.events.register("after-call-error", _sentry_after_call_error)
+
+    def sentry_patched_make_api_call(
+        self: "BaseClient", operation_name: str, api_params: "Any"
+    ) -> "Any":
+        """
+        Track a single API call, including retries, serialization, and endpoint
+        resolution. For streaming responses, keep the span open until the
+        response body is consumed or closed.
+        https://github.com/boto/botocore/blob/develop/botocore/client.py
```
ericapisani, line 63: I would change this to a permalink to that if the authors move this file (or something similar) we won't lose what you are referencing here


```suggestion
        https://github.com/boto/botocore/blob/2891dd413e8b644590042d8ff226fc2181711985/botocore/client.py
```

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -19,145 +19,63 @@
...
-                    "sentry.op": OP.HTTP_CLIENT,
-                    "sentry.origin": Boto3Integration.origin,
-                    SPANDATA.RPC_METHOD: f"{service_id}/{operation_name}",
-                },
+        return None
+
+    # use unknown if `service_id_hyphenized` so span name can still be created.
+    # e.g. "aws.unkown.GetObject"
+    service_name = ctx.service_id_hyphenized or "unknown"
+    span_name = "aws.%s.%s" % (service_name, ctx.operation_name)
```
ericapisani, line 44: f-strings are generally the preferred way to construct strings like this in modern python these days

```suggestion
    span_name = f"aws.{service_name}.{ctx.operation_name}"
```

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -19,145 +19,63 @@
...
+    if has_span_streaming_enabled(client.options):
+        if sentry_sdk.traces.get_current_span() is None:
+            return None
+
+        attributes: "Attributes" = {
+            SPANDATA.SENTRY_OP: span_op,
+            SPANDATA.SENTRY_ORIGIN: span_origin,
+        }
+        if ctx.service_id:
+            attributes[SPANDATA.RPC_METHOD] = "%s/%s" % (
```
ericapisani, line 57: Same f-string comment here

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -194,8 +112,8 @@ def _instrument_streaming_body(
             parent_span=span,
             active=False,
             attributes={
-                "sentry.op": OP.HTTP_CLIENT_STREAM,
-                "sentry.origin": Boto3Integration.origin,
+                SPANDATA.SENTRY_OP: OP.HTTP_CLIENT_STREAM,
+                SPANDATA.SENTRY_ORIGIN: Boto3Integration.origin,
```
ericapisani, line 116: Nice change👌🏻

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -274,28 +194,146 @@ def sentry_raw_stream_close(*args: "Any", **kwargs: "Any") -> None:
...
+            if has_span_streaming_enabled(client.options)
+            else sentry_sdk.get_current_span()
+        )
+        if span is None:
+            return
+
+        # an ignored streamed span is not activated; avoid enriching its parent.
+        if isinstance(span, StreamedSpan) and (
+            span.get_attributes().get(SPANDATA.SENTRY_ORIGIN) != Boto3Integration.origin
+        ):
```
ericapisani, line 281: Because of the length of this conditional, I think it'd be a bit cleaner to pull this into a variable

```suggestion
        is_span_origin_from_boto = span.get_attributes().get(SPANDATA.SENTRY_ORIGIN) == Boto3Integration.origin
        if isinstance(span, StreamedSpan) and is_span_origin_from_boto:
```

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -274,28 +194,146 @@ def sentry_raw_stream_close(*args: "Any", **kwargs: "Any") -> None:
...
+        if request.url is None or not should_propagate_trace(client, request.url):
+            return
+
+        def _replace_header(request: "AWSRequest", key: str, value: str) -> None:
+            """
+            Botocore's `HTTPHeaders` inherits from `email.message.Message`, where:
+                headers["foo"] = "old"
+                headers["foo"] = "new"
+            produces two fields: {"foo": "old", "foo": "new"}. So delete existing
+            fields before assigning replacement.
```
ericapisani, line 313: Great comment, thanks for adding this 👍🏻 🙏🏻

sentry_sdk/integrations/stdlib.py
```diff
@@ -300,6 +307,9 @@ def putrequest(
                         "sentry.op": OP.HTTP_CLIENT,
                         SPANDATA.HTTP_REQUEST_METHOD: method,
                     },
+                    # boto3 integration owns span's lifecycle; keep child inactive so it
+                    # can't restore boto3 span later on.
+                    active=not is_inactive_boto3_span,
```
ericapisani, line 312: Apologies for this slight nitpick, but can we add a space after the `=` and before the `not`? 

The "not" almost blends in with the `=`, especially because the syntax highlighting of Github and IDEs use the same colour for the two. 😭

ericapisani (commented):
Minor things but overall looking good. Will take a look at tests on my 2nd pass

sentry_sdk/integrations/boto3/_instrumentation.py
pabloDeputter: yeah it's really weird behavior 😆

sentry_sdk/integrations/boto3/_instrumentation.py
pabloDeputter: yupp, I took over the old code, but I usually also use f-strings.

sentry_sdk/integrations/boto3/_instrumentation.py
pabloDeputter: I agree, but then you get the issue that the legacy span doesn't support `get_attributes()`; I fixed it like this:
```python
# an ignored streamed span is not activated; avoid enriching its parent.
if isinstance(span, StreamedSpan):
    if not (span.get_attributes().get(SPANDATA.SENTRY_ORIGIN) == ORIGIN):
        return
```

sentry_sdk/integrations/stdlib.py
pabloDeputter: I agree, but the linter doesn't really like that 🤣 I fixed it by adding `fmt: off` and `fmt: on` for that specific block; I haven't seen it being used anywhere else in the codebase, so not sure whether this is correct.

sentry_sdk/integrations/boto3/_client.py
```diff
@@ -1,44 +1,115 @@
...
+        except BaseException as error:
+            _finish_span(span, error)
+            raise
+
+        streaming_body_instrumented = False
+        with capture_internal_exceptions():
+            streaming_body_instrumented = _instrument_streaming_body(span, parsed)
+
+        # `StreamingBody`s finish their span when consumed or closed.
+        if not streaming_body_instrumented:
```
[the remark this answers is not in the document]
pabloDeputter: This can be fixed, but then we either go back to the previous behavior where we finish the stream span after the boto span; or we would need to wrap `Transaction.finish()`, etc

PR #7540, opened 2026-09-18
fix(boto3): Finish `StreamingBody` span correctly

pabloDeputter:
### Description
Finish boto streaming spans when a `StreamingBody` is consumed, closed or fails. Previously, streaming spans where only finished when a read returned no data or `StreamingBody.close()` was called.


sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -160,26 +160,39 @@ def _replace_header(request: "AWSRequest", key: str, value: str) -> None:
...
+        else:
+            span.__exit__(type(error), error, error.__traceback__)
 
-    span.__exit__(None, None, None)
+
+def _instrument_streaming_body(
+    span: "Union[Span, StreamedSpan]", parsed: "Dict[str, Any]"
+) -> bool:
+    if isinstance(span, NoOpStreamedSpan):
+        return False
```
ericapisani, line 184: Why return a boolean here when we're not using the result?

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -160,26 +160,39 @@ def _replace_header(request: "AWSRequest", key: str, value: str) -> None:
...
     if not isinstance(body, StreamingBody):
-        return
+        return False
 
     streaming_span: "Union[Span, StreamedSpan]"
     if isinstance(span, StreamedSpan):
         streaming_span = sentry_sdk.traces.start_span(
             name=span.name,
             parent_span=span,
+            active=False,
```
ericapisani, line 195: Because this is an unusual decision (creating a span that we're marking as inactive) and then finishing it at a later point, I'd document why this was done for future readers.

It's not clear why this is happening if I were to read the code outside of this pull request.

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -194,35 +207,86 @@ def _sentry_after_call(
...
             ret = orig_read(*args, **kwargs)
-            if ret:
-                return ret
-
-            if isinstance(streaming_span, StreamedSpan):
-                streaming_span.end()
-            else:
-                streaming_span.finish()
+            with capture_internal_exceptions():
+                amount = args[0] if args else kwargs.get("amt")
```
ericapisani, line 235: Couple of variable name suggestions to improve readability:

ret => `read_return_value`
amount => `amount_of_bytes_requested`

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -194,35 +207,86 @@ def _sentry_after_call(
 
     orig_read = body.read
     orig_close = body.close
+    raw_stream = body._raw_stream  # type: ignore[attr-defined]
+    orig_raw_close = raw_stream.close
+    finished = False
+
+    def finish(error: "Optional[BaseException]" = None) -> None:
```
ericapisani, line 214: Not specific to your changes and more of a general note on why I'm suggesting this (and other) renames:

We have more than a few vague names within the SDK - both within the "working code" and our tests. This makes it difficult to understand what exactly is being invoked when calling a method, what data is represented by a variable, or what behaviour we're looking to test.

I'd like us to try and move a bit more in the direction of being more specific so that someone reading this in the future doesn't have to jump to function definitions or ask a clanker what pieces of data mean or what's being tested (at least as often as we may have to now).

```suggestion
    def finish_span(error: "Optional[BaseException]" = None) -> None:
```

tests/integrations/boto3/test_client.py
```diff
@@ -1,6 +1,132 @@
...
     assert Boto3Integration.__module__ == "sentry_sdk.integrations.boto3"
     assert Boto3Integration.identifier == "boto3"
+
+
+@pytest.fixture
+def client_factory(sentry_init, monkeypatch, span_streaming):
+    sentry_init(
+        traces_sample_rate=1.0,
+        integrations=[Boto3Integration()],
+        trace_lifecycle="stream" if span_streaming else "static",
```
ericapisani, line 28: When these changes land, we're going to have to make sure that we remove the branching on `span_streaming` in the `major/3.0` branch 😅

pabloDeputter: Yupp, there are quite a lot of these cases 😬

sentry_sdk/integrations/boto3/_instrumentation.py
pabloDeputter: good to know, I'll try to keep this in mind in the future.

sentry_sdk/integrations/boto3/_instrumentation.py
pabloDeputter: I probably messed up when splitting the large PR in smaller ones; in the next PR of the stack, we'll use this return value to check whether there was 1. a `StreamingBody` as response and 2. it was instrumented; since we only want to delay closing the boto span if both conditions are met. 

I'll document this better in the next PR.

sentry_sdk/integrations/boto3/_instrumentation.py
pabloDeputter: Yeah, I agree. I added a comment with some examples where it might fail when it's not initialized with `active=False`.

sentry_sdk/integrations/boto3/_instrumentation.py
```diff
@@ -194,35 +207,86 @@ def _sentry_after_call(
 
     orig_read = body.read
     orig_close = body.close
+    raw_stream = body._raw_stream  # type: ignore[attr-defined]
+    orig_raw_close = raw_stream.close
+    finished = False
+
```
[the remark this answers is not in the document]
pabloDeputter: I'll check this out in the next PR, cause I don't want to have merge issues

PR #7560, opened 2026-09-21
ref(data-collection): Promote `data_collection` from `_experiments` o top-level

### LINKED ISSUE #7557: Move data_collection setting to top-level of config
This should continue to remain an optional, opt-in property.

ericapisani:
Move `data_collection` out of `_experiments` into a documented top-level `sentry_sdk.init()` option. Update `has_data_collection_enabled` to check `provided_by_user` on the resolved dict instead of key presence under `_experiments`, and update all integration tests to pass `data_collection` at the top level.

Fixes PY-2796
Fixes #7557


sentry_sdk/consts.py
```diff
@@ -89,7 +89,6 @@ class CompressionAlgo(Enum):
                 Callable[[SpanJSON, Hint], Optional[SpanJSON]]
             ],
             "suppress_asgi_chained_exceptions": Optional[bool],
-            "data_collection": Optional[DataCollectionUserOptions],
```
sentrivana, line 92: Leave it in experiments, too, please, so that it doesn't stop working for folks that are using it at the moment

sentrivana: Might need a helper that retrieves it from the right place (experiments or top-level, with top-level taking precedence) to use in code. (Or just inline, looks like it's only two places?)

sentry_sdk/data_collection.py
```diff
@@ -293,8 +293,12 @@ def _resolve_data_collection(options: "Dict[str, Any]") -> "DataCollection":
     concrete values for every field.
 
     ``data_collection`` must be a plain ``dict``.
+
+    Must be called exactly once per options dict, before ``client._get_options``
+    overwrites ``options["data_collection"]`` with the resolved result. Feeding an
+    already-resolved dict back in would flip ``provided_by_user`` to ``True``.
     """
-    user_dc = options.get("_experiments", {}).get("data_collection")
```
sentrivana, line 297: Flagging this -- we will need to use the [helper](https://github.com/getsentry/sentry-python/pull/7560/changes#r4063760897) here

sentry_sdk/utils.py
```diff
@@ -2109,7 +2110,13 @@ def has_data_collection_enabled(options: "Optional[dict[str, Any]]") -> bool:
     if options is None:
         return False
 
-    return "data_collection" in options.get("_experiments", {})
+    data_collection: "Optional[DataCollection]" = options.get("data_collection")
```
sentrivana, line 2113: Here, too

sentrivana (approved):
Approving to not block, but please see comments.

alexander-alderman-webb (approved):

